PR #226, opened 2015-01-29
Topic/filter and match

Neverlord:
- user-defined types only need to be announced when using `message_builder` or `libcaf_io`
- no more static `types_array` for each match expression
- give type numbers for builtin types to speed things up
- generate type tokens based on the type numbers
- new `message::filter` function
- got rid of the last `cppa` remnants; `message::filter_cli` replaces `cppa::options_description` API


Neverlord:
Seems like Jenkins is finally happy with the changes. By the way @josephnoir, this patch removes the static type arrays previously used by the pattern matching, all access to `std::type_info` when using builtin types only, and all calls to `dynamic_cast` in the core library. Should definitely help porting CAF to RIOT-OS and maybe reduces binary size along the way (mind checking?).

PR #241, opened 2015-02-24
Add a profiled scheduler for workers and jobs.

mavam:
The new` profiled<Policy>` scheduling policy wraps an existing policy and injects its own instrumentation calls to report to the `profiled_coordinator` scheduler. This coordinator enables fine-grained profiling of workers and jobs (i.e., actors) by calling getrusage before and after each job resumption. It logs its measurements at a given
resolution to a file, with absolute and cumulative timings. One can activate the new scheduler as follows:

    set_scheduler(
      new scheduler::profiled_coordinator<>{
        "/tmp/caf.log", std::chrono::milliseconds{1000}, threads, thruput});

This would log to `/tmp/caf.log` at a resolution of 1 second.


mavam:
test this please

PR #244, opened 2015-03-03
Fix build on MSVC

Neverlord:


Neverlord:
I'd say we wait for the next MSVC compiler update before retrying. Maybe we can reduce the amount of necessary changes as well, keeping the code base more stable.

Issue #141, opened 2014-07-01
Provide abstraction for pool/set of actors 


mavam:
I scenario I face quite common is to maintain a group of actors. At first I always consider using the actual group class, but then quickly decide against it in favor of a `std::vector<actor>`. The reason is that I *(i)* want to manage group membership myself, as opposed the actor its memberships, and *(ii)* want to iterate over the members (or just determine the number of members) to drive a decision in the managing actor. 

Since an STL container of `actor` handles does not transparently act as an actor itself, there are many annoying explicit loops. Moreover, with an explicit container it is now necessary to monitor every actor and remove it from the container when it terminates. This is a lot of boilerplate which a libcppa primitive could nicely abstract away.

Thus, I suggest to introduce an actor "bag" abstraction: call it `pool` because it has thread pool semantics, or `client_set` to allude to the semantics of managing a group of actors.

Neverlord:
Closed via f42f143eca801fdf3a198a66a31f9f0d06fb97fc.

PR #247, opened 2015-03-04
Provide abstraction for pool of worker actors

Neverlord:
See #141


mavam:
It's Christmas! :gift: 

Some questions:
- In order to provide local failure isolation (e.g., by removing a failed worker from the set), the parent (or any other actor knowing the pool) must monitor *all* workers and use the system messages to remove them from the pool, right? If failed actors continue to live in the pool, not simply one, but all future messages to this actor get lost. This does not seem to be a problem for broadcast (other than the dead actor "leaking memory"), but for round-robin or random dispatching. Would it be possible to have a "safer" pool which removes workers as they die?
- Out of curiosity, is the only reason you need the spinlocks because the pool can be passed around and accessed concurrently? It looks like it has message-based interface to control it, which made me wonder why it is needed.

Neverlord:
> It's Christmas! 
> 
> Some questions:
> 
> In order to provide local failure isolation (e.g., by removing a failed worker from the set), the parent (or any other actor knowing the pool) must monitor all workers and use the system messages to remove them from the pool, right? If failed actors continue to live in the pool, not simply one, but all future messages to this actor get lost. This does not seem to be a problem for broadcast (other than the dead actor "leaking memory"), but for round-robin or random dispatching. Would it be possible to have a "safer" pool which removes workers as they die?
You're of course right about failure isolation, but how would you address this? Monitoring workers would also mean to have a queue in the pool, never give a worker more than one item, and intercept response messages to give it its next job. This just breakes as soon as actors forward messages... Alternatively, the pool would need to be able to "inject" pending messages whenever a worker is about to become blocked. This also breaks: if the worker sync_sends something and waits for a response before completing a task.

Only monitoring the workers would mean to loose all messages sitting in its mailbox regardless. It would reduce the loss via removing it from the set eventually, but that's not a good solution either.

> Out of curiosity, is the only reason you need the spinlocks because the pool can be passed around and accessed concurrently? It looks like it has message-based interface to control it, which made me wonder why it is needed.
The Pool dispatches directly in the thread of the sender. It's neither scheduled nor does it have its own thread. That's what makes it more efficient than any "custom actor" users could come up with.

mavam:
> Monitoring workers would also mean to have a queue in the pool, never give a worker more than one item, and intercept response messages to give it its next job. 

Why the lock-step? If there are no messages in the queue, it works perfectly. If there are *N* messages in the queue, yeah, you lose O(*N*). That's still much better than losing *all* future messages, especially when considering distributed systems that run continuously. Since a pool is often just a dispatcher and doesn't do the work itself, it is a natural fit for such a long-running component. 

Speaking about sync send: would "the right thing happen" if I sync-send message to a pool with either random or round-robin dispatching?

> Only monitoring the workers would mean to loose all messages sitting in its mailbox regardless. It would reduce the loss via removing it from the set eventually, but that's not a good solution either.

Per the above argument, I think this would give significant mileage compared to continuous message dropping.

Neverlord:
> Speaking about sync send: would "the right thing happen" if I sync-send a message to a pool with either random or round-robin dispatching?

Yes.

> Per the above argument, I think this would give significant mileage compared to continuous message dropping.

Convinced. The latest commit adds monitoring for all workers. I gave it some thought as well and I think a more restrictive policy could `sync_send` each task to the worker while caching the original message as well as the original sender and message ID. On a sync failure, the policy could remove the worker and dispatch the message to a different one. Maybe I'll add such a `reliable_anycast` policy later on. Right now I'm not quite sure if it wouldn't end up being too slow and complicated.

mavam:
> The latest commit adds monitoring for all workers. 

:+1: 

> I gave it some thought as well and I think a more restrictive policy could sync_send each task to the worker while caching the original message as well as the original sender and message ID. On a sync failure, the policy could remove the worker and dispatch the message to a different one. Maybe I'll add such a reliable_anycast policy later on. Right now I'm not quite sure if it wouldn't end up being too slow and complicated.

This sounds quite cool, as a separate policy. In this case it would be quite important that workers don't exhibit a high variance in their processing, as it could result in undesired buffering in the pool's sender context. But then we could also give the pool a maximum buffer size for messages to hold. As I think about this this, flow-control sounds like a more appropriate mechanism to handle this orthogonally.

Neverlord:
test this please

Neverlord:
Merged with squash + rebase: f42f143eca801fdf3a198a66a31f9f0d06fb97fc

PR #248, opened 2015-03-04
Add new decorator for concatenated messages

### LINKED ISSUE #173: Allow for constructing nested messages from other messages
It would be nice to have the ability to next messages. Here is a fictitious example:

```cpp
message_handler h =
  [=](expensive const& e) {
     send(sink, make_message(unwrap(last_dequeued()), 42));
  };
```

The scenario is that a data structure with an expensive copy constructor exists already as a message (and thus internally as a reference-counted object). Then I'd like to be able to still reuse it in differently typed messages. It appears to be already possible to nest messages and extract them via `apply`, but as a user, I don't want to work with this extra level of type-erased indirection. Rather, I'd like to directly reuse the value. In the above example, the send statement with the fictitious `unwrap` would create a message of type `<expensive, int>` as opposed to `<message, int>`. I'm not saying that an `unwrap` function is the right primitive, but I hope it illustrates the point.

Hamdor:
Its now possible to concat messages, to do so you can use the
`message::concat(message other)` or the `+` operator.
Closes  #173

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -35,6 +35,7 @@ set (LIBCAF_CORE_SRCS
      src/binary_serializer.cpp
      src/blocking_actor.cpp
      src/channel.cpp
+     src/concatenated_tuple.cpp
      src/continue_helper.cpp
      src/decorated_tuple.cpp
      src/default_attachable.cpp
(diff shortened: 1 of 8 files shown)


mavam:
Saaweeet, been looking for this for a while! :+1:

Neverlord:
test this please

Hamdor:
Thank you for the feedback. :)

PR #252, opened 2015-03-15
Add workaround for a bug in `compare_exchange_weak`

Hamdor:
Instead of `compare_exchange_weak` we use `comare_exchange_strong`
on older compilers to prevent race conditions.
This bug is present in GCC <= 4.8.2 and Clang <= 3.4.0.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60272
https://llvm.org/bugs/show_bug.cgi?id=18899

### DIFF
--- a/libcaf_core/caf/detail/caf_weak.hpp
+++ b/libcaf_core/caf/detail/caf_weak.hpp
@@ -17,90 +17,27 @@
  * http://www.boost.org/LICENSE_1_0.txt.                                      *
  ******************************************************************************/
 
+#ifndef CAF_WEAK_HPP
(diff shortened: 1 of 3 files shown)


Neverlord:
Just a bit after-the-fact feedback: `CAF_WEAK_HPP => CAF_DETAIL_CAS_WEAK_HPP` and `caf_weak.hpp => cas_weak.hpp`.

Neverlord:
Just for future reference, final commit is https://github.com/actor-framework/actor-framework/commit/913b8233fa4a4481a99621e2a865926da159e86d

PR #257, opened 2015-03-26
no __thread on iOS and android. Use pthread_getspecific instead.

araste:
I am a mobile app developer and also a big fan of CAF. I have tried to build CAF on iOS and almost succeeded except one source file: libcaf_core/src/logging.cpp . It is because __thread keyword is unsupported on iOS and android. I have written a patch that uses pthread_key_create and pthread_getspecific / setspecific to store actor ids into TLS on iOS and android.

### DIFF
--- a/libcaf_core/src/logging.cpp
+++ b/libcaf_core/src/logging.cpp
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <pthread.h>
 #include <condition_variable>
+#include <stdint.h>
 
 #ifndef CAF_WINDOWS
 #include <unistd.h>
@@ -36,6 +37,14 @@
 # include <cxxabi.h>
 #endif
 
+#if defined(__APPLE__)
+# include <TargetConditionals.h>
+# if TARGET_OS_IPHONE == 1
+#  define CAF_USE_PTHREAD_KEY
+# endif
+#elif defined(__ANDROID__)
+# define CAF_USE_PTHREAD_KEY
+#endif
 
 #include "caf/string_algorithms.hpp"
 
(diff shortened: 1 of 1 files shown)


Neverlord:
Welcome aboard! :)

Thank you for contributing to CAF. Im curious, did you have to tweak the build setup / CMake files to get CAF compiling?

The PR has some minor issues: By calling `self_id_init();` only from `initialize()`, you initialize the key for the very first thread only. The call to `pthread_once` belongs to `get_self_id()`. It also does not follow the CAF coding convention (never use C-style casts).

The bigger problem is that it introduces code duplication (not your fault). This would be the second time we make use of the thread-specific POSIX API, so we should refactor the existing code from `memory.hpp` and wrap it properly. It makes sense to adopt the interface of good ol' `boost::thread_specific_ptr` to allow replacing one with the other (for boost.actor).

I'll refactor the existing code (see #258) and then merge your patch with the new interface.

araste:
Thank you for this quick reply!

I am happy that you reckon my need to use CAF on mobile devices. I was just to tell you that CAF is very close to be used for mobile app development. As far as I know in mobile devices there is very small room for pthread thread specific storage; limited to 59 for android devices. (See http://mobilepearls.com/labs/native-android-api/ .) It will be preferable to have pthread_key_t as few as possible.

I am using Cocoapods to make an XCode build system of CAF for Mac and iOS. The source tree of your project is so clean that I did little work to build it on iOS. As soon as you release a new version without __thread keyword I will publish a cocoapods podspec file that every iOS developer can easily integrate CAF into their project and take advantage of your great work.

PR #259, opened 2015-03-29
configure and CMakeLists.txt changes for iOS build

araste:
Here I made some changes to build CAF for iOS. 

configure file options:
 --sysroot : set system root dir for clang. set iphoneos for iphone devices, and iphonesimulator for ios simulators
 --ios-min-ver : set the minimum iOS version supported by the compiled binary

and made changes to CMakeLists.txt to support these config options.

Only works with Xcode generator.

I have written scripts/build-ios-universal.sh to build universal static library targeting both device and simulator. You can run this script to check whether CAF builds well for iOS devices.

Surely the build fails this time, because of logging.cpp `__thread` keyword problem. ;)

Tested on CMake 3.1.3, Mac OS X 10.10 and Xcode 6.2.


Neverlord:
Thank you for your contribution to CAF. I'll fix the `__thread` issue later this week. ;)

I mislike the hardcoded paths (you assume `../configure`). We should at least validate if `SRCROOT` is correct and add a comment to the `.sh` script to explain its purpose.

mavam:
Great to see iOS porting underway.

I wonder whether the build script should be a Makefile target instead, so that simply typing `make` would build according to the CMake parameters specified via `configure`. Then we could move the shell script into the `cmake` folder because it's not really a script CAF users would interact with directly, that is, it is an implementation detail of the build procedure.

araste:
@Neverlord: build-ios-universal.sh is a 'meta' make script. For iOS developers it is convenient to have a universal library that can both be used by iphone device and ios simulator. But cmake can only target one and universal library should be made outside cmake build process.

it locates the configure file relative to the script file directory, not the current working directory. So you can run the script anywhere and it will locate the source root correctly unless you move the shell script into another location in the repository. I will write more description in the script and verify whether it points to the correct directory.

Neverlord:
Hm. I don't really like the "meta-build-script", but I don't see a better way to handle this with CMake either. Can you please squash the commits into a single one and give the commit a good description? Something like "Add iOS-related build options". Thanks. :)

araste:
I squashed the commits and editted the commit message. please check this out.

https://github.com/araste/actor-framework/commit/3bcadac59c53df08886017affe1c331f6ba3ac4e

Neverlord:
Thank you for the patch. I rebased your branch to current `develop` and slightly edited the commit message to conform to our Git style: https://github.com/actor-framework/actor-framework/commit/1ea21d7cdf8caf4a10adf8303f4192172c033965. Hope you're ok with that.

PR #260, opened 2015-03-29
Add biicode support

outro56:
- Modified the CMakeList.txts to delegate to biicode.cmake when a biicode build is detected
- Added copies of the project license to the sub projects so that biicode publishes them also. I did a copy instead of symlinking because of git windows support for symbolic links. If you do not want this, we can simply eliminate having biicode also publish the project license information


Neverlord:
Thank you for you contribution. We're always glad to have CAF run in more build environments, but this PR is quite intrusive and has a lot of redundancies. Why copy the license files? The `bii.cmake` files are identical, why not move it to the other cmake files and include it properly? The `biicode.conf` files have mostly empty groups, can't this be omitted?

I would prefer a setup that does not add files to the main repository. Can Biicode resolve (Git) links, so that we can move the biicode setup to a separate repository and link all library components together there?

libcaf_core/bii.cmake
```diff
@@ -0,0 +1,45 @@
+# build a shared library
+set(BII_LIB_TYPE SHARED)
```
outro56: This forces biicode to create a dynamic shared library instead of its default static library.

outro56:
> The ``bii.cmake`` file are identical

Yes they are, because logically of the way I think CAF is structure. ``libcaf_core`` and ``libcaf_io`` a two different modules libraries and the later depends on the former. The way biicode works i generally tries to build static or dynamic libraries (based on configuration) and so that they could be linked against an executable. 

biicode has its own separate build process and automatically resolves module dependencies (using information in ``biicode.conf`` and ``biicode.cmake``). This is the reason why I split off biicode's make file instruction into a different file. 

Note though the contents of the ``bii.cmake`` file could be dumped into the existing make files, but would need to short circuit evaluation and wont look pretty IMHO

> The biicode.conf files have mostly empty groups, can't this be omitted?

You could probably leave off empty sub groups in this file. Note the ``conf`` file in the libcaf_io folder references a version for libcaf_core. 

Alternatively I could probably try re-factoring the whole thing so biicode considers both libcaf_core and libcaf_io as one module and thus builds and publish only one static/share library when libcaf is being referenced.

> Why copy the license files

Like I mentioned in the PR comment, if its not important to you to have biicode also copy down the license information when users reference the library then it can be removed from the PR

> Can Biicode resolve (Git) links, so that we can move the biicode setup to a separate repository and link all library components together 

The beauty of biicode is that it makes it totally pain free for consumers of your library to not have to bother with knowing what git repository, make file settings to use for compiling your library or knowing what version is stable to use. This is all controlled by you publishing to biicode. 

If you are familiar with maven (java world), nuget (c# .net world) or npm (javascript/node) then I would say biicode works basically the same way. I would highly suggest working though the sample (http://docs.biicode.com/c++/gettingstarted.html). 

From your side publishing just as easy as executing ``bii publish`` on the command line. biicode automatically figures out which files in the project directory would be needed by library users publishes them out to your feed on their website

My attempt in this PR was to show an initial approach to achieving this for libcaf. You can take a look at my feed to see which files are published by biicode. Actually in my feed the CMakefile.txt, and license files are actually not need to build the project.

https://www.biicode.com/outro56/libcaf_core

libcaf_core/biicode.conf
```diff
@@ -0,0 +1,42 @@
...
+[parent]
+	outro56/libcaf_core: 1
+[paths]
+    # Local directories to look for headers (within block)
+    /
+
+[dependencies]
+    # Manual adjust file implicit dependencies, add (+), remove (-), or overwrite (=)
+    *.hpp + *.cpp 
+    * + LICENSE LICENSE_ALTERNATIVE
```
outro56: If you dont want to have biicode publish out the projects license then can remove this line. Also can remove the license files from the PR

libcaf_core/biicode.conf
```diff
@@ -0,0 +1,42 @@
+# Biicode configuration file
```
outro56: The biicode project folder for this file is ``libcaf_core`` so bii code expects all reference files needed to build this module within that folder.

libcaf_core/biicode.conf
```diff
@@ -0,0 +1,42 @@
...
+
+[requirements]
+    # Blocks and versions this block depends on e.g.
+    # user/depblock1: 3
+    # user2/depblock2(track) @tag
+
+[parent]
+	outro56/libcaf_core: 1
```
outro56: I should probably remove the parent section here since it points to my published feed in biicode. It should be empty or point to your own feed published feed for libcaf

libcaf_io/biicode.conf
```diff
@@ -0,0 +1,44 @@
+# Biicode configuration file
+
+[requirements]
+	 biicode/cmake: 3
+	 outro56/libcaf_core: 1
```
outro56: building libcaf_io requires libcaf_core
currently this is pointed to my published feed,

Neverlord:
Thank you for the detailed response. Logically, CAF consists of multiple libraries, yes. By bypassing the main CMake file, however, you also create problems. For example, your shared libs won't have the correct version.

From a maintainer perspective, this approach requires a lot of copies that individually need to be touched if something changes in the future and---even worse---adds another dependency graph. CMake already knows that `libcaf_io` depends on `libcaf_core` and that `libcaf_riac` on the former two. If at some point we change this in the CMake file (e.g. by introducing a new library) than we have to update multiple places instead of a single one (the main CMake file).

Ideally, Biicode should simply use the existing knowledge of the CMake setup. If we can boil the Biicode support down to a single Biicode config file (plus maybe some extra boilerplate in the main CMake file) than I would be happy to merge the additions to the main repository.

I'm not familiar with Biicode, but can't the config file expose what modules are provided (`libcaf_core`, `libcaf_io`, ...) and figure the dependencies between the components out via CMake? This might be a bit less efficient on the user's end if Biicode does not work well enough with CMake to disable unwanted components. Having a maintainable solution is more important, though.

outro56:
So like I mentioned in my previous comment, I can try to rework the PR so that it would only contain the files necessary for biicode support (namely biicode.conf and bii.cmake). I ran into problems that way earlier but can investigate further and let you know. Does that sound ok?

The only drawback I see with this approach is that the include definition for your library users would not be so nice out of the box (i.e. ``#include "libcaf_core/caf/xxx.hpp``, ``#include "libcaf_io/caf/io/xxx.hpp``). There is a way on their end they could strip out the root directory listing via the ``[includes]`` section in their ``biicode.conf`` file; this is because biicode publishes the same directory structure as original.

> this approach requires a lot of copies that individually need to be touched if something changes in the future and---even worse---adds another dependency graph. CMake already knows that libcaf_io depends on libcaf_core and that libcaf_riac on the former two

If you notice the ``bii.cmake``is pretty empty apart from some plumbing required by biicode to configure compiler options etc. Biicode ***auto-magically*** resolve all dependencies for library. The little magic that makes this happen is in the ``[dependencies]`` section in ``biicode.conf``. Also note this is how biicode knows which files need to get published. What I proposed above should take care of your concern about changing dependencies

NOTE: Although biicode uses CMake under the covers it explicitly does not depend on your hand-rolled CMakefile.txt that's why it short circuits it. I can move the short circuit code after the point that determines the library version.

IMHO biicode is quite simply a really easy platform for sharing your library. From your part once you have the ``biicode.conf`` and ``bii.cmake`` all setup, you only need to execute ``bii build && bii publish`` to publish out a version. See more here http://docs.biicode.com/c++/publishing.html. You can even publish from a git commit http://docs.biicode.com/c++/integration/git.html#publish-from-git-commit

libcaf_io/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+# check for BIICode build
+if (BIICODE)
```
outro56: This line here detects if this make file is being evaluated as a result of running ``bii build`` and then short circuits evaluation to only include biicode specific make targets i.e. allow biicode uses its own auto generated CMakefile.txt

Neverlord:
Thank you for the links, especially to the Git integration. I think setting up the correct include paths on the user's end is acceptable.

> So like I mentioned in my previous comment, I can try to rework the PR so that it would only contain the files necessary for biicode support (namely biicode.conf and bii.cmake). I ran into problems that way earlier but can investigate further and let you know. Does that sound ok?

That would be great. With only two files and the Git integration, I'd be happy to merge the changes to the main repository (and "officially" publish CAF on Biicode).

ignore.bii
```diff
@@ -0,0 +1,9 @@
+build/**
```
outro56: This file excludes certain directories from being published to the biicode repository

outro56:
Hi Dominik, I have updated the PR and simplified the biicode integration. The ignore file was added so biicode does not publish certain directories that would not be needed for library users. I excluded the "cash" executable from being published because it depend on "sash" which is not in the libcaf repository.

outro56:
For a look at how it would look like published see my feed here 
https://www.biicode.com/outro56/actor-framework

Neverlord:
Thanks, this is looking much better now.

I hadn't the time quite yet to read the Biicode documentation and tell Jenkins to do the publishing. I'll merge this if I have all the pieces together.

outro56:
Sweet let me know if you need anything. 
PS: I would like to start volunteering for on the caf project, where/what would be a good task to start on? 
--
Chike 
On Apr 15, 2015 8:42 AM, Dominik Charousset <notifications@github.com> wrote:Thanks, this is looking much better now.

I hadn't the time quite yet to read the Biicode documentation and tell Jenkins to do the publishing. I'll merge this if I have all the pieces together.

Neverlord:
> would be a good task to start on?

Depends on your field of interest: networking parts, debugging (nexus/cash), documentation... :)

Best drop me a mail or join Gitter.

outro56:
Well granted I only have like 3hr/week to dedicate to the project. 
I am on gitter, and email is outro56@yahoo.com 
Networking seems fine to me. I haven't played with the other parts yet but would like to. 
--
Chike 
On Apr 15, 2015 10:40 AM, Dominik Charousset <notifications@github.com> wrote:
would be a good task to start on?


Depends on your field of interest: networking parts, debugging (nexus/cash), documentation... :)

Best drop me a mail or join Gitter.

Neverlord:
Obsoleted by #288

PR #275, opened 2015-04-24
Rename CMake include files camel case

firegurafiku:
Do the following renames to match their file names to the style which CMake uses for packages in its standard distribution:

 * `FindLibcaf.cmake` -> `FindLibCAF.cmake`
 * `FindOPENCL.cmake` -> `FindOpenCL.cmake`

More specifically, there is the file pattern to follow:
```
$ ls -1 /usr/share/cmake/Modules/FindOpen*
/usr/share/cmake/Modules/FindOpenAL.cmake
/usr/share/cmake/Modules/FindOpenGL.cmake
/usr/share/cmake/Modules/FindOpenMP.cmake
...
$ ls -1 /usr/share/cmake/Modules/FindLib*
/usr/share/cmake/Modules/FindLibArchive.cmake
/usr/share/cmake/Modules/FindLibLZMA.cmake
/usr/share/cmake/Modules/FindLibXml2.cmake
...
```


firegurafiku:
I am experiencing some problems building `libcaf_opencl`, could you please test this comprehensively?

Hamdor:
test this please

Hamdor:
Seems jenkins doesn't like me, anyway looks good to me and works with OpenCL. :+1:

Neverlord:
`FindLibcaf.cmake` -> `FindLibCAF.cmake`

The libraries of CAF are all named `libcaf_<module>`, so this change is actually inconsistent with the naming scheme. Either one of "FindCAF" or "FindLibcaf" makes sense, "FindLibCAF" doesn't. I prefer "FindLibcaf" since the script explicitly is looking for the library components. And please don't start a discussion about renaming the library components...

If you change the PR to include only the OpenCL-change I would merge it. Though I really cannot fathom your obsession with CMake.

Neverlord:
@firegurafiku I hope you don't mind if I pull in a few lines from the mail conversation we had recently.

> The original intention was to fix FindLibcaf.cmake (...) I wanted to see it more consistent with the CMake's naming scheme. I have never seen such FindLiblowercase.cmake files before (...)

You've also mentioned that we should make it easy for the CMake guys to add this file to the standard distribution eventually. Yes... but the current file contains what @mavam called "too much magic" and I think I have to agree on that point (at long last). The script tries to find non-installed CAF versions sitting around somewhere near. That's convenient for us, since we never install CAF on our development machines. It's probably not a bright idea on anyone else's machine, though.

As far as the naming is concerned, I think we should go with "FindCAF.cmake", then. CAF is the official name. To the content of the file, I think we should simply replace it with https://github.com/mavam/vast/blob/master/cmake/FindCAF.cmake ... and life with the fact that @mavam did it right from the beginning and I didn't :P

Neverlord:
@firegurafiku I've copied the `FindCAF` from mavam [1], could you update the PR to include only the OpenCL change?

[1] https://github.com/actor-framework/actor-framework/commit/456d50a75001cfae80d6964a8eaf87c79d71f3cb

firegurafiku:
Sorry for forgetting about this PR. Now I've removed everything but a single file rename and a corresponding `CMakeLists.txt` fix.

Neverlord:

PR #278, opened 2015-05-02
Make the operator << of actor_ostream call caf::to_string automatically.

ufownl:
After this change, we can print some caf types using caf::aout without calling caf::to_string.

such as:

from:
```C++
caf::aout(self) << caf::to_string(self->current_message()) << std::endl;
```
to:
```C++
caf::aout(self) << self->current_message() << std::endl;
```

from:
```C++
caf::aout(self) << caf::to_string(self->address()) << std::endl;
```
to:
```C++
caf::aout(self) << self->address() << std::endl;
```

and so on.

### DIFF
--- a/examples/aout.cpp
+++ b/examples/aout.cpp
@@ -15,16 +15,16 @@ using std::endl;
 int main() {
   for (int i = 1; i <= 50; ++i) {
     spawn<blocking_api>([i](blocking_actor* self) {
-      aout(self) << "Hi there! This is actor nr. "
-                 << i << "!" << endl;
+      aout(self) << "Hi there! This is actor nr. " << i << "! "
+                 << "Actor address: " << self->address() << endl;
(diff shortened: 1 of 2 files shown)


Neverlord:
test this please

ufownl:
I had modified the example of aout to print the actor address and current message, it works fine.

Neverlord:
I'm sorry, my "test this please" comment was merely to trigger a test build. I didn't mean for you to add an extra commit.

Anyway, the change makes perfect sense. Thank you for the patch. The only complain I have about it is that we don't accept PRs to `master`. Please base future PRs on `develop`. I cherry-picked the original PR to the develop branch: https://github.com/actor-framework/actor-framework/commit/3ed5bf7c959cec2290c1bb92724810aaacd79a7b. Will get merged back to `master` with the next release.

PR #289, opened 2015-05-09
Add kill exit reason

ufownl:
The kill exit reason kill actors unconditionally, even if they
trap exit messages.


Neverlord:
ok to test

Neverlord:
Awesome. Thanks for the patch. :)

PR #296, opened 2015-05-14
Correct the mistake of group chat example

ufownl:
BTW, I didn't find anything who send `group_down_msg`. Is it deprecated?


Neverlord:
ok to test

Neverlord:
Hm, the `group_down_msg` should be send in case a remote group became unreachable. I'll make a new ticket. ;)

PR #299, opened 2015-05-18
Add boost asio as an optional multiplexer backend

josephnoir:
This is enabled via the `--with-asio` option of the configure script and requires boost to be installed.
Further, a new test ensures the multiplexer can handle an ipv6 connection to the localhost.


Neverlord:
retest this please

PR #303, opened 2015-05-22
Use `thread_local` keyword

Hamdor:
Replace pthread specific calls with `thread_local` keyword.


Hamdor:
I will close this for now due to lag of compiler support.

Issue #300, opened 2015-05-18
Remove obsolete GCC 4.7 hacks


Hamdor:
CAF no longer support GCC 4.7, due to this we can remove the obsolete compiler hacks.

ufownl:
Really good news!

Neverlord:
@Hamdor, can we close this ticket via #310?

Hamdor:
Yes

PR #310, opened 2015-05-28
Remove GCC 4.7 workarounds

Hamdor:
relates #300

### DIFF
--- a/examples/brokers/simple_http_broker.cpp
+++ b/examples/brokers/simple_http_broker.cpp
@@ -13,7 +13,7 @@ using namespace caf::io;
 
 using tick_atom = atom_constant<atom("tick")>;
 
-const char http_ok[] = R"__(HTTP/1.1 200 OK
+constexpr const char http_ok[] = R"__(HTTP/1.1 200 OK
 Content-Type: text/plain
 Connection: keep-alive
 Transfer-Encoding: chunked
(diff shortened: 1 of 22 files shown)


Neverlord:
Ok. My only complain is the weird change in `behavior_impl.hpp`. Revert this detail and I'll merge.

Hamdor:
Done ;)

firegurafiku:
So, there is a new compiler requirement: GCC newer then 4.8?

Neverlord:
Yes. The readme file in `develop` already states this new fact: https://github.com/actor-framework/actor-framework/blob/develop/README.md

Our policy is to support compilers for two years in order to eventually catch up to C++14/17 features and keep the maintenance overhead manageable.

firegurafiku:
Okay. I've asked because I have to know it to specify compiler dependency for package building in RPM spec file.
Also, I feel that most libraries are going to stay with C++11 for very long time, like they did with C++99. I'd expect more or less wide adoption of C++14 up to year 2020.

Neverlord:
Yes, some developers are extremely conservative and slow moving compilers stifled progress in the industry for a long time. C++98/03 took ages until it was reasonably well supported by all major compilers. Luckily, it seems like this slow adaption rate is a thing of the past. As long as GCC, Clang, and MSVC are keeping the pace, I see no reason why the C++ community should hold back. There will be libraries that still use C++98 in 2020, that's a safe bet. We will be using C++17 in CAF until then. :)

PR #314, opened 2015-06-17
Fix bug about typed actor/broker

ufownl:
Correct the parameter error of `send` and `delayed_send`.

I'm so sorry that I made these mistakes.


Neverlord:
@Inet-Jenkins ok to test

Neverlord:
Hm, the fail on Windows seems to be a misconfiguration.

Issue #309, opened 2015-05-28
Typed Brokers


Neverlord:
Brokers are currently always dynamically typed, which is only due to historic reasons (brokers existed before typed actors). The interface to the MM needs to be re-shaped to allow typed brokers as well.

Neverlord:
Closed via #311.

Neverlord:
I gave this some thought and the brokers are currently not "100% type safe". Brokers can receive any message type from the [system_messages header](https://github.com/actor-framework/actor-framework/blob/master/libcaf_io/caf/io/system_messages.hpp). So either we make it mandatory for all typed brokers to handle *all* of these messages or we restrict what a typed broker is allowed to to based on its type.

The former is straightforward to implement, but would mean that an average typed broker would have quite a lot of empty message handlers, which is not idea. The latter requires probably quite a lot of tweaking... For example, `add_doorman()` would need a static assertion that the broker is able to receive `new_connection_msg`. The `spawn_io_*` family would need similar checks. So this might turn out to be quite a lot of work.

In any case, it is probably a good idea to move the typed brokers to an `experimental` namespace in the meantime.

ufownl:
Good suggestion! I'll look into and try to improve it. :)

ufownl:
@Neverlord There is a problem.
If the broker is used as client, it only need to handle `new_data_msg` and `connection_closed_msg`. But if it's used as server, I can't determine whether or not it should handle `new_data_msg` and `connection_closed_msg`.

Neverlord:
Yes, it's definitely tricky. My idea was to restrict access to `add_broker` and `add_doorman` based on the defined interface. An actor without `new_data_msg` would not be allowed to add any scribe. This only works if we have a CRTP base for the typed broker or if we move those function into the broker directly. It's not going to be pretty in any case.

ufownl:
> An actor without `new_data_msg` would not be allowed to add any scribe.

When the broker is used as server, adding scribes is inevitable. But if we fork all of the connections, it doesn't need handle `new_data_msg` and `connection_closed_msg`.

Neverlord:
That's right. So we might end up just enforcing all message handlers on all typed actors.

Neverlord:
@ufownl do you still see any need for action here or can we close this via #317?

ufownl:
I think we can close this. :)

Neverlord:
:+1:

PR #317, opened 2015-06-18
Achieve complete type safe of `typed_broker`

ufownl:
Relates #309 .

We can just enforcing all message handlers on the typed actors that they are used as server.

### DIFF
--- a/libcaf_io/caf/io/experimental/typed_broker.hpp
+++ b/libcaf_io/caf/io/experimental/typed_broker.hpp
@@ -17,18 +17,19 @@
  * http://www.boost.org/LICENSE_1_0.txt.                                      *
  ******************************************************************************/
 
-#ifndef CAF_IO_TYPED_BROKER_HPP
-#define CAF_IO_TYPED_BROKER_HPP
+#ifndef CAF_IO_EXPERIMENTAL_TYPED_BROKER_HPP
+#define CAF_IO_EXPERIMENTAL_TYPED_BROKER_HPP
 
 #include <map>
 #include <vector>
+#include <type_traits>
 
 #include "caf/none.hpp"
 #include "caf/config.hpp"
 #include "caf/make_counted.hpp"
 #include "caf/spawn.hpp"
 #include "caf/extend.hpp"
-#include "caf/typed_behavior.hpp"
+#include "caf/typed_actor.hpp"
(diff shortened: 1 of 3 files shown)


ufownl:
I have tried `std::is_convertible`, it didn't work.

Neverlord:
It does, my example code was a bit incorrect. It used to be `handle::extend<...>::type`, which is a bit verbose for my taste anyway. I changed it to accept `handle::extend<...>` instead [1] and also verified that `std::is_convertible` works as expected in the `typed_spawn` unit tests.

So the preferred syntax should work for you after a rebase.

[1] https://github.com/actor-framework/actor-framework/commit/0dc600510a1d4f85cb884035f6f151e08b55b5aa

ufownl:
```C++
// check invariants of type system
using dummy1 = typed_actor<reacts_to<int, int>,
                           replies_to<double>::with<double>>;

//using dummy2 = dummy1::extend<reacts_to<ok_atom>>;
using dummy2 = typed_actor<reacts_to<ok_atom>>;

static_assert(std::is_convertible<dummy2, dummy1>::value,
              "handle not assignable to narrower definition");
```
It has no compilation error on my computer, does it work as expected?

Neverlord:
> It has no compilation error on my computer, does it work as expected?

Crap. The handles are always assignable and only check via `static_assert` if the assignment is correct. I'm working on a fix.

Neverlord:
I'm really sorry for not double-checking this prior to pushing the code to GitHub. I've added a counter-test to the unit test and use SFINAE to make assigning invalid handles impossible: https://github.com/actor-framework/actor-framework/commit/b6a18114fe374460a17f065f04dc09dbdf2f0037

Thank you for your patience (and double-checking the nonsense I wrote). :)

ufownl:
Now, it works fine! And the newest commit has been pushed. :)

Neverlord:
@Inet-Jenkins ok to test

ufownl:
Jenkins didn' work, what's wrong with it?

Neverlord:
Sometimes Jenkins picks up the wrong hash for some reason.

@Inet-Jenkins retest this please

ufownl:
Seems Jenkins hates this PR.

Neverlord:
Looks like it..

PR #318, opened 2015-07-01
thread_local implementation of memory_cache and PC build fix

cmaughan:
A thread_local implementation of memory_cache, doesn't require pthreads. Should work on everything except the Clang compiler.  It's tested on PC, but not GCC.
Change the makelists to supply CMAKE_CXX_FLAGS if they are not already set (which on windows they typically are by the toolchain)


Neverlord:
Since the thread-local data is never shared, `unique_ptr` is indeed a better fit. Also, please reformat the code according to our [code convention](https://github.com/actor-framework/actor-framework/blob/develop/CONTRIBUTING.md) and squash the commits.

cmaughan:
Will do - though I'm out for 2 weeks.

Neverlord:
Obsoleted by 36077a3c2c2bc39aaa3357af8e3b6af649a0892d

cmaughan:
Thanks for integrating it; sorry I didn't get around to updating the pull request!

Neverlord:
No problem! @Hamdor fixed several issues on MSVC anyway, so it wasn't much of a problem to fix this along the way.

PR #323, opened 2015-07-07
Fix `sender_signature_checker`

ufownl:
Make it support `replies_to<…>::with_either<…>::or_else<…>`.

### DIFF
--- a/libcaf_core/caf/detail/typed_actor_util.hpp
+++ b/libcaf_core/caf/detail/typed_actor_util.hpp
@@ -235,16 +235,25 @@ struct sender_signature_checker {
     using dest_output_types =
       typename deduce_output_type<
         DestSigs, ArgTypes
-      >::type::first;
+      >::type;
     sender_signature_checker<
       DestSigs, OrigSigs,
-      dest_output_types
+      typename dest_output_types::first
+    >::check();
+    sender_signature_checker<
+      DestSigs, OrigSigs,
+      typename dest_output_types::second
     >::check();
   }
 };
 
 template <class OrigSigs, class DestSigs>
 struct sender_signature_checker<OrigSigs, DestSigs, detail::type_list<void>> {
+  static void check() {}
+};
+
+template <class OrigSigs, class DestSigs>
+struct sender_signature_checker<OrigSigs, DestSigs, detail::type_list<>> {
   static void check() {}
 };
 


Neverlord:
Can you add a test for this (that failed previously) to one of the unit tests? Having a fix is nice, but we also need to make sure we don't brake it again by accident in the future. :)

ufownl:
Hmm...This patch will prevent from compiling the code as follow:
```C++
using trigger_atom = atom_constant<atom("trigger")>;
using foo_atom = atom_constant<atom("foo")>;
using bar_atom = atom_constant<atom("bar")>;

using foo_type = typed_actor<replies_to<trigger_atom>::with_either<foo_atom>::or_else<bar_atom>>;
using foo_result_type = either<foo_atom>::or_else<bar_atom>;

using bar_type = typed_actor<reacts_to<foo_atom>>;

foo_type::behavior_type foo_action(foo_type::pointer self) {
	return {
		[] (trigger_atom) -> foo_result_type {
			return {bar_atom::value};
		}
	};
}

bar_type::behavior_type bar_action(bar_type::pointer self) {
	auto foo = self->spawn_typed<linked>(foo_action);
	self->send(foo, trigger_atom::value);
	return {
		[=] (foo_atom) {
			self->quit(exit_reason::user_shutdown);
		}
	};
}
```
And I have no idea how to test it.

Neverlord:
That's pretty much a unit test already. ;)

Testing whether something compiles & terminates sometimes is just enough. I'll merge and use the provided snipped as test.

Neverlord:
Erm, while implementing this unit test: the snipped you provided should not compile. Your `bar_action` should not be able to send `foo` a `trigger`, because it does not handle `bar_atom`. Was this a mistake on your side or do I miss something? I've implemented the unit test by adding the second handler to `bar` (see https://github.com/actor-framework/actor-framework/commit/7ede7f2af53a0a1b0f4d148f8197e4d92a2e92b5).

ufownl:
Maybe I didn't describe clearly. Actually I mean that the code I pasted above shouldn't be compiled. Because `foo_type` may return `foo_atom` or `bar_atom`, so `bar_type` must handle both `foo_atom` and `bar_atom`:
```C++
using trigger_atom = atom_constant<atom("trigger")>;
using foo_atom = atom_constant<atom("foo")>;
using bar_atom = atom_constant<atom("bar")>;

using foo_type = typed_actor<replies_to<trigger_atom>::with_either<foo_atom>::or_else<bar_atom>>;
using foo_result_type = either<foo_atom>::or_else<bar_atom>;

using bar_type = typed_actor<reacts_to<foo_atom>, reacts_to<bar_atom>>;

foo_type::behavior_type foo_action(foo_type::pointer self) {
    return {
        [] (trigger_atom) -> foo_result_type {
            return {bar_atom::value};
        }
    };
}

bar_type::behavior_type bar_action(bar_type::pointer self) {
    auto foo = self->spawn_typed<linked>(foo_action);
    self->send(foo, trigger_atom::value);
    return {
        [=] (foo_atom) {
            self->quit(exit_reason::user_shutdown);
        },
        [=] (bar_atom) {
            self->quit(exit_reason::user_shutdown);
        }
    };
}
```

Neverlord:
Ah, so the code did compile previously but shouldn't. Got it. We currently have no setup to check if something does *not* compile. Maybe we could test the template tools directly and check via `static_assert`.

ufownl:
But `sender_signature_checker` exactly do some checks via `static_assert`, so that's really hard to add unit test for it directly.

Neverlord:
Yes, we would need to test several `detail` tools in the `metaprogramming` unit test individually. But I agree that it's out of scope for this PR.

Issue #319, opened 2015-07-02
Forwarding to and from typed actors


Neverlord:
The current facility for forwarding messages (`forward_to`) does only cover dynamically typed actors. This should be extended to allow forwarding to and from typed actors, e.g. with an API like this:

```cpp
// inside a typed actor
[](int a, int b, int c) -> typed_response_promise<int> {
  return delegate(buddy, a + b, b + c);
}
```

Note that the forwarded message is not necessarily the received message. This enables a much more flexible way of forwarding. The return type of `delegate` is a typed response promise to verify the guarantee type safety in statically typed actors.

mavam:
How does this forward the current message? It seems you're taking the data from an existing message to construct a new one. I assume `delegate` passes through the message ID so that responses make it back to the original sender. Perhaps a call to `delegate(buddy)` would simply forward `current_message()`?

Neverlord:
> I assume `delegate` passes through the message ID so that responses make it back to the original sender.

Yes.

> Perhaps a call to `delegate(buddy)` would simply forward `current_message()`?

This is not possible in a type-safe way. Otherwise we could just tweak `forward_to` to behave this way. The problem is that the calling side of `forward_to` does not have the necessary type information and only sees the type-erased `current_message()`. `delegate` would add an alternative that works for type-safe actors as well and also adds more flexibility, since it allows actors to "enrich" a message before forwarding it. With `move`-aware data, this can also be quite cheap.

PR #324, opened 2015-07-07
Forwarding to and from typed actors

ufownl:
Relates #319.

### DIFF
--- a/libcaf_core/caf/detail/ctm.hpp
+++ b/libcaf_core/caf/detail/ctm.hpp
@@ -24,6 +24,7 @@
 #include "caf/typed_response_promise.hpp"
 
 #include "caf/detail/type_list.hpp"
+#include "caf/detail/delegate_helper.hpp"
 #include "caf/detail/typed_actor_util.hpp"
 
 namespace caf {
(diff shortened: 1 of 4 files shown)


Neverlord:
I think this PR is a start, but it would make sense to grant dynamically-typed actors access to the new `delegate` functionality as well.

ufownl:
I have moved `delegate` into `local_actor` and added a test for it. Now the dynamically-typed actors can access to the new `delegate` functionality.

Neverlord:
@Inet-Jenkins ok to test

Neverlord:
It still misses an overload for `delegate(const actor& dest, ...)`, though.

ufownl:
The new solution you offered has been implemented, and the overload for `delegate(const actor& dest, ...)` also has been added.

Neverlord:
Amazing work, as always. I cleaned up the code a bit and renamed the `detail::delegate_helper` to `delegated`. This is because users need to be able to type this name. Also, delegating to a dynamically typed actor returns `void` (there's really nothing that could be inferred here).

PR #333, opened 2015-08-17
Fix `has_name`

ufownl:
Make `has_name` support scalar types.

The code like below will be compiled and output 0:
```C++
std::cout << has_name<int>::value << std::endl;
std::cout << has_name<double>::value << std::endl;
std::cout << has_name<void*>::value << std::endl;
```


Neverlord:
That's the problem of having "primitive data types" in the language...

PR #341, opened 2015-08-26
Typed stateful broker

ufownl:
Add necessary template aliases and template specialization, so that we can
implement typed_stateful_broker using stateful_actor.
```C++
using foo = typed_actor<...>;

class foo_state {
  ...
};

foo::behavior_type foo_impl(foo::stateful_broker_pointer<foo_state> self, ...) {
  ...
}
```


Neverlord:
Pretty much a no-brainer to accept. Thanks for contributing to CAF, you rock!

PR #351, opened 2015-09-17
Support array members in `announce`

ufownl:
Relates #350

Issue #350, opened 2015-09-16
Support array members in `announce`


Neverlord:
From the mailing list:

> I'm attempting to send messages including Boost uuids over a network using CAF:
    caf::announce<uuid>("uuid", &ID::data);
However, ID::data is a uint8_t[], and I get the error:
libcaf_core/caf/uniform_type_info.hpp:62:39: error: array used as initializer
         value_(std::forward<Ts>(xs)...) {

tbenthompson:
Awesome. Thanks Dominik. I'm thinking I will actually work on this. I'll submit a pull request once I have something working. I'll post again if I get stuck or decide to abandon the problem.

tbenthompson:
Just in case it wasn't clear, I am the poster from the mailing list...

tbenthompson:
Further discussion in #351.

Neverlord:
Fixed via https://github.com/actor-framework/actor-framework/commit/1cf9ca9bb644805b22064a094610ff43211d9ba9

PR #352, opened 2015-09-17
Announcement mechanisms for C-style arrays.

tbenthompson:
See #350 and #351 for the issues addressed. Since this is my first commit to the project, let me know if there are things that are not as they should be. As far as I can figure out, C-style arrays require a non-default uniform_type_info. Trying to announce an array type with default_uniform_type_info appears to create an infinite loop...

### DIFF
--- a/libcaf_core/caf/announce.hpp
+++ b/libcaf_core/caf/announce.hpp
@@ -32,6 +32,7 @@
 #include "caf/detail/safe_equal.hpp"
 #include "caf/detail/type_traits.hpp"
 #include "caf/detail/default_uniform_type_info.hpp"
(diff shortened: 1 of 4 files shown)


ufownl:
it has no necessary to define `array_uniform_type_info` and specified `announce` separate, see the latest commit in #351

tbenthompson:
Beautiful. I figured there was a better way of doing this. Thanks for figuring it out. It was fun to learn about the internals of caf nonetheless.

ufownl:
> It was fun to learn about the internals of caf nonetheless.

Really agree with you. :)

Neverlord:
So... I suppose this is obsoleted by #351 then. :)

Issue #340, opened 2015-08-20
Add INI parser for Configuring CAF Applications


Neverlord:
Currently, configuring a scheduler, setting up Nexus/Probes for RIAC or enabling automatic connections in the middleman requires API access. This should be made possible by simply providing a config file, e.g. `actor-framework.config`.

For example:

```ini
[scheduler]
policy=work-sharing
max-threads=2

[middleman]
automatic-connections=true

[nexus]
host="127.0.0.1"
port=4242
```

Why INI-Syntax? We need a format that is very simple to parse *without* external dependencies or regexes (as long as we support GCC 4.8, unfortunately). A parser for INI is, even without regexes, very simple and easy to maintain. It is also a widely known format (though it isn't exactly pretty).

An open question is whether CAF should always try to read that config or if we require users to explicitly call a function for running this step, e.g., `void initialize(const char* config_file = nullptr)` or `void startup(const char* config_file = nullptr)`. The default argument would mean: try finding a file named `actor-framework.config` in path.

Further down the lines, this would also allow setting up default nodes to connect to, enable/disable migration or load balancing, etc.

The INI parser should auto-detected integers, booleans, and strings.

Neverlord:
This parser is going to be used in the new `actor_system`-based approach (see #370).

PR #354, opened 2015-09-22
Add INI parser

1blankz7:
This topic relates to issue #340 and introduces a INI parser with no external dependencies.

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -61,6 +61,8 @@ set (LIBCAF_CORE_SRCS
      src/message_data.cpp
      src/message_handler.cpp
      src/node_id.cpp
+     src/parse_config.cpp
+     src/parse_ini.cpp
      src/ref_counted.cpp
      src/response_promise.cpp
      src/replies_to.cpp
--- a/libcaf_core/caf/detail/parse_ini.hpp
+++ b/libcaf_core/caf/detail/parse_ini.hpp
@@ -1,0 +1,42 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright (C) 2011 - 2015                                                  *
+ * Dominik Charousset <dominik.charousset (at) haw-hamburg.de>                *
+ *                                                                            *
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#ifndef CAF_DETAIL_PARSE_INI_HPP
+#define CAF_DETAIL_PARSE_INI_HPP
+
+#include <istream>
+
+#include "caf/parse_config.hpp"
+
+namespace caf {
+
+namespace detail {
+
+/// Parse the given input stream as INI formatted data and calls the consumer
+/// with every key-value pair.
+/// @param raw_data the INI formatted input stream
+/// @param errors a stream of all errors which occure while parsing
+/// @param consumer a function that consums the key-value pairs
+void parse_ini(std::istream& raw_data, std::ostream& errors,
+               config_consumer consumer);
+
+} // namespace detail
+} // namespace caf
+
+#endif // CAF_DETAIL_PARSE_INI_HPP
--- a/libcaf_core/caf/parse_config.hpp
+++ b/libcaf_core/caf/parse_config.hpp
@@ -1,0 +1,47 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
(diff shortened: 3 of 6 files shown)


Neverlord:
So... sorry for the long silence. While there's nothing "wrong" with the INI class, it is quite heavyweight and inflexible. Such an interface makes it hard to add support for something else, like JSON for example, in the future.

You also store *a lot* of strings unnecessarily. The map is good idea when having it persist for the duration of the program, but that's not the case in CAF. Here is a snipped how CAF would want to use the parser:

```cpp
// converts a `variant<string, double, int64_t, bool>` to a `message`
struct to_message_visitor : static_visitor<message> {
  message operator()(string& value) {
    return make_message(std::move(value));
  }
  message operator()(int64_t value) {
    return make_message(value);
  }
  message operator()(bool value) {
    return make_message(value);
  }
};

enum class config_format {
  auto_detect,
  ini,
  json
};

using config_value = variant<string, double, int64_t, bool>;

using config_consumer = std::function<void (string, config_value)>;

void parse_config(const string& file_name, 
                  config_format cf = config_format::auto_detect) {
  if (cf == config_format::auto_detect) {
    if (file_name.size() < 5)
      return;
    if (file_name.compare(file_name.size() - 4, 4, ".ini"))
      parse_config(file_name, config_format::ini);
    // else if ... json ...
    else
      std::cerr << "unknown config file format" << std::endl;
    return;
  }
  struct consumer {
    consumer() {
      config_server_ = whereis(atom("ConfigServ"));
    }

    consumer(const consumer&) = default;
    consumer& operator=(const consumer&) = default;

    void operator()(string key, config_value value) const {
      to_message_visitor v;
      anon_send(config_server_, put_atom::value, std::move(key),
                apply_visitor(value, v));
    }

    actor config_server_;
  };
  switch (cf) {
    case config_format::ini:
      detail::parse_ini(file_name, consumer);
      break;
    case config_format::json:
      detail::parse_json(file_name, consumer);
      break;
    // ...
  }
}
```

This code (properly split) would belong to `libcaf_core/caf/parse_config.hpp` and `libcaf_core/src/parse_config.cpp`. 

Ideally, your parser consists of one function: `parse_ini(config_consumer f)`. This removes a lot of verbose boilerplate code on your end. Since the parser is not meant for users, it would belong to `libcaf_core/caf/detail/parse_ini.hpp`.

You also don't need additions such as the trim functions to the `string_algorithms.hpp` header. It is much cheaper to not add unwanted characters to the string in the first place. Here's a trimmed down version of your current `parse` member function that also doesn't store all lines but handles them immediately:

```cpp
void ini_parser::parse() {
  using std::find_if;
  using std::isspace;
  using iterator = string::iterator;
  auto find_non_ws = [](iterator first, iterator end) -> iterator {
    return find_if_not(first, end, isspace);
  };
  string group;
  string line;
  size_t ln = 0; // line number
  while (getline(raw_data_, line)) {
    ++ln;
    // get begin-of-line (bol) and end-of-line (eol), ignoring whitespaces
    auto eol = find_if_not(line.rbegin(), line.rend(), isspace).base();
    auto first = find_if_not(line.begin(), eol, isspace);
    // ignore empty lines and lines starting with ';' (comment)
    if (bol == eol || *bol == ';')
      continue;
    // do we read a group name?
    if (*bol == '[') {
      if (*(eol - 1) != ']')
        cerr << "error in line " << ln << ": missing ] at end of line" << endl;
      else
        group.assign(bol + 1, eol - 1);
      // skip further processing of this line
      continue;
    }
    // do we have a group name yet? (prohibit ungrouped values)
    if (group.empty()) {
      cerr << "error in line " << ln << ": value outside of a group" << endl;
      continue;
    }
    // position of the equal sign
    auto eqs = find(bol, eol, '=');
    if (eqs == eol) {
      cerr << "error in line " << ln << ": no '=' found" << endl;
      continue;
    }
    if (bol == eqs) {
      cerr << "error in line " << ln << ": line starting with '='" << endl;
      continue;
    }
    if ((eqs + 1) == eol) {
      cerr << "error in line " << ln << ": line ends with '='" << endl;
      continue;
    }
    // our keys have the format "<group>.<config-name>"
    auto key = group;
    key += '.';
    // ignore any whitespace between config-name and equal sign
    key.insert(key.end(), bol, find_if(bol, eqs, isspace));
    // begin-of-value, ignoreing whitespaces after '='
    auto bov = find_if_not(eqs + 1, eol, isspace);
    // auto-detect what we are dealing with
    const char* true_str = "true";
    const char* false_str = "false";
    if (std::equal(bov, eol, true_str)) {
      f(std::move(key), true);
    } else if (std::equal(bov, eol, false_str)) {
      f(std::move(key), false);
    } else if (*bov == '"') {
      // found a string, start escaping string sequence
    } else // ... check whether we have a double or integer value ...
  }
}
```

It is of course OK to store everything into a map for your unit test.

```cpp
namespace {

class fixture {
  void parse(const char* cstr) {
    std::stringstream ss;
    ss << cstr;
    auto f = [&](std::string key, config_value value) {
      values.emplace(std::move(key), std::move(value));
    };
  }
  std::map<std::string, config_value> values;
};

template <class T>
bool value_is(const config_value& cv, const T& what) {
  const T* ptr = get<T>(&cv);
  return ptr != nullptr && *ptr == what;
}

constexpr const char* case1 = R"__(
[scheduler]
policy="work-sharing"
max-threads=2

[middleman]
automatic-connections=true

[nexus]
host="127.0.0.1"
port=4242

[cash]
greeting="Hi there, this is \"CASH\"!\n~~~ use at your own risk ~~~\n"
)__";

} // namespace <anonymous>

CAF_TEST_FIXTURE_SCOPE(parse_ini_tests, fixture)

CAF_TEST(test_case1) {
  parse(case1);
  CAF_CHECK(values.count("nexus.port") > 0);
  CAF_CHECK(value_is(values["nexus.port"], int64_t{4242}));
  // ...
}

```

1blankz7:
Would it be better if I squash the three commits together?

Neverlord:
Yes. This makes reviewing much easier and also keeps the history clean.

1blankz7:
Okay, it's done.

Neverlord:

PR #363, opened 2015-10-13
Fix timeout handler

ufownl:
Issue which commented yesterday on gitter.


Neverlord:
Wow, that was quick. :)

PR #364, opened 2015-10-16
Simplify obs-commit-version.sh

firegurafiku:
I've noticed that current Jenkins setup issues builds onto wrong OBS repository, thus breaking stable builds. In order to completely separate release packages and nightly packages, I've modified `obs-commit-version.sh` to take no parameters and be as much self-contained as possible. New behaviour is the following:

* if branch is 'master', then trigger release build to 'devel:libraries:caf/caf' with version like '15.0.1''
* if branch is 'develop', then trigger nightly build to 'devel:libraries:caf:nightly/caf' with long version '15.0.1_20170325`
* on other branches do nothing.

So, Jenkins should just issue the same command sequence regardless the build type:
```text
$ make
$ make doc
$ scripts/obs-commit-version.sh
```

### DIFF
--- a/scripts/obs-commit-version.sh
+++ b/scripts/obs-commit-version.sh
@@ -7,17 +7,15 @@
 #
 #     $ make
 #     $ make test
-#     $ make doc                          <-- this is a necessary step!
-#     $ obs-commit-version.sh --release   <-- for release build
-#     -or-
-#     $ obs-commit-version.sh --nightly   <-- for nightly build
+#     $ make doc                       <-- this is a necessary step!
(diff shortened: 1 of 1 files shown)


Neverlord:

Aside from this issue, all lights are red currently. Some packages (Debian 7) should not try to build in the first place because of an unsupported compiler (GCC 4.7).

firegurafiku:
> Some packages (Debian 7) should not try to build in the first place because of an unsupported compiler (GCC 4.7).

If so, I think it wouldn't be unfair to just forget about Debian 7. I'll disable its build configuration for now.

PR #366, opened 2015-10-22
Rename `sync_send` to `request`

### LINKED ISSUE #365: Rename sync_send
The "sync" in `sync_send` really seems to confuse people. Regular feedback is "why is this thing called sync_send if it doesn't use synchronization"? This comes up in every review phase, so it seems that there are strong associations to the word "sync". After discussing this with @josephnoir and @mavam, the best replacement we have found is `request`.

This should give developers the immediate association to request/response protocols, which this function actually models. This also plays nice with the already existing `response_promise` and the proposed `respond` for returning a "ready promise". It also has the nice property that a request includes sending a message but is more than that (in contrast to use "distinct" words that could give the idea that the operations do completely separated things).

Hamdor:
This deprecates `sync_send`, `on_sync_failure` and
`has_sync_failure_handler`, close #365


Neverlord:
I'm closing this for now, since there's currently too much refactoring going on with #367, #369, and #370.

PR #377, opened 2015-12-10
Fix the return type of `add_tcp_doorman`

ufownl:
The function with the same name and parameters in `abstract_broker` return a
`std::pair<accept_handle, uint16_t>` object.


Neverlord:
Thanks. Ultimately, I/O functions that currently throw exceptions could return a `maybe<T>` instead. Errors are likely for I/O operations, so I think exceptions are not a good approach here. I haven't figured out best practices for `maybe<T>` yet, though.

ufownl:
Actually, I just catched the exceptions and returned a maybe<T> as you said.

PR #385, opened 2015-12-15
Fix possible segfault when forwarding messages

ufownl:
It occurs when using GCC.


Neverlord:

PR #387, opened 2015-12-16
Fix `actor_ostream`

ufownl:
To avoid the extra quotation marks when printing `std::string`.


Neverlord:

Issue #378, opened 2015-12-10
Requests multiplexing in event-based actors


Neverlord:
A common design pattern in CAF is to have an actor supervise a set of actors by delegating incoming jobs. These supervisors cannot make use of `request().then()`, because this would "block" the supervisor until a worker has finished its job by suspending its regular behavior.

A solution to this problem is to give requests a second, non-blocking alternative to `.then` that does not suspend its regular behavior.  For example `self->request(...).multiplex(...)` or `self->request(...).async(...)`.

Since 0.15 already makes quite a lot of breaking changes, we also could rename `.then` to have `.sync()` and `.async()`. Alternatively, we could use `.then` for the multiplexing and `.await` for the current suspend-until-result-arrives semantics (we should then rename `.await` for blocking actor to something else, e.g. simply `request(...).receive(...)`).

I think `.then` and `.await` make the most sense, but could be confusing for users when upgrading to 0.15. Opinions?

mavam:
In general, I quite like this proposal. In places where I use the forward-to-spawn idiom and do not really need the concurrency, a non-blocking `.then` would come in handy and make the code much easier to understand.

> A solution to this problem is to give requests a second, non-blocking alternative to .then that does not suspend its regular behavior. For example `self->request(...).multiplex(...)` or `self->request(...).async(...)`.

Could you elaborate on the messaging semantics? IIRC, `.then` created a one-shot handler for a specific message ID. What happens when multiple messages are inflight with `.multiplex`? Would the runtime maintain a mapping from message IDs to callbacks?
 
> I think `.then` and `.await` make the most sense, but could be confusing for users when upgrading to 0.15. 

Naming-wise, this makes most sense to me. I would not re-introduce the word `sync`, which we just got rid of due to confusion. Using `receive` instead of `await` as blocking one-shot continuations would also be consistent with the top-level `self->receive`, i.e., users know that when a `receive` is involved, there's true blocking.

Neverlord:
> Would the runtime maintain a mapping from message IDs to callbacks?

Yes. An actor would have three "sets" of behaviors: (1) its regular behavior stack, (2) the stack of awaited responses, and with this proposal (3) a map `{message id => callback}` for "concurrent" response handlers.

> In places where I use the forward-to-spawn idiom (...) make the code much easier to understand.

Absolutely. This idiom needs to die. :)

> I would not re-introduce the word `sync`, which we just got rid of due to confusion.

[...]

PR #393, opened 2015-12-26
Fix and improve sync response handling of `local_actor`

Lingxi-Li:
In the previous version, handling of sync responses by `local_actor` is done in reverse order of their corresponding `sync_send()` calls. For example,

~~~~~C++
void exp_behavior(caf::event_based_actor* self,
                  const caf::actor& responder_1,
                  const caf::actor& responder_2) {
  ...
  self->sync_send(responder_1, ...).then(handler_1);
  self->sync_send(responder_2, ...).then(handler_2);
}
~~~~~

Regardless of the order response messages are enqueued to the mailbox, `handler_2` is going to be invoked before `handler_1`. I don't think this is correct `sync_send()` semantics. This PR fixes this issue, and makes sure that handling of sync responses follows the order of their corresponding `sync_send()` calls.

It turns out that this issue only exists among sync responses. The behavior between sync and async responses is correct. For example,

~~~~~C++
caf::behavior exp_behavior(caf::event_based_actor* self,
                           const caf::actor& responder_1,
                           const caf::actor& responder_2) {
  ...
  self->sync_send(responder_1, ...).then(handler_1);
  self->send(responder_2, ...);
  return handler_2;
}
~~~~~

It's guaranteed that `handler_1` be invoked before `handler_2`, even when the async response arrived before the sync one.

Besides the fix, an improvement is also made to the removal of invoked sync response handlers. Now the removal takes constant time instead of linear time to the number of pending sync response handlers.

### DIFF
--- a/libcaf_core/caf/local_actor.hpp
+++ b/libcaf_core/caf/local_actor.hpp
@@ -24,7 +24,7 @@
 #include <cstdint>
 #include <exception>
 #include <functional>
-#include <forward_list>
+#include <list>
 
 #include "caf/fwd.hpp"
 
@@ -646,7 +646,8 @@ protected:
   message_id last_request_id_;
 
   // identifies all IDs of sync messages waiting for a response
-  std::forward_list<pending_response> pending_responses_;
+  // arranged in chronological order, and used as a FIFO queue
+  std::list<pending_response> pending_responses_;
 
   // points to dummy_node_ if no callback is currently invoked,
   // points to the node under processing otherwise
--- a/libcaf_core/caf/local_actor.hpp
+++ b/libcaf_core/caf/local_actor.hpp
@@ -599,7 +599,10 @@ public:
 
   message_id new_request_id(message_priority mp);
 
-  void mark_arrived(message_id response_id);
+  // sync response handler invokation follows a chronological FIFO order
+  // the handler to remove should be at the front the of list
+  // call `pending_responses_.pop_front()` directly
+  // void mark_arrived(message_id response_id);
 
   bool awaits_response() const;
 
--- a/libcaf_core/src/local_actor.cpp
+++ b/libcaf_core/src/local_actor.cpp
@@ -348,6 +348,7 @@ maybe<message> post_process_invoke_res(local_actor* self,
                                           Handle hdl = Handle{}) {
   CAF_LOGF_TRACE(CAF_MARG(mid, integer_value) << ", " << CAF_TSARG(res));
   if (! res) {
+    // error occurred
     return none;
   }
   if (res->empty()) {
@@ -366,6 +367,8 @@ maybe<message> post_process_invoke_res(local_actor* self,
     return res;
   }
   if (handle_message_id_res(self, *res, none)) {
+    // the responsibility to respond is handed over to a pending `sync_send()`
+    // response handler
     return message{};
   }
   // respond by using the result of 'fun'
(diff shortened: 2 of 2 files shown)


Lingxi-Li:
Added a follow-up to what is done.

Since the order of pending response handlers have been reversed, the search direction of `find_pending_response()` should be reversed as well for better performance. This is due to common usage pattern and the principle of locality. In general, the match, if any, is more likely to be recently added, and thus reside to the rear of the list of pending response handlers. I will give one example to illustrate the idea. `response_handle::then()` relies on `find_pending_response()` to find the corresponding response handler to set. With the `sync_send().then()` usage pattern, the match would be exactly at the end of the list. Even if `then()` is not called immediately after `sync_send()`, it should not be far from that, and the match should be close to the rear of the list instead of front. Besides, `local_actor::handle_message_id_res()` relies on `find_pending_response()` to find the corresponding response handler to delegate the responsibility of replying to. The situation should be similar.

`local_actor::mark_arrived()` is also removed, since it will no longer be of any use. The handler to remove will now always be at the front of the list. So, just call `pending_responses_.pop_front()` directly. I guess one of the reasons it was needed before is that the target handler might actually be at the second position of the list. Within `invoke_message()`, we see that the current handler is invoked before it is removed from the list. So, in the previous version, if the handler invocation adds a new handler to the list, it is added before the current one, leaving the handler to remove be at the second position. In the current version, any new handler would be added to the rear. Consequently, the current handler will always be at the front of the list, and a simple call to `pending_responses_.pop_front()` should be sufficient under all circumstances.

Neverlord:
> Regardless of the order response messages are enqueued to the mailbox, `handler_2` is going to be invoked before `handler_1`. I don't think this is correct `sync_send()` semantics.

If you are calling `become()` several times with `keep_behavior` (and `unbecome()`), only the last behavior is active, i.e., behaviors are organized in a stack. This allows users to nest behaviors. Requests (sync messages) model a second behavior stack that has priority over the behavior stack for asynchronous messages. This is the reasoning behind the design decision to have the last `sync_send` active: consistency. It also allows you to nest sync messages in the same way and prioritize (a sequence of) messages in this way.

The new request multiplexing for 0.15 (#378) will improve performance for use cases where you don't mind ordering and don't need responses to have priority over asynchronous messages.

So before discussing the PR itself, why do you think a change of semantics from LIFO to FIFO is necessary? With #378 in the picture, I could get behind a decision to give all "awaited" responses the same priority, i.e., do impose neither FIFO nor LIFO order. This would make `request(...).then(...)` and `request(...).await(...)` orthogonal in the sense that the latter adds only priority over asynchronous messages, but does not add ordering constraints.

Lingxi-Li:
I guess I have misunderstood the motivation/intention behind `sync_send()`. I'm expecting that it does not just send the message but also has additional ordering semantics, similar to the memory order thing. In the standard atomic library, you specify atomic operations along with ordering options. I intend to use `sync_send()` as a synchronization point with somewhat `memory_order_seq_cst` semantics. Something like a memory barrier. Given two `sync_send()` calls in a row, I'm expecting behavior as if the latter does not happen at all before the former's response handler is invoked. Do you think this makes sense?

Neverlord:
> I'm expecting that it does not just send the message but also has additional ordering semantics, similar to the memory order thing.

This assumption is not correct. Ordering semantics only matter if more than one thread concurrently operate on the same data. The whole point of decoupling actors via message passing is to avoid such data races by design.

>  I'm expecting behavior as if the latter does not happen at all before the former's response handler is invoked.

Event-based actors never block. `sync_send` simply sends the message with a unique request ID that allows the runtime to unambiguously identify the response message and to invoke the given one-shot handler (instead of the normal behavior).

You are not the first one to be mislead by the name `sync_send`, though. The only synchronization that takes place is on a logical messaging layer, but that has nothing to do with blocking, memory ordering, etc. That's why we rename the function to `request` with 0.15 (it is already renamed in the `actor_system` topic branch, btw).

Lingxi-Li:
After renaming `sync_send()` to `request()`, the behavior makes sense now.

PR #394, opened 2015-12-27
Fix exception handling in `local_actor::resume()`

Lingxi-Li:
The previous version does not correctly handle the case when the call to `self->act()` throws an exception that is not successfully handled by the call to `self->handle(eptr)`.


Neverlord:

Issue #370, opened 2015-10-29
Replace singleton-based environment with actor_system class


Neverlord:
Discussion and original proposal can be found in the mailing list: https://groups.google.com/forum/#!topic/actor-framework/XjkOyHTa588

Neverlord:
Some issues that came up while implementing this:

- `to_string(message)` and `from_string()` need access to unified type information (UTI) map
  + messages do not have (or need) a reference to an actor system
  + no access to UTI -> no access to (de)serialization code
  + possible solution: make `to_string` and `from_string` member functions of `actor_system`
  + impact: actors would need to call `self->to_string(...)` instead of simply `to_string`
-  how to best deal with modules?
  + `io::middleman` and `opencl::metainfo` only exists if the module has been loaded on startup
  + possible solution: provide `middleman` getter in `actor_system` that throws a `std::logic_error` if the I/O module is not available and users can query whether this would throw using `has_middleman()`
  + impact: `remote_actor()` becomes `system().middleman().remote_actor()`
  + alternative: we could use a `maybe`-based API instead of a throwing getter [1]
    * requires more typing by the user
    * I'm actually in favor of a throwing getter in this case, since users are unlikely to have a reasonable fallback strategy if the middleman does not exist

[1] example API using `maybe`
```cpp
// inside of an actor
auto mm = self->system().middleman();
// sys has type maybe<io::middleman>
if (mm) {
  auto buddy = mm->remote_actor(...);
  // ...
} else {
  // ...
}
```

mavam:
> to_string(message) and from_string() need access to unified type information (UTI) map

That's too bad. `to_string` is a customization point by convention. Users expect it to be a free function using ADL.

> possible solution: make to_string and from_string member functions of actor_system

Alternative: `to_string(x, as)`.

> I'm actually in favor of a throwing getter in this case, since users are unlikely to have a reasonable fallback strategy if the middleman does not exist

:+1:

Neverlord:
@mavam this is well underway with https://github.com/actor-framework/actor-framework/commit/0e33f68ec0ae4b7cfcbad5412b463be48785cba1. I'm waiting for your green light for merging to `develop`.

mavam:
Good to know it's ready! Gimme 1-2 weeks, then I can start the transition.

PR #395, opened 2015-12-29
Factor TLS utility

Lingxi-Li:
Provides portable interface to thread local storage (TLS). The implementation uses `thread_local` if supported, and pthreads otherwise. Code that uses TLS can now be written cleanly. See the modified `detail::get_cache_map()` implementation for a usage example.

### DIFF
--- a/libcaf_core/caf/detail/thread_specific.hpp
+++ b/libcaf_core/caf/detail/thread_specific.hpp
@@ -1,0 +1,168 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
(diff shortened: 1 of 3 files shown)


Neverlord:
This looks like a solid contribution with a lot of effort... *but*, unfortunately it contributes to legacy code.

With 0.15, we're going to remove all implicitly shared state, in particular singletons (see #370). This also means we no longer need any thread-local state...

It really feels bad to turn down this PR, since we welcome and appreciate contributions... If you want to have a look at the direction CAF is going, please check out the [`actor_system` topic branch](https://github.com/actor-framework/actor-framework/tree/topic/actor-system). The branch will get merged to `develop` after some remaining tweaks* are done and documentation updates are underway.

I hope this is not too much of a disappointment to you. It would simply not make sense to accept code that is going to be removed again soon. Sorry. Please feel free to ask in the chat or to drop me an email before starting on a new contribution. I know its frustrating to have worked on something that it is in vain... Outside of holiday seasons, I'm usually quick to respond and I would really like to see and accept more contributions from you in the future.

* For example, removing the last thread-local state used for memory management.

PR #398, opened 2016-01-02
Improve robustness of `actor_system` destruction

Lingxi-Li:
This modification adds the check that no user-spawned actors (including hidden ones) are running upon destruction of the hosting actor system. If the check fails, `std::terminate()` is called, following the convention of `std::thread` destruction. The added restriction and check should help with user code regulation and debugging. The performance overhead amounts to one atomic read-modify-write operation per actor start/stop, which is negligible.


libcaf_core/src/actor_system.cpp
```diff
@@ -135,12 +139,20 @@ actor_system::actor_system(actor_system_config&& cfg)
...
   for (auto i = modules_.rbegin(); i != modules_.rend(); ++i)
     if (*i)
       (*i)->stop();
   logger_.stop();
   CAF_SET_LOGGER_SYS(nullptr);
+  CAF_ASSERT(num_running_actors_ == 0);
+  // calls `std::terminate()` if running actors exist here
+  // follows the convention of `std::thread` destruction
+  if (num_running_actors_ > 0) {
+    CAF_LOG_ERROR("actors running upon hosting actor system destruction");
```
ufownl: The logger has been stopped.

Lingxi-Li: Thanks. I didn't notice this. But, fortunately, it turns out that `CAF_LOG_ERROR()` does not use the logger of the destructing actor system. See [here](https://github.com/actor-framework/actor-framework/blob/topic/actor-system/libcaf_core/caf/logger.hpp#L193) for the implementation of `CAF_LOG_ERROR()`. Macros like this and `CAF_ASSERT()` should be usable outside the context of an actor system. As a result, they do not depend on anything of it.

Neverlord: @Lingxi-Li `CAF_LOG_ERROR` calls `CAF_LOG_IMPL` and `CAF_LOG_IMPL` (which calls `caf::logger::log_static` [L265@logger.cpp](https://github.com/actor-framework/actor-framework/blob/topic/actor-system/libcaf_core/src/logger.cpp#L265)).

Neverlord:
Throwing an exception or, even worse, calling `std::terminate` makes only sense if the application would otherwise run into undefined behavior. Is this the case here? I don't think so. What happens in the destructor of `actor_system` can and should be defined by CAF properly.

The destructor of `actor_system` executes what was previously called `shutdown` in 0.14. There are legitimate use cases where one would want the system to shut down whether or not all actors have been terminated (user-defined `CTRL+C` action on the command line, for example).

As I mentioned on the mailing list, the destructor of `actor_system` could simply call `await_all_actors_done` implicitly. Looking at the unit tests, benchmark programs, and all examples, this function is always called before exiting main. This makes a strong case for making this behavior the default. The system could provide a flag to disable this implicit wait, in which case the developer is responsible for calling (or not calling) it.

If there are still actors alive, will there be undefined behavior? Not necessarily. Event-based actors will never be resumed again after the scheduler has been stopped. This means there might be some actors that are never deleted, but that's not an issue in the mentioned `CTRL+C` scenario. Detached actors on the other hand could cause undefined behavior in case they access the system after it has been destroyed. We could extend the registry to always store all detached actors to allow the system to kill them even in a "dirty shutdown". This adds more overhead to detached actors, but they are heavyweight already and not recommended to be spawned in high numbers. So I think that's a reasonable approach.

While I appreciate contributions to a discussion, I think the behavior proposed in this PR would in fact decrease robustness CAF-based applications and take away design space.

PR #400, opened 2016-01-04
Add reuse_addr argument in publish_local_groups

sjappig:
Enable setting socket option SO_REUSEADDR also when
publish_local_groups is used.


Neverlord:

Issue #381, opened 2015-12-10
Implement "actor decorators"


Neverlord:
Decorators for actor handles allows CAF to model `bind` (see #379) or "dot operator" compositions (see #380). Locally, decorators are a layer of indirection to perform additional steps during message passing. When sending a decorator via the network, CAF needs to make sure decorators are treated differently, i.e., a decorator needs to be decomposed during serialization and rebuild during deserialization to keep compositions network-transparent.

Neverlord:
@Lingxi-Li thanks again for discussion and contribution on this topic. :)

Issue #380, opened 2015-12-10
Provide a "dot operator" for actor handles


Neverlord:
Combining two actors handles using "dot operator" semantics is an important building block for building pipelines of actors. The current proposal is to use `operator*`, since `operator.` cannot be overloaded in C++ and `operator|` is reserved for UNIX-like pipelining in future CAF releases.

Lingxi-Li:
Currently, a composed actor is impersonating the identify of its first constituent. I don't quite understand this design. Maybe, it's because `composed_actor` is intended to be an actor decorator. If so, then I don't think it should be intended as such. Impersonations like this could lead to confusions when the lifetime of a composed actor does not match that of the impersonated actor, i.e., the first constituent. Specifically, a composed actor exits when either constituent exits or it could be exited all alone, independent of its constituents. Because of this, it is appropriate to give a composed actor its own identify.

Also, it seems possible to trim `composed_actor` by deriving from `monitorable_actor` instead of `local_actor`. To summarize, we should have composed actors as independent light-weight actors with own identifies. If you think the idea is good, I'm going to work on a PR.

Neverlord:
> Currently, a composed actor is impersonating the identify of its first constituent. I don't quite understand this design. Maybe, it's because composed_actor is intended to be an actor decorator. If so, then I don't think it should be intended as such. Impersonations like this could lead to confusions when the lifetime of a composed actor does not match that of the impersonated actor, i.e., the first constituent. Specifically, a composed actor exits when either constituent exits or it could be exited all alone, independent of its constituents. Because of this, it is appropriate to give a composed actor its own identify.

In the first draft of this feature, composed actors used to have their own identity. The reason I changed it to the current behavior is twofold. First, the composed actor really is a decorator for `g`. You send `g` a message that says the response goes to `f` (which in turn will respond to the original sender). The fact that `h`  is internally linked to both `f` and `g` is merely an optimization. Second, identical compositions should compare equal, i.e., `h = f*g ^ j = f*g => h = j`. I don't see a way to ensure that each composition gets the same actor ID assigned. Comparing actors should remain a lightweight operation that keeps a total order.

[...]

Issue #379, opened 2015-12-10
Provide a std::bind-like facility for actor handles


Neverlord:
A `bind` facility is an important building block for actor composition that enables currying, re-ordering, etc. In particular useful when building a pipeline of actors where the result of actor `A` is the input of actor `B` (but users usually need to bind atom constants to `B` in order to get this working).

Neverlord:
The current `bind` approach turns out to be not idea. It also has different semantics than `std::bind`, where you get a "function-like" bind expression back that isn't in itself a function. Types are checked at call time, not when constructing the bind expression. Since we want to return an actual, type-checked handle, `bind` is not the right abstraction. Partial function application alone, on the other hand, is not powerful enough. I'd like to keep the use cases for swapping or multiplying input arguments.

Neverlord:
This feature is in experimental stage for 0.15.

PR #401, opened 2016-01-05
Redesign actor decorators

Lingxi-Li:
1. Give decorators their own identities
2. Fix dot composition behavior to match `f.g(x)=f(g(x))`
3. Make decorators depend on decorated actors, but not vice versa

Relates to #379, #380, #381


Neverlord:

Lingxi-Li:
Honestly, I think the decorators deserve their own test suits. Since some decorator tests are already put in `typed_spawn`, I just tried to fit in there also.

Neverlord:
I agree. Putting the decorator test in a separate suite is a good idea.

Lingxi-Li:
Revised as instructed. Also factored and improved decorator unit tests.

Lingxi-Li:
Revised according to 2nd review. The code is much better now :)

Neverlord:
I merged the PR and applied some streamlining + coding style nitpicks [1] to it: https://github.com/actor-framework/actor-framework/commit/af4f3410c8cd159fce4744917482ce4074660c53

[1] for example:
- `: member(value)` instead of `: member { value }`
- `foo tmp{bar}` instead of `foo tmp { bar };`

Neverlord:
@Lingxi-Li by the way, the new unit tests are pretty good. Please feel free to improve other unit tests in CAF as well. ;)

PR #404, opened 2016-01-08
Refine memory order on `monitorable_actor::exit_reason_`

Lingxi-Li:
May improve performance on certain architectures.

### DIFF
--- a/libcaf_core/caf/monitorable_actor.hpp
+++ b/libcaf_core/caf/monitorable_actor.hpp
@@ -51,11 +51,19 @@ public:
 
   size_t detach(const attachable::token& what) override;
 
+  /// Returns the exit reason of this actor or
+  /// `exit_reason::not_exited` if it is still alive.
+  inline exit_reason get_exit_reason() const {
+    return exit_reason_.load(std::memory_order_relaxed);
+  }
+
   /// @cond PRIVATE
 
-  // Returns `exit_reason_ != exit_reason::not_exited`.
+  // returns `exit_reason_ != exit_reason::not_exited`;
+  // it's possible for user code to call this method
+  // without acquiring `mtx_`
   inline bool exited() const {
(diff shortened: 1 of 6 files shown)


Neverlord:
"Improve performance" is a bold statement. On x86 it actually makes no difference, because the platform has only atomics + memory barriers and no acquire/release semantics (other comments see review to https://github.com/Lingxi-Li/actor-framework/commit/3b0953a0e2c6ccf54a932c90fc9fffe515c0e726).

Unfortunately this PR hit while I was rewriting the decorator part and adding a new one. So you'd have to rebase.

Lingxi-Li:
I was understanding memory order from a quite high (and thus coarse) level. I was expecting that unless release-acquire constructs are in place, a reader thread may not catch (see) the result of a writer thread, and then period, without any finer analysis. You seem to be working on a much lower level than me with the understanding of underlying mechanisms, and are thus able to give a more refined and relaxed memory order solution. Learned a lot. Gonna update the PR as suggested.

mavam:
> Improves performance.

Have you measured it?

Lingxi-Li:
Revised. Also edited the bold statement.
@mavam As you may have guessed, I haven't done any performance profiling. Should have been careful with my wording.

mavam:
> @mavam As you may have guessed, I haven't done any performance profiling. Should have been careful with my wording.

I'm curious how you started looking into this particular code branch. Did it show in a profiler on a critical path?

Lingxi-Li:
Not really. I'm mostly wandering about, learning the code. From time to time, things come to me that may turn out to be an improvement to the code. So I propose it. Have learned a lot from the code and from the discussions with you guys.

Lingxi-Li:
Revised according to 2nd review.

Neverlord:

Issue #353, opened 2015-09-17
Improve support for constructing typed response promises


mavam:
`type_response_promise` lacks a viable constructor to support this pattern:
```cpp
    using P = typed_response_promise<either<T>::or_else<U>>;

    [=](X) -> P {
      return T{};
      // ...
      return U{};
    }
```
The workaround involves creating a `P p` and invoking `p.deliver(T{})`. However, CAF should support the above more idiomatic syntax.

Neverlord:
Automatic conversion from values to promises can get a bit tricky. Especially when considering:

```cpp
using P = typed_response_promise<either<int, double>::or_else<bool, float>>;

[=](X) -> P {
  return ???;
};
```

The "standard" workaround is to return a `std::tuple`, but this gets really ugly quickly.

What about a `make_ready_promise` function converting `T => typed_response_promise<T>`? By allowing implicit conversion from both `typed_response_promise<T>` and `typed_response_promise<U>` to `typed_response_promise<either<T>::or_else<U>>`, this would get close to your proposed syntax:

```cpp
    using P = typed_response_promise<either<T>::or_else<U>>;

    [=](X) -> P {
      return make_ready_promise(T{});
      // ...
      return make_ready_promise(U{});
    }
```

A bit more typing, but at the same time more flexible and deals with multiple return values. I'm not too keen on the name, though. Maybe you have a good idea for a better name.

mavam:
Per our discussion, `return response(x, y, z)` would be my favorite for a name.

Neverlord:
Does anything change for this issue when removing `either<...>::or_else<...>` from the picture?

mavam:
`typed_response_promise` would probably get a non-explicit construct overload for an `error_condition`, no?

PR #406, opened 2016-01-11
Fix and improve response promise

Lingxi-Li:
1) Fix issue that promise doesn't get invalidated after delivering in non-chained cases
2) Fix issue that typed promise delivers error response message to non-request senders
3) Add move semantics to typed promise
4) Add `typed_response_promise::valid()`
5) Add `local_actor::make_ready_response_promise()`
6) Improve unit test

IMHO, `pending()` would be a better name than `valid()`. I also thought about the name `make_fulfilled_response_promise()`. But then, I realised that the standard library has been using the term `ready` (see [here](http://en.cppreference.com/w/cpp/thread/promise)), and we should follow existing practice. There is `std::future::valid()`, but no similar things for `std::promise`. The name `valid()` seems a little bit odd and counter-intuitive on response promise. A fulfilled promise should be a valid promise. It's just closed an no longer pending. Users are even tempted to expect a valid promise to be a fulfilled promise, given the semantics of `std::future::valid()`.

Relates to #353

### DIFF
--- a/libcaf_core/caf/detail/split_join.hpp
+++ b/libcaf_core/caf/detail/split_join.hpp
@@ -58,7 +58,7 @@ public:
           this->send(x.first, std::move(x.second));
         this->become(
           // collect results
-          others >> [=] {
+          others >> [=]() mutable {
             join_(value_, this->current_message());
             if (--awaited_results_ == 0) {
               rp.deliver(make_message(value_));
--- a/libcaf_core/caf/local_actor.hpp
+++ b/libcaf_core/caf/local_actor.hpp
@@ -24,6 +24,7 @@
 #include <cstdint>
 #include <exception>
 #include <functional>
+#include <type_traits>
 #include <forward_list>
 #include <unordered_map>
 
@@ -55,6 +56,7 @@
 #include "caf/check_typed_input.hpp"
 #include "caf/monitorable_actor.hpp"
 #include "caf/invoke_message_result.hpp"
+#include "caf/typed_response_promise.hpp"
 
 #include "caf/detail/disposer.hpp"
 #include "caf/detail/behavior_stack.hpp"
@@ -340,6 +342,16 @@ public:
 
   /// Creates a `response_promise` to response to a request later on.
   response_promise make_response_promise();
+
+  /// Creates a `response_promise` and responds immediately.
+  /// Non-requests do not get an error response message.
+  template <class... Ts>
+  typed_response_promise<typename std::decay<Ts>::type...>
+  response(Ts&&... xs) {
+    auto promise = make_response_promise();
+    promise.deliver(std::forward<Ts>(xs)...);
+    return promise;
+  }
 
   /// Sets a custom exception handler for this actor. If multiple handlers are
   /// defined, only the functor that was added *last* is being executed.
--- a/libcaf_core/caf/response_promise.hpp
+++ b/libcaf_core/caf/response_promise.hpp
@@ -34,49 +34,42 @@ namespace caf {
 /// to the client (i.e. the sender of the request).
 class response_promise {
 public:
-  response_promise() = default;
+  /// Constructs an invalid response promise.
+  response_promise();
+  response_promise(local_actor* self, mailbox_element& src);
+
   response_promise(response_promise&&) = default;
   response_promise(const response_promise&) = default;
   response_promise& operator=(response_promise&&) = default;
   response_promise& operator=(const response_promise&) = default;
 
-  using forwarding_stack = std::vector<actor_addr>;
-
-  //response_promise(local_actor* self, actor_addr source,
-  //                 forwarding_stack stages,
-  //                 message_id response_id);
-
-  response_promise(local_actor* self, mailbox_element& src);
-
-  /// Queries whether this promise is still valid, i.e., no response
-  /// was yet delivered to the client.
-  inline bool valid() const {
-    // handle is valid if it has a receiver or a next stage
-    return source_ || ! stages_.empty();
-  }
-
-  inline explicit operator bool() const {
-    return valid();
-  }
-
-  /// Sends the response_message and invalidate this promise.
+  /// Satisfies the promise by sending a non-error response message.
   template <class T, class... Ts>
   typename std::enable_if<
+    (sizeof...(Ts) > 0) ||
     ! std::is_convertible<T, error>::value
   >::type
-  deliver(T&&x, Ts&&... xs) const {
+  deliver(T&&x, Ts&&... xs) {
     deliver_impl(make_message(std::forward<T>(x), std::forward<Ts>(xs)...));
   }
 
-  /// Sends an error as response unless the sender used asynchronous messaging
-  /// and invalidate this promise.
-  void deliver(error x) const;
+  /// Satisfies the promise by sending an error response message.
+  /// For non-requests, nothing is done.
+  void deliver(error x);
+
+  /// Queries whether this promise is a valid promise that is not satisfied yet.
+  inline bool pending() const {
+    return ! stages_.empty() || source_;
+  }
 
 private:
-  void deliver_impl(message response_message) const;
+  using forwarding_stack = std::vector<actor_addr>;
+
+  void deliver_impl(message response_message);
+
   local_actor* self_;
-  mutable actor_addr source_;
-  mutable forwarding_stack stages_;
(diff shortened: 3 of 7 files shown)


Neverlord:
Regarding the name, we already settled on `response` [1] as counterpart to `request`. Further, this function should return a typed promise instead. By making the typed promise convertible to an untyped one, users can always use the same function in both cases. The conversion between typed and untyped promises is quite messy, but that's because `make_response_promise` has no access to the surrounding context in C++ to return the proper type, unfortunately.

Btw, the whole "invalidate promise after deliver" is ultimately pointless (not your fault), because you are allowed to copy promises. The correct solution would be to make promise a move-only type, but that approach fails because C++11 does not have move-capture syntax for lambdas. So, you could send any number of responses you want by simply copying the promises that many times. Just as a side-node, you don't have to change any code for this. We have to wait until we can drop C++11 support in CAF and move on to C++14.

[1] https://github.com/actor-framework/actor-framework/issues/353

Lingxi-Li:
How about making promises non-copyable and non-movable? And then, users can only play with shared pointers to promises. With respect to the naming. `response` is a nice counterpart to `request`. The thing is, would `response_promise::pending()` be a better name than `response_promise::valid()`?

Neverlord:
> users can only play with shared pointers to promises

Making stuff slow via heap indirection is not a good option in my opinion. It's a temporary flaw anyways, since it will disappear once we switch to C++14.

Regarding the name: it's intuitively clear that operator `bool` checks vor `valid`, while the correlation to `pending` would not be intuitive. That might speak against the change. That said, I don't really mind the name too much. We could as well go with `operator bool` only.

mavam:
Do we have a concrete use case where checking for validity is crucial? Assuming that response promises will eventually only be moveable, I don't see directly a situation where you would want to check for the promise validity. (I just haven't come across this situation in my own usage, that doesn't mean there exists a useful one.)

I'm reluctant because the validity semantics are not intuitive, but appear rather an arbitrary definition. A "valid" promise could in my eyes either one that has shipped or one that's still ready to go. Therefore, I'd vote at least against `operator bool` if there exists a use case where checking the promise status is necessary.

Lingxi-Li:
It's an error to satisfy a promise that has already been satisfied. In the standard library, if `std::promise::set_value()` is called on an already satisfied promise, `promise_already_satisfied` is thrown. So, users may not need to perform the check themselves. The library, however, does perform such check under the hood to enforce consistency.

mavam:
> The library, however, does perform such check under the hood to enforce consistency.

Okay, I see the use case and agree that the runtime needs to perform this check. In this case I'd say we simply hide this functionality from the public response promise interface.

Lingxi-Li:
Agreed. `std::promise` is a good reference for our design. We could as well perform the check under the hood, and throw a suitable exception if failed. Also note that `std::promise` is movable only, just as @Neverload has planned for `response_promise` :-)

Neverlord:
Well then. `pending` is fine, but make it private. I'm not in favor of throwing exceptions, though. `CAF_LOG_ERROR("response_promise::deliver called twice")` does a better job at telling developers they're doing something wrong. Otherwise the actor simply gets killed and the developer has to figure out why an actor died--and which--for no obvious reason.

mavam:
This could even raise an assertion in my eyes, as it's violating usage of response promises. Fail early, fail often. (An exception would have get caught be the runtime and just silently kill the actor.)

Neverlord:
I don't think `assert` is necessary. Killing a whole application should be the last resort before entering undefined behavior (which we don't).

Lingxi-Li:
I'm confused about the `deliver()` method being marked `const` and having `mutable` data members.

Neverlord:
That's a C++11-hack. If you're copying the promise into a lambda, the copy inside the lambda is `const`. In C++14 we can declare the lambda itself mutable, but C++11 doesn't have this feature. The alternative is to use `const_cast` everywhere we are copying promises into lambdas, which is even worse than the current workaround.

ufownl:
It seems that we can declare mutable lambda in C++11. See: [lambda functions](http://en.cppreference.com/w/cpp/language/lambda)

Lingxi-Li:
GCC and clang are both happy with `mutable` with `-std=c++11`.

I'm also considering implementing a custom destructor to ensure that a response message is sent anyway for requests, indicating the error that the promise has been abandoned.

Neverlord:
Ah, if that's the case than this was a bug in GCC 4.7. Another legacy workaround we can remove, then.

Neverlord:
Regarding the destructor: this does not work properly as long as we allow copying promises. It's a good idea, but we have to wait until we move on to C++14.

Lingxi-Li:
Really uncomfortable that we have to live with a C++ bug. I'm thinking about formulating a workaround that at leat keeps correct semantics. For example, develop a class template named `move_helper` or something that actually moves the contained object given a copy syntax. In using the helper, its lifetime is kept to the minimum. We get rid of it as soon as the contained object is extracted within a lambda. This workaround should not be that confusing given the name of the template, its usage pattern plus documentation/annotation.

Neverlord:
> I'm thinking about formulating a workaround 

Please don't. Maintaining code that is going to be removed anyway is pointless. The compiler that's currently holding us back is GCC 4.8, which originally released 2013. It might well be that we drop support for this compiler some time this year.

Lingxi-Li:
Revised. Following are some additional notes.

- `pending()` is made public. It is used by `local_actor`. It can also be used for debugging purpose.
- I don't see the reason that non-requests do not receive error messages. Given a non-request, you invoke `deliver()` to send an error message, but nothing happens. If the promise is pending, it remains so afterwards. This is somewhat weird. Both requests and non-requests should be able to receive error messages, e.g., when we have a broken promise. Although for non-requests, the error response message may not simply contain an `error`.
- Both ordinary response messages and error response messages are delivered using `deliver()`. I suggest using different names for the two purposes. For example, `deliver()` and `deliver_error()`. For `typed_response_promise`, the template arguments specify the type of normal response messages. Error response messages can have a different type. So, `deliver()` does the type-match check, while `deliver_error()` does not.

Neverlord:
> pending() is made public.

No objections to that. :)

> I don't see the reason that non-requests do not receive error messages.

Asynchronous messages are "fire and forget" operations. You send the message and move on. What good would an error message be without any context? Just think about it: what should an actor do that receives an `error`? How should it know what this error means or what was causing it? This would also violate type interfaces of actors, unless we treat `error` messages differently and implicitly add them to all interfaces, just like we do with `down_msg` and `exit_msg`. But what would be the design goal here? The error only makes sense when preserving the original context. This is what requests are there for in CAF. In a sense, requests behave a bit like "ordinary" function calls, where you transfer control to another entity and wait for whatever is returned or whatever error/exception might arise.

> I suggest using (...) deliver() and deliver_error().

This information is already encoded in the type system, why repeat it in the function name? It's the same reason why we don't use Hungarian notation.

To give a second reason: this design is orthogonal. When working without promises, actors can have the message handler return `maybe<T>` instead of `T`. In this case, you also use the same statement for actual results and errors: `return ...;`.

Lingxi-Li:
2nd revision done~

Lingxi-Li:
3rd revision done.

Lingxi-Li:
4th revision done. Guess I really messed things up this time ~>_<~

Neverlord:
I don't think you did. :)

Thanks again for the great work and your patience.

Lingxi-Li:
I guess delivering an error response message to non-requests is an error. Printing an error log message with `CAF_LOG_ERROR()` in such case seems appropriate.

Neverlord:
No, this is not an error. Actors have no knowledge whether or not the message they are processing was sent asynchronously or not (and should not have). This is the equivalent of returning `maybe<T>` from a message handler: sending a message asynchronously indicates that the sender is simply not interested in possible errors on the callee's side.

Sending a request means I am preserving the original context of the message and need to compute something based on the result. That's why actors receive errors in this case.

Lingxi-Li:
I see. But we should make the promise satisfied even though the error response message is not actually sent. Currently, nothing is done at all. This does not seem correct when viewed from the responder's side.

Lingxi-Li:
I would rather treat this as an optimization. Conceptually, the error response message is sent, but is then ignored by the receiver. Since the promise has been responded, it should be left in a satisfied state.

PR #407, opened 2016-01-12
Fix compound Boolean expressions in unit tests

Lingxi-Li:
It turns out that the CAF unit testing framework does not support compound Boolean expressions like `x && y` and `x || y`. The tricky part is that no diagnostic was ever issued for such use. So, you are doing it wrong, and you don't know it. I've seen such use in several unit test suits. So I fixed them and re-ran the tests. Guess what? The once-OK `io_dynamic_remote_actor` test suite failed. The mis-use did cover up problems! To prevent future mis-use, I modified the testing framework a bit, so that code with compound Boolean expressions is rejected.

### DIFF
--- a/libcaf_core/test/actor_lifetime.cpp
+++ b/libcaf_core/test/actor_lifetime.cpp
@@ -73,7 +73,7 @@ behavior tester(event_based_actor* self, const actor& aut) {
       // must be still alive at this point
       CAF_CHECK_EQUAL(s_testees.load(), 1);
       CAF_CHECK_EQUAL(msg.reason, exit_reason::user_shutdown);
-      CAF_CHECK_EQUAL(self->current_message().vals()->get_reference_count(), 1);
+      CAF_CHECK_EQUAL(self->current_message().vals()->get_reference_count(), 1u);
       CAF_CHECK(&msg == self->current_message().at(0));
       // testee might be still running its cleanup code in
       // another worker thread; by waiting some milliseconds, we make sure
--- a/libcaf_core/test/actor_pool.cpp
+++ b/libcaf_core/test/actor_pool.cpp
@@ -167,7 +167,7 @@ CAF_TEST(broadcast_actor_pool) {
       CAF_TEST_ERROR("didn't receive a result");
     }
   );
-  CAF_CHECK_EQUAL(results.size(), 25);
+  CAF_CHECK_EQUAL(results.size(), 25u);
   CAF_CHECK(std::all_of(results.begin(), results.end(),
                         [](int res) { return res == 3; }));
   self->send_exit(w, exit_reason::user_shutdown);
--- a/libcaf_core/test/adapter.cpp
+++ b/libcaf_core/test/adapter.cpp
@@ -132,7 +132,7 @@ CAF_TEST(request_response_promise) {
       CAF_CHECK(false);
     },
     [](error err) {
-      CAF_CHECK(err.code() == sec::request_receiver_down);
+      CAF_CHECK(err.code() == static_cast<uint8_t>(sec::request_receiver_down));
     }
   );
   anon_send_exit(dbl, exit_reason::kill);
--- a/libcaf_core/test/aout.cpp
+++ b/libcaf_core/test/aout.cpp
@@ -64,7 +64,7 @@ CAF_TEST(redirect_aout_globally) {
     }
   );
   self->await_all_other_actors_done();
-  CAF_CHECK_EQUAL(self->mailbox().count(), 0);
+  CAF_CHECK_EQUAL(self->mailbox().count(), 0u);
 }
 
 CAF_TEST(global_and_local_redirect) {
@@ -92,5 +92,5 @@ CAF_TEST(global_and_local_redirect) {
   );
   CAF_CHECK(std::is_permutation(lines.begin(), lines.end(), expected.begin()));
   self->await_all_other_actors_done();
-  CAF_CHECK_EQUAL(self->mailbox().count(), 0);
+  CAF_CHECK_EQUAL(self->mailbox().count(), 0u);
 }
--- a/libcaf_core/test/atom.cpp
+++ b/libcaf_core/test/atom.cpp
@@ -83,7 +83,7 @@ CAF_TEST(receive_atoms) {
     self->receive(
       [&](foo_atom, uint32_t value) {
         matched_pattern[0] = true;
-        CAF_CHECK_EQUAL(value, 42);
+        CAF_CHECK_EQUAL(value, 42u);
       },
       [&](abc_atom, def_atom, const std::string& str) {
         matched_pattern[1] = true;
--- a/libcaf_core/test/dynamic_spawn.cpp
+++ b/libcaf_core/test/dynamic_spawn.cpp
@@ -310,11 +310,11 @@ public:
     for (int i = 0; i < 100; ++i) {
       send(this, ok_atom::value);
     }
-    CAF_CHECK_EQUAL(mailbox().count(), 100);
+    CAF_CHECK_EQUAL(mailbox().count(), 100u);
     for (int i = 0; i < 100; ++i) {
       send(this, ok_atom::value);
     }
-    CAF_CHECK_EQUAL(mailbox().count(), 200);
+    CAF_CHECK_EQUAL(mailbox().count(), 200u);
     return {};
   }
 };
--- a/libcaf_core/test/extract.cpp
+++ b/libcaf_core/test/extract.cpp
@@ -73,12 +73,12 @@ CAF_TEST(cli_args) {
     {"in-file,i", "read from file", input_file},
     {"verbosity,v", "1-5", verbosity}
   });
-  CAF_CHECK_EQUAL(res.remainder.size(), 0);
+  CAF_CHECK_EQUAL(res.remainder.size(), 0u);
   CAF_CHECK(res.remainder == message{});
-  CAF_CHECK_EQUAL(res.opts.count("no-colors"), 1);
-  CAF_CHECK_EQUAL(res.opts.count("verbosity"), 1);
-  CAF_CHECK_EQUAL(res.opts.count("out-file"), 1);
-  CAF_CHECK_EQUAL(res.opts.count("in-file"), 0);
+  CAF_CHECK_EQUAL(res.opts.count("no-colors"), 1u);
+  CAF_CHECK_EQUAL(res.opts.count("verbosity"), 1u);
+  CAF_CHECK_EQUAL(res.opts.count("out-file"), 1u);
+  CAF_CHECK_EQUAL(res.opts.count("in-file"), 0u);
   CAF_CHECK_EQUAL(output_file, "/dev/null");
   CAF_CHECK_EQUAL(input_file, "");
 }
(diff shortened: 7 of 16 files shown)


Lingxi-Li:
For example, `CAF_CHECK(true && false)` will succeed while `CAF_CHECK(false || true)` will fail. Looking into the code, I find that the testing framework simply doesn't support such use. It's possible to make it work, but with the loss of the nice overloaded operator `==` and `!=` for floating-point values. I don't yet see a way to combine the best of both worlds.

Lingxi-Li:
Revised to make the diagnostic more accurate.

Neverlord:
Hm, I'm not very familiar with the guts of the unit test framework. @mavam do you think there's a straightforward fix to this issue? I'd like to keep `CAF_CHECK(x() && y())` syntax (that is, as long it does the right thing).

mavam:
I'm not yet seeing the root issue. Let's walk through this with an example of a macro expansion of `CAF_CHECK(a == b && c != d)`:

```cpp
expr ->* a == b && c != d
```
The compiler evaluates this conjunction as:

```cpp
((expr ->* a) == b) && (c != d)
```

Assume the following types: `a -> T`, `b -> U`, `c -> V`, `d -> W`. Then we have:

```cpp
lhs<T> == U && V != W
```

@Lingxi-Li, does anything speak against beefing up the expression templates here, instead of confining them?

Lingxi-Li:
After `lhs<T> == U && V != W`, comes `bool && bool`. The latter boolean value is not checked and so is the `&&` result, and hence the issue. More generally, the current design does not support chaining operations. For example, `CAF_CHECK(a > c == b > c)` which tests that both `a` and `b` are either greater than `c` or not.

In my opinion, one of the biggest differences (and advantages) of the CAF testing framework from (over) others is its special treatment of floating-point comparisons. For example, with `CAF_CHECK(3.14 == 3.14159...)`, you are not actually comparing the values for an exact match, but that they are sufficiently close to be considered equal. The treatment is completely transparent and avoids the common floating-point comparison issue. I really like this feature. With compound expressions in the picture, things get difficult. Consider `CAF_CHECK(3.14 == 3.14159... && 2.718 == 2.718...)`. It's non-trivial to make both comparisons get the special treatment. And for consistency, one cannot really do it for one but not the other. So it's either all or none. It's also non-trivial to check for the floating-point comparison issue in such case.

Another concern regards information output and easy troubleshooting. Consider `CAF_CHECK(a == b)`. If the check failed, both operands can be printed. This gives users a chance to figure out what's going wrong right away from the output instead of setting up an environment and firing up a live debugging. Again, with compound expressions in place. This is difficult to achieve. Just think about the case `CAF_CHECK(a == b && c == d)`. Breaking compound expressions into simple pieces means designing more refined tests, allowing the testing framework to supervise the code more closely. In this regard, compound expressions seem to be a lazy programmer's choice (just saying, not seriously).

If we cannot combine the best of both worlds, a decision has to be made. Dropping these two benefits and having the testing framework support compound expressions makes it more resemble other testing frameworks. And I'm thinking that this may well defeat the goal for CAF to have its own testing framework in the first place. All those facilities (and efforts) there are made just to enable the two advantages.

mavam:
> After `lhs<T> == U && V != W`, comes `bool && bool`. The latter boolean value is not checked and so is the `&&` result, and hence the issue.

Right, the crux is the part `V != W` part, because we cannot lift either operand type. But that's different from the compound expression problem, no? Let's say that we overload the operators for boolean algebra (`&&`, `||`, `!`), just like in standard expression template approaches to define a DSL. Maybe I'm missing something, but I don't see why we can't achieve compound expressions in this fashion. I understand does not fix the floating point and printing issues you point out, but let's keep that a separate discussion.

> For example, `CAF_CHECK(a > c == b > c)` which tests that both a and b are either greater than c or not.

Did you really mean to write `==` here?

> If we cannot combine the best of both worlds, a decision has to be made. Dropping these two benefits and having the testing framework support compound expressions makes it more resemble other testing frameworks. And I'm thinking that this may well defeat the goal for CAF to have its own testing framework in the first place.

We wrote our own unit testing framework mainly because it did not see the point in committing to an external dependency. In general, we avoid any external dependency for the core functionalities. So we started with a simple, Boost-inspired API, but much slimmer than others. But the goal was never to offer superior features to other unit testing frameworks, they just arose naturally. In fact, we want to stay Boost API-compatible for our boost.actor effort.

That said, I hear you that the two features of safe floating-point comparison and operand printing come in really handy, but they actually already fail with boost.actor, because `BOOST_CHECK` does plain eager evaluation. And that is indeed a problem, because currently compound expressions fail in CAF but would work with boost.actor. To remain API-compatible, it would be great if we could support compound expressions. The main problem I see with potential [expression-template-based solution](http://cpptruths.blogspot.com/2014/09/short-circuiting-overloaded-and-using.html) is short-circuiting. Consider `p && *p == 42`. It seems that even lazy evaluation would not prevent the dereference in the second operand.

edwardcapriolo:
Sorry for this from the outfield comment. I have used C-UNIT on c/cpp projects before. I can understand CAF wanting it's own specialized asserts for custom objects, but I wonder for boolean asserts like this would it be better to defer to a more standard library? IE cunit supplies this template with this vard implementation and anything else is something build on top.

Lingxi-Li:
> Maybe I'm missing something, but I don't see why we can't achieve compound expressions in this fashion. I understand does not fix the floating point and printing issues you point out, but let's keep that a separate discussion.

You get it right. Compound expressions can be achieved, but at the cost of raising the floating-point and printing issues.

> Did you really mean to write `==` here?

Yes. It's a shorthand for `(a > c && b > c) || (a <= c && b <= c)`.

> In general, we avoid any external dependency for the core functionalities.

That's a good point. In fact, I've already benefitted from it. Building CAF is a really nice and smooth experience, which would be impossible with tons of external dependencies.

> The main problem I see with potential expression-template-based solution is short-circuiting.

If we are to support compound expressions and just ignore the floating-point and printing issues, I don't see there is a need for expression templates. A trivial solution should do the job flawlessly. Say, we don't interfere with the expression evaluation, but just take its result and act accordingly. We really don't need to care the details of the evaluation process. It's the evaluation result that matters.

Neverlord:
@edwardcapriolo no need to apologize. More input is always welcome. :)
Still, @mavam's right that we don't want dependencies for core components.

Hm, the only thing I take away from the discussion is that "it's complicated". For now, I rather take a bit of inconvenience over a broken framework.

Lingxi-Li:
OK, here comes the new PR. Basically:
1) Add support for chained operations
2) Use `test::equal_to` and `test::not_equal_to` for floating point comparisons

Neverlord:
Actually, I was about to merge your original PR... The "complicated but a bit more correct" float comparison is not really necessary in a unit testing framework IMO, since compared values are usually small. I also prefer minimal changes to the code base, because now I basically have to re-read the whole code.

Lingxi-Li:
Your argument is reasonable. I made this new PR mainly because of the discussion with @mavam. I think he is more in favor of supporting chained operations in the expressions. I still have the original PR locally. So, the decision is yours :-)

If you want to go with this new PR, I want to say something more about floating point comparison (if not, tl;dr). There is no general solution that works in all cases. Otherwise, the standard library should have provided such a thing in `<numeric>` or something. But I do believe there is a nice answer for the particular case of CAF unit testing. And I come up with the following algorithm. Simple, yet effective. I don't think anyone will ever write something in CAF unit tests that may break it :-)
~~~C++
bool equal_to(double x, double y) {
  auto max = std::max(std::abs(x), std::abs(y));
  auto dif = std::abs(x - y);
  return dif <= max * 1e-5;
}
~~~

mavam:
Correct, @Lingxi-Li. I'm in favor of compound predicates and would live with not having fancy floating point comparison and pretty-printing in such cases. In my eyes, this results in the least amount of surprise on the user side while retaining natural semantics.

To move forward, I think we should first agree on the right prioritization of concerns and then find the technical means to get there.

Neverlord:
@mavam to be quite frank, I'm not very interested in details of the unit test framework. As long as it works, of course. So, I would merge the original PR only to have a working solution and don't invest any more time in it. However, I don't want to stifle progress either. I'd be happy to leave this PR to you - unless you don't have the cycles for it either.

mavam:
I'll take care of this. (Alas, I currently I can only advise and don't have the spare cycles to touch code.) In alignment with your lets-just-make-it-work desire, we should simply make compound expressions work and avoid the extra frills. This doesn't require a lot of work: overload the boolean operators accordingly. The end result: we can keep the existing unit testing code as-is.

@Lingxi-Li does this sound like something you could put together?

Lingxi-Li:
Just revised the comparison algorithm. @mavam I think the current version quite meets your design goals. Support compound predicates and with simple yet effective floating point comparison. Also, it presents no surprise to the user since no explicit cast is performed under the hood. However, since no explicitly cast is done automatically, some changes to existing testing code may be needed. But I think this is well justified. Just a simple example will make things clear.
~~~C++
if (err.code() == sec::request_receiver_down)
~~~
The code won't compile, then should the following do?
~~~C++
CAF_CHECK(err.code() == sec::request_receiver_down);
~~~
Still, I can retain the original behavior and perform explicit cast under the hood. Then, no change to existing testing code is needed.

> This doesn't require a lot of work: overload the boolean operators accordingly.

Then, the comparison operators should not simply return a `bool`.

libcaf_core/test/actor_lifetime.cpp
```diff
@@ -73,7 +73,8 @@ behavior tester(event_based_actor* self, const actor& aut) {
       // must be still alive at this point
       CAF_CHECK_EQUAL(s_testees.load(), 1);
       CAF_CHECK_EQUAL(msg.reason, exit_reason::user_shutdown);
-      CAF_CHECK_EQUAL(self->current_message().vals()->get_reference_count(), 1);
+      CAF_CHECK_EQUAL(self->current_message().vals()->get_reference_count(),
+                      static_cast<size_t>(1));
```
mavam: In these cases, you could avoid the verbose cast by using `CAF_CHECK`. Or, my preferred solution, just use an unsigned literal and write `1u` instead of `static_cast<size_t>(1)`.

mavam:
Looks good to me, just a few inline comments left. Now Boost users can feel at home with `CAF_CHECK`, no more surprises.

> Then, the comparison operators should not simply return a bool.

(Yeah, my bad. I was still stuck in the old design, but that would have run into fatal failures for predicate evaluation, such as `p && *p == 42`.)

Too bad the printing is gone, that was really helpful during the debugging process. Could we keep the functionality from your first pull request version and under a separate macro (say `CAF_PRETTY_CHECK`, better name needed), which fails for compound expression with a static assertion? This would give us the best of both worlds.

libcaf_core/test/actor_lifetime.cpp
Lingxi-Li: Indeed. What an oversight I've made!

Lingxi-Li:
Sounds a nice idea. Should be done by tonight :-)

Lingxi-Li:
Finally~ This version should make everyone happy :-)

I added two families of macros: the check family with `CAF_CHECK_EQUAL(x, y)`, `CAF_CHECK_LESS(x, y)`, `CAF_CHECK_GREATER_EQUAL(x, y)`, and so on, and similarly, the require family. Using these macros produces pretty output automatically. Because of the syntax, there is no way to make compound check expressions with these macros.

By the way, someone should look into [this](https://github.com/actor-framework/actor-framework/blob/topic/actor-system/libcaf_io/test/remote_actor.cpp#L554) check. It fails for `system.node() != serv2->node()`.

mavam:
Thanks for bearing with us here and keeping up the great contributions, you rock!

> By the way, someone should look into this check. It fails for `system.node() != serv2->node()`

@Neverlord Can you debug this? Hopefully it's a minor fix that doesn't warrant a new ticket.

PR #410, opened 2016-01-13
Improve `intrusive_ptr`

Lingxi-Li:
As per [this](http://scottmeyers.blogspot.com/2014/06/the-drawbacks-of-implementing-move.html) article. It also puts an object in a determined state after being moved, which is a good thing.

### DIFF
--- a/libcaf_core/caf/intrusive_ptr.hpp
+++ b/libcaf_core/caf/intrusive_ptr.hpp
@@ -30,7 +30,7 @@
 
 namespace caf {
 
-/// An intrusive, reference counting smart pointer impelementation.
+/// An intrusive, reference counting smart pointer implementation.
 /// @relates ref_counted
 template <class T>
 class intrusive_ptr : detail::comparable<intrusive_ptr<T>>,
@@ -66,9 +66,8 @@ public:
   }
 
   ~intrusive_ptr() {
-    if (ptr_) {
+    if (ptr_)
       intrusive_ptr_release(ptr_);
-    }
   }
 
   void swap(intrusive_ptr& other) noexcept {
@@ -92,14 +91,8 @@ public:
   void reset(pointer new_value = nullptr, bool add_ref = true) {
     auto old = ptr_;
     set_ptr(new_value, add_ref);
-    if (old) {
+    if (old)
       intrusive_ptr_release(old);
-    }
-  }
-
-  template <class... Ts>
-  void emplace(Ts&&... xs) {
-    reset(new T(std::forward<Ts>(xs)...));
   }
 
   intrusive_ptr& operator=(pointer ptr) {
@@ -107,21 +100,8 @@ public:
     return *this;
   }
 
-  intrusive_ptr& operator=(intrusive_ptr&& other) {
+  intrusive_ptr& operator=(intrusive_ptr other) {
     swap(other);
-    return *this;
-  }
-
-  intrusive_ptr& operator=(const intrusive_ptr& other) {
-    intrusive_ptr tmp{other};
-    swap(tmp);
-    return *this;
-  }
-
-  template <class Y>
-  intrusive_ptr& operator=(intrusive_ptr<Y> other) {
-    intrusive_ptr tmp{std::move(other)};
-    swap(tmp);
     return *this;
   }
 


Neverlord:
Ok, I was about to write a review to the original PR, but you've removed the unnecessary rvalue/lvalue split in the meantime. The only remaining change is that move-from pointers now always are null, which is a reasonable change.

Three requests: 1) remove the comment above `intrusive_ptr& operator=(intrusive_ptr&& other)`, 2) remove `intrusive_ptr::emplace` (it's dead code), and 3) remove repetitions in the code:

```
intrusive_ptr& operator=(intrusive_ptr&& other) {
    intrusive_ptr tmp{std::move(other)};
    swap(tmp);
    return *this;
  }

  intrusive_ptr& operator=(const intrusive_ptr& other) {
    intrusive_ptr tmp{other};
    swap(tmp);
    return *this;
  }

  template <class Y>
  intrusive_ptr& operator=(intrusive_ptr<Y> other) {
    intrusive_ptr tmp{std::move(other)};
    swap(tmp);
    return *this;
  }
```

This repeats the same code three times now. Maybe we could streamline the code to something like this:

```
  intrusive_ptr& operator=(intrusive_ptr&& other) {
    return assign(std::move(other));
  }

  intrusive_ptr& operator=(const intrusive_ptr& other) {
    return assign(intrusive_ptr{other});
  }

  template <class Y>
  intrusive_ptr& operator=(intrusive_ptr<Y> other) {
    return assign(std::move(other));
  }

// ... in private block ...
  template <class Y>
  intrusive_ptr& assign(intrusive_ptr<Y>&& other) noexcept {
    set_ptr(other.detach(), false);
    return *this;
  }
```

Lingxi-Li:
Yes, yes, the unnecessary rvalue/lvalue split. I just realized that I had been thinking and doing it wrong the whole time! Just failed to shift my mind with the introduction of move semantics. I'm so excited about this finding that I messaged @ufownl (my schoolmate, btw) right away about it. And it turns out that he has done a better job in evolving his mind to catch up with the rapid development of C++ :-) So, let's push it further. Could there be just a single `operator=` defined as follows:

~~~~~C++
intrusive_ptr& operator=(intrusive_ptr other) {
  intrusive_ptr tmp{std::move(other)};
  swap(tmp);
  return *this;
}
~~~~~

And with this in place, is it still necessary to factor an `assign()` method? Actually, I would say programmers may expect and be accustomed to see the pattern on the spot. It's just so widely used. When seeing `return assign(std::move(other))`, one may just think "what the heck is this `assign()` thing" when he is expecting a pattern that has been seen a thousand times.

Neverlord:
> @ufownl my schoolmate, btw

Now I'm really curious. Are you guys doing this in your spare time and happen to have the same hobby (contributing to CAF, apparently) or does your school teaches you C++ and motivates you to contribute to open source projects?

If only the committee would think things through... Try for yourself: only defining `template <class T> intrusive_ptr& operator=(intrusive_ptr<T> other)` will give you compiler errors, because the default assign operators are implicitly deleted. This is because "the standard says so" and the compiler is simply not allowed to take the other signature instead. Even though it would make perfect sense. For the same reason you still have to implement copy and move constructors, even if you could simply implement this with a single constructor taking a sink argument.

Lingxi-Li:
We graduated from the same university and are now professionals (in different cities though). So, the contributions are done in spare time. We both like modern (post-modern?) C++ very much, and want to fully exercise (embrace) it in some serious projects. ufownl found CAF and told me about it. And I would say he has made a right decision. CAF is the way to go. It rocks :-)

Just revised the code and passed all the tests.

Lingxi-Li:
Note that `intrusive_ptr<Y>` is implicitly convertible to `intrusive_ptr<T>`. So no need for an extra `operator=(intrusive_ptr<Y>)`.

ot32em:
I think removing the curly bracket for one-line for loop could be error-prone.

ufownl:
@Neverlord Not really spare time. I'm working for a MMO server now, and I plan to use CAF in the future. ;)

Lingxi-Li:
I think I've figured out the whole thing. Usually, C++ plays with `operator=` overloads under the hood by either defaulting or deleting them. This behavior can be suppressed by providing a user-defined **non-template** overload. Providing a template does not suppress this behavior. It simply adds to the overload set, and the compiler-defined ones are still there.

So, if one just provide a single
~~~C++
template <class U>
intrusive_ptr& operator=(intrusive_ptr<U>);
~~~
What we have ultimately is actually
~~~C++
intrusive_ptr& operator=(intrusive_ptr&&) = default/delete;
intrusive_ptr& operator=(const intrusive_ptr&) = default/delete;

template <class U>
intrusive_ptr& operator=(intrusive_ptr<U>);
~~~
This is not what we want. With some observation, one could find that we don't really need a template, a non-template will work just fine
~~~C++
intrusive_ptr& operator=(intrusive_ptr);
~~~
Since a user-defined non-template overload is provided, the compiler does not generate any definitions under the hood. Even though the assignment operator takes a single concrete `intrusive_ptr` type, you can still assign whatever `intrusive_ptr<U>` to whatever `intrusive_ptr<U>`, since they are implicitly convertible to each other, thanks to this ctor template overload
~~~C++
template <class Y>
intrusive_ptr(intrusive_ptr<Y> other) : ptr_(other.detach()) {
  static_assert(std::is_convertible<Y*, T*>::value,
                "Y* is not assignable to T*");
}
~~~

Neverlord:
@ufownl good to know! And thanks for the recruiting! :)

@ot32em adding curly braces to single-statement `if` and `for` blocks adds code noise. You are right that one has to be careful when refactoring the code later, but three different compilers [with most warnings turned on](https://jenkins.inet.haw-hamburg.de/view/CAF%20Dashboard/), unit testing and static analysis tools (cppcheck + Coverity) spot simple mistakes sooner rather than later.

@Lingxi-Li looks good to me. The less code the better.

Lingxi-Li:
Just a side note. Such idiom doesn't seem OK in general (see [here](http://stackoverflow.com/q/34804343/1348273)), but should be fine in this particular case.

Neverlord:

PR #414, opened 2016-01-21
Fix compilation error on gcc4.8

ufownl:


Neverlord:
Thanks for the patch! :)

PR #416, opened 2016-01-22
Improve typed/untyped response promise cast

Lingxi-Li:
Implicit conversion from untyped promise to typed promise is forbidden. Use `response_promise_cast` to cast between typed and untyped promises explicitly. To cast from typed to untyped, do
~~~C++
auto y = response_promise_cast(std::move(x));
~~~
To cast from untyped to typed, do
~~~C++
auto y = response_promise_cast<int, int>(std::move(x));
~~~
or equivalently
~~~C++
auto y = response_promise_cast<typed_response_promise<int, int>>(std::move(x));
~~~
The second form is to accommodate usage like
~~~C++
using typed_promise = typed_response_promise<int, int>;
auto y = response_promise_cast<typed_promise>(std::move(x));
~~~
Added an `local_actor::make_response_promise()` overload to return typed promise. To return untyped promise, do
~~~C++
auto x = make_response_promise();
~~~
To return typed promise, do
~~~C++
auto x = make_response_promise<int, int>();
~~~
or equivalently
~~~C++
auto x = make_response_promise<typed_response_promise<int, int>>();
~~~

### DIFF
--- a/libcaf_core/caf/all.hpp
+++ b/libcaf_core/caf/all.hpp
@@ -77,7 +77,6 @@
 #include "caf/abstract_channel.hpp"
 #include "caf/may_have_timeout.hpp"
 #include "caf/message_priority.hpp"
-#include "caf/response_promise.hpp"
 #include "caf/binary_serializer.hpp"
 #include "caf/event_based_actor.hpp"
 #include "caf/primitive_variant.hpp"
@@ -86,6 +85,7 @@
 #include "caf/binary_deserializer.hpp"
 #include "caf/scoped_execution_unit.hpp"
 #include "caf/typed_continue_helper.hpp"
+#include "caf/typed_response_promise.hpp"
 #include "caf/typed_event_based_actor.hpp"
 
 #include "caf/decorator/adapter.hpp"
--- a/libcaf_core/caf/local_actor.hpp
+++ b/libcaf_core/caf/local_actor.hpp
@@ -22,6 +22,7 @@
 
 #include <atomic>
 #include <cstdint>
+#include <utility>
 #include <exception>
 #include <functional>
 #include <type_traits>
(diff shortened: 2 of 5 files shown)


Neverlord:
What use case do you see for `response_promise_cast`? If you create the correct response promises via template, there is no more need to convert between the two (in the current use cases).

Lingxi-Li:
`local_actor::response()` always returns a typed promise. So, if one wants to create an untyped promise and respond at the same time, he has to take advantage of an explicit promise cast. If there is a cast from typed promise to untyped, there should also be a cast back for symmetry and completeness.

Through the discussions, I've seen a flaw in my design philosophy. I've always been trying to make things complete and provide support for as many usages as possible, while paying little attention to whether they are actually necessary. When things grow big and become way too flexible, it allows more room for misuse. Take the promise cast for example. A good design should be creating the right promise type from the start and staying with it. When casts are supported, users are then allowed to create a wrong promise type and only to perform a cast later.

So, if you think casts are really not necessary at present time, I will remove them and leave only the improvement on `local_actor::make_response_promise()`. If for some reason, the need for cast does show up in the future, we can add it at that time.

mavam:
> Through the discussions, I've seen a flaw in my design philosophy. I've always been trying to make things complete and provide support for as many usages as possible, while paying little attention to whether they are actually necessary.

I like that realization. There's a term for it: *feature creep*. It is the enemy of slim, use-case driven architecture. If you look at successful projects, you find that their leads make an active effort to combat feature creep.

Neverlord:
Speaking of use-case driven architectures:

@Lingxi-Li I think the design for `make_response_promise` is reasonable, but your design falls flat on its face for the use case `return response(...);`. I don't think anyone would enjoy writing an abomination like `return response_promise_cast(response(...));`, nor is it well readable.

The current design gets away with returning the result of `response` by allowing typed promises to be assigned to untyped ones. While this is a hack, I think it yields much better code on the caller side. After thinking a bit about it, I think we can remove the auto-conversion from untyped handles to typed ones (which originally was a hack for getting `make_response_promise` to work with typed promises), but leave it legal for other way around. The alternative would be to have two different versions of `response()`, but the caller-side code would suffer for it. I always think in terms of "what code would I like to write as a user of this library". Everything else is secondary.

Lingxi-Li:
Revision done ;-]

Lingxi-Li:
2nd revision done.

Neverlord:
Thank you once more for contributing to CAF! :)

PR #417, opened 2016-01-22
Add stream `<<` support for types with `to_string()`

Lingxi-Li:
Now, you can `os << x` instead of `os << to_string(x)`. What's more, in the future, you implement `to_string()` only, and you will have `os << x` support for free :-)

### DIFF
--- a/libcaf_core/caf/all.hpp
+++ b/libcaf_core/caf/all.hpp
@@ -35,6 +35,7 @@
 #include "caf/channel.hpp"
 #include "caf/message.hpp"
 #include "caf/node_id.hpp"
+#include "caf/utility.hpp"
 #include "caf/anything.hpp"
 #include "caf/behavior.hpp"
 #include "caf/duration.hpp"
--- a/libcaf_core/caf/utility.hpp
+++ b/libcaf_core/caf/utility.hpp
@@ -17,75 +17,24 @@
  * http://www.boost.org/LICENSE_1_0.txt.                                      *
  ******************************************************************************/
 
-#include <sstream>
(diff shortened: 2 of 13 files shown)


ufownl:
In my opinion the overload operators you added are unnecessary. Because in most cases, the objects of these classes should be printed using `actor_ostream` rather than `std::ostream'.

Lingxi-Li:
The unit testing framework uses `std::ostringstream` for output (see [here](https://github.com/actor-framework/actor-framework/blob/topic/actor-system/libcaf_test/caf/test/unit_test.hpp#L187)). Besides, CAF should allow easy use with the standard library. One should use `actor_stream` for direct output, but what if he or she wants to use `std::ostringstream` to compose custom strings?

Lingxi-Li:
Just found a better solution (see [here](http://stackoverflow.com/a/34946808/1348273)). PR updated.

Neverlord:
This PR has too much (repetitive) code for such a little addition. Even the current solution for `to_string` is not ideal. We discussed this with @mavam a while ago. A modified version of his concept-based solution in VAST is the way to go going forward. This could also be easily extended to include `operator<<` etc.

Speaking about going forward, maybe we can safe some time if we discuss this in chat before you work on PRs. ;)

Lingxi-Li:
It's OK. No time wasted really. I've learned a lot through the process :-) In fact, I think this should be something to be done by the standard library, and have opened a [topic](https://groups.google.com/a/isocpp.org/d/topic/std-proposals/q3WCpLp1xmQ/discussion) on the std discussion group.

PR #427, opened 2016-02-17
Fix compilation error on GCC4.8

ufownl:


Neverlord:
Thanks a lot! :)

One item less on my todo list.

Issue #426, opened 2016-02-05
There are two tests failed on CentOS7


ufownl:
LastTestsFailed.log:
```
53:io_basp
54:io_basp_asio
```

LastTest.log:
```
...
53/54 Testing: io_basp
53/54 Test: io_basp
Command: "/home/ufownl/Softwares/actor-framework/build/bin/caf-test" "-n" "-v" "5" "-s" "io_basp"
Directory: /home/ufownl/Softwares/actor-framework/build
"io_basp" start time: Feb 06 00:09 CST
Output:
----------------------------------------------------------
+----------------------------------------------------------------------+
                                io_basp
+----------------------------------------------------------------------+

- empty_server_handshake
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:113  mpx_ != nullptr
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:114  &system == &mpx_->system()
  -> this node: 95bef94272608bf17d55c0d64a08bee8ff309e00:5173
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:454  basp::valid(hdr)
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:455  basp::is_handshake(hdr)
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:456  ::caf::test::equal_to(to_string(hdr), to_string(expected))
  -> 5 checks took 1 ms

- non_empty_server_handshake
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:113  mpx_ != nullptr
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:114  &system == &mpx_->system()
  -> this node: 95bef94272608bf17d55c0d64a08bee8ff309e01:5173
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:472  hexstr(buf) == hexstr(expected_buf)
  -> 3 checks took 712 us

- remote_address_and_port
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:113  mpx_ != nullptr
  ** /home/ufownl/Softwares/actor-framework/libcaf_io/test/basp.cpp:114  &system == &mpx_->system()
  -> this node: 95bef94272608bf17d55c0d64a08bee8ff309e02:5173
[...]

PR #428, opened 2016-02-21
Fix thread sync issue in test `io_basp[_asio]`

ufownl:
Relates #426 .


Neverlord:
Big thumbs up! Thanks a lot.

PR #429, opened 2016-02-26
Make `function_view` support void result

ufownl:


ufownl:
Revision done. :)

Neverlord:
Thanks for the improvement! :)

Issue #431, opened 2016-03-01
Actively probe connections to remote nodes


Neverlord:
Currently, CAF does use periodic keepalive messages and relies on TCP. When remote nodes fail, this means that CAF will recognize failed peers only after a TCP timeout occurs, which can take a few minutes.

At the very least, CAF should offer a configuration parameter that allows developers to have down and exit messages triggered quicker.

Neverlord:
@ufownl thank you again for implementing this

PR #433, opened 2016-03-03
Add the periodic heartbeat mechanism of BASP

ufownl:
Relates #431.

### DIFF
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -122,6 +122,7 @@ public:
   atom_value middleman_network_backend;
   bool middleman_enable_automatic_connections;
   size_t middleman_max_consecutive_reads;
+  size_t middleman_basp_heartbeat_interval;
 
   // System parameters that are set while initializing modules.
   node_id network_id;
--- a/libcaf_core/caf/atom.hpp
+++ b/libcaf_core/caf/atom.hpp
@@ -139,6 +139,9 @@ using spawn_atom = atom_constant<atom("spawn")>;
 /// Used for migrating actors to other nodes.
 using migrate_atom = atom_constant<atom("migrate")>;
 
+/// Used for triggering periodic operations.
+using tick_atom = atom_constant<atom("tick")>;
+
 } // namespace caf
 
 namespace std {
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -142,6 +142,7 @@ actor_system_config::actor_system_config() {
   middleman_network_backend = atom("default");
   middleman_enable_automatic_connections = false;
   middleman_max_consecutive_reads = 50;
+  middleman_basp_heartbeat_interval = 0;
   nexus_port = 0;
 }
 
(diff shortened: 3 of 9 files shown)


Neverlord:
I discussed this with @josephnoir a bit and we both wonder why you do this ping/pong. A heartbeat message usually is one-way only. Doing this request/response-style heartbeat doubles the amount of periodic messages. If you have 10 nodes, a usual heartbeat mechanism would send 90 heartbeat messages per cycle (each node sends its heartbeat to the other nine). You send 180 messages instead.

The purpose of heartbeats is to signalize to *others* that we are still alive. By receiving their heartbeats in turn, we can determine whether or not they are still alive. Since CAF is currently only using TCP, there is not even a need to count how many heartbeats we received per second. In a TCP setting, heartbeats are only there to trigger periodic communication in order to propagate node failures quicker if no actual actor communication is happening.

For BASP, this would mean to add one new message type (`heartbeat`) that is recognized by receivers but doesn't trigger any other code. Or maybe add a new `instance::callee::handle_heartbeat(const node_id& nid)` event handler is currently empty in `basp_broker_state`. This would allow us to add some custom code later that perhaps tries to detect weird behaviors of nodes (heavy jitter between received heartbeats could indicate an overloaded network for example).

ufownl:
> I discussed this with @josephnoir a bit and we both wonder why you do this ping/pong. 

Hmm... I think I trapped in a thinking pattern. In the project of our company, the client periodically send a heartbeat-ping to the server and the server response a heartbeat-pong. Then both of them can determine   whether or not peer endpoint is alive and measure the delay between them.

> In a TCP setting, heartbeats are only there to trigger periodic communication in order to propagate node failures quicker if no actual actor communication is happening.

So, the pattern above does not suit CAF. I'll do the revision soon.

ufownl:
Revision done.

ufownl:
> Seems correct, but the commit only says "Add the periodic heartbeat mechanism of BASP". Streamlining code is fine, but it needs to be documented and properly motivated to be transparent to other developers. :)
> 
> It also makes reviewing PRs tedious if changes are seemingly unrelated to the commit description.

You are right, I'll pay attention afterward. :)

ufownl:
2nd revision done.

Neverlord:
I have one complain regarding the commit itself. It breaks in the summary line. Here's the relevant part from our contribution guide:

Commit Message Style
--------------------

- The first line succintly summarizes the changes in no more than 50
  characters. It is capitalized and written in and imperative present tense:
  e.g., "Fix bug" as opposed to "Fixes bug" or "Fixed bug".

- The first line does not contain a dot at the end. (Think of it as the header
  of the following description).

- The second line is empty.

- Optional long descriptions as full sentences begin on the third line,
  indented at 72 characters per line.


A short commit message in this case could be: "Add heartbeats to BASP; streamline basp_broker". That you streamlined handling of `connection_closed_msg` would then belong into the long description.

Two minor issues with the patch itself:

- "heartbeat period" sounds odd, the common term is "heartbeat interval"
- heartbeats currently have second resolution, which is rather long and quite coarse; usually, systems have something like 250ms intervals, so the parameter should use milliseconds instead

ufownl:
Tweaks done.

Neverlord:

Neverlord:
Thank you once more for contributing to CAF, you rock! :)

PR #438, opened 2016-03-07
Cleanup scheduling-related code

Lingxi-Li:


Neverlord:
As mentioned in `scheduler_policy.hpp`, the class is a `concept`, i.e., not intended as base class. It exists solely for documentation. Using a template-heavy class like this one as base makes no sense. It has no virtual member functions, not even a virtual destructor. And avoiding virtual dispatch was the point to get as much inlined as possible. Once we finally have concepts in C++, `scheduler_policy` would become just that.

Other than that, there are a couple of changes I'd be happy to merge such as simply calling `enqueue` in `work_sharing::internal_enqueue` and moving `after_resume` out of the switch-case in `worker.hpp` (and removing dead code, etc.).

Lingxi-Li:
I see that there are a number of customization points, namely `before_shutdown()`, `before_resume()`, `after_resume()`, `after_completion()`. Currently, both `work_sharing` and `work_stealing` do not use any of them. So I guess they are usually not needed by actual policies. As such, wouldn't it be nice if we provide a default implementation for these customization points that does nothing? As a result, the customization points are made optional, actual policy classes only need to implement the customization points they need. We can even provide a default empty definition for `coordinator_data` and `worker_data`. This simplifies the implementation of actual policy classes. Since no virtual dispatch is involved, there is no performance penalty. Of course, by doing this, `scheduler_policy` not only serves as a concept but also as a base class to provide default implementations and a place to factor common functionalities like 'd()'. Would this be a fair deal?

Neverlord:
> As such, wouldn't it be nice if we provide a default implementation for these customization points that does nothing?

Sure. You could use something like `unprofiled` as base class (with virtual destructor) that provides empty implementations for all profiling related member functions. That gives readers of the code a clear idea what the base type does (and does not).

Lingxi-Li:
Revised. I named the base class `trivial` for it provides a trivial empty implementation for the customization points.

Neverlord:
`trivial` is not a good name in my opinion. It is too generic and says nothing about the *intent* of this class. You have to read the documentation of `trivial` in order to find out what is so trivial about it.

Why not go with the proposed `unprofiled` instead? `class work_stealing : public trivial` doesn't give any context as what `trivial` provides or does. `class work_stealing : public unprofiled` on the other hand is clear and makes sense in an object-oriented world with is-a relationship. The work stealing policy is not trivial, but it is unprofiled.

Lingxi-Li:
Renamed `trivial` to `unprofiled` :-)

Neverlord:

Neverlord:
Thanks for your patience with me. :)

Issue #439, opened 2016-03-07
BASP handshake issue


Neverlord:
Reported on the mailing list.

---

Hi Folks,

 This issue is a follow-up on a previous discussion: https://groups.google.com/forum/#!topic/actor-framework/wxWrEyoSGMs . I have opened this new discussion. because the topic has been shifted from the old one.

 During our evaluation we have found that basp handshake can be problematic in many cases.

- If we're connecting to a port which is not transmitting data, then the caf::io::remote_actor() call hangs without any sensible timeout (as written on the previous)
- My colleage also made some tests connecting towards a remote netcat instance, which fed random junk  (from /dev/random) as handshake

Let me quote his findings:
"
cat /dev/random | nc -l 5555 | hexdump -C 

and in another shell run a program that executes a 
caf::io::remote_actor("localhost", 5555); 

> ./bin/remotebroker localhost 
2016-03-02 19:06:46,426 DEBUG [connector] [caf::behavior 
connector::connecting()] [../connector.cpp:43] connecting 

and there we stand. CPU is spinning at 100%, random data is travelling 
over the network and memory usage is hitting the 4GB limit here after a 
few seconds: 

8629 user 20 0 4878160 3,987g 4048 R 47,33 25,56 0:12.20 
connector 

pmap output: 

4336152K writable-private, 541796K readonly-private, 212K shared, and 
4182232K referenced 

The last thing that I see in the debugger is remote_actor.cpp:50 (git 
master branch 5097f69c7ac29fb51f32d33d31ccfe5e4763dc13), which is doing a 

self->sync_send(mm, connect_atom::value, std::move(host), port).await( 
... 
} 

which however is never happening (the await) 

The offending class seems to be basp_worker, since I see exceptions 
being thrown from there in case I do a 

cat /dev/zero | nc -l 5555 

At least in that case, the remote_actor does not get stuck. 

AFAICS, non-blocking read/write is needed here, along with a simple 
checker for protocol correctness. "

Unfortunately these issues are pretty much showstoppers for us, any help or advice is appreciated.

PR #440, opened 2016-03-09
Fix BASP handshake issue

ufownl:
Relates #439 .


Neverlord:
Awesome, thanks for the patch!

PR #392, opened 2015-12-21
Improve `policy::work_stealing::try_steal()`

Lingxi-Li:
Victim selection is made collision-free. A try-loop is no longer needed. In previous version, with two workers, the probability of collision is 50%!

PR #447, opened 2016-03-16
Implement true uniform random work stealing

Lingxi-Li:
In general, `rand() % n` does not generate a true discrete uniform random distribution among `[0..(n - 1)]`, unless `(RAND_MAX + 1) % n == 0`. As a result, in the previous version, workers with smaller indices are more likely to be victims.

### DIFF
--- a/libcaf_core/caf/policy/work_stealing.hpp
+++ b/libcaf_core/caf/policy/work_stealing.hpp
@@ -57,23 +57,30 @@ public:
 
   // The coordinator has only a counter for round-robin enqueue to its workers.
   struct coordinator_data {
-    std::atomic<size_t> next_worker;
-    inline coordinator_data() : next_worker(0) {
+    inline explicit coordinator_data(scheduler::abstract_coordinator*)
+        : next_worker(0) {
       // nop
     }
+
+    std::atomic<size_t> next_worker;
   };
 
   // Holds job job queue of a worker and a random number generator.
   struct worker_data {
+    inline explicit worker_data(scheduler::abstract_coordinator* p)
+        : rengine(std::random_device{}()),
+          // no need to worry about wrap-around; if `p->num_workers() < 2`,
+          // `uniform` will not be used anyway
+          uniform(0, p->num_workers() - 2) {
+      // nop
+    }
+
     // This queue is exposed to other workers that may attempt to steal jobs
     // from it and the central scheduling unit can push new jobs to the queue.
     queue_type queue;
     // needed to generate pseudo random numbers
     std::default_random_engine rengine;
-    // initialize random engine
-    inline worker_data() : rengine(std::random_device{}()) {
-      // nop
-    }
+    std::uniform_int_distribution<size_t> uniform;
   };
 
   // Goes on a raid in quest for a shiny new job.
@@ -85,7 +92,7 @@ public:
       return nullptr;
     }
     // roll the dice to pick a victim other than ourselves
-    size_t victim = d(self).rengine() % (p->num_workers() - 1);
+    auto victim = d(self).uniform(d(self).rengine);
     if (victim == self->id())
       victim = p->num_workers() - 1;
     // steal oldest element from the victim's queue
--- a/libcaf_core/caf/policy/scheduler_policy.hpp
+++ b/libcaf_core/caf/policy/scheduler_policy.hpp
@@ -21,6 +21,7 @@
 #define CAF_POLICY_SCHEDULER_POLICY_HPP
 
 #include "caf/fwd.hpp"
+#include "caf/scheduler/abstract_coordinator.hpp"
 
 namespace caf {
 namespace policy {
@@ -30,10 +31,14 @@ namespace policy {
 class scheduler_policy {
 public:
   /// Policy-specific data fields for the coordinator.
-  struct coordinator_data { };
+  struct coordinator_data {
+    explicit coordinator_data(scheduler::abstract_coordinator*);
+  };
 
   /// Policy-specific data fields for the worker.
-  struct worker_data { };
+  struct worker_data {
+    explicit worker_data(scheduler::abstract_coordinator*);
+  };
 
   /// Enqueues a new job to coordinator.
   template <class Coordinator>
--- a/libcaf_core/caf/policy/unprofiled.hpp
+++ b/libcaf_core/caf/policy/unprofiled.hpp
@@ -20,6 +20,8 @@
 #ifndef CAF_POLICY_UNPROFILED_HPP
 #define CAF_POLICY_UNPROFILED_HPP
 
+#include "caf/scheduler/abstract_coordinator.hpp"
+
 namespace caf {
 namespace policy {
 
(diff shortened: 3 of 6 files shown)


libcaf_core/caf/policy/work_stealing.hpp
```diff
@@ -85,7 +85,8 @@ class work_stealing : public unprofiled {
       return nullptr;
     }
     // roll the dice to pick a victim other than ourselves
-    size_t victim = d(self).rengine() % (p->num_workers() - 1);
+    std::uniform_int_distribution<size_t> uniform(0, p->num_workers() - 2);
+    size_t victim = uniform(d(self).rengine);
```
mavam: (Nit: you could use `auto` on the LHS here.)

libcaf_core/caf/policy/work_stealing.hpp
```diff
@@ -85,7 +85,8 @@ class work_stealing : public unprofiled {
       return nullptr;
     }
     // roll the dice to pick a victim other than ourselves
-    size_t victim = d(self).rengine() % (p->num_workers() - 1);
+    std::uniform_int_distribution<size_t> uniform(0, p->num_workers() - 2);
```
mavam: This is has nothing to do with your patch (which looks valid to me), but I wonder why we are not drawing a value from *[0, N)* where *N* is the number of workers. Why are we using *N-1*? There is a check in place afterwards where we take an arbitrary worker in case we picked ourself.

Not sure if I am missing something obvious here.

ufownl: @mavam It's a trick to avoid collision. See #392 .

Lingxi-Li: `N` workers excluding self are `N - 1` workers. We should select from these `N - 1` workers, not `N` workers, otherwise we may have a chance to select self, and thus a collision. Among `[0..(N - 2)]` are the `N - 1` indices we need. The self index  may be within `[0..(N - 2)]` or not. In the latter case, things are fine trivially. In the former case, suppose self index is `X`, things are fine if the generated index is not `X`. If it is `X`, it should actually be `N - 1`. This is because we are not supposed to generate `X`, and `N - 1` would never be generated for it is out of the range `[0..(N - 2)]`.

Lingxi-Li:
Updated.

Lingxi-Li:
Optimized according to [this](http://stackoverflow.com/a/36048705/1348273).

libcaf_core/caf/policy/work_stealing.hpp
mavam: > N workers excluding self are N - 1 workers. 

If the idea is to exclude self, then I don't think this is statistically correct, because the two distributions `Unif(0, N-1)` and `Unif(0, N-2)` have two different probabilities for each victim. Using `Unif(0, N-2)` would be correct if we can ensure that the current worker is always `N-1`. But I don't see the code doing that. Instead, there is a candidate check that checks if the selection hit ourselves:

```cpp
    if (victim == self->id())
      victim = p->num_workers() - 1;
```

What we want is a random variable that describes all indices *but* the current one. Since there's not straight-forward distribution for that, we use a candidate check. In fact, I think the candidate check should not pick always the last element, but something dynamic. Let's hold that thought. Ultimately, we want to sample from the remaining workers, but I think we have to do it with probability `1/N` instead of `1/(N-1)`, because the currently executing worker cannot be assumed to be `N`.

mavam: > @mavam It's a trick to avoid collision. See #392 

I see that this could be an issue. Instead of repeatedly resampling from the same distribution (which #392 changed), another approach would be to fix the candidate check. Say if `i` is the index of self, then upon collision one simple way would be to choose `i + 1 % N`.

To be frank, I would really like to measure this to see the concrete effects, but I lack the cycles to do so.

Neverlord: @mavam I'm not sure if I get what you are trying to get to. We have N workers, but need to choose one of N-1 (any worker other than self).

Now, we get a random number for N-1. If the ID of self is  N, than we have no problem. Otherwise, we exclude the worker holding the last ID but have a change of hit ourselves. The fix is to map our ID to the omitted one... This is straightforward mapping and each ID has the same chance of getting picked while making sure our ID isn't in the pool.

Did I overlook something?

Neverlord:
> workers with smaller indices are more likely to be victims.

Do you have some numbers to back this up? The uniform distribution comes at a (runtime) price, so I first would like to see if it's worth it. How many cores do you need for this to become significant? 4? 32? 128? 10k?

libcaf_core/caf/policy/work_stealing.hpp
```diff
@@ -57,23 +57,32 @@ class work_stealing : public unprofiled {
...
+
+    std::atomic<size_t> next_worker;
   };
 
   // Holds job job queue of a worker and a random number generator.
   struct worker_data {
+    inline explicit worker_data(scheduler::abstract_coordinator* p)
+          // when `p->num_workers() < 2`,
+          // this saves us from the `std::random_device{}()` call
+        : rengine(p->num_workers() < 2 ? 0 : std::random_device{}()),
```
Neverlord: This micro optimization isn't useful. The few cycles we "waste" when running with a single worker isn't worth the code IMO.

libcaf_core/caf/policy/work_stealing.hpp
mavam: > The fix is to map our ID to the omitted one...

Ah, right. I finally could put the pieces together. What we're doing indeed works correctly. Thanks for clarifying!

libcaf_core/caf/policy/work_stealing.hpp
Lingxi-Li: Indeed. After a second thought, it's yet another evil premature optimization.

Lingxi-Li:
The number generated by `random_device` is in the range `[0..UINT_MAX]`. `UINT_MAX + 1` is a power of 2. So, the issue happens whenever we have an odd number of workers. Let's say 3. Then workers 0 and 1 will have a higher probability to be victims than worker 2. Whether this makes a notable difference depends on a number of things. Concerning probability, one such thing is time, specifically, how long the CAF program is going to run uninterrupted. As time goes by, the probability difference will eventually manifest itself. Alas, a detailed analysis is going to be difficult.

I agree that this modification may not be useful in typical use cases (still, a hard-to-analyze argument). But nevertheless, I think it's worth it to write theoretically sound/regular programs that are easy to analyze and with guaranteed performance (so, sleep tight ever since ^_^).

> The uniform distribution comes at a (runtime) price, so I first would like to see if it's worth it.

The price paid is constant runtime overhead added to a single operation. It's not going to be a matter under almost all circumstances. However, when the probability difference manifests itself, the effect may be drastic.

Lingxi-Li:
Any news on this PR? If it seems no good, we can just close it. Maybe, it's better that we look into this issue when problems do arise from it in practice.

Neverlord:
Sorry for the silence. I'm going to benchmark this PR against the topic branch to see if it has any measurable effect on our 64 core machine before making any decision on it. However, this is not going to happen before 4/4.

Lingxi-Li:
Thanks for the reply. Since 64 is a power of 2, you may consider making an odd number of logical cores. During the benchmarking, counting the number of times work stealing happens may also help.

Neverlord:
@Lingxi-Li Looks good to merge, but please squash the commits first (preferably using the second commit message).

Lingxi-Li:
Squashing is done :)

Neverlord:
Thank you once more for contributing to CAF. You rock! :)

PR #448, opened 2016-03-30
Improve the use of random distributions

Lingxi-Li:
As per [this](http://stackoverflow.com/a/36048705/1348273), we should store a random distribution object and reuse it if possible. This is the intended usage. Otherwise, the facility would have been designed as free functions instead of classes.

### DIFF
--- a/libcaf_core/src/actor_pool.cpp
+++ b/libcaf_core/src/actor_pool.cpp
@@ -78,13 +78,14 @@ actor_pool::policy actor_pool::random() {
     }
     void operator()(actor_system&, uplock& guard, const actor_vec& vec,
                     mailbox_element_ptr& ptr, execution_unit* host) {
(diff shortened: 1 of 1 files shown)


mavam:
In general, this optimization makes sense. I once looked at the implementation of the normal distribution when I created a std-compliant interface for a Pareto distribution. Distribution objects can have several levels of optimizations and generate multiple values at once, where `operator()` only retrieves a value from the cache.

Neverlord:
The reason why the object is not re-used is that `operator()` must be implemented thread-safe. I'm actually not quite sure if the current implementation with `random_device` as member could cause UB.

Lingxi-Li:
Yes, thread-safety may be a problem. But in this particular case, the relevant code is guarded by the lock `guard`. So, I think it shouldn't be a problem here. No?

Neverlord:
Silly me. Sorry for the noise.

PR #451, opened 2016-04-09
Cleanup `actor_factory` test code

Lingxi-Li:
- Remove redundant code which gives readers the false impression that a user-defined `make_behavior()` is required;
- Use [inheriting constructor](http://en.cppreference.com/w/cpp/language/using_declaration#Inheriting_constructors) (a C++11 feature) to make code cleaner.


Neverlord:
@Inet-Jenkins test this please.

PR #452, opened 2016-04-10
Enhance serialization framework

mavam:
This patch makes the version parameter in the serialization API optional. CAF doesn't use it anyway right now and the majority of use cases won't need it either. It currently exists only for Boost compatibility, which this patch doesn't affect.

### DIFF
--- a/libcaf_core/caf/detail/delegate_serialize.hpp
+++ b/libcaf_core/caf/detail/delegate_serialize.hpp
@@ -52,9 +52,19 @@ namespace detail {
 // to enable both ADL and picking up existing boost code.
 
 template <class Processor, class U>
-void delegate_serialize(Processor& proc, U& x) {
+auto delegate_serialize(Processor& proc, U& x)
+  -> decltype(serialize(proc, x, 0)) {
   using namespace boost::serialization;
   serialize(proc, x, 0);
+}
+
+// Calls `serialize(...)` without the unused version argument, which CAF
+// ignores anyway.
+
+template <class Processor, class U>
+auto delegate_serialize(Processor& proc, U& x)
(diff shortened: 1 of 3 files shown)


mavam:
I've also added a simple byte counter to the binary (de)serializer. This change assumes that all function call funnel into `apply_raw`, which seems to be the case (but I haven't manually checked all code paths exhaustively).

Neverlord:
What is the motivation behind the byte count? I don't see any useful application to be honest. The serializer always writes into a buffer. Why not simply query the size of that buffer directly?

mavam:
> What is the motivation behind the byte count?

I use it for various performance counters. When compute compression ratios, for examples, which I want to let drive control flow eventually. The counter also serves as direct proxy for output size when you write into a file.

> Why not simply query the size of that buffer directly?

That would assume you always fill the buffer entirely, which is not always the case. If you reuse an existing buffer, it usually doesn't shrink (unless you call `shrink_to_fit`).

Neverlord:
> That would assume you always fill the buffer entirely, which is not always the case. If you reuse an existing buffer, it usually doesn't shrink (unless you call `shrink_to_fit`).

Still, your write count in the serializer will return the same value as `vector::size`. That's not the same as `vector::capacity`, but that's true in both cases, isn't it?

mavam:
> Still, your write count in the serializer will return the same value as `vector::size`

Yeah, right, I guess that's the case in the current design, where you let the `vector` allocate its needed space and then only work with `size`. You could equally assume that you allocate a fixed block a priori and then just write in it, which was my mental model. Then you'd need the counter to know where you're at.

I'm not feeling too strongly about the counter patch. There are indeed other ways to get these numbers. I'll remove the patch.

PR #453, opened 2016-04-13
Make (de)serializer more flexible

mavam:
First, this change moves the context to the parameter to the end in order to make it optional. The context is only necessary when serializing messages. Since CAF's serialization framework also serves also other application's need, this CAF-specific detail now no longer constitutes a mandatory part of the API.

This change also unifies the constrution of caf::serializer from a function that writes a pair of const char* pointer and size into an arbirtrary destination.

Status:

- [x] Make CAF context parameter optional
- [x] Add new streambuffer types
- [x] Make serializer API streambuffer-based
- [x] Make deserializer API streambuffer-based
- [x] Perform rough viability performance comparison
- [x] Perform end-to-end performance comparison


mavam:
Here's the more comprehensive benchmark:

### Infrastructure

```cpp
#include <caf/all.hpp>

using namespace std;
using namespace std::chrono;
using namespace caf;

using stopwatch = high_resolution_clock;

// Measure overhead of data serialization performance.
template <class T>
microseconds benchmark(T x) {
  vector<char> buffer;
  // Serialize.
#ifdef NEW
  vectorbuf<char> vb{buffer};
  binary_serializer bs{vb};
#else
  scoped_execution_unit ctx;
  binary_serializer bs{&ctx, buffer};
#endif
  auto start = stopwatch::now();
  for (auto i = 0u; i < 100000; ++i)
    bs << x;
  // Deserialize.
#ifdef NEW
  charbuf cb{buffer};
  binary_deserializer bd{cb};
#else
  binary_deserializer bd{&ctx, buffer};
#endif
  T y;
  for (auto i = 0u; i < 100000; ++i)
    bd >> y;
  auto stop = stopwatch::now();
  return duration_cast<microseconds>(stop - start);
}

// Measure overhead of construction, since serializers are often used in a
// ad-hoc fashion.
template <class T>
microseconds benchmark2(T x) {
#ifndef NEW
  scoped_execution_unit ctx;
#endif
  auto start = stopwatch::now();
  for (auto i = 0u; i < 100000; ++i) {
    vector<char> buffer;
#ifdef NEW
    vectorbuf<char> vb{buffer};
    binary_serializer bs{vb};
#else
    binary_serializer bs{&ctx, buffer};
#endif
    bs << x;
  // Deserialize.
#ifdef NEW
    charbuf cb{buffer};
    binary_deserializer bd{cb};
#else
    binary_deserializer bd{&ctx, buffer};
#endif
    T y;
    bd >> y;
  }
  auto stop = stopwatch::now();
  return duration_cast<microseconds>(stop - start);
}

auto main() -> int {
  auto str = std::string{"the quick brown fox jumps over the lazy dog"};
  auto large = str;
  for (auto i = 0; i < 100; ++i)
    large += str;
  auto u64 = uint64_t{42};
  cout << benchmark(str).count() << "us" << endl;
  cout << benchmark(large).count() << "us" << endl;
  cout << benchmark(u64).count() << "us" << endl;
  cout << "-------" << endl;
  cout << benchmark2(str).count() << "us" << endl;
  cout << benchmark2(large).count() << "us" << endl;
  cout << benchmark2(u64).count() << "us" << endl;
}
```

System: Mac OS X (Yosemite)
Compiler: Clang "7.02"

Compiled as follows:
- `c++ -O3 -g -std=c++11 -stdlib=libc++ -I ~/code/caf/libcaf_core -L ~/code/caf/build/lib -DNEW benchmark.cpp -o bench-new -lcaf_core`
- `c++ -O3 -g -std=c++11 -stdlib=libc++ -I ~/code/caf/libcaf_core -L ~/code/caf/build-as/lib benchmark.cpp -o bench-old -lcaf_core`

### Results

Old:
```
32811us
2755070us
2822us
-------
51900us
1964340us
10060us
```

NEW:
```
17125us
1077772us
2891us
-------
90098us
877076us
32766us
```

Multiple runs give numbers in the same range.

Difference Benchmark 1:
- 1.92
- 2.56
- 0.98

Difference Benchmark 2:
- 0.58
- 2.24
- 0.31

The difference represents the fraction of *old* divided by *new*, i.e., the relative change to the existing code base.

### Interpretation

The new change increases the throughput up to a factor of 2.5, at a slight setup cost. As soon as the system pumps a reasonable amount of larger data through the pipeline, this overhead is quickly amortized.

However, the "construction overhead" represents a worst-case scenario. A real-world scenario does not construct a serializer/deserializer in a loop of 100K iterations.

Overall, I conclude that the huge performance gains of abstraction and throughput outweigh the somewhat more expensive setup costs.

mavam:
(@Neverlord I've assigned you on this one to indicate that you're the one who controls the merge.)

Neverlord:
I think the new interface is quite clumsy and repetitive on the caller side. I would prefer an approach where the caller side doesn't change. Instead of inverting the order on the ctor, you could as well introduce a new constructor that takes a single argument. It's one extra ctor, but it keeps the interface stable.

Since CAF will always use the same buffer types with `binary_serializer` and `binary_deserializer`, I would prefer having the buffer simply as member (again, to keep the interface as-is). This would allow us to keep the caller-side stable and get around the tedious "make a buffer from the vector, then initialize the (de)serializer".

You could template the current `binary_serializer` implementation as `stream_serializer<T>` to configure what member it has and then `using binary_serializer = stream_serializer<vectorbuf<char>>`. Users that wish to store the streambuf separately can simply have `using my_serializer = stream_serializer<std::streambuf&>` instead.

Looking at your (pretty good!) evaluation, I think we should consider storing the serializers as members in the long run instead of the current ad-hoc approach to safe us the ctor/dtor overhead.

mavam:
> I think the new interface is quite clumsy and repetitive on the caller side.

Could you elaborate on the clumsiness? Is it the 2-step initialization? I find it actually more elegant because it decouples the source/sink aspect from the rest of the serializer initialization. In my opinion, the previously hard-coded ctors made the serializers very inflexible. Additionally, Boost.Serialization's binary serializers also take a `std::streambuf` as ctor argument, which seemed to be an important consideration in the overall framework design. I do agree that it's more code and thus a bit more verbose, if that's what you mean. Conversely, it's more explicit. Since this is targeting 0.15, it seemed a reasonable design choice to me.

> Since CAF will always use the same buffer types with `binary_serializer` and `binary_deserializer`, I would prefer having the buffer simply as member (again, to keep the interface as-is). This would allow us to keep the caller-side stable and get around the tedious "make a buffer from the vector, then initialize the (de)serializer".

I was designing the interface such that other applications (in particular, VAST and Broker) can directly using the means provided by CAF. I have various other streambuffers that I'd like to plug in the binary serializers. Hence the decoupling.

>You could template the current binary_serializer implementation as stream_serializer<T> to configure what member it has and then using binary_serializer = stream_serializer<vectorbuf<char>>. Users that wish to store the streambuf separately can simply have using my_serializer = stream_serializer<std::streambuf&> instead.

I'm a bit confused here. First, I'm surprised that *you* are advocating for a template-based design where it's actually not necessary :wink:. Second, `std::streambuf` is a virtual class, so the serializer only needs a reference to the base class. But since the instantiation paths are known at compile-time, the compiler can apply devirtualization optimizations. Consequently, we can keep the serializers generic without resorting to templates.

> [..] I think we should consider storing the serializers as members in the long run instead of the current ad-hoc approach to safe us the ctor/dtor overhead.

Agreed.

Another aspect: I am still thinking about a `char` version for `vectorbuf<char>`. It would make sense to have one such that CAF's common case of `vector<char>` serialization is symmetric. Currently, there's an `arraybuf<T>` with an alias called `charbuf = arraybuf<char>`. I want the same for the streambuffer `vectorbuf<T>` which has an unbounded put area. Some suggestions:

- `vectorbuf = containerbuf<std::vector<char>>>
- `uvectorbuf = containerbuf<std::vector<uint8_t>>>` (not necessarily in CAF)

Or just `bufferbuf = containerbuf<std::vector<char>>>`, since I see `buffer = std::vector<char>` in the BASP implementation.

Neverlord:
> Is it the 2-step initialization?

Yes, I call an interface where you are always *forced* to do two steps instead of one clumsy. It leads to code repetition if you simply don't need the flexibility.

Going through the code base of CAF (the PR currently changes 16 files) and apply this two-step phase to each and every occurrence of `binary_serializer` and `binary_deserializer` is tedious, hard to refactor, and ultimately unnecessary. Because of the changes to the caller side, your patch already has conflicts to the actor-system branch although you only change internals that should be opaque to the caller side. If CAF *always* takes these two initialization steps, then there should be an abstraction in place tailored for this use case.

> I was designing the interface such that other applications (in particular, VAST and Broker) can directly using the means provided by CAF. I have various other streambuffers that I'd like to plug in the binary serializers. Hence the decoupling.

I'm not saying the design choice is bad or unreasonable. I would just like to see a streamlined version that fits the use case of CAF, where you don't need this flexibility and consequently don't want to have the repetitive initialization steps.

By adding the proposed `stream_serializer`, you could give users a choice: reference member or value member? Picking the first one gives you this two-step initialization and allows you to manage your stream manually (which does have its merit for other use cases, don't get me wrong). Picking the second one moves the stream into the serializer and you can initialize the whole thing with a vector, i.e., the use case of CAF. You could as well pick any other approach, as long as it is reasonable to maintain and keeps the current interface for `binary_serializer` and `binary_deserializer` stable. The `stream_serializer` template was merely an idea how to get the best of both worlds with minimal effort.

> I'm surprised that you are advocating for a template-based design where it's actually not necessary.

The entire template stuff in this PR is utterly useless for CAF itself. ;)

However, I see that it has value when opening up the internals of our frameworks for third-parties like VAST and Bro. We should still be able to re-build the existing interface on top of the new internals, though.

mavam:
"Green" light on my end (minus the known `message_lifetime` failures), PR ready to merge after fixing annoying libstdc++ bugs.

Neverlord:
Great work, as usual. 👍

PR #457, opened 2016-04-21
add an atom-decoder script

nq-ebaratte:
This is a script I use when debugging caf programs; it may be useful for others. This is a direct port of `std::string to_string(const atom_value& what)` (string_serialization.cpp:586)

Typical use case: when analysing a core dump, `atom` values are printed as a raw (meaningless) integer. It proves useful to be able to quickly decode them to something more meaningul.

### DIFF
--- a/scripts/atom.py
+++ b/scripts/atom.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import sys
+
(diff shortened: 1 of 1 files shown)


Neverlord:
Indeed I can see this being useful to others. Thanks a lot for sharing. :)

PR #471, opened 2016-06-14
Fix function view

ufownl:
Fix the compilation errors when the result of `function_view` is an
actor handle.


Neverlord:
Thanks a lot! 👍

PR #475, opened 2016-06-19
Disable the invalid error report

ufownl:
The message below will be printed in the console when calling `delayed_send` if the log is enabled.
```
[ERROR] CAF_PUSH_AID called in file
/home/ubuntu/Softwares/actor-framework/libcaf_core/src/local_actor.cpp
on line 726 without thread-local logger pointer
```


Neverlord:
👍

PR #476, opened 2016-06-20
Logging enhancements

mavam:
This change set adds a new column to the log file that represents the *component*. It is a static variable that users can change by defining `CAF_LOG_COMPONENT` prior to including `caf/logger.hpp`. If not defined, the logger will use the default component `caf`.

Moreover, one can now log any type that provides `operator<<` compatible with a `std::ostream`. The semantics are the same as with the arg wrappers: when using `<<` to stream into the log, elements are white-space separated.


Neverlord:
I'll merge in the next couple of days as soon as Jenkins gives green light for the `develop` branch.

Issue #478, opened 2016-06-21
Support user-defined magic cookies


Neverlord:
Add a new config parameter to the middleman to support magic cookies. This allows applications to fail during handshake, e.g., to prevent errors due to incompatible versions.

ufownl:
It means the magic number?

Neverlord:
No, maybe "app identifier" is a better name. It is simply a user-defined string that is used to authenticate nodes during connect. During handshake, each node sends the CAF magic number, the BASP version, and (new) the app identifier / cookie. If any of those mismatches, connection is refused. This allows users to make sure incompatible versions of an application fail to connect, rather than failing later with some obscure error.

PR #479, opened 2016-06-22
Add app identifier config parameter

ufownl:
Relate #478.

### DIFF
--- a/examples/caf-application.ini
+++ b/examples/caf-application.ini
@@ -39,6 +39,8 @@ relaxed-sleep-duration=10000
 enable-automatic-connections=false
 ; accepted alternative: 'asio' (only when compiling CAF with ASIO)
 network-backend='default'
+; application identifier of this node
+app-identifier=''
 ; maximum number of consecutive I/O reads per broker
 max-consecutive-reads=50
 ; heartbeat message interval in ms (0 disables heartbeating)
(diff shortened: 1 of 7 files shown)


Neverlord:
Wow, thanks for the effort. Looks solid, too. I only have two minor requests:

* the identifier should be a string instead with empty-string as default
  + default strings have 4 bytes on the wire (size with value 0 as `uint32_t`)
  + the `atom` is simply too limited in this case
  + I think there's no reason to forbid names like "my_app_v1.0"
  + only serialized during handshake, so no performance implications anyways
* `basp::instance` does not need the `app_identifier_` field
  + you can access the config directly: `callee_.system().config(). middleman_app_identifier`
  + `basp_broker` should not initialize instance state in `make_behavior`

ufownl:
Tweaks done. And I found that the size of a string in the payload has variant length, the size with value 0 only takes 1 byte.

examples/caf-application.ini
```diff
@@ -39,6 +39,8 @@ relaxed-sleep-duration=10000
 enable-automatic-connections=false
 ; accepted alternative: 'asio' (only when compiling CAF with ASIO)
 network-backend='default'
+; application identifier of this node
+app-identifier=''
```
Neverlord: `''` is parsed as atom, an empty string is `""` (will fix on my end, just FYI).

Neverlord:
Thank you once more. You rock! :)

Issue #488, opened 2016-07-13
CAF_LOG_LEVEL not handled correctly


alexeiz:
As of commit 5e68580, setting CAF_LOG_LEVEL to any number (>0 I suppose) enables all levels.  For example, I configure and compile with CAF_LOG_LEVEL=1 which should enable ERROR and WARNING log levels.  But instead, in addition to ERROR and WARNING I also get logging for INFO, DEBUG and TRACE.

PR #489, opened 2016-07-13
Respect CAF_LOG_LEVEL

alexeiz:
This should fix issue #488 

Wrt logger start/stop log entries, I think INFO is a more appropriate log level there instead of TRACE.


Neverlord:
Thanks for contributing to CAF. You rock! :)

PR #490, opened 2016-07-13
Add flare actor

mavam:
This change set introduces a "flare", a UNIX pipe that exposes a file descriptor to signal readiness, as well as a "flare actor", which is a blocking actor that exposes the flare's descriptor for testing mailbox emptiness. This enables use of blocking actors in combination with existing descriptor-based event loops, e.g.,:

```cpp
auto fd = get_descriptor_from_flare_actor();
pollfd p = {fd, POLLIN, {}};
auto n = ::poll(&p, 1, -1); // block until a message arrives
if (n == 1)
  assert(p.revents & POLLIN);
```

### DIFF
--- a/libcaf_io/CMakeLists.txt
+++ b/libcaf_io/CMakeLists.txt
@@ -12,6 +12,8 @@ set (LIBCAF_IO_SRCS
      src/broker.cpp
      src/default_multiplexer.cpp
      src/doorman.cpp
+     src/flare.cpp
+     src/flare_actor.cpp
      src/middleman.cpp
      src/middleman_actor.cpp
      src/hook.cpp
--- a/libcaf_io/caf/io/detail/flare.hpp
+++ b/libcaf_io/caf/io/detail/flare.hpp
@@ -1,0 +1,66 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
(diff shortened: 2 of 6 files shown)


Neverlord:
This does not belong into the core library. I might be convinced to accept it as part of `libcaf_io` if it compiles on all platforms (including Windows).

mavam:
I've moved the code into `caf::io::detail`. Jenkins is happy on all UNIX platforms.

I don't have the cycles to port the polling to Windows. Since we're keeping it in namespace `detail`, I would have thought that's fine. If that's a show-stopper, feel free to reject the PR.

Neverlord:
Adding this to `detail` makes no sense. The `detail` namespace contains implementation details for CAF. I don't see a use case for flare actors in CAF, so it only makes sense in the public API. And I'm sorry to say, not compiling on all platforms is indeed a showstopper.

All in all I think pushing this to CAF at the moment is rushed. Adding a flare actor is generally a good idea, but only with proper documentation and tutorial how to integrate those actors in custom event loops. These actors also would need some alternative/overload to `spawn` that also returns the socket descriptor. Currently the only possible way to get the socket is to cast the actor to a pointer and then upcast the pointer. That is not a good user-facing API.

There's currently too much to do on my end to get this in good shape. I believe the same is true for you. So I would say we delay this until we have a few cycles to spare.

mavam:
> I believe the same is true for you. So I would say we delay this until we have a few cycles to spare.

Yes.

I like the descriptor-based spawn overload you describe. Let's keep it in mind for the future.

Issue #498, opened 2016-08-05
link_to() is broken for remote actors


CrazyIvan32:
When trying to use link_to on an actor obtained via remote_actor() I get the following message on remote actor side:
`*** unexpected message [id: 7, name: scheduled_actor]: ('link', actor(9, node_id(9428, BCF04264617F1058A5E67BF9DEDCDBADF934D400)))`
although it works for local actors

PR #499, opened 2016-08-07
Fix `link_to()` for remote actors

ufownl:
Relates #498.


CrazyIvan32:
It works somehow but not the way it should be.
Local actor can call link_to() with remote actor and remote actor does receive exit_msg when local actor quits but it receives it twice and doesn't quit as it should (maybe because it's published?). Local actor receives no exit_msg messages.

Neverlord:
The patch also tries to fix this at the wrong layer. Link messages are supposed to be handled by the BASP broker. I'm working on a fix.

PR #500, opened 2016-08-11
Fix broker quit flow

ufownl:
Fix the bug that broker can not exit normally when calling quit in
the `connection_closed_msg` handler.


Neverlord:
Thanks a lot! :)

PR #503, opened 2016-08-31
Fix compile error

ufownl:
The compile error could be reproduced using the code fragment below:
```C++
using foo_handle = caf::typed_actor<caf::reacts_to<std::string>>;

using bar_handle = caf::typed_actor<caf::reacts_to<int>>;

foo_handle::behavior_type foo(foo_handle::pointer self) {
  return {
    [=](const std::string& str) {
      caf::aout(self) << str << std::endl;
      self->quit();
    }
  };
}

bar_handle::behavior_type bar(bar_handle::pointer self) {
  auto f = self->spawn<caf::linked>(foo);
  self->send(f, "Hello, world!");
  return {
    [=](int) {
      // nop
    }
  };
}
```


Neverlord:
Thanks a lot! 👍

PR #518, opened 2016-10-14
Port to Cygwin

sourcedelica:
Added Cygwin support for #517 

I tested with `./configure` and no options.

There is one TODO so far - in `profiled_coordinator.hpp` I hacked around the fact that Cygwin doesn't have an `RUSAGE_THREAD`.  I haven't looked in detail into what that class is doing and how best to handle the Cygwin situation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.4)
 project(caf C CXX)
 
 # be nice to vim users
@@ -266,7 +266,7 @@ endif(CAF_ENABLE_ADDRESS_SANITIZER)
 if(NOT APPLE AND NOT WIN32)
   set(EXTRA_FLAGS "${EXTRA_FLAGS} -pthread")
 endif()
-# -fPIC generates warnings on MinGW plus extra setup steps needed on MinGW
+# -fPIC generates warnings on MinGW and Cygwin plus extra setup steps needed on MinGW
 if(MINGW)
   add_definitions(-D_WIN32_WINNT=0x0600)
   add_definitions(-DWIN32)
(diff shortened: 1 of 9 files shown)


sourcedelica:
Even though I ran the tests with shared libraries, the Cygwin shared libraries are DLLs and may (probably?) have the same limitations as you found with Windows shared libraries.  I don't know the specifics about the issues you ran into and I'm also not very knowledgeable about Windows development unfortunately.  I can disable the shared libs for Cygwin if you want.  

Side note: I'm hoping to learn more about Windows development - I installed VS 2015 and I'm going to try to build CAF and ultimately get it into the Conan recipe.  I'm actually a fairly new C++ developer in general.  The CAF codebase is very sophisticated and I'm learning a lot from it.  I did a ton of C/Unix in the 90s, and some C++, but have been doing mostly JVM stuff in the last decade, including a number of Akka projects.

CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 2.8)
+set(CMAKE_LEGACY_CYGWIN_WIN32 0) # Remove when CMake >= 2.8.4 is required
```
Neverlord: I don't mind bumping minimal required CMake version to 2.8.4 instead (which is 5 years old at this point).

libcaf_core/CMakeLists.txt
```diff
@@ -108,7 +108,9 @@ if (NOT CAF_BUILD_STATIC_ONLY)
                         SOVERSION ${CAF_VERSION}
                         VERSION ${CAF_VERSION}
                         OUTPUT_NAME caf_core)
-  if(NOT WIN32)
+  if (CYGWIN)
+    install(TARGETS libcaf_core_shared RUNTIME DESTINATION bin)
```
Neverlord: What is the reason to install libraries into bin?

sourcedelica:
In Windows they don't have a `LD_LIBRARY_PATH`, just `PATH`.  So DLLs are
installed into a `bin` dir that's in the path.  In the case of Cygwin its
`/usr/local/bin`.  I checked Protobuf and it does the same thing.

I'll bump to 2.8.4 and get rid of that `set(CMAKE_LEGACY_CYGWIN_WIN32 0)`.
I agree it's very annoying.


-- Eric

sourcedelica:
CMake version changed to 2.8.4 and I removed that `CMAKE_LEGACY_CYGWIN_WIN32` flag.

sourcedelica:
One thing that isn't working with Cygwin is find_package(CAF).

Neverlord:
LGTM. Thanks for contributing to CAF. You rock! :)

> One thing that isn't working with Cygwin is find_package(CAF).

It's probably not working with the MSVC build either. No reason to hold this PR back.

sourcedelica:
Thanks!  My pleasure.


-- Eric

PR #523, opened 2016-11-07
Make logger more configurable

mavam:
With this patch, users can change the following aspects of the logger via `actor_system_config`:

1. Set the name of the logger log file (`--caf#logger.filename=/path/to/file`)
2. Adjust the logger verbosity (`--caf#logger.verbosity=1`)
3. Redirect message to `std::clog`  (`--caf#logger.console=true`)
4. Colorize output according to severity level (ignored on Windows) (`--caf#logger.colorize=true`)

### DIFF
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -252,6 +252,13 @@ public:
   size_t work_stealing_relaxed_steal_interval;
   size_t work_stealing_relaxed_sleep_duration_us;
 
+  // -- config parameters for the logger ---------------------------------------
+
+  std::string logger_filename;
+  std::string logger_verbosity;
+  bool logger_console;
+  bool logger_colorize;
+
   // -- config parameters of the middleman -------------------------------------
 
   atom_value middleman_network_backend;
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -255,9 +255,8 @@ public:
   // -- config parameters for the logger ---------------------------------------
 
   std::string logger_filename;
-  std::string logger_verbosity;
-  bool logger_console;
-  bool logger_colorize;
+  atom_value logger_verbosity;
+  atom_value logger_console;
 
   // -- config parameters of the middleman -------------------------------------
 
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -257,6 +257,7 @@ public:
   std::string logger_filename;
   atom_value logger_verbosity;
   atom_value logger_console;
+  std::string logger_filter;
 
   // -- config parameters of the middleman -------------------------------------
 
--- a/libcaf_core/caf/logger.hpp
+++ b/libcaf_core/caf/logger.hpp
@@ -190,18 +190,6 @@ private:
 #define CAF_LOG_LEVEL_DEBUG 3
 #define CAF_LOG_LEVEL_TRACE 4
 
-#define CAF_PRINT_ERROR_IMPL(nclass, nfun, message)                            \
-  do {                                                                         \
-    caf::logger::line_builder lb;                                              \
-    lb << message;                                                             \
-    if (nclass.empty())                                                        \
-      printf("[ERROR] in %s:%d %s::%s: %s\n", __FILE__, __LINE__,              \
-             nclass.c_str(), nfun, lb.get().c_str());                          \
-    else                                                                       \
-      printf("[ERROR] in %s:%d %s: %s\n", __FILE__, __LINE__,                  \
-             nfun, lb.get().c_str());                      \
-  } while (false)
-
 #define CAF_ARG(argument) caf::logger::make_arg_wrapper(#argument, argument)
 
 #ifdef CAF_MSVC
@@ -289,6 +277,8 @@ inline caf::actor_id caf_set_aid_dummy() { return 0; }
 #  define CAF_LOG_WARNING(output) CAF_LOG_IMPL(CAF_LOG_LEVEL_WARNING, output)
 #endif
 
+#define CAF_LOG_ERROR(output) CAF_LOG_IMPL(CAF_LOG_LEVEL_ERROR, output)
+
 #endif // CAF_LOG_LEVEL
 
 #ifndef CAF_LOG_INFO
@@ -303,11 +293,9 @@ inline caf::actor_id caf_set_aid_dummy() { return 0; }
 #  define CAF_LOG_WARNING(output) CAF_VOID_STMT
 #endif
 
-#define CAF_LOG_ERROR(output)                                                  \
-  do {                                                                         \
-    CAF_PRINT_ERROR_IMPL(CAF_GET_CLASS_NAME, __func__, output);                \
-    CAF_LOG_IMPL(CAF_LOG_LEVEL_ERROR, output);                                 \
-  } while (false)
+#ifndef CAF_LOG_ERROR
+#  define CAF_LOG_ERROR(output) CAF_VOID_STMT
+#endif
 
 #ifdef CAF_LOG_LEVEL
 
--- a/libcaf_core/caf/logger.hpp
+++ b/libcaf_core/caf/logger.hpp
@@ -64,8 +64,10 @@ public:
   struct event {
     event* next;
     event* prev;
+    int level;
+    std::string prefix;
     std::string msg;
-    explicit event(std::string log_message = "");
+    explicit event(int l = 0, std::string p = "", std::string m = "");
   };
 
   template <class T>
@@ -167,6 +169,7 @@ private:
   void stop();
 
   actor_system& system_;
+  int level_;
   detail::shared_spinlock aids_lock_;
   std::unordered_map<std::thread::id, actor_id> aids_;
   std::thread thread_;
--- a/libcaf_core/caf/logger.hpp
+++ b/libcaf_core/caf/logger.hpp
@@ -65,9 +65,11 @@ public:
     event* next;
     event* prev;
     int level;
+    const char* component;
     std::string prefix;
     std::string msg;
-    explicit event(int l = 0, std::string p = "", std::string m = "");
+    explicit event(int l = 0, char const* c = "", std::string p = "",
+                   std::string m = "");
   };
 
   template <class T>
--- a/libcaf_core/caf/logger.hpp
+++ b/libcaf_core/caf/logger.hpp
@@ -17,8 +17,8 @@
  * http://www.boost.org/LICENSE_1_0.txt.                                      *
  ******************************************************************************/
 
-#ifndef CAF_DETAIL_LOGGING_HPP
-#define CAF_DETAIL_LOGGING_HPP
+#ifndef CAF_LOGGER_HPP
+#define CAF_LOGGER_HPP
 
 #include <thread>
 #include <cstring>
@@ -320,4 +320,4 @@ inline caf::actor_id caf_set_aid_dummy() { return 0; }
 
 #endif // CAF_LOG_LEVEL
 
-#endif // CAF_DETAIL_LOGGING_HPP
+#endif // CAF_LOGGER_HPP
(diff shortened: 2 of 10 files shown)


mavam:

Neverlord:
Something's wrong with the GitHub feedback. Jenkins itself works fine: https://jenkins.inet.haw-hamburg.de/job/CAF%20Contributions/job/pull%20requests/318/

mavam:
I'd also like the to give the user complete control over the output CAF produces. The logger prints errors unconditionally. I'm proposing to let the user decide what to log, no exception for errors:
```diff
diff --git a/libcaf_core/caf/logger.hpp b/libcaf_core/caf/logger.hpp
index ce0b235..5a2c6ff 100644
--- a/libcaf_core/caf/logger.hpp
+++ b/libcaf_core/caf/logger.hpp
@@ -190,18 +190,6 @@ private:
 #define CAF_LOG_LEVEL_DEBUG 3
 #define CAF_LOG_LEVEL_TRACE 4
 
-#define CAF_PRINT_ERROR_IMPL(nclass, nfun, message)                            \
-  do {                                                                         \
-    caf::logger::line_builder lb;                                              \
-    lb << message;                                                             \
-    if (nclass.empty())                                                        \
-      printf("[ERROR] in %s:%d %s::%s: %s\n", __FILE__, __LINE__,              \
-             nclass.c_str(), nfun, lb.get().c_str());                          \
-    else                                                                       \
-      printf("[ERROR] in %s:%d %s: %s\n", __FILE__, __LINE__,                  \
-             nfun, lb.get().c_str());                      \
-  } while (false)
-
 #define CAF_ARG(argument) caf::logger::make_arg_wrapper(#argument, argument)
 
 #ifdef CAF_MSVC
@@ -289,6 +277,8 @@ inline caf::actor_id caf_set_aid_dummy() { return 0; }
 #  define CAF_LOG_WARNING(output) CAF_LOG_IMPL(CAF_LOG_LEVEL_WARNING, output)
 #endif
 
+#define CAF_LOG_ERROR(output) CAF_LOG_IMPL(CAF_LOG_LEVEL_ERROR, output);
+
 #endif // CAF_LOG_LEVEL
 
 #ifndef CAF_LOG_INFO
@@ -303,11 +293,9 @@ inline caf::actor_id caf_set_aid_dummy() { return 0; }
 #  define CAF_LOG_WARNING(output) CAF_VOID_STMT
 #endif
 
-#define CAF_LOG_ERROR(output)                                                  \
-  do {                                                                         \
-    CAF_PRINT_ERROR_IMPL(CAF_GET_CLASS_NAME, __func__, output);                \
-    CAF_LOG_IMPL(CAF_LOG_LEVEL_ERROR, output);                                 \
-  } while (false)
+#ifndef CAF_LOG_ERROR
+#  define CAF_LOG_ERROR(output) CAF_VOID_STMT
+#endif
 
 #ifdef CAF_LOG_LEVEL
```

libcaf_core/caf/actor_system_config.hpp
```diff
@@ -252,6 +252,13 @@ class actor_system_config {
   size_t work_stealing_relaxed_steal_interval;
   size_t work_stealing_relaxed_sleep_duration_us;
 
+  // -- config parameters for the logger ---------------------------------------
+
+  std::string logger_filename;
+  std::string logger_verbosity;
```
Neverlord: A `string` for one of 6 possible options feels quite overkill to me. Why not use an `atom` here? Atoms also clean up the `if-else` trees by allowing `switch` instead.

libcaf_core/caf/actor_system_config.hpp
```diff
@@ -252,6 +252,13 @@ class actor_system_config {
   size_t work_stealing_relaxed_steal_interval;
   size_t work_stealing_relaxed_sleep_duration_us;
 
+  // -- config parameters for the logger ---------------------------------------
+
+  std::string logger_filename;
```
Neverlord: Why support only `[NODE]` as dynamic placeholder? Seems inconsequential to me. We could instead give users more fine-grained control over the filename and set the default to `actor_log_[PID]_[TIMESTAMP]_[NODE].log`.

libcaf_core/caf/actor_system_config.hpp
```diff
@@ -252,6 +252,13 @@ class actor_system_config {
...
+  std::string logger_verbosity;
+  bool logger_console;
```
Neverlord: These two bools team up to model three possible states, which seems a bit fiddly to me. We could again use atoms: `console = 'NONE' | 'COLORED' | 'UNCOLORED'`, whereas `'NONE'` is the default.

libcaf_core/src/logger.cpp
```diff
@@ -55,12 +54,35 @@ constexpr const char* log_level_name[] = {
   "TRACE"
 };
 
+#ifndef CAF_MSVC
```
Neverlord: This code is copy-pasted from caf_test. Please wrap this in a header (`libcaf_core/caf/color.hpp`) and remove the dups. Also, I think we should simply set all colors to `""` on MSVC and get rid of all other `#ifdef CAF_MSVC`. Paying two `<< ""` on Windows isn't a big deal and is easily avoided by setting the terminal to `UNCOLORED`.

libcaf_core/src/logger.cpp
```diff
@@ -223,7 +247,9 @@ void logger::log(int level, const char* component,
                  const std::string& class_name, const char* function_name,
                  const char* c_full_file_name, int line_num,
                  const std::string& msg) {
-  CAF_ASSERT(level >= 0 && level <= 4);
+  CAF_ASSERT(level <= 4);
```
Neverlord: Why does the assert no longer check for negative numbers?

libcaf_core/src/logger.cpp
```diff
@@ -273,11 +300,15 @@ logger::logger(actor_system& sys) : system_(sys) {
...
 
 void logger::run() {
-  std::ostringstream fname;
-  fname << "actor_log_" << detail::get_process_id() << "_" << time(0)
-        << "_" << to_string(system_.node())
-        << ".log";
-  std::fstream out(fname.str().c_str(), std::ios::out | std::ios::app);
+  auto filename = system_.config().logger_filename;
+  // Replace node ID placeholder.
+  auto placeholder = std::string{"[NODE]"};
```
Neverlord: Creating a heap-stored string here is unnecessary.

```cpp
const char needle[] = "[NODE]";
auto i = std::search(filename.begin(), filename.end(), std::begin(needle), std::end(needle) - 1);
```

-1 because the char array is null-terminated.

libcaf_core/src/logger.cpp
```diff
@@ -273,11 +300,15 @@ logger::logger(actor_system& sys) : system_(sys) {
...
+  auto i = filename.find(placeholder);
+  if (i != std::string::npos) {
+    auto nid = to_string(system_.node());
+    filename.replace(i, placeholder.size(), nid);
```
Neverlord: Accordingly, `filename.replace(i, i + ((sizeof(needle) - 1), nid);`.

libcaf_core/src/logger.cpp
```diff
@@ -289,25 +320,71 @@ void logger::run() {
...
-#if defined(CAF_LOG_LEVEL) && CAF_LOG_LEVEL >= CAF_LOG_LEVEL_INFO
-  log(CAF_LOG_LEVEL_INFO, "caf", "caf::logger", "run", __FILE__, __LINE__, "EXIT");
+#if defined(CAF_LOG_LEVEL)
+  if (!thread_.joinable())
+    return;
+  log(CAF_LOG_LEVEL_INFO, "caf", "caf::logger", "run", __FILE__, __LINE__,
+      "EXIT");
   // an empty string means: shut down
-  queue_.synchronized_enqueue(queue_mtx_, queue_cv_, new event{""});
+  queue_.synchronized_enqueue(queue_mtx_, queue_cv_, new event{});
```
Neverlord: If you already drop `""`, you can as well drop the remaining `{}`.

libcaf_core/caf/logger.hpp
```diff
@@ -64,8 +64,10 @@ class logger {
   struct event {
     event* next;
     event* prev;
+    int level;
+    std::string prefix;
```
Neverlord: Where does this prefix come from? What is the purpose of this? I don't see it used anywhere. All it currently does is padding each single log entry with a whitespace.

libcaf_core/caf/actor_system_config.hpp
mavam: Cool, didn't know you can switch on these things. Will change.

libcaf_core/caf/actor_system_config.hpp
mavam: ✅

libcaf_core/src/logger.cpp
mavam: Okidoke.

libcaf_core/src/logger.cpp
mavam: That was a left-over. I temporarily switch the level to `size_t`. Will undo that change.

libcaf_core/src/logger.cpp
mavam: It's a short string, and with SSO common these days. It's also only payed during logger startup, so far away from the critical path. I'm happy to polish it, though.

libcaf_core/src/logger.cpp
mavam: Fair point.

libcaf_core/caf/logger.hpp
mavam: The prefix is all the logger-generated goop in front of the actual message. Only the log file contains this verbose representation. The console, which is meant for human consumption, merely contains the message.

For the console, I'm inclined to add the log facility/category filter, because often users of the CAF logging framework only want to see their own logs.

libcaf_core/caf/actor_system_config.hpp
mavam: Can you give an example on how you see the switch working on an `atom_value`? The obvious one doesn't work:
```cpp
    switch (system_.config().logger_console) {
      case atom("NONE"):
...
```

mavam:
I've added one more feature, which I deem important for users (in particular Broker and VAST). The ability to filter based on log category on the console. For example, in VAST I would configure the command line application as follows:

```cpp
  struct configuration : caf::actor_system_config {
    configuration() {
      logger_filter = "vast";
      logger_filename = "/var/log/vast.log";
      logger_console = caf::atom("COLORED");
    }
  };
```
Now the console can be used to display application output from actors. Example output from VAST's unit tests:
```
task#8 notifies actor#7 about task completion
task#8 notifies actor#7 about task completion
task#8 registers actor actor#9
task#8 registers actor actor#10
task#8 manually completed actor with address 10
task#8 completed actor#9
task#8 notifies actor#7 on task status change
task#8 notifies actor#7 about task completion
task#8 registers actor actor#9
task#8 registers actor actor#10
task#8 registers actor actor#11
task#11 registers actor actor#12
task#11 registers actor actor#13
task#11 registers actor actor#14
task#11 completed actor#14
task#11 completed actor#13
task#11 completed actor#12
task#8 completed actor#9
task#8 completed actor#10
```

Without this feature, I would be incredibly tedious to come up with meaningful console output for a CAF-enabled application, because the console itself must be passed around to all actors that may log some details.

libcaf_core/caf/actor_system_config.hpp
Neverlord: You need to convert the atoms back to `uint64_t`. I've pushed a bit syntactic sugar for this with d729bd2af8b506eae1e9ddefe32c3d87307b9ad6.

```cpp
switch (to_u64(...)) {
  case to_u64(atom("NONE")):
    ...
}
```

Neverlord: Nevermind, turns out some compiler don't think `constexpr` functions qualify for constant expressions. 😒

Here's a workaround with `atom_constants`: https://github.com/actor-framework/actor-framework/blob/master/libcaf_io/src/basp_broker.cpp#L239

PR #535, opened 2017-02-20
fix version extract.

gaoxiaojun:
example:
    #define CAF2_VERSION 1405
    VERSION_INT = 2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ file(READ "libcaf_core/caf/config.hpp" CONFIG_HPP)
 # get line containing the version
 string(REGEX MATCH "#define CAF_VERSION [0-9]+" VERSION_LINE "${CONFIG_HPP}")
 # extract version number from line
-string(REGEX MATCH "[0-9]+" VERSION_INT "${VERSION_LINE}")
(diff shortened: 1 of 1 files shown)


Neverlord:
I try to wrap my head around what you are trying to fix here. The version extract works fine, as documented in the Build Summary of each CMake run. Your change would only come into effect if we were to change the name of the version macro name. Frankly, `CAF2_VERSION` does not make any sense to me as changing the version macro name (let alone start numbering it) defeats its sole purpose.

Furthermore, even if we were to introduce a `CAF2_VERSION`: this line would simply be ignored by the CMake script. So there is no potential for ambiguity. Can you add context to as to why you needed this patch in CAF?

gaoxiaojun:
I start a new project yesterday, copy & paste some lines from CMakeLists.txt, replace CAF_ to project name "H9",  in my case,
 version.h:
"#define H9_VERSION 100"
CMakeLists.txt:
 string(REGEX MATCH "#define H9_VERSION [0-9] VERSION_LINE "${VERSION_H}")
# extract version number from line
string(REGEX MATCH " [0-9]+" VERSION_INT "${VERSION_LINE}")
and found the version is always "0.0.9",so i pull this request to make script more robust.

PR #543, opened 2017-03-27
Scoped_actor loses messages after skipping them

aufgang001:
A scoped_actor seems to lose messages under certain circumstances.

For example: 
Messages 3, 1, 2 and then 0 are send to a scoped_actor and it is necessary to receive them in order 0, 1, 2, 3. All messages, which don't match the receive order are skipped for later processing.
In this example the scoped_actor loses message 2 and 3.

I added a unit test to identify the problem and proposed a solution.

### DIFF
--- a/libcaf_core/src/blocking_actor.cpp
+++ b/libcaf_core/src/blocking_actor.cpp
@@ -396,13 +396,16 @@ void blocking_actor::receive_impl(receive_cond& rcc,
         }
       }
     } while (skipped && !timed_out);
-    if (timed_out)
+    if (timed_out) {
       bhvr.handle_timeout();
-    else
-      seq.erase_and_advance();
-    // check loop post condition
-    if (!rcc.post())
-      return;
+      if (!rcc.post())
+        return;
+    } else {
+      if (rcc.post())
+        seq.erase_and_advance();
+      else
+        return;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


Neverlord:
It seems to me that you didn't unterstand what the `marked` flag does. This flag allows an actor to nest `receive` statements, i.e., start traversing the cache/mailbox while another traversal (or multiple, in fact) are already underway.

You didn't actually fix a bug. You simply traded one for another. By removing `marked` from the mailbox element you are no longer able to detect nested `receive` statements. This will cause undefined behavior in any actor that nests `receive` blocks that aren't mutually exclusive.

Without the flag, one would need to remove elements from the cache in order to prevent sub-functions to re-iterate it again. However, there's is no feasible way to determine where to put it back, because predecessor and successor might get removed from nested functions. The flag exists for a reason.

libcaf_core/src/blocking_actor.cpp
```diff
@@ -396,12 +398,14 @@ void blocking_actor::receive_impl(receive_cond& rcc,
...
       }
     } while (skipped && !timed_out);
-    if (timed_out)
+    bool post_condition = rcc.post();
+    if (timed_out) {
       bhvr.handle_timeout();
-    else
-      seq.erase_and_advance();
+    } else {
+      seq.erase(post_condition);
```
Neverlord: Calling `erase(false)` does nothing in both implementations. Isn't a more straightforward fix `if (!post_condition) seq.erase_and_advance();`? That wouldn't require any changes outside of `receive_impl`. Am I missing something or are you overcomplicating the solution?

PR #553, opened 2017-04-25
Add SSL/TLS support via OpenSSL

rsmmr:
I've filled in the SSL code and that all seems to work. 

Couple of questions:

1. I can't get the test in openssl_authentication_failure to work. That test makes sure SSL session fail if certificates aren't right. However, I didn't manage to catch the runtime exception thrown in that case to then let the test pass. When I try a try/catch around the code, the whole test hangs. What's the right way to handle an expected exception in a test?

2. I'd like to pass the OpenSSL error stack (essentially a string) back when one of the i/o methods fail so that the user can understand what's going on. Is there a way to do that? Right now it looks like I can only set the return value of read_some/etc. to false.

Any other feedback welcome.


libcaf_openssl/caf/openssl/ssl_session.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "caf/io/network/native_socket.hpp"
+
+#include <openssl/err.h>
+#include <openssl/ssl.h>
+
+namespace caf {
+namespace openssl {
+using native_socket = io::network::native_socket;
+
+class ssl_session {
```
Neverlord: I think the name `caf::openssl::ssl_session` is redundant. Any reason not to change the class name to simply `session`?

libcaf_openssl/caf/openssl/ssl_session.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#ifndef CAF_OPENSSL_SSL_SESSION_HPP
+#define CAF_OPENSSL_SSL_SESSION_HPP
+
+#include "caf/actor_system.hpp"
+#include "caf/io/network/native_socket.hpp"
+
+#include <openssl/err.h>
```
Neverlord: Are forward declarations sufficient here? If so, we could move `openssl` includes to the `.cpp` files instead.

libcaf_openssl/src/manager.cpp
```diff
@@ -55,7 +58,19 @@ void manager::stop() {
 
 void manager::init(actor_system_config&) {
   CAF_LOG_TRACE("");
-  // TODO: initialize OpenSSL state
+
```
Neverlord: [Vertical whitespaces separate functions and are not used inside functions: use comments to document logical blocks.](https://github.com/actor-framework/actor-framework/blob/master/CONTRIBUTING.md#general)

libcaf_openssl/src/manager.cpp
```diff
@@ -66,6 +81,13 @@ void* manager::subtype_ptr() {
   return this;
 }
 
+bool manager::authentication_enabled() {
+  auto& cfg = system().config();
+  return cfg.openssl_certificate.size() || cfg.openssl_key.size()
```
Neverlord: `cfg.openssl_certificate.size() > 0 || ...` is a tiny bit more verbose but a whole lot better to read. :)

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -48,34 +50,53 @@ using native_socket = io::network::native_socket;
 using default_mpx = io::network::default_multiplexer;
 
 struct ssl_policy {
-  /// Reads up to `len` bytes from an OpenSSL socket.
-  static bool read_some(size_t& result, native_socket fd, void* buf,
+  ssl_policy(std::shared_ptr<ssl_session> ssl) : ssl_(ssl) {
```
Neverlord: First, `ssl_(std::move(ssl))` safes two needless operations on the reference count. Second, we prefer `intrusive_ptr<...>` in CAF whenever feasible. All you need to do is to derive `session` from `ref_counted`.

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -48,34 +50,53 @@ using native_socket = io::network::native_socket;
...
-  static bool try_accept(native_socket& result, native_socket fd) {
+  bool try_accept(native_socket& result, native_socket fd) {
     CAF_LOG_TRACE(CAF_ARG(fd));
-    // TODO: implement me
-    CAF_RAISE_ERROR("unimplemented function: openssl::try_accept");
+
+    sockaddr_storage addr;
+    memset(&addr, 0, sizeof(addr));
+    socklen_t addrlen = sizeof(addr);
+    result = ::accept(fd, reinterpret_cast<sockaddr*>(&addr), &addrlen);
```
Neverlord: Is there a conflicting declaration for `accept` that makes the `::` prefix necessary?

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -48,34 +50,53 @@ using native_socket = io::network::native_socket;
...
 
 class scribe_impl : public io::scribe {
   public:
-    scribe_impl(default_mpx& mpx, native_socket sockfd)
+    scribe_impl(default_mpx& mpx, native_socket sockfd,
+                std::shared_ptr<ssl_session> ssl)
         : scribe(io::network::conn_hdl_from_socket(sockfd)),
           launched_(false),
-          stream_(mpx, sockfd) {
+          stream_(mpx, sockfd, ssl) {
```
Neverlord: `std::move(ssl)`

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -146,9 +167,11 @@ class scribe_impl : public io::scribe {
 
 class doorman_impl : public io::doorman {
 public:
-  doorman_impl(default_mpx& mx, native_socket sockfd)
+  doorman_impl(default_mpx& mx, native_socket sockfd,
+               std::shared_ptr<ssl_session> ssl)
       : doorman(io::network::accept_hdl_from_socket(sockfd)),
-        acceptor_(mx, sockfd) {
+        acceptor_(mx, sockfd, ssl),
```
Neverlord: `std::move(ssl)`

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -202,6 +226,7 @@ class doorman_impl : public io::doorman {
 
 private:
   io::network::acceptor_impl<ssl_policy> acceptor_;
+  std::shared_ptr<ssl_session> ssl_;
```
Neverlord: `intrusive_ptr<session>`

libcaf_openssl/src/middleman_actor.cpp
```diff
@@ -214,16 +239,29 @@ class middleman_actor_impl : public io::middleman_actor_impl {
 protected:
   expected<io::scribe_ptr> connect(const std::string& host,
                                    uint16_t port) override {
-    native_socket fd;
-    // TODO: implement me
-    return make_counted<scribe_impl>(mpx(), fd);
+    auto fd = io::network::new_tcp_connection(host, port);
+
+    if (!fd)
```
Neverlord: I know this usually works, but I'd still prefer `if (fd == io::network::invalid_socket)`.

libcaf_openssl/test/auth_success.cpp
```diff
@@ -0,0 +1,125 @@
...
+public:
+  config() {
+    load<io::middleman>();
+    load<openssl::manager>();
+    add_message_type<std::vector<int>>("std::vector<int>");
+    actor_system_config::parse(test::engine::argc(),
+                               test::engine::argv());
+  }
+
+  static std::string data_dir() {
```
Neverlord: Navigating the file system makes testing quite brittle. The location of the binary shouldn't matter. Neither the directory from which you are calling `caf-test`. Is there a way around this in OpenSSL? Can we give it a static string instead of files for this test or maybe use temporary files instead?

Neverlord:
Thanks for the great work!

One minor nitpick: you missed 33bc1ce while squashing/rebasing.

> However, I didn't manage to catch the runtime exception thrown in that case to then let the test pass.

Do you mean the exception thrown in the ctor of `session`? Throwing an exception in a constructor is usually a bad idea in the first place. In this case, this exception will kill the `middleman_actor`, because you are not catching the exception. Exceptions remain local in CAF, i.e., exceptions are caught by the runtime system if an actor doesn't take care of it.

Since the middleman actor is supposed to be alive for the lifetime of the process, no SSL connections can be established after hitting the first exception (because there's no middleman actor anymore that takes care connecting to remotes).

> I'd like to pass the OpenSSL error stack (essentially a string) back when one of the i/o methods fail so that the user can understand what's going on.

We could change the signature of `read_some` and `write_some` to return `error`. However, there's not really a channel to pass errors on this level back to the broker. Maybe we could extend [`connection_closed_msg`](https://github.com/actor-framework/actor-framework/blob/master/libcaf_io/caf/io/system_messages.hpp#L81) with an `error` to allow brokers to distinguish between ordinary shutdowns and errors.

libcaf_openssl/caf/openssl/ssl_session.hpp
rsmmr: Doesn't really work well, as the types are typedefs to internal structs. I've removed the err.h include though, don't need that in the header.

libcaf_openssl/caf/openssl/ssl_session.hpp
rsmmr: changed

libcaf_openssl/src/manager.cpp
rsmmr: Removed (reluctantly :-)

libcaf_openssl/src/manager.cpp
rsmmr: Done

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Done.

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Removed.

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Done.

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Done (actually not there, but for the next line)

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Done.

libcaf_openssl/src/middleman_actor.cpp
rsmmr: Done.

libcaf_openssl/test/auth_success.cpp
rsmmr: This one is tricky. I don't want add extra openssl options just for testing. I could hardcode the certs into the C code and write to tmp files first, but that doesn't sound great either. Any chance to get a well-defined way to refer to the test source directory to pull files from there? Seems something that could be more generally useful.

rsmmr:
> One minor nitpick: you missed 33bc1ce while squashing/rebasing.

Not quite sure what you mean, I merged develop in first to get the branch up to date. Let me know what you want me to do.

> However, I didn't manage to catch the runtime exception thrown in that case to then let the test pass.
> Do you mean the exception thrown in the ctor of session?

I've moved the exceptions out of the ctor, but that's not the actual problem. Take a look at openssl_authentication_failure, it's now how I would like it to work (pass if there's an exception; fail if not), but it just hangs this way.

> Maybe we could extend connection_closed_msg with an error to allow brokers to distinguish 
> between ordinary shutdowns and errors.

Yeah, something like that would be helpful. OpenSSL can fail for all kinds of reasons, and it can be hard to understand what's going on without access to the error stack.

libcaf_openssl/test/auth_success.cpp
mavam: CMake is a good fit here. Here's how I generate the necessary includes for VAST unit tests:

```cmake
set(test_data "#ifndef TEST_DATA\n#define TEST_DATA")
set(test_data "${test_data}\n\nnamespace bro {\n")
file(GLOB logs test/logs/bro/*)
foreach (log ${logs})
  get_filename_component(log_basename ${log} NAME_WE)
  set (this "constexpr auto ${log_basename} = \"${log}\";")
  set (test_data "${test_data}\n${this}")
endforeach ()
set(test_data "${test_data}\n\n} // namespace bro")
set(test_data "${test_data}\n\n#endif")

configure_file(test/data.hpp.in ${CMAKE_CURRENT_BINARY_DIR}/test/data.hpp)
```

mavam: The discussion for this item is now in #555.

rsmmr:
What's the state here now?

Neverlord:
I cleaned up the code and merged `auth_success` and `auth_failure` into a single unit test. However, after fixing the original `auth_failure` unit test the suite no longer terminates.

rsmmr:
So what does that mean? Are you looking into that, or are you assigning it back to me?

Neverlord:
I will have a look at it. I will come back to you if it turns out to be an issue with the OpenSSL code, but I don't think that's the case.

Neverlord:
@rsmmr  I took me longer because I had to fix the session management first. Sharing the session between all connections and acceptors is asking for trouble, because one SSL context only manages a single socket. Besides, the acceptor doesn't need SSL state in the first place. SSL handshaking occurs *after* a TCP connection is open. I also addressed the termination issue in the authentication failure case (that was actually a bug in CAF, not in the module).

Anyway, all tests pass now.

rsmmr:
Yeah, thanks!

PR #554, opened 2017-04-27
Allow socket creation on IPv4 only hosts

josephnoir:
Use `getaddrinfo` to discover available ip version and address info instead of direct socket creation. This (hopefully) addresses the failure to create server sockets on hosts that only have IPv4 available, [see mailing list](https://groups.google.com/d/msg/actor-framework/lcbPO4JtHkk/pLAUqmk7AQAJ).

Adding `ipv6.disable=1` to the `GRUB_CMDLINE_LINUX_DEFAULT` options in `/etc/default/grub` disabled IPv6 on my test machine ([related post](https://askubuntu.com/questions/440649/how-to-disable-ipv6-in-ubuntu-14-04), don't forget to update the config `sudo update-grub` and reboot).

### DIFF
--- a/libcaf_io/src/default_multiplexer.cpp
+++ b/libcaf_io/src/default_multiplexer.cpp
@@ -1316,10 +1316,20 @@ expected<void> set_inaddr_any(native_socket fd, sockaddr_in6& sa) {
 }
 
 template <int Family>
-expected<void> new_ip_acceptor_impl(native_socket fd, uint16_t port,
-                                        const char* addr) {
+expected<native_socket> new_ip_acceptor_impl(uint16_t port, const char* addr,
+                                             bool reuse_addr) {
   static_assert(Family == AF_INET || Family == AF_INET6, "invalid family");
   CAF_LOG_TRACE(CAF_ARG(port) << ", addr = " << (addr ? addr : "nullptr"));
+  CALL_CFUN(fd, cc_valid_socket, "socket", socket(Family, SOCK_STREAM, 0));
+  // sguard closes the socket in case of exception
+  socket_guard sguard{fd};
+  if (reuse_addr) {
+    int on = 1;
+    CALL_CFUN(tmp1, cc_zero, "setsockopt",
+              setsockopt(fd, SOL_SOCKET, SO_REUSEADDR,
+                         reinterpret_cast<setsockopt_ptr>(&on),
+                         static_cast<socklen_t>(sizeof(on))));
+  }
   using sockaddr_type =
     typename std::conditional<
       Family == AF_INET,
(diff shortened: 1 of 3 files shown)


marcoben73:
I've tested your patch and in my use case it works good.
I had to change only one line of your code (due to a compilation error)
```
- return std::make_pair(buffer, family == AF_INET ? protocol::ipv4 : protocol::ipv6);
+ return std::make_pair(std::string{buffer}, family == AF_INET ? protocol::ipv4 : protocol::ipv6);
```

Just a note. Using spawn_io_server(server, 25661) and check with netstat -lnt
Old behavior 

- system with IPv6 - process bind to :::25661 (so both IPv4 and IPv6)
- system with no IPv6 - error "could not create server socket: Address family not supported by protocol 

[errno: 97]"
New behaviour

- system with IPv6 - process bind to 0.0.0.0:25661 (this is strange, should be :::25661) 
- system with no IPv6 - process bind to 0.0.0.0:25661 
-

josephnoir:
It seems that `getaddrinfo`  returns the addresses in no specific order and in the presence of IPv6, the first address can be a v4 address as well. Hence your observation. So, instead of just returning the first one, CAF will now prefer v6 over v4, if available. My problem is that my Linux VM returns a v6 address in the list even when configured without IPv6 through GRUB.

That might be different on your system where IPv6 is not part of the kernel at all. Could you test the commit of your side?

Alternative solutions seem to be usage of the `AI_ADDRCONFIG` flag--which might introduce other side effects (?)--or going back to the develop branch and skipping the `getaddrinfo` if no address is passed explicitly. Instead we could just create a v4 socket if creation of the v6 socket fails. That seems to be the wrong approach though ...

josephnoir:
Sorry for changing things so much. Looking at [Beej's guide](http://beej.us/guide/bgnet/output/html/multipage/clientserver.html#simpleserver), looping over the available interfaces seems the right way to go. And, this works as well on my VM with IPv6 disabled at runtime and not in the kernel.

libcaf_io/src/default_multiplexer.cpp
```diff
@@ -1316,10 +1316,20 @@ expected<void> set_inaddr_any(native_socket fd, sockaddr_in6& sa) {
...
 template <int Family>
-expected<void> new_ip_acceptor_impl(native_socket fd, uint16_t port,
-                                        const char* addr) {
+expected<native_socket> new_ip_acceptor_impl(uint16_t port, const char* addr,
+                                             bool reuse_addr) {
   static_assert(Family == AF_INET || Family == AF_INET6, "invalid family");
   CAF_LOG_TRACE(CAF_ARG(port) << ", addr = " << (addr ? addr : "nullptr"));
+  CALL_CFUN(fd, cc_valid_socket, "socket", socket(Family, SOCK_STREAM, 0));
+  // sguard closes the socket in case of exception
+  socket_guard sguard(fd);
```
Neverlord: Style: `sguard{fd}`

libcaf_io/src/default_multiplexer.cpp
```diff
@@ -1339,35 +1349,33 @@ expected<void> new_ip_acceptor_impl(native_socket fd, uint16_t port,
...
-              setsockopt(fd, SOL_SOCKET, SO_REUSEADDR,
-                         reinterpret_cast<setsockopt_ptr>(&on),
-                         static_cast<socklen_t>(sizeof(on))));
+  auto addrs = interfaces::server_address(port, addr);
+  if (addrs.empty())
+    return make_error(sec::cannot_open_port, "No local interface available",
+                      addr);
+  int fd = invalid_native_socket;
+  for (auto& elem : addrs) {
+    const char* addr = elem.first.c_str();
```
Neverlord: Don't repeat yourself: `auto addr = ...`

libcaf_io/src/default_multiplexer.cpp
```diff
@@ -1339,35 +1349,33 @@ expected<void> new_ip_acceptor_impl(native_socket fd, uint16_t port,
...
+    auto p = elem.second == ipv4
+           ? new_ip_acceptor_impl<AF_INET>(port, addr, reuse_addr)
+           : new_ip_acceptor_impl<AF_INET6>(port, addr, reuse_addr);
+    if (!p) {
+      CAF_LOG_ERROR(p.error());
```
Neverlord: This is not an error, since the loop tries all alternatives. `CAF_LOG_DEBUG`, if any.

libcaf_io/src/default_multiplexer.cpp
```diff
@@ -1339,35 +1349,33 @@ expected<void> new_ip_acceptor_impl(native_socket fd, uint16_t port,
...
+    }
+    fd = *p;
+    break;
   }
-  auto p = proto == ipv4 ? new_ip_acceptor_impl<AF_INET>(fd, port, addr)
-                         : new_ip_acceptor_impl<AF_INET6>(fd, port, addr);
-  if (!p)
-    return std::move(p.error());
+  if (fd == invalid_native_socket)
+    return make_error(sec::cannot_open_port,
```
Neverlord: This is where I would put the log output. However, this is not a hard error in the program. Either of `CAF_LOG_WARNING` or `CAF_LOG_INFO` is a better fit.

libcaf_io/src/interfaces.cpp
```diff
@@ -24,6 +24,7 @@
 #include <cerrno>
 #include <cstdlib>
 #include <cstring>
+#include <iostream>
```
Neverlord: I don't see any use of `cout` or `cerr` in this file. Leftover of debug output?

libcaf_io/src/interfaces.cpp
```diff
@@ -239,6 +240,43 @@ interfaces::native_address(const std::string& host,
...
+                           family == AF_INET ? protocol::ipv4
+                                             : protocol::ipv6);
+    }
+  }
+  std::stable_sort(std::begin(results), std::end(results),
+                   [](const addr_pair& lhs, const addr_pair& rhs) {
+                     return lhs.second > rhs.second;
+                   });
+  return results;
+
```
Neverlord: Empty line in function.

josephnoir:

PR #558, opened 2017-05-15
remove OSX/Clang specific pthread code

S4H:
The compiler shipped with XCode 8 now supports the `thread_local` specifier, this PR drops the workaround code that was previously required.

The only caveat to this PR is that the minimum compiler for OSX will need to be upgraded ( but I am unaware of any breaking reason that would block this ).

### DIFF
--- a/libcaf_core/caf/config.hpp
+++ b/libcaf_core/caf/config.hpp
@@ -101,6 +101,9 @@
 #  define CAF_ANNOTATE_FALLTHROUGH [[clang::fallthrough]]
 #  define CAF_COMPILER_VERSION                                                 \
     (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)
+#  if !__has_feature(cxx_thread_local)
+#    define CAF_NO_THREAD_LOCAL
+#  endif
 #elif defined(__GNUC__)
 #  define CAF_GCC
 #  define CAF_DEPRECATED __attribute__((__deprecated__))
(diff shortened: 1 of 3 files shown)


Neverlord:
I'd love to get rid of this pthread nonsense, but this would break CAF for many users without any benefit. It would even break CAF in our own Jenkins installation. Is there a way to detect the Clang version shipped with XCode 8 or never with macros and then gracefully fall back to the pthread code for older versions?

Hamdor:
Maybe we can check if the compiler supports thread local using `__has_feature(cxx_thread_local)` [1]. This is a Clang specific extension and GCC does not support it.

[1] https://clang.llvm.org/docs/LanguageExtensions.html

S4H:
It is possible to use cmake `cxx_known_features` feature [1] to detect supported language features. I'll have a dig around when i get chance.

[1] https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html?highlight=cxx_thread_local

Hamdor:
If we use `cxx_known_features` from CMake, this would require CMake 3.1 or higher. Distributions like Debian (jessie, stable) are current on CMake 3.0 and would no longer work out of the box.

grahamreeds:
And some of us are running ubuntu 14_04 LTS - 2.8.12.2.

GR

S4H:
Even if just for mailing list sanity I can see the advantage of it working out of the box. This issue is exclusively build related and Kitware does provide portable cmake binaries for download, somewhat reducing the possible fallout from the change in it's current form.

I will have a look at reworking the PR based on @Hamdor's original suggestion as I think that is probably the best practical solution

Neverlord:
Please squash the commits and rebase to latest `develop`.

Neverlord:

Neverlord:
Thanks for contributing to CAF. You rock!

Issue #542, opened 2017-03-22
VS 2017 - actor_control_block.hpp - null pointer


jonface:
First off, are you interested in VS2017 / win related issues? If so I'll continue to report anything I find.

After compiling caf_core_static, io_static, with some manipulation of the VS project files (mainly remove GCC options, switch to the debug dll, enable debug info), I've compiled the example group_chat, all in debug mode.

I can start the group_chat server, but when the client connects, the server crashes and the client crashes. The client crashes at,

```
>	group_chat.exe!caf::inspect::__l2::<lambda>() Line 220	C++
 	group_chat.exe!caf::data_processor<caf::serializer>::operator()<caf::error <lambda>(void) >(caf::meta::save_callback_t<caf::error <lambda>(void) > x) Line 473	C++
 	group_chat.exe!caf::data_processor<caf::serializer>::operator()<caf::error <lambda>(void),caf::meta::save_callback_t<caf::error <lambda>(void) > >(caf::meta::load_callback_t<caf::error <lambda>(void) > x, caf::meta::save_callback_t<caf::error <lambda>(void) > && <xs_0>) Line 482	C++
 	group_chat.exe!caf::data_processor<caf::serializer>::operator()<caf::node_id &,caf::meta::load_callback_t<caf::error <lambda>(void) >,caf::meta::save_callback_t<caf::error <lambda>(void) > >(caf::node_id & x, caf::meta::load_callback_t<caf::error <lambda>(void) > && <xs_0>, caf::meta::save_callback_t<caf::error <lambda>(void) > && <xs_1>) Line 513	C++
 	group_chat.exe!caf::data_processor<caf::serializer>::operator()<caf::node_id &,caf::meta::load_callback_t<caf::error <lambda>(void) >,caf::meta::save_callback_t<caf::error <lambda>(void) > >(const caf::meta::annotation & __formal, caf::node_id & <xs_0>, caf::meta::load_callback_t<caf::error <lambda>(void) > && <xs_1>, caf::meta::save_callback_t<caf::error <lambda>(void) > && <xs_2>) Line 487	C++
[...]

PR #560, opened 2017-05-27
Fix null pointer exception

hardliner66:
As described in Issue #542, there is a null pointer exception in the inspect method.
The exception was caused by capturing a local variable as reference in a lambda and passing that lambda as a callback to another function. At the time the lambda was called, the captured references are allready gone.

Never capture local variables by ref, if you pass the lambda lives longer than the scope it is created in.

### DIFF
--- a/libcaf_core/caf/actor_control_block.hpp
+++ b/libcaf_core/caf/actor_control_block.hpp
@@ -216,8 +216,8 @@ typename Inspector::result_type inspect(Inspector& f, strong_actor_ptr& x) {
     aid = x->aid;
     nid = x->nid;
   }
-  auto load = [&] { return load_actor(x, context_of(&f), aid, nid); };
-  auto save = [&] { return save_actor(x, context_of(&f), aid, nid); };
+  auto load = [&f, &x, aid, nid] { return load_actor(x, context_of(&f), aid, nid); };
+  auto save = [&f, &x, aid, nid] { return save_actor(x, context_of(&f), aid, nid); };
   return f(meta::type_name("actor"), aid,
            meta::omittable_if_none(), nid,
            meta::load_callback(load), meta::save_callback(save));


Neverlord:
Are you still using a beta version of VS 2017 or is this on the release build?

If this change fixes the bug for you than this is clearly a compiler bug. Both lambdas are called by the inspector `f` *before* the function returns. The standard guarantees that `nid` and `aid` live until the scope ends. In other words, until `f(...)` produces the result of the function. Is this a known deficiency in VC 2017? Otherwise, please file a bug to Microsoft.

hardliner66:
I'm using a stable version Visual Studio 2017 (Community Edition).

I played around with the code a bit and it doesn't seem to be related to the scope of the variables. Maybe it is related to the move of the lambda, but I can't reproduce it outside of caf.

But it would still be better to explicitly define the capture instead of an implicit ref capture.

I will investigate a bit more in the evening and report back.

hardliner66:
I looked into the issue a bit more and found out, that the problem also goes away if you wrap the lambdas inside a std::function. But I still haven't found out why exactly it happens.

It seems that the capture of the lambda gets lost somewhere. Maybe an optimization from the compiler or mayber some UB in the Code. I'm still not sure if this is a compiler bug, because I can't reproduce this behaviour outside of caf.

TurkishTlg:
I've just run into this bug too.  I've not spent a huge amount of time on it, but it looks like a compiler bug to do with copying lambdas with reference captures.  If you examine the callstack where the parameters are manually unpacked, and then examine the locals for <xs_n, you can see the contents of the lambdas are fine.  When they hit the overloads on lines 470 and 478 of data_processor.hpp, they become corrupt.  My guess is because they are being passed by value at that stage in the callstack rather than by reference like the previous stages, the generated constructor for meta::save_callback_t<F> and meta::load_callback_t<F> does not correctly copy the F.

I'll try and create a stripped down stand-alone version for the VS team to have a look at.

P.S.  I've only just come across this project and I'm massively excited by it - keep up the fantastic work!

Hamdor:
Maybe we should add a `#ifdef` workaround for VS 2017 and capture those variables by value.

hardliner66:
@TurkishTlg I would be really interested in the stripped down version if you wouldn't mind sharing it. We are currently shifting our development towards c++ and I would like to know more about this error.

@Hamdor If this gets wrapped inside an ifdef, we still need to deciede if it will be better to make a copy of the object or to wrap the lambda in a std::function.

I ran the benchmarks with both versions and it seemed like the performance of the std::function version was a little better. I only ran the benchmarks on my personal computer and only on a smaller scale, so there could be other things messing with the results, but after a few runs the std::function version was a bit faster.

TurkishTlg:
OK.  So it looks like there is a Visual Studio bug.  From what I can tell, the copy constructors for save_callback_t and load_callback_t sometimes go wrong because the compiler doesn't generate them.  Taking an equivalent struct:

```C++
template <typename T>
struct test
{
  test(T && t) {}
  test(test<T> &&) = default;
};

int main()
{
  test<int> t1(1);
  test<int> t2(t1); // disallowed - attempting to reference a deleted function
}
```
According to the standard:
"The move constructor is auto-generated if there is no user-declared copy constructor, copy assignment operator or destructor, and if the generated move constructor is valid (e.g. if it wouldn't need to assign constant members) (§12.8/10)."

I don't think = default constitutes a user-declared copy constructor, so the copy constructor should be generated, but regardless, VS disallows it.  If we looks at the overloads in data_processor:

```C++
template <class F, class... Ts>
  error operator()(meta::load_callback_t<F> x, Ts&&... xs) {
    error e;
    if (Derived::writes_state)
      e = x.fun();
    return e ? e : (*this)(std::forward<Ts>(xs)...);
  }
```
If VS is sticking with its own rules, x is passed by value so the compiler should reject it as with the test struct... but the code is not rejected..

The fix is simple enough though.  You can either have the overloads take a reference or you can define the copy constructor.  The former is likely to be faster whilst the later is safer.  If as @Neverlord says the lambdas are **always** called before going out of scope, change the signature to be 

```C++
error operator()(meta::load_callback_t<F> & x, Ts&&... xs)
error operator()(meta::save_callback_t<F> & x, Ts&&... xs)
```

The problem goes away.

Or you can define the copy constructors:
```C++
template <class F>
struct save_callback_t : annotation {
  save_callback_t(F&& f) : fun(f) {
    // nop
  }

  save_callback_t(save_callback_t&&) = default;
  save_callback_t(const save_callback_t &) = default;

  F fun;
};

template <class F>
struct load_callback_t : annotation {
  load_callback_t(F&& f) : fun(f) {
    // nop
  }

  load_callback_t(load_callback_t&&) = default;
  load_callback_t(const load_callback_t &) = default;

  F fun;
};
```
The problem goes away.

I've attached [this](https://github.com/actor-framework/actor-framework/files/1061760/caf.txt) code file that is a standalone demo of the bug.  I'll be submitting to MS presently, if I can't find an existing bug report.

TurkishTlg:
Just an FYI, the caf-test project runs fine with the copy constructors defined but fails with the pass-by-reference fix.  It doesn't crash, but it looks like some junk is left in a buffer and when comparing expected to actual, the test fails.

Neverlord:
@TurkishTlg Thank you so much for investigating this issue. I've added the copy constructors with https://github.com/actor-framework/actor-framework/commit/b7e222369c97390fc0605aa2a1c5fe9544874c36. @hardliner66 does this fix the issue for you as well?

timontheweb:
@hardliner66 I've changed my github username from TurkishTlg to timontheweb (someone else set up the account for me), so if there are any other problems, please mention my new name and I'll take a look.

hardliner66:
Just tried it with your commit and it fixes the issue.

Neverlord:
Ok, thanks for checking. I close this PR then.

PR #561, opened 2017-06-01
Minor fixes on Windows.

Chartas:
FindCAF.cmake: Fix to also find static only libs.
optional.hpp: Replace union with struct to avoid msvc "implict destructor call" error in conjunction with ue4.

### DIFF
--- a/cmake/FindCAF.cmake
+++ b/cmake/FindCAF.cmake
@@ -67,7 +67,8 @@ foreach (comp ${CAF_FIND_COMPONENTS})
       endif ()
       find_library(CAF_LIBRARY_${UPPERCOMP}
                    NAMES
-                     "caf_${comp}"
+                      "caf_${comp}"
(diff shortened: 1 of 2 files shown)


Neverlord:
Changing `union` to `struct` defeats the whole purpose of `optional`. A `struct` always constructs its members, a union does not (which is what we want here). What is ue4?

Chartas:
unreal engine 4. And i seem to have broken something anyway. And i didn't know that about unions. Thanks for teaching me something new. =)

Neverlord:
> Thanks for teaching me something new. =)

You're welcome. :)

Please squash the commits for the PR.

Chartas:
Hmmm, no static lib fix? I removed the union change.

Chartas:
Never mind. You'll probably be able to change that one line yourself. ^^ (I finally got it to run =D wuhu!)

PR #563, opened 2017-06-12
fix profile_coordinator not recording

nq-ebaratte:


Neverlord:
Thanks for the patch! :)

PR #564, opened 2017-06-15
Ignore CLI args after --

aufgang001:


libcaf_core/test/message.cpp
```diff
@@ -107,26 +107,36 @@ CAF_TEST(extract2) {
 }
 
 CAF_TEST(extract_opts) {
-  auto f = [](std::vector<std::string> xs) {
-    std::string filename;
-    size_t log_level;
+  auto f = [](std::vector<std::string> xs, std::vector<std::string> remainder) {
+    std::string filename = "";
```
Neverlord: The default constructor of string takes care of this more efficiently.

PR #567, opened 2017-06-20
Improve convenience for FindCAF

aufgang001:
find_package(CAF) will now find the component core 
and is equal to find_package(CAF COMPONENTS core)

### DIFF
--- a/cmake/FindCAF.cmake
+++ b/cmake/FindCAF.cmake
@@ -2,7 +2,7 @@
 #
 # Use this module as follows:
 #
-#     find_package(CAF)
+#     find_package(CAF [COMPONENTS <core|io|opencl|...>*] [REQUIRED])
 #
 # Variables used by this module (they can change the default behaviour and need
 # to be set before calling find_package):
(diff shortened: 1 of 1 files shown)


mavam:
What's the rationale behind choosing only `core`? As a CMake user, I would want the script to locate as many available components as possible.

Neverlord:
> I would want the script to locate as many available components as possible.

The rationale is to not introduce unnecessary dependencies. If I use CAF for concurrency I only care about the `core` module. You don't want to force users to install OpenCL to get `libcaf_opencl` although you actually don't need this library. Picking only the minimal set of required components is IMO good practice. On the flipside, if your application does need `libcaf_openssl` then you want CMake to produce an error if it can't find it.

Neverlord:
@aufgang001 I'm all for convenience, but I'm also for not obfuscating interfaces. This saves lazy users 16 characters, but it hides the fact that CAF is separated into modules.

I would much rather prefer a hard CMake error raised by the script if users attempt to use the script without component specification.

mavam:
> I would much rather prefer a hard CMake error raised by the script if users attempt to use the script without component specification.

Agreed.

aufgang001:
Ok, sounds good to me.

Neverlord:
I'll merge if you fix the commit message. It has nothing to do with the PR's content.

Maybe you find fatal errors convenient, but most people don't. ;)

PR #569, opened 2017-06-23
Remove duplicate .add_message_type<>() from middleman::init()

grahamreeds:
Removes 4 duplicate messages from middleman::init().


Neverlord:
Thanks for the patch! 👍🏻

PR #576, opened 2017-07-05
fix local_group::erase_subscriber

### LINKED ISSUE #574: Joining a group may prevent proper actor release
I believe there is an issue with the reference counting on class actors that join a group.

Here is the smallest reproduction case I could put together:

    #include <iostream>
    #include <vector>
    
    #include "caf/all.hpp"
    
    
    using namespace std;
    using namespace caf;
    
    
    using atomTerminate = atom_constant<atom("plz_die")>;
    using atomInit = atom_constant<atom("plz_init")>;
    
    class B: public event_based_actor {
    public:
        B(actor_config& cfg) : event_based_actor(cfg) {}
    
        ~B() {
            std::cout << "B::dtor" << std::endl;
        }
    protected:
        behavior make_behavior() override {
            join(system().groups().get_local("some_group"));
            return {
                [=](atomTerminate) {
                    quit();
                }
            };
        }
    
        void on_exit() override {
            leave(system().groups().get_local("some_group"));
        }
    };
    
    class A: public event_based_actor {
    public:
        A(actor_config& cfg) : event_based_actor(cfg) {}
    
        ~A() {
            std::cout << "A::dtor" << std::endl;
        }
        
    protected:
        behavior make_behavior() override {
            return {
                [=](atomInit) {
                    b = spawn<B>();
                },
                [=](atomTerminate) {
                    send(b, atomTerminate::value);
                    quit();
                }
            };
        }
    
        void on_exit() override {
            destroy(b);
        }
    
    private:
        actor b;
    };
        
    int main(){
        actor_system_config cfg;
        actor_system sys(cfg);
    
        std::vector<actor> a;
        for (int i = 0; i < 10; ++i) {
            a.push_back(sys.spawn<A>());
            anon_send(a.back(), atomInit::value);
        }
    
    
        for (int i = 0; i < 10; ++i)
            anon_send(a[i], atomTerminate::value);
    
        a.clear();
    
        actor dont_terminate_because_an_actor_is_still_alive = sys.spawn<A>();
        sys.await_all_actors_done();
        return 0;
    }

I expect to see 10 calls to ~A, and 10 calls to ~B; yet, when running this, only 0 to 5 B objets are deleted effectively (all As seem to be deleted properly).

- if B actors don't join any group, the issue does not occur
- leaving the group before exiting does not prevent the issue
- if B are spawned directly from the actor_system (instead of A), the issue does not occur.

I reproduce the issue on the `master `branch, as well as `develop`.

nq-ebaratte:
Fixes #574.

This PR makes use of a functionnality introduced in C++14: the ability to use a type that "compares equivalent" to the keys of an` std::set<>` in its `find()` method.

I added the `operator<` to compare an `intrusive_ptr<T>` to a raw `T*`

As an added bonus, the deletion is now O(log(N)), instead of O(N)

### DIFF
--- a/libcaf_core/caf/intrusive_ptr.hpp
+++ b/libcaf_core/caf/intrusive_ptr.hpp
@@ -232,6 +232,17 @@ bool operator<(const intrusive_ptr<T>& x, const intrusive_ptr<T>& y) {
   return x.get() < y.get();
 }
 
+/// @relates intrusive_ptr
+template <class T>
+bool operator<(const intrusive_ptr<T>& x, const T* y) {
+  return x.get() < y;
+}
+/// @relates intrusive_ptr
+template <class T>
+bool operator<(const T* x, const intrusive_ptr<T>& y) {
+  return x < y.get();
+}
+
 template <class T>
 std::string to_string(const intrusive_ptr<T>& x) {
   auto v = reinterpret_cast<uintptr_t>(x.get());
--- a/libcaf_core/src/group_manager.cpp
+++ b/libcaf_core/src/group_manager.cpp
@@ -87,12 +87,16 @@ public:
   std::pair<bool, size_t> erase_subscriber(const actor_control_block* who) {
     CAF_LOG_TRACE(""); // serializing who would cause a deadlock
     exclusive_guard guard(mtx_);
-    auto cmp = [](const strong_actor_ptr& lhs, const actor_control_block* rhs) {
-      return actor_addr::compare(lhs.get(), rhs) < 0;
-    };
     auto e = subscribers_.end();
-    auto i = std::lower_bound(subscribers_.begin(), e, who, cmp);
-    if (i == e || actor_addr::compare(i->get(), who) != 0)
+#if __cplusplus > 201103L
+    auto i = subscribers_.find(who);
+#else
+    auto cmp = [&](const strong_actor_ptr& lhs) {
+      return lhs.get() == who;
+    };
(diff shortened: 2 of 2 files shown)


Neverlord:
It's too bad, but I have to turn this PR down. CAF uses C++11, because we still support GCC 4.8. I look forward to drop support for this relatively old compiler eventually, but now isn't the time unfortunately.

Neverlord:
Thinking a bit more about it: we shouldn't discard your work entirely. Can you surround the critical code piece with an `#if __cplusplus > 201103L` guard and provide a C++11 fallback?

nq-ebaratte:
I updated the PR.

Please note that I dropped the actor_addr::compare() test, and use only address comparison. Is there a case where two different actor_control_block would describe the same actor ?

Neverlord:
Thanks a lot! 👍🏻

Neverlord:
Sorry, I forgot to answer your question.

> Is there a case where two different actor_control_block would describe the same actor?

For local actors: no.

I think there is only one way this could happen in CAF: storing an `actor_addr` to a remote actor. If the remote actor terminates, all actors receive down/exit messages and after that point CAF can drop the handle to it from the proxy registry. If we receive a message with a handle to the terminated actor then we re-create the handle. However, sending messages to the newly created handle will have no effect (actor already dead). Having the old weak pointer laying around therefore is not an issue, as you need to cast it to a strong pointer before doing anything with it (which will fail).

Relevant to this PR: the group only stores handles to local actors. So there's no way two control blocks identify the same actor.

Issue #540, opened 2017-03-21
Cannot Serialise std::vector<bool>


S4H:
Hitting compile errors when trying to build on OSX (XCode 8.0 + Clang 4.0.0). The issue is with serialising std::vector<bool>'s, CAF's automatic translation seems to be failing due the iterator proxies that the container type uses.

The culprit appears to be in a for loop in `detail/stringification_inspector.hpp` taking an `auto&` instead of a universal reference as it can't extend the lifetime of the PR value returned.

Relevant SO Article: http://stackoverflow.com/questions/34079390/range-for-loops-and-stdvectorbool

minimal example:
```cpp
#include <caf/typed_event_based_actor.hpp>

#include <caf/actor_system.hpp>
#include <caf/actor_system_config.hpp>
#include <caf/send.hpp>

using FooActor = caf::typed_actor< caf::reacts_to<std::vector<bool> >>;

FooActor::behavior_type make_actor(FooActor::pointer self)
{
    return [](std::vector<bool> foo)
    {};
}

int main(int argc, char* argv[])
{
    caf::actor_system_config cfg;
    caf::actor_system system{cfg};
    auto a = system.spawn(make_actor);
    std::vector<bool> foo(10, false);
    caf::anon_send(a, foo);
}
```

mavam:
As a side note: serializing large `vector<bool>` instances is very inefficient because you end up inflating a single bit into a byte on the wire, resulting in a 8x blow-up. I'd recommend use some simple packing scheme, such as stuffing bits into a `uint32_t` or `uint64_t`.

Given that `vector<bool>` is a part of the standard library, it's something CAF could do as well. It should be straight-forward to add. (PRs always welcome.)

Neverlord:
Yes, the 8x size overhead is unfortunate. A proper size-optimized serialization would be a simple, straightforward side project for any potential contributor. Maybe we should start documenting such "low hanging fruit" optimizations for people that are interested in contributing to CAF.

mavam:
We could introduce something along the lines of http://www.firsttimersonly.com, with the emphasis on complexity as opposed to seniority.

mavam:
I like the idea of a `low-hanging-fruit` label. Shall we?

Neverlord:
Yes, but please let's move the offtopic discussion someplace else. ;)

PR #577, opened 2017-07-07
Serialize vector<bool> efficiently

nq-ebaratte:
This is a more efficient serialization for vector<bool>; it is transformed to a vector<uint8>, each boolean encoded on a single bit (see #540)
At worst, 7 bits will be unused, when (size() % 8 == 1).

Note that the size of the vector is written ~twice (the size of the original vector<bool>, then the size of the vector<uint8_t>), because we need to know the number of unused bits in the last uint8_t. I guess it could still be optimized: instead of writing the full (uint64_t) size, we could only write its value mod. 8, shaving a few more bytes.

### DIFF
--- a/libcaf_core/caf/data_processor.hpp
+++ b/libcaf_core/caf/data_processor.hpp
@@ -226,26 +226,38 @@ public:
     return convert_apply(dref(), x, tmp, assign);
   }
 
-  // no better way around this abomination
-  error consume_range(std::vector<bool>& xs) {
-    auto i = xs.begin();
-    auto e = xs.end();
-    using proxy_iterator = decltype(i);
+  // Special case to avoid using 1 byte per bool
+  error apply(std::vector<bool>& x) {
+    uint64_t len = x.size();
+    auto err = begin_sequence(len);
+    if (err || len == 0)
+      return err;
     struct {
-      void operator()(proxy_iterator& lhs, bool& rhs) const {
-        *lhs = rhs;
+      size_t len;
+      void operator()(std::vector<bool>& lhs, std::vector<uint8_t>& rhs) const {
+        lhs.resize(len, false);
+        size_t cpt = 0;
+        for (auto v: rhs) {
+          for (int k = 0; k < 8; ++k) {
+            lhs[cpt] = ((v & (1 << k)) != 0);
+            if (++cpt >= len)
+              return;
+          }
+        }
       }
-      void operator()(bool& lhs, proxy_iterator& rhs) const {
-        lhs = *rhs;
+      void operator()(std::vector<uint8_t>& lhs, std::vector<bool>& rhs) const {
+        size_t k = 0;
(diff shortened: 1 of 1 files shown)


mavam:
Here's an idea for better CPU and space utilization: 

1. Write the number of bits as VarInt-encoded value
2. Use a `uint64_t` as block type for better CPU utilization
3. Write the last dirty block again as VarInt-encoded value to avoid wasting space

nq-ebaratte:
That's smart. I pushed a new commit implementing that.
Note that it does *no* make the code any clearer :(

mavam:
>  I pushed a new commit implementing that.

Great, would you mind adding a unit test as well? Ideally one with a `vector<bool>` instance that has a dirty last word and one that lines up exactly with the block so that we get full coverage for the complicated logic.

> Note that it does not make the code any clearer

Yeah, it's more logic after all. However, we [already have dedicated routines for VarInt encoding in the stream serializers](https://github.com/actor-framework/actor-framework/commit/a2418241e673e8948f01e1b636a366f48f7b1649#diff-d349abcfd3addf1bf4a14792166491ccR99). Ideally we'd refactor them. The problem is, that [decoding takes place one byte at a time](https://github.com/actor-framework/actor-framework/commit/a2418241e673e8948f01e1b636a366f48f7b1649#diff-5b51a1cdb09a57907e8de43817d74753R102), which tightly couples the implementation to the stream serializer. 

Still, I think we should factor the implementation because it quickly gets brittle when using it with different integer widths, plus factoring it makes it obvious to the reader what's going on behind the bit twiddling. To this end, all we need is a generic VarByte/VarInt decoding along the lines of [what we have in VAST](https://github.com/vast-io/vast/blob/master/libvast/vast/detail/varbyte.hpp#L56). Just substitute the `void*` source with an Iterator template et voilà, we can use VarByte in both code paths.

Neverlord:
Thanks a lot! 👍🏻

Neverlord:
@nq-ebaratte I had to rewind this in `develop` with dfea8b591894ef0135fc6e81922221de21a8bb47 to your initial `uint8_t` solution. The `uint64_t` solution fails on several systems because it assumes `size_t` and `uint64_t` are the same.

Neverlord:
The calculation of the remainder does not work if `size_t` does not have the same size as `uint64_t`, because you are passing `encoded` to `begin_sequence` [1], which expects a `size_t`.

[1] https://github.com/actor-framework/actor-framework/pull/577/commits/463c97dd95e44c452ad4696245672dadbb03ab15#diff-cbc92d374ecd5213f915f0f69e38b0cfR274

nq-ebaratte:
OK. I pushed a new commit that appears to fix the issue: https://github.com/nq-ebaratte/actor-framework/commit/7b2275ee656454e17d8a95044c6cfe62aa499a39

Or should we keep a uint8_t implementation, as "good enough" ?

PR #578, opened 2017-07-07
Fix build when CAF_LOG_LEVEL is not #defined

nq-ebaratte:


Neverlord:
Thanks for the fix! 👍🏻

PR #579, opened 2017-07-14
Fix bug in default_multiplexer

grahamreeds:
This fixes a signature bug in default_multiplexer that was preventing compiling completing.


grahamreeds:
borked this.  Sorry.

Neverlord:
@grahamreeds The issue is not GCC related and occurs when using the `epoll` backend. The new `try_run_once` and `run_once` functions are no longer platform-dependent and thus should be moved into the platform-agnostic part of the default multiplexer to avoid code duplication. I fixed this with  adc11fec9bbf1d1357abf991ee511c28786550a3.

PR #585, opened 2017-08-08
Add "--with-openssl" option to the "configure" script

dnthayer:
Added a new option "--with-openssl" to the "configure" script so
that a user can specify the location of the OpenSSL library and
headers.  This is needed in cases where OpenSSL is not found
automatically or if the wrong version is found.  For example, in order
to build CAF on macOS, we need to specify "--with-openssl=/opt/local"
when using MacPorts, or "--with-openssl=/usr/local/opt/openssl" when
using Homebrew.


mavam:
Thanks, that's an important switch to have in the configure script.

PR #586, opened 2017-08-09
Use cmake provide module to find OpenCL

josephnoir:
Our shipped `FindOpenCL.cmake` does not reliably set the `OpenCL_FOUND`. The variable is sometimes `true` even though OpenCL is not completely available (e.g. the headers are missing). In my test this solved the problems.

Alternatively (or additionally) I'd argue to disable OpenCL by default, i.e., exchanging the `--no-opencl` option to an `--with-opencl` option.


josephnoir:
A `FindOpenCL.cmake` is shipped with Cmake since version 3.1, we [currently require](https://github.com/actor-framework/actor-framework/blob/develop/CMakeLists.txt) 2.8.4.

Is this a thing we can (and want) to change?
Alternative ideas would be to include the FindOpenCL script from the newest CMake Version or just change the minimum Cmake version of [libcaf_opencl](https://github.com/actor-framework/actor-framework/blob/develop/libcaf_opencl/CMakeLists.txt) to 3.1?

Hamdor:
Since Debian (whezzy, jessie) and CentOS does not provide this CMake version we should keep the requirement of CMake 2.8. I think the simplest solution would be to put the `FindOpenCL.cmake` in the repository.

josephnoir:
I'll see if cmake 2.8.4 works with the newer find script.

Neverlord:
> Alternative ideas would be to include the FindOpenCL script from the newest CMake Version

I think that's the best option. Guard the OpenCL-related path with (something like) `if (CMAKE_VERSION >= 3.1)`, only printing a status message on older CMake releases to inform users that the OpenCL module requires a more up-to-date CMake.

PR #589, opened 2017-08-21
Enable compilation with VS 2017 (15.3.1)

josephnoir:
Makes the compiler happy, but there are parts that I simply commented out (in the streaming related files).

### DIFF
--- a/libcaf_core/caf/detail/stream_distribution_tree.hpp
+++ b/libcaf_core/caf/detail/stream_distribution_tree.hpp
@@ -201,6 +201,7 @@ public:
     in_.assign_credit(out_.credit());
   }
 
+/*
   error close(const stream_id& sid, const actor_addr& hdl) override {
     CAF_LOG_TRACE(CAF_ARG(sid) << CAF_ARG(hdl));
     if (in_.remove_path(sid, hdl, none, true))
@@ -229,6 +230,7 @@ public:
       return policy_.path_force_dropped(sid, hdl, std::move(reason));
     return none;
   }
+  */
 
   scheduled_actor* self() {
     return self_;
--- a/libcaf_core/caf/detail/intrusive_partitioned_list.hpp
+++ b/libcaf_core/caf/detail/intrusive_partitioned_list.hpp
@@ -76,6 +76,10 @@ public:
       return res;
     }
 
+    const value_type& operator*() const {
+      return *ptr;
+    }
+
     value_type& operator*() {
       return *ptr;
     }
--- a/libcaf_core/caf/typed_actor.hpp
+++ b/libcaf_core/caf/typed_actor.hpp
@@ -302,25 +302,25 @@ bool operator!=(const typed_actor<Xs...>& x,
 }
 
 /// @relates typed_actor
-template <class... Xs, class... Ys>
+template <class... Xs>
 bool operator==(const typed_actor<Xs...>& x, std::nullptr_t) noexcept {
   return actor_addr::compare(actor_cast<actor_control_block*>(x), nullptr) == 0;
 }
 
 /// @relates typed_actor
-template <class... Xs, class... Ys>
+template <class... Xs>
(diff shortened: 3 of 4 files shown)


josephnoir:

libcaf_core/caf/detail/stream_distribution_tree.hpp
```diff
@@ -229,6 +230,7 @@ class stream_distribution_tree : public stream_manager {
       return policy_.path_force_dropped(sid, hdl, std::move(reason));
     return none;
   }
+  */
```
Hamdor: What is the compile error for these functions?

josephnoir: IIRC something about the functions not being available for a policy type.

Neverlord:
This patch bundles unrelated things. Why the `using` changes to `caf-vec.cpp`?

The changes in `typed_actor.hpp` have nothing to do with MSVC. These are straight up fixes that you can push in a single commit to `develop` right away.

Adding the const overload to `intrusive_partioned_list`: if there a code path where this is needed then you can push this addition to `develop` as well.

I will look into the issue with `stream_distribution_tree` next Monday. You can remove the failing unit test for the tree (not the header itself) for now to get things compiling again.

josephnoir:
Kk.

> Adding the const overload to intrusive_partioned_list: if there a code path where this is needed then you can push this addition to develop as well.

This is only required by MSVC and fixes the bug reported in Gitter:

```
3>local_actor.cpp
3>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\algorithm(564): error C2678: binary '': no operator found which takes a left-hand operand of type 'const caf::detail::intrusive_partitioned_list<caf::mailbox_element,caf::detail::disposer>::iterator' (or there is no acceptable conversion)
3>D:\skorokhod\photon\thirdparty\decode\thirdparty\caf\libcaf_core\caf/detail/intrusive_partitioned_list.hpp(79): note: could be 'caf::mailbox_element &caf::detail::intrusive_partitioned_list<caf::mailbox_element,caf::detail::disposer>::iterator::operator (void)'
3>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\algorithm(564): note: while trying to match the argument list '(const caf::detail::intrusive_partitioned_list<caf::mailbox_element,caf::detail::disposer>::iterator)'
3>D:\skorokhod\photon\thirdparty\decode\thirdparty\caf\libcaf_core\src\local_actor.cpp(149): note: see reference to function template instantiation '_FwdIt std::partition_point<caf::detail::intrusive_partitioned_list<caf::mailbox_element,caf::detail::disposer>::iterator,caf::local_actor::push_to_cache::<lambda_0ef6105c7853238a21b080e72f554130>>(_FwdIt,_FwdIt,_Pr)' being compiled
3> with
3> [
3> _FwdIt=caf::detail::intrusive_partitioned_list<caf::mailbox_element,caf::detail::disposer>::iterator,
3> _Pr=caf::local_actor::push_to_cache::<lambda_0ef6105c7853238a21b080e72f554130>
3> ]
3>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\include\algorithm(564): error C2100: illegal indirection
```

> Why the using changes to caf-vec.cpp?

Only solution i found to make MSVC accept that `cerr << "..." << endl` is acceptable code in the file. Maybe there is another problem here, but the error message complained about missing the `<<` operation ... or something like that.

libcaf_core/caf/detail/stream_distribution_tree.hpp
josephnoir: Happens in clang based builds [in Jenkins](https://jenkins.inet.haw-hamburg.de/job/CAF/job/develop%20branch%20clang/label_exp=Linux%20&&%20clang%20&&%20LeakSanitizer/680/console) as well.

PR #590, opened 2017-08-22
Support for static runtime

iggu:
I added an ability to link with std c++ library in a totally static way so that there would be no need to have msvcrt near the final executable (/MT link flag under Windows and -static-libstdc++ for GCC and -static-libgcc for clang).


Neverlord:
LGTM. One more thing: please rebase & squash into a single commit for merging.

Neverlord:
Thanks for contributing to CAF. You rock! :)

PR #610, opened 2017-10-15
Add possibility to register thread hooks

### LINKED ISSUE #603: Hooks for new threads in CAF
(feature request from a private conversion)

CAF starts new threads as scheduler workers or whenever a thread is detached. Adding a hook that gets executed whenever a new thread gets started by CAF would give users more control. For example, users could wish to log new threads or set signal handlers.

Hamdor:
Example:
```
class my_thread_hooks : public thread_hook {
public:
  my_thread_hooks(/*arguments...*/) {
    // ...
  }
  /// Called by the actor system once before starting any threads.
  virtual void init(actor_system&) {
    // ...
  }

  /// Called whenever the actor system has started a new thread.
  /// To access a reference to the started thread use `std::this_thread`.
  /// @warning must the thread-safe
  virtual void thread_started_cb(actor_system&) {
    // ...
  }

  /// Called whenever a thread is about to quit.
  /// To access a reference to the terminating thread use `std::this_thread`.
  /// @warning must the thread-safe
  virtual void thread_terminates_cb(actor_system&) {
    // ...
  }
};
// ...
actor_system_config cfg{};
cfg.add_add_thread_hook<my_thread_hooks>(/*arguments for constructor*/);
// ...
```

closes #603

### DIFF
--- a/libcaf_core/caf/actor_system.hpp
+++ b/libcaf_core/caf/actor_system.hpp
@@ -491,6 +491,14 @@ public:
   /// Blocks the caller until all detached threads are done.
   void await_detached_threads();
 
+  /// Calls all thread started hooks
+  /// @warning must be called by thread which is about to start
+  void thread_started();
+
+  /// Calls all thread terminates hooks
+  /// @warning must be called by thread which is about to terminate
+  void thread_terminates();
+
   /// @endcond
 
 private:
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -30,6 +30,7 @@
 
 #include "caf/fwd.hpp"
 #include "caf/stream.hpp"
+#include "caf/thread_hook.hpp"
 #include "caf/config_value.hpp"
 #include "caf/config_option.hpp"
 #include "caf/actor_factory.hpp"
@@ -50,6 +51,8 @@ public:
   using hook_factory = std::function<io::hook* (actor_system&)>;
 
   using hook_factory_vector = std::vector<hook_factory>;
+
+  using thread_hooks = std::vector<std::unique_ptr<thread_hook>>;
 
   template <class K, class V>
   using hash_map = std::unordered_map<K, V>;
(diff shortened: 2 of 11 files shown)


libcaf_core/caf/actor_system.hpp
```diff
@@ -491,6 +491,14 @@ class actor_system {
   /// Blocks the caller until all detached threads are done.
   void await_detached_threads();
 
+  /// Calls all thread started hooks
+  /// @warning must be called by thread which is about to start
+  void call_thread_started_cb();
```
Neverlord: The past or continuous tense already implies a callback nature and I feel like the `call_` prefix is redundant. I'd simply rename this to `thread_started()` and the function below to `thread_terminates`.

libcaf_core/caf/actor_system_config.hpp
```diff
@@ -224,6 +227,15 @@ class actor_system_config {
     });
   }
 
+  /// Adds a hook type to the scheduler.
+  template <class Hook, class... Ts>
+  actor_system_config& add_thread_hook(Ts&&... ts) {
+    thread_hooks_.push_back(
```
Neverlord: Can be simplified to `thread_hooks_.emplace_back(new Hook(...))`.

libcaf_core/caf/thread_hook.hpp
```diff
@@ -0,0 +1,48 @@
...
+    // nop
+  }
+
+  /// Called by the actor system once before starting any threads.
+  virtual void init(actor_system&) = 0;
+
+  /// Called whenever the actor system has started a new thread.
+  /// To access a reference to the started thread use `std::this_thread`.
+  /// @warning must the thread-safe
+  virtual void thread_started_cb(actor_system&) = 0;
```
Neverlord: The `actor_system&` argument is redundant, because of the `init` member function. Any implementation can simply store a pointer to the actor system and than have all the context available at any time.

Additionally, I would strip the `_cb` suffix. You've probably adapted this from the `io::hook` class, but the only reason the suffix is used in that class is to avoid name clashes with the enum values of `io::hook::event_type`.

Neverlord:
Merged with 15d04e37cae2567dc8a52299d31783181b4c0658. I've also removed the hardcoding of magic numbers with 35405ce6272b74d4e10f8136622ef1ae582638c6.

Thanks a lot for the patch!

PR #611, opened 2017-10-17
Add check for `result` type to `deliver`

### LINKED ISSUE #592: make_response_promise and response shoukd accept caf::result
now _self->response(caf::result<type>) trys to send result as it is insted of extracting error or good resonse from caf::result<...>

Hamdor:
This commit adds a `static_assert` to `response_promise::deliver` which
prevents faulty usage of `result<...>` and `deliver`.

closes #592

### DIFF
--- a/libcaf_core/caf/detail/type_traits.hpp
+++ b/libcaf_core/caf/detail/type_traits.hpp
@@ -607,6 +607,12 @@ constexpr bool can_insert_elements() {
   return can_insert_elements_impl<T>(static_cast<T*>(nullptr));
 }
 
+template <class Test, template <class...> class Ref>
+struct is_specialization : std::false_type { };
+
+template <template <class...> class Ref, class... Args>
+struct is_specialization<Ref<Args...>, Ref> : std::true_type { };
+
 } // namespace detail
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/detail/type_traits.hpp
```diff
@@ -607,6 +607,12 @@ constexpr bool can_insert_elements() {
   return can_insert_elements_impl<T>(static_cast<T*>(nullptr));
 }
 
+template <class Test, template <class...> class Ref>
+struct is_specialization : std::false_type { };
+
+template <template <class...> class Ref, class... Args>
+struct is_specialization<Ref<Args...>, Ref> : std::true_type { };
+
 } // namespace detail
```
Neverlord: Having the template template argument as second argument seems a bit counterintuitive to me. Did you consider doing it this way?

```cpp
/// -- add docu here ---
template <template <class...> class Tpl, class T>
struct is_specialization : std::false_type { };

template <template <class...> class T, class... Ts>
struct is_specialization<T, T<Ts...>> : std::true_type { };
```

PR #617, opened 2017-10-29
Add OpenSSL header/library version mismatch check

### LINKED ISSUE #616: Disable OpenSSL per default
We should think about disable the OpenSSL module per default. The reason is, that the FindOpenSSL provided by CMake is not doing the correct thing. For example, on MacOS 10.12.6 we have OpenSSL libs in the default library path (/usr/lib). This shipped lib is outdated (OpenSSL 0.9.8zh 14 Jan 2016). If we now install an current version with brew, we will have a separate lib/include dir in (usr/local/Cellar).
If we run the configure script, this will lead to the following:
* CMake selects the outdated system libraries (0.9.8)
* CMake detects this as version 1.0.2l which is the version I installed with brew.

```
-- Found OpenSSL: /usr/lib/libcrypto.dylib (found version "1.0.2l") 
found openssl: 1.0.2l. /usr/lib/libssl.dylib;/usr/lib/libcrypto.dylib /usr/local/include
```

This is problematic since this links against the outdated library and also this lib does not match the selected includes. On my system (MacOs 10.12.6) this leads to undefined symbols for TLS. Which is reasonable since TLS1.2 support was introduced with OpenSSL 1.0.1:
```
Undefined symbols for architecture x86_64:
  "_TLSv1_2_method", referenced from:
      caf::openssl::session::create_ssl_context() in session.cpp.o
  "caf::io::network::new_tcp_connection(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned short, caf::optional<caf::io::network::protocol>)", referenced from:
      caf::openssl::(anonymous namespace)::middleman_actor_impl::connect(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned short) in middleman_actor.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
An version check in our CMake files would not help, since this only checks the version of the headers, which are for some reason the correct version. If we make the module not build by default, and only enable it, if a user provide the path to the OpenSSL installation, we would not have this problem. We could then also add a check for a OpenSSL version >= 1.0.1.

Hamdor:
closes #616

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 2.8.4)
 project(caf C CXX)
 
+include(CheckCSourceCompiles)
+include(CheckCSourceRuns)
+
 # be nice to vim users
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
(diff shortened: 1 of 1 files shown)


Hamdor:
I am not sure if I should put the check to the CMakeLists.txt of the libcaf_openssl subfolder. What do you think?

Neverlord:
Should remain in the main `CMakeLists`, because the main file is responsible for resolving dependencies. Thanks for the fix!

PR #619, opened 2017-10-30
actor_system_config::options_ member access

lovemycatt:
The option_ member was moved from private to protected, in order to enable derived class access (e.g. for dump purposes).
Maybe a better solution would be to have a proper const accessor function.

### DIFF
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -359,6 +359,7 @@ private:
 
   std::string extract_config_file_name(message& args);
 
+protected:
   option_vector options_;
 };
 


Neverlord:
This member should remain private. Subclasses only should manipulate options via `custom_options_`.

If iterating is your use case, what about a `for_each_option` member function that iterates both `custom_options_` and `options_`?

Neverlord:
@lovemycatt does the `for_each_option` convenience function (see 5634c340f1a72dcedc3295de93801ae29bb3d087) cover your use case?

lovemycatt:
Yes it seems exactly what we need, thanks for the commit.

PR #621, opened 2017-11-10
Modify relaxed work-stealing to reduce latency

massimotorquati:
This patch modifies the work-stealing scheduler to let the runtime
worker threads sleep on a wait_for call instead of using a sleep_for
call, only involving the relaxed strategy.
The relaxed-sleep-duration value is used as the timeout value for
the wait_for call.
This patch improves message latency at low message rates without
introducing significant extra overheads at high message rate.

### DIFF
--- a/libcaf_core/caf/policy/work_stealing.hpp
+++ b/libcaf_core/caf/policy/work_stealing.hpp
@@ -25,6 +25,8 @@
 #include <thread>
 #include <random>
 #include <cstddef>
+#include <mutex>
+#include <condition_variable>
 
 #include "caf/resumable.hpp"
 #include "caf/actor_system_config.hpp"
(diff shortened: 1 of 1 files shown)


massimotorquati:
This is the link to the benchmark code we used for testing the latency patch:
https://github.com/ATS-Advanced-Technology-Solutions/caf_latency_benchmarks

Neverlord:
Thanks for the PR!

The diff looks very suspicious, unfortunately, as it claims you changed every single line in the file. Are you using Windows file endings? Please fix the line endings and force-push the fixed commit to your branch to update this PR.

I'm not using Windows, so I cannot provide you with the instructions to fix the endings. Maybe this SO discussion is of some use to you: https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

massimotorquati:
I'm fixing the line endings. Apologize.

massimotorquati:
Done (I hope)

Neverlord:
Thanks a lot! 👍

Hamdor:
Is there more work to be done or why this got merged to a separate branch?

Neverlord:
> Neverlord merged commit 05f70d8 into `actor-framework:topic/latency`  2 days ago

It's in a different branch because we want to do some more experiments based on the PR.

PR #627, opened 2017-11-18
Fix unit tests with OpenSSL 1.1.0

### LINKED ISSUE #625: Unit tests fail with OpenSSL versions newer than 1.0.1
./configure
make
make test
give me the following result:

> 60/63 Test #60: io_typed_remote_actor ............   Passed    0.01 sec
>       Start 61: io_unpublish
> 61/63 Test #61: io_unpublish .....................   Passed    0.01 sec
>       Start 62: openssl_authentication
> 62/63 Test #62: openssl_authentication ...........   Passed    0.02 sec
>       Start 63: openssl_dynamic_remote_actor
> 63/63 Test #63: openssl_dynamic_remote_actor .....***Exception: Other  0.48 sec
> 
> 98% tests passed, 1 tests failed out of 63
> 
> Total Test time (real) =   1.17 sec
> 
> The following tests FAILED:
> 	 63 - openssl_dynamic_remote_actor (OTHER_FAULT)
> Errors while running CTest
> make[1]: *** [Makefile:130: test] Error 8
> make[1]: Leaving directory '/home/felix/github/actor-framework/build'
> make: *** [Makefile:7: test] Error 2
>

Hamdor:
With OpenSSL 1.1.0 a security level specifier was introduced.
By default its set to 1. Our default settings does not fulfill level 1.

With this patch we don't use a specific curve anymore. OpenSSL will
select a curve automatically. In OpenSSL 1.1.0 and higher this is
already the default.

closes #625

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ if(CAF_BUILD_STATIC_RUNTIME)
       endif()
     endif()
   endforeach()
+else()
+  set(CAF_BUILD_STATIC_RUNTIME no)
 endif()
 
 # add helper target that simplifies re-running configure
@@ -537,7 +539,7 @@ if(NOT CAF_NO_OPENSSL)
     # Check OpenSSL version >= 1.0.1
     if (OPENSSL_VERSION VERSION_LESS 1.0.1)
       message(STATUS
-              "Disable OpenSSL. Required >= 1.0.1 due to TLSv2 support.")
+              "Disable OpenSSL. Required >= 1.0.1 due to TLSv1.2 support.")
       set(CAF_NO_OPENSSL yes)
     else()
       # Check if openssl headers and library versions match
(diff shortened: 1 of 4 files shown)


josephnoir:
Works on my Archlinux VM. (OpenSSL 1.1.0g  2 Nov 2017)

Hamdor:
@josephnoir with or without my patch? ;)

josephnoir:
With your patch ^__^
Without it, I encountered the same problem as described in #625.

Neverlord:
@Hamdor thanks a lot for the PR!

@rsmmr I would like to hear your opinion, since I believe you are better at judging OpenSSL-related code. Do you think this could be done even more elegantly or would you agree with Hamdor's approach?

rsmmr:
not sure either here, @0xxon  what do you think?

Neverlord:
I think we can go ahead and merge it since nothing obvious seems to come up.

@Hamdor Thanks a lot once more! 👍

josephnoir:
Did we ever run this through Jenkins? Several builds fail with the error `SSL_CTX_set_ecdh_auto was not declared in this scope`. I am not sure if something changed in the Jenkins setup or the failures started after the merge. OpenSSL version is 1.0.1f 6 Jan 2014 on a Linux VM with 3.19.0-68-generic kernel.

(We could also disable SSL for Jenkins builds of contributions.)

Hamdor:
I just checked the manpage for `SSL_CTX_set_ecdh_auto`:
"These functions were first added to OpenSSL 1.0.2."
I will provide a fix for this later this weekend. :-(

PR #628, opened 2017-12-08
Fix build on MSVC 15.5.0.

dfedin:
Fixed problem with C1903 error in static_assert, used in constructor with
variadic template arguments

### DIFF
--- a/libcaf_core/caf/actor_storage.hpp
+++ b/libcaf_core/caf/actor_storage.hpp
@@ -42,6 +42,11 @@ public:
   template <class... Us>
   actor_storage(actor_id x, node_id y, actor_system* sys, Us&&... zs)
       : ctrl(x, y, sys, data_dtor, block_dtor) {
+    // construct data member
+    new (&data) T(std::forward<Us>(zs)...);
+  }
+
+  ~actor_storage() {
     // 1) make sure control block fits into a single cache line
     static_assert(sizeof(actor_control_block) < CAF_CACHE_LINE_SIZE,
                   "actor_control_block exceeds a single cache line");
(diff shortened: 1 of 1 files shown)


Neverlord:
Welcome to CAF and thank you for contributing!

Is there any reason not to just move the static asserts to the constructor in order to avoid the `ifdef` blocks and code duplication?

dfedin:
Do you mean "move the static asserts to the destructor"?
Because this code cannot be compile in constructor with variadic template parameters on MSVC 15.5.0.
I can move all the code to the destructor, but I need some clarifications for this snippet of code:
```c++
// 4) make sure static_cast and reinterpret_cast  
//    between T* and abstract_actor* are identical  
     constexpr abstract_actor* dummy = nullptr;  
     constexpr T* derived_dummy = static_cast<T*>(dummy);  
     static_assert(derived_dummy == nullptr,  
                    "actor subtype has illegal memory alignment "  
                    "(probably due to virtual inheritance)");  
```
Should it be moved to destructor and enable for clang on Linux only?
This code causes internal compiler error on MSVC 15.5.0, in the destructor and in the constructor both.

Neverlord:
> Do you mean "move the static asserts to the destructor"?

Yes, sorry.

Just for my understanding: 1, 2, and 3 are working on MSVC in the destructor, but 4 doesn't. However, currently check 4 is enabled on Clang/Linux only anyways, so your patch works as-is. Correct?

dfedin:
Yes, correct. Just have moved all static asserts to destructor in a last version of the patch

Neverlord:

Neverlord:
@josephnoir mobi10 failed for some reason, I removed it from the builds for now.

Neverlord:
@dfedin Thanks for contributing to CAF. You rock! 🤘

PR #635, opened 2017-12-21
Fix compatible problems with ASIO in boost 1.66

ufownl:
ASIO of Boost 1.66 used `io_context` instead of `io_service` and added a typedef for backwards compatibility. So the forward declaration at [multiplexer.hpp](https://github.com/actor-framework/actor-framework/blob/master/libcaf_io/caf/io/network/multiplexer.hpp#L42) lead compilation errors when using ASIO of Boost 1.66.


Hamdor:

Neverlord:
Thanks for the patch!

PR #636, opened 2017-12-21
Fix caf::io::remote_actor free function

rkfg:
The remote_actor() method of the middleman probably has been refactored some time ago and the corresponding free function wasn't, so I fixed that.


Neverlord:
Thanks for the patch!

PR #637, opened 2017-12-23
Add templated put/get registry methods

rkfg:
If I understand this right, we need to cast actors to and from `strong_actor_ptr` when working with the registry. I think it could be moved to the registry itself to make the code more readable. Now when I put I use this:
```cpp
system.registry().put(main_loop_atom::value, actor_cast<strong_actor_ptr>(self));
```
and when I get I have to cast again:
```cpp
auto ml = actor_cast<actor>(self->system().registry().get(main_loop_atom::value));
```
Casting on put seems to be inevitable anyway so better make `put` a template that accepts any type and casts it to `strong_actor_ptr` internally and allow specifying the desired result type in `get`. The type information is lost anyway thus we need to know what type to cast the result to. Now the syntax is a bit shorter and more intuitive (`get` without the template parameter still returns just `strong_actor_ptr` to retain compatibility):
```cpp
self->system().registry().put(main_loop_atom::value, self);
auto ml = self->system().registry().get<actor>(main_loop_atom::value);
```
This must work for all types of actors, I tried with scoped_actor, dynamically and statically typed ones. Please take a look at the `put` signature, I'm getting the value by reference so that putting `scoped_actor` to the registry isn't any different from other actor types. If it would be by value, we'd need to use something like `self.ptr()` for scoped actors which is ugly. I'm a bit worried about reference counting in that case, please check if it works fine. I suppose since these operations are synchronous and we're calling the other `put` method that accepts by value (incrementing ref) it shouldn't be a problem. Also the reference prevents calling this templated version from the regular `put()` (which results in compile error because `actor_cast` doesn't work on `nullptr`) when it clears the entry in the functor:
```cpp
  if (value)
    value->get()->attach_functor([=] {
      system_.registry().put(key, nullptr);
    });
```

...or else we'd need to cast `nullptr` to `strong_actor_ptr`. So we're killing two birds with one stone.

### DIFF
--- a/libcaf_core/caf/actor_registry.hpp
+++ b/libcaf_core/caf/actor_registry.hpp
@@ -29,6 +29,7 @@
 
 #include "caf/fwd.hpp"
 #include "caf/actor.hpp"
+#include "caf/actor_cast.hpp"
 #include "caf/abstract_actor.hpp"
 #include "caf/actor_control_block.hpp"
 
@@ -49,10 +50,16 @@ public:
   ~actor_registry();
 
   /// Returns the local actor associated to `key`.
-  strong_actor_ptr get(actor_id key) const;
+  template<class T = strong_actor_ptr>
+  T get(actor_id key) const {
+      return actor_cast<T>(get_impl(key));
+  }
 
   /// Associates a local actor with its ID.
-  void put(actor_id key, strong_actor_ptr val);
+  template<class T>
+  void put(actor_id key, const T& val) {
+      put_impl(key, actor_cast<strong_actor_ptr>(val));
+  }
 
   /// Removes an actor from this registry,
   /// leaving `reason` for future reference.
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/actor_registry.hpp
```diff
@@ -51,9 +52,21 @@ class actor_registry {
   /// Returns the local actor associated to `key`.
   strong_actor_ptr get(actor_id key) const;
 
+  /// Returns the local actor associated to `key`.
+  template<typename T>
```
Hamdor: please change to `template <class T>`, also for the others. :)

libcaf_core/caf/actor_registry.hpp
```diff
@@ -74,9 +87,23 @@ class actor_registry {
...
+
   /// Associates given actor to `key`.
   void put(atom_value key, strong_actor_ptr value);
 
+  /// Associates given actor to `key`.
+  template<typename T>
+  void put(atom_value key, T& value) {
+      // using reference here and before to allow putting a scoped_actor without calling .ptr()
+      // also prevents calling this version from non-templated put as (key, nullptr)
+      put(key, actor_cast<strong_actor_ptr>(value));
```
Hamdor: I am not sure if we should allow putting `scoped_actors` in there. Maybe @Neverlord can give some advise.

rkfg: I think it wouldn't hurt as when the actor is deconstructed it's removed from the registry anyway. But communicating back to an "initial" actor, say, created in main() is a nice use case.

libcaf_core/caf/actor_registry.hpp
rkfg: Ok, fixed!

rkfg:
Done and done. Sorry about forgetting `const`, I know it's important.

Neverlord:
Thanks for contributing to CAF! 👍

Issue #638, opened 2017-12-23
ASIO multiplexer does not compile in master


josephnoir:
The boost asio-based multiplexer does not compile in the master branch as the new UDP-related multiplexer functions are missing here. I see two ways forward:

1. Add stubs that raise an error when trying to use ASIO with UDP
2. Implement UDP in the ASIO multiplexer

To get things working quickly, I'd suggest adding the stubs but I would like to do things "the right way" eventually.

josephnoir:
The first one won't work because the io test are automatically generated for ASIO if it is available and that will lead to failures.

josephnoir:
Should we keep this open until UDP is implemented for the asio multiplexer?

Neverlord:
> Should we keep this open until UDP is implemented for the asio multiplexer?

No, it's still possible that we drop ASIO support alltogether.

PR #639, opened 2017-12-23
Fix asio multiplexer compilation with UDP

josephnoir:
Although this allows compilation of the asio multiplexer with UDP it does not enable to use UDP with the multiplexer. The related stubs will throw runtime errors when called. Addresses the fist suggestion in #638


josephnoir:
Not sure if Jenkins even tests with ASIO for contributions ...

Neverlord:
> Not sure if Jenkins even tests with ASIO for contributions ...

I'm afraid not. That's something we should work on in 2018. :)

PR #640, opened 2017-12-24
Fix build with logging enabled

Hamdor:


libcaf_io/src/connection_helper.cpp
```diff
@@ -61,16 +61,15 @@ behavior datagram_connection_broker(broker* self, uint16_t port,
     after(autoconnect_timeout) >> [=]() {
       CAF_LOG_TRACE(CAF_ARG(""));
       // nothing heard in about 10 minutes... just a call it a day, then
-      CAF_LOG_INFO("aborted direct connection attempt after 10min:"
-                    << CAF_ARG(nid));
```
Hamdor: I had to remove `nid` since we do not have it anymore. I have not seen a alike state so I removed it complete for now... This message is might be not that helpful anymore...

PR #641, opened 2017-12-24
Fix build on msvc

Hamdor:
Unit tests on windows will fail in some UDP unit test. This needs further investigation.

### DIFF
--- a/libcaf_io/src/default_multiplexer.cpp
+++ b/libcaf_io/src/default_multiplexer.cpp
@@ -683,7 +683,7 @@ bool read_datagram(size_t& result, native_socket fd, void* buf, size_t buf_len,
   CAF_LOG_TRACE(CAF_ARG(fd));
   memset(ep.address(), 0, sizeof(sockaddr_storage));
   socklen_t len = sizeof(sockaddr_storage);
-  auto sres = ::recvfrom(fd, buf, buf_len, 0, ep.address(), &len);
+  auto sres = ::recvfrom(fd, static_cast<socket_recv_ptr>(buf), buf_len, 0, ep.address(), &len);
   if (is_error(sres, true)) {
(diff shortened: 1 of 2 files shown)


libcaf_io/src/default_multiplexer.cpp
```diff
@@ -683,7 +683,7 @@ bool read_datagram(size_t& result, native_socket fd, void* buf, size_t buf_len,
   CAF_LOG_TRACE(CAF_ARG(fd));
   memset(ep.address(), 0, sizeof(sockaddr_storage));
   socklen_t len = sizeof(sockaddr_storage);
-  auto sres = ::recvfrom(fd, buf, buf_len, 0, ep.address(), &len);
+  auto sres = ::recvfrom(fd, static_cast<char*>(buf), buf_len, 0, ep.address(), &len);
```
Neverlord: Please use our typedef `socket_recv_ptr` instead of the platform-dependent `char*`.

libcaf_io/src/ip_endpoint.cpp
```diff
@@ -36,6 +36,10 @@
 # include <netinet/ip.h>
 #endif
 
+#ifdef CAF_WINDOWS
+using sa_family_t = unsigned int;
```
Neverlord: Microsoft uses a `short` for the `sin_family` and `sin6_family` fields: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740496(v=vs.85).aspx

Hamdor: Done

libcaf_io/src/default_multiplexer.cpp
Hamdor: Done

Neverlord:
Thanks for the patch! 👍

PR #643, opened 2017-12-27
Fix build for mingw (mxe)

rkfg:
This fixes building CAF in [MXE](https://github.com/mxe/mxe) for the Windows target. The cURL example doesn't build yet (cURL linkage and signals issues), the Qt example doesn't link but the tests are fine and 100% passed using Wine. Please test on MSVC before merge, I don't have it and can't check if everything is good.

The command to configure using MXE (64 bit, assuming you've created and switched to a new build directory inside CAF):
```
/path/to/mxe/usr/bin/x86_64-w64-mingw32.static.posix-cmake -DCMAKE_BUILD_TYPE=Release -DOPENSSL_CORRECT_VERSION_NUMBER=1 -DCAF_BUILD_STATIC_ONLY=yes -DCAF_NO_CURL_EXAMPLES=yes -DCAF_NO_QT_EXAMPLES=yes ..
```
Building a debug version takes ages for me, YMMV. The OpenSSL hack is required and reported by CMake itself.

### DIFF
--- a/libcaf_io/src/ip_endpoint.cpp
+++ b/libcaf_io/src/ip_endpoint.cpp
@@ -24,9 +24,9 @@
 
 #ifdef CAF_WINDOWS
 # include <winsock2.h>
+# include <windows.h>
 # include <ws2tcpip.h>
 # include <ws2ipdef.h>
-# include <windows.h>
 #else
 # include <unistd.h>
 # include <cerrno>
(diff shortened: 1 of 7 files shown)


libcaf_io/src/ip_endpoint.cpp
```diff
@@ -209,12 +209,20 @@ std::string host(const ip_endpoint& ep) {
   switch(ep.caddress()->sa_family) {
     case AF_INET:
       inet_ntop(AF_INET,
+#ifdef CAF_WINDOWS
```
Neverlord: A `const_cast` doesn't impose any runtime cost. Only Windows requires a mutable pointer, but it doesn't hurt on other OSes. So, please remove the `ifdef` trees, in both cases keeping the `const_cast` version.

rkfg: Done!

rkfg:
I also changed the original message here, the Qt example currently can't be linked also, requires lots of libs like libpng, libpcre, libharfbuzz etc. to be specified explicitly. I'm not sure it's that important right now so I added another flag, `-DCAF_NO_QT_EXAMPLES=yes`, to the sample `cmake` command.

Neverlord:
Thanks for contributing! 👍

I wouldn't bother with the Qt examples on Windows. Maybe we should consider making them opt-in anyways.

PR #655, opened 2018-01-25
Improve reordering latency

josephnoir:
Improve the latency in case UDP has to reorder packets. In addition to a timeout, the queue with pending messages has a length (currently 10). Pending messages will be when it runs full. We'll still need a timeout in case we received the end of a message stream and our queue will never run full, but a message is lost.

I don't see a problem with dropping messages in this case as there is no reliability guarantee at the moment. The queue length ... is a good question, 10 will reorder things that arrive a bit twisted but messages with a longer delay will be dropped. If you have a steady stream of messages the queue would need to be rather long to wait for messages that have a significant delay. I'm not sure how to determine a good queue size atm. In any case, the length is configurable via the actor system config.

### DIFF
--- a/libcaf_io/src/basp_broker.cpp
+++ b/libcaf_io/src/basp_broker.cpp
@@ -77,7 +77,8 @@ basp_broker_state::basp_broker_state(broker* selfptr)
                              static_cast<proxy_registry::backend&>(*this)),
       self(selfptr),
       instance(selfptr, *this),
-      max_buffers(self->system().config().middleman_cached_udp_buffers) {
+      max_buffers(self->system().config().middleman_cached_udp_buffers),
+      max_pending_messages(self->system().config().middleman_max_pending_msgs) {
   CAF_ASSERT(this_node() != none);
 }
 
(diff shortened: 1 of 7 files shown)


josephnoir:

libcaf_io/src/instance.cpp
```diff
@@ -162,20 +162,20 @@ bool instance::handle(execution_unit* ctx, new_datagram_msg& dm,
       return err();
     }
   }
-  // Handle FIFO ordering of datagrams
```
Neverlord: Why switch to lowercase? New code in CAF usually comments with whole sentences, i.e., capitalized first letter (and period at the end). All lowercase comments often indicate old code in CAF.

libcaf_io/src/basp_broker.cpp
```diff
@@ -533,37 +534,49 @@ basp_broker_state::next_sequence_number(datagram_handle hdl) {
...
 
 bool basp_broker_state::deliver_pending(execution_unit* ctx,
-                                        basp::endpoint_context& ep) {
-  if (!ep.requires_ordering)
+                                        basp::endpoint_context& ep,
+                                        bool force) {
+  if (!ep.requires_ordering || ep.pending.empty())
     return true;
   std::vector<char>* payload = nullptr;
-  auto itr = ep.pending.find(ep.seq_incoming);
```
Neverlord: Any reason not to use simply `i` as variable name?

libcaf_io/src/instance.cpp
josephnoir: I somehow thought it was the other way around. What about comments for members?

Neverlord: > What about comments for members?

Same, except that they use `///` for doxygen (unless private, but even then I recommend using the doxygen style; that makes shuffling code around easier in case the access changes later).

PR #657, opened 2018-01-29
Fixed GCC 4.8 issue with lack of space between operator and literals …

grahamreeds:
…in durations


Neverlord:
Thanks for the fix!

grahamreeds:
No problem.  Hopefully one day I will have the time to sit down and actually work a prototype with the UDP I waited so long for! :-)

PR #658, opened 2018-01-31
Update ordering test to have control over timeouts

josephnoir:
Some Jenkins test failed because the timeout to deliver pending message was triggered before all messages were in the queue. This behavior is addressed in this PR by using the `test_coordinator` gain control over timeout delivery.

Additionally, an additional tests checks if a timeout messages skips missing messages.

### DIFF
--- a/libcaf_io/test/basp_udp.cpp
+++ b/libcaf_io/test/basp_udp.cpp
@@ -110,14 +110,16 @@ struct node {
 
 class fixture {
 public:
-  fixture(bool autoconn = false)
+  fixture(bool autoconn = false, bool use_test_coordinator = false)
       : sys(cfg.load<io::middleman, network::test_multiplexer>()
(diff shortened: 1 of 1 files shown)


libcaf_io/test/basp_udp.cpp
```diff
@@ -685,7 +701,7 @@ CAF_TEST(publish_and_connect_udp) {
 
 CAF_TEST(remote_actor_and_send_udp) {
   constexpr const char* lo = "localhost";
-  CAF_MESSAGE("self: " << to_string(self()->address()));
+  CAF_MESSAGE("self:    " << to_string(self()->address()));
```
Neverlord: Please don't add whitespaces to the output. You probably tried to align some of the output, but it's simply not worthwhile doing this by hand.

josephnoir:
K, fixed that.

PR #662, opened 2018-02-14
Replace routing in CAF

josephnoir:
This PR makes it easier to review the work of replacing the routing in CAF. It allows review along the way while I continue work in the [topic branch](https://github.com/actor-framework/actor-framework/tree/topic/replace-routing).

### DIFF
--- a/libcaf_io/caf/io/basp/instance.hpp
+++ b/libcaf_io/caf/io/basp/instance.hpp
@@ -45,11 +45,12 @@ namespace basp {
 /// Describes a protocol instance managing multiple connections.
 class instance {
 public:
+  using endpoint_handle = routing_table::endpoint_handle;
   /// Provides a callback-based interface for certain BASP events.
   class callee {
   protected:
     using buffer_type = std::vector<char>;
-    using endpoint_handle = variant<connection_handle, datagram_handle>;
+    using endpoint_handle = instance::endpoint_handle;
   public:
     explicit callee(actor_system& sys, proxy_registry::backend& backend);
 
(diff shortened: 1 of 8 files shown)


Neverlord:
Can you target this PR to an integration branch? In this way I can merge it while you work in background until all is done (and we continue merging chunks into the integration branch as we go along).

josephnoir:
Don't think I get the concept. Should I make PRs from my topic branch into the integration branch? That way I'd have to wait to push new things until you merged the previous commits? ... or do we need three branches? ... or are you simply much faster than me?

mavam:
The idea is as follows: you work in your topic branch as usual, but don't do the PR against master. Instead, you use a separate *integration* branch, which is merged into master at a later point. A common way to do this is create a branch `integration/name-of-topic-branch`. Then perform a no-ff merge into this branch from your topic branch to make it clear you're operating in the integration branch.

Usually, the integration branch acts like a "staging" branch for more complex features. Another way to use an integration branch is if a maintainer wants to perform a substantial pass over the code that requires one or more commits in addition.

josephnoir:
@mavam Thanks for the clarification! I'll close this PR, reset the integration branch to match the master, and open a PR from the topic to the integration branch. (Still think that I can't push new thins into my topic branch while the PR is open, but ... we will see how that works.)

mavam:
>  I'll close this PR, reset the integration branch to match the master, and open a PR from the topic to the integration branch. 

Ah, doesn't github allow you to change the target of the PR? My thought was re-using this PR, but instead of the `master` target, use `integration/replace-routing`.

PR #664, opened 2018-02-15
Remove indirect routing

josephnoir:
First PR on the way to replace routing in CAF.

### DIFF
--- a/libcaf_io/caf/io/basp/instance.hpp
+++ b/libcaf_io/caf/io/basp/instance.hpp
@@ -45,11 +45,12 @@ namespace basp {
 /// Describes a protocol instance managing multiple connections.
 class instance {
 public:
+  using endpoint_handle = routing_table::endpoint_handle;
   /// Provides a callback-based interface for certain BASP events.
   class callee {
   protected:
     using buffer_type = std::vector<char>;
-    using endpoint_handle = variant<connection_handle, datagram_handle>;
+    using endpoint_handle = instance::endpoint_handle;
   public:
     explicit callee(actor_system& sys, proxy_registry::backend& backend);
 
@@ -83,14 +84,8 @@ public:
                          std::vector<strong_actor_ptr>& forwarding_stack,
                          message& msg) = 0;
 
-    /// Called whenever BASP learns the ID of a remote node
-    /// to which it does not have a direct connection.
-    virtual void learned_new_node_directly(const node_id& nid,
-                                           bool was_known_indirectly) = 0;
-
-    /// Called whenever BASP learns the ID of a remote node
-    /// to which it does not have a direct connection.
-    virtual void learned_new_node_indirectly(const node_id& nid) = 0;
+    /// Called whenever BASP learns the ID of a remote node.
+    virtual void learned_new_node(const node_id& nid) = 0;
 
     /// Called if a heartbeat was received from `nid`
     virtual void handle_heartbeat(const node_id& nid) = 0;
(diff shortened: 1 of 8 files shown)


libcaf_io/caf/io/basp/instance.hpp
```diff
@@ -320,37 +315,31 @@ class instance {
...
-        if (tcp_based) {
-          auto ch = get<connection_handle>(path->hdl);
-          write_client_handshake(ctx, callee_.get_buffer(ch),
-                                 hdr.source_node);
-        }
-        callee_.learned_new_node_directly(hdr.source_node, was_indirect);
+        // Add this node to our contacts.
+        CAF_LOG_INFO("new endpoint:" << CAF_ARG(hdr.source_node));
+        tbl_.add(hdl, hdr.source_node);
+        // TODO: Add status, addresses, ... ?
```
Neverlord: This TODO would add code that wasn't there previously. Why do we need to store additional state now?

libcaf_io/caf/io/basp/instance.hpp
```diff
@@ -368,49 +357,35 @@ class instance {
...
-                          && !tbl_.lookup_direct(hdr.source_node));
-          if (new_node) {
-            // add direct route to this node and remove any indirect entry
-            CAF_LOG_INFO("new direct connection:" << CAF_ARG(hdr.source_node));
-            tbl_.add_direct(hdl, hdr.source_node);
-          }
+          // Add this node to our contacts.
+          CAF_LOG_INFO("new endpoint:" << CAF_ARG(hdr.source_node));
+          tbl_.add(hdl, hdr.source_node);
+          // TODO: Add status, addresses, ... ?
```
Neverlord: Same question here. What additional state do we need now that's not in the routing table?

libcaf_io/caf/io/basp/routing_table.hpp
```diff
@@ -126,16 +99,11 @@ class routing_table {
...
   }
 
-  using node_id_set = std::unordered_set<node_id>;
-
-  using indirect_entries = std::unordered_map<node_id,      // dest
-                                              node_id_set>; // hop
-
   abstract_broker* parent_;
   std::unordered_map<endpoint_handle, node_id> direct_by_hdl_;
+  // TODO: Do we need a list as a second argument as there could be
```
Neverlord: For now I think restricting to a single path is the only sane option. A list wouldn't do anyways, since we would have to specify what technologies are preferred (e.g. `Infiniband > TCP > UDP`).

"The first path wins" seems like a reasonable strategy for now. We can expect users to make intelligent deployment decisions and not mix UDP and TCP, for example.

libcaf_io/test/basp.cpp
```diff
@@ -52,7 +52,7 @@ constexpr uint64_t no_operation_data = 0;
 
 constexpr auto basp_atom = caf::atom("BASP");
 constexpr auto spawn_serv_atom = caf::atom("SpawnServ");
-constexpr auto config_serv_atom = caf::atom("ConfigServ");
+// constexpr auto config_serv_atom = caf::atom("ConfigServ");
```
Neverlord: Please don't comment-out things. Either remove or don't.

libcaf_io/test/basp_udp.cpp
```diff
@@ -56,7 +56,7 @@ constexpr uint64_t no_operation_data = 0;
 
 constexpr auto basp_atom = caf::atom("BASP");
 constexpr auto spawn_serv_atom = caf::atom("SpawnServ");
-constexpr auto config_serv_atom = caf::atom("ConfigServ");
+// constexpr auto config_serv_atom = caf::atom("ConfigServ");
```
Neverlord: Same here: remove or don't.

libcaf_io/test/basp_udp.cpp
```diff
@@ -829,6 +828,7 @@ CAF_TEST(actor_serialize_and_deserialize_udp) {
           std::vector<actor_id>{}, msg);
 }
 
+/*
```
Neverlord: Remove if it's obsolete now. Otherwise, please provide comments when this block gets "re-activated".

libcaf_io/caf/io/basp/instance.hpp
```diff
@@ -294,11 +289,11 @@ class instance {
   bool handle(execution_unit* ctx, const Handle& hdl, header& hdr,
               std::vector<char>* payload, bool tcp_based,
               optional<endpoint_context&> ep, optional<uint16_t> port) {
-    // function object for checking payload validity
+    // Function object for checking payload validity.
```
Neverlord: It's good to see your attention to detail. However, all of these changes are ideally wrapped into a single commit. Mixing style changes with functional changes blows up this commit and also blurs the step-by-step evolution a Git history ideally should represent.

libcaf_io/caf/io/basp/instance.hpp
josephnoir: * *Status:* Track if the node is reachable. Can probably be avoided if we assume that its entry in the table means it is known and an associated handle means it is reachable.
* *Address:* I'd like to buffer the addresses of known nodes to make the propagation easier and avoid asking each time a third node requests them.

libcaf_io/caf/io/basp/routing_table.hpp
josephnoir: Sounds reasonable.

PR #667, opened 2018-03-10
Fix actor_system_config move ctor

jsiwek:
Example test case that summarizes the basic sequence of events of how Bro/Broker configure CAF:

```c++
#include <string>
#include <iostream>

#include "caf/all.hpp"
#include "caf/io/all.hpp"

using std::endl;
using std::string;

using namespace caf;

behavior mirror(event_based_actor* self) {
  return {
    [=](const string& what) -> string {
      aout(self) << what << endl;
    }
  };
}

void hello_world(event_based_actor* self, const actor& buddy) {
  self->request(buddy, std::chrono::seconds(10), "Hello World!").then(
    [=](const string& what) {
      aout(self) << what << endl;
    }
  );
}

int main() {
  actor_system_config tmp_cfg;
  tmp_cfg.load<io::middleman>();
  actor_system_config cfg(std::move(tmp_cfg));
  actor_system system{cfg};

  auto res = system.middleman().publish(mirror_actor, 9999, "127.0.0.1", true);

  if ( ! res )
    printf("failed to publish: %s\n", to_string(res).data());

  system.spawn(hello_world, mirror_actor);
  // system will wait until both actors are destroyed before leaving main
}
```

Without the patch:
```
$ ./a.out
failed to publish: !error(41, 'system')
Hello World!
!dlroW olleH
```

With the patch:
```
$ ./a.out
Hello World!
!dlroW olleH
^C
```

(Bro/Broker are basically broken on CAF 0.15.6 or master without this patch.)


Neverlord:
Thanks for the patch!

PR #673, opened 2018-03-22
Fix error propagation of publish and connect

### LINKED ISSUE #672: Fatal error listening on already-in-use port (using openssl)
Using CAF 0.15.7 (or older ones, too) on Linux and this test case:

```c++
#include <string>
#include <iostream>

#include "caf/all.hpp"
#include "caf/io/all.hpp"
#include "caf/openssl/all.hpp"

using std::endl;
using std::string;

using namespace caf;

behavior mirror(event_based_actor* self) {
  return {
    [=](const string& what) -> string {
      aout(self) << what << endl;
    }
  };
}

int main(int argc, char** argv) {
  auto use_ssl = argc > 1;
  printf("using ssl: %d\n", use_ssl);

  actor_system_config cfg;
  cfg.load<io::middleman>();
  cfg.load<caf::openssl::manager>();
  actor_system system{cfg};

  const char* addr = "";
  expected<uint16_t> res = caf::error{};

  if ( use_ssl )
    res = caf::openssl::publish(mirror_actor, 9999, addr, true);
  else
    res = system.middleman().publish(mirror_actor, 9999, addr, true);

  if ( res )
  	  printf("published actor, waiting for peer...\n");
  else
  	  printf("failed to publish actor\n");
}
```

Expected behavior after starting two separate processes of that:

```
./listener 
using ssl: 0
published actor, waiting for peer...
```

```
./listener
using ssl: 0
failed to publish actor
```

Unexpected fatal error when using the openssl path:

```
./listener asdf
using ssl: 1
published actor, waiting for peer...
```

```
/listener asdf
using ssl: 1
published actor, waiting for peer...
epoll_ctl() failed: Bad file descriptor
[FATAL] ../libcaf_io/src/default_multiplexer.cpp:415: critical error: 'epoll_ctl() failed'
Aborted
```

Not a particularly bothersome problem for our actual use-case (we'd just be waiting around for the port to free anyway), though wanted to report in case there's more going on with it.

Hamdor:
closes #672

### DIFF
--- a/libcaf_openssl/src/middleman_actor.cpp
+++ b/libcaf_openssl/src/middleman_actor.cpp
@@ -214,7 +214,7 @@ protected:
                                    uint16_t port) override {
     CAF_LOG_TRACE(CAF_ARG(host) << CAF_ARG(port));
     auto fd = io::network::new_tcp_connection(host, port);
-    if (fd == io::network::invalid_native_socket)
+    if (!fd)
       return std::move(fd.error());
     io::network::nonblocking(*fd, true);
     auto sssn = make_session(system(), *fd, false);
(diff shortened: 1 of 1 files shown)


Neverlord:
Awesome, thanks!

PR #680, opened 2018-03-31
Fix implicit conversion to result<unit_t>

### LINKED ISSUE #679: No implicit conversion of typed_response_promise<unit_t> to result<unit_t>
In 0.15.7, I think due to the changes brought by #652, it is no longer possible to implicitely convert `caf::typed_response_promise<caf::unit_t>` to `caf::result<caf::unit_t>`. This means that this common pattern is no longer possible:

```
caf::behavior {
	[] () -> caf::result<caf::unit_t> {
		auto rp = make_response_promise<caf::unit_t>();
		// ... do something with rp ...
		return rp;
	}
}
```

In 0.15.7, you need to replace the return with an explicit cast `return caf::result<caf::unit_t>(rp);`

I fixed the issue locally for us by commenting out the `caf::result<unit_t>` template specialization in result.hpp and letting the generic `caf::result<Ts...>` template take care of `unit_t`s, but in doing so I might have broken some assumptions about the implicit convertibility between `caf::result<void>` and `caf::result<unit_t>` which we don't use.

Hamdor:
closes #679

### DIFF
--- a/libcaf_core/test/unit.cpp
+++ b/libcaf_core/test/unit.cpp
@@ -28,17 +28,23 @@
 
 using namespace caf;
 
-using unit_res_atom = atom_constant<atom("unitRes")>;
-using void_res_atom = atom_constant<atom("voidRes")>;
-using unit_raw_atom = atom_constant<atom("unitRaw")>;
-using void_raw_atom = atom_constant<atom("voidRaw")>;
+using unit_res_atom   = atom_constant<atom("unitRes")>;
+using void_res_atom   = atom_constant<atom("voidRes")>;
+using unit_raw_atom   = atom_constant<atom("unitRaw")>;
+using void_raw_atom   = atom_constant<atom("voidRaw")>;
+using typed_unit_atom = atom_constant<atom("typedUnit")>;
 
-behavior testee(event_based_actor*) {
+behavior testee(event_based_actor* self) {
   return {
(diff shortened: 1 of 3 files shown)


Hamdor:
@Neverlord, I added some more constructors to `result<void>`, could you please check if this is correct?

libcaf_core/test/unit.cpp
```diff
@@ -28,17 +28,23 @@
...
 
-using unit_res_atom = atom_constant<atom("unitRes")>;
-using void_res_atom = atom_constant<atom("voidRes")>;
-using unit_raw_atom = atom_constant<atom("unitRaw")>;
-using void_raw_atom = atom_constant<atom("voidRaw")>;
+using unit_res_atom   = atom_constant<atom("unitRes")>;
+using void_res_atom   = atom_constant<atom("voidRes")>;
+using unit_raw_atom   = atom_constant<atom("unitRaw")>;
+using void_raw_atom   = atom_constant<atom("voidRaw")>;
+using typed_unit_atom = atom_constant<atom("typedUnit")>;
```
Neverlord: While aligning the equal signs certainly looks a bit nicer, it's an unnecessary change at the end of the day that only clutters the Git history. Just for future reference.

Neverlord:
LGTM, thanks a lot!

PR #685, opened 2018-04-24
Improve workaround for older libstdc++ versions

jsiwek:
Checking just for GCC version is not sufficient as the problem lies
with whether libstdc++ (usable from Clang) implements the required
vector::erase(const_iterator).

The easiest way to check for a valid libstdc++ version at the moment
seems to be via a compilation check at CMake's configure-time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,6 +406,20 @@ else()
   set(CAF_USE_ASIO_INT -1)
 endif()
 
+# GCC (libstdc++) < 4.9 has a broken STL: vector::erase accepts iterator
+# instead of const_iterator.
+# TODO: remove when dropping support for GCC 4.8.
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles("
+#include <vector>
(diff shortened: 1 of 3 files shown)


Neverlord:
Involving CMake in this workaround is really heavyweight. Let me try if I can solve this with a lightweight SFINAE workaround instead.

Neverlord:
A lightweight solution to me fixes an issue in the nearest possible scope. This solution escalates a fix all the way up to the build system and touches three files. A SFINAE solution can work around a broken STL version with standard C++11 and without touching another file.

Workarounds like this are a pain to maintain. I annotate any code smell like this with a `TODO` that states when the line can get removed. With light- or heavyweight I essentially mean how far a patch spreads throughout the code base and whether or not it introduces dependencies that make maintaining the code base (and keeping it lean and clean) harder.

https://github.com/actor-framework/actor-framework/commit/e339df03ab24159d8060c0fd913fea7579311f64 solves this with a simple SFINAE check now.

PR #687, opened 2018-04-27
Improve cross compilation support

jsiwek:
It's not possible to do CMake checks that require executing something
on the host that was built for the target platform.


Neverlord:
Thanks for the patch!

PR #690, opened 2018-05-08
Pull in a missing lib file on Windows

machonky:
Fix missing lib linking error when building with caf_core_static.lib

### DIFF
--- a/libcaf_core/src/get_mac_addresses.cpp
+++ b/libcaf_core/src/get_mac_addresses.cpp
@@ -167,6 +167,8 @@ std::vector<iface_info> get_mac_addresses() {
 #include <iterator>
 #include <algorithm>
 
+#pragma comment(lib, "Iphlpapi.lib")
+
 namespace {
 
 constexpr size_t working_buffer_size = 15 * 1024; // 15kb by default


Neverlord:
We deliberately don't use `pragma comment`. Managing dependencies is the responsibility of the build system.

We already [add Iphlpapi in CMake](https://github.com/actor-framework/actor-framework/blob/master/CMakeLists.txt#L328) as a link-time dependency. What issues do you have and what is your setup?

machonky:
Windows CMake build was super smooth and uneventful - nice work! After successfully building libs and examples with CMake on VS2017, I got linking errors with my own applications requiring Iphlapi.lib. Since CAF always required it for windows, I thought it best to automatically link it, rather than explicitly add it every time in my own projects. Just making a better windows development experience.

Neverlord:
> I thought it best to automatically link it, rather than explicitly add it every time in my own projects

I know that building C++ projects is a pain. Still, resolving dependencies is the job of the build system and should not leak into the code. I think we should add a Wiki and/or manual entry dedicated to this topic.

machonky:
Agreed. Documentation would be the best solution on this.

PR #691, opened 2018-05-08
Fix update message in config server

josephnoir:
If I'm not mistaken the `msg` should be sent to the subscriber not the subscriber set. Since this has probably been there forever here is a PR for that and not just a push.


Neverlord:
Seems like this is unused. Do you have a test for this or how did you stumble across it?

josephnoir:
For the peer address exchange with automatic connections I need similar functionality. A `get` that creates a sub if no entry is present and unsubs after an entry was added and delivered to the sub. Also, it should sent the same message in both cases. I used the config server implementation as a basis and noticed that it delivered unexpected messages.

Neverlord:
> I need similar functionality

An actor for modeling a key/value store strikes me as something we should eventually factor out into some utility library (such as the envisioned `cabb`). Anyway, thanks for the fix.

PR #694, opened 2018-05-10
Fix a file descriptor leak in get_mac_addresses()

jsiwek:
Probably even better to re-use the socket_guard class to ensure it gets closed on all code paths, including exceptions, though unsure how exactly you would have liked to have that factored out to make it more widely usable.


Neverlord:
> Probably even better to re-use the socket_guard class

Agreed. That's something we can address later, though. Thanks for the patch!

PR #700, opened 2018-06-15
Add OpenSSL locking functions

jsiwek:
The OpenSSL docs for CRYPTO_set_locking_callback seem to describe what I
observe: "Multi-threaded applications will crash at random if it is not
set".

### DIFF
--- a/libcaf_openssl/src/manager.cpp
+++ b/libcaf_openssl/src/manager.cpp
@@ -23,6 +23,9 @@ CAF_PUSH_WARNINGS
 #include <openssl/ssl.h>
 CAF_POP_WARNINGS
 
+#include <vector>
+#include <mutex>
+
 #include "caf/expected.hpp"
 #include "caf/actor_system.hpp"
 #include "caf/scoped_actor.hpp"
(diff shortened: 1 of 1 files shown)


jsiwek:
For a more specific reproduction case, I had been testing this code (actually trying to reproduce a different problem, so it's maybe not minimal):

```c++
#include "caf/all.hpp"
#include "caf/io/all.hpp"
#include "caf/openssl/all.hpp"

using namespace caf;
using namespace std;

behavior mirror(event_based_actor* self)
	{
	return {
	[=](const string& what) -> string
		{
		//aout(self) << "got msg " << what << endl;
		}
	};
	}


behavior core(event_based_actor* self)
	{
	return {
	[=](int i)
		{
		},
	[=](const string& what)
		{
		//aout(self) << "exit core" << endl;
		self->quit(exit_reason::user_shutdown);
		}
	};
	}

int main()
	{
	actor_system_config cfg;
	cfg.load<caf::io::middleman>();
	cfg.load<caf::openssl::manager>();
	actor_system system{cfg};
	auto core_actor = system.spawn(core);
	//auto res = system.middleman().publish(core_actor, 9999,
	//		                              "127.0.0.1", true);
	auto res = openssl::publish(core_actor, 9999, "127.0.0.1", true);

	if ( ! res )
		{
		cout << "failed to publish" << endl;
		return 1;
		}

	auto peer = openssl::remote_actor(system, "127.0.0.1", 9999);

	if ( ! peer )
		{
		cout << "failed to connect" << endl;
		return 1;
		}

	for ( int i = 0; i < 500; ++i )
		{
		anon_send(*peer, i);
		}

		{
		scoped_actor self{system};
		self->send(mirror_actor, "hello");
		self->receive(
			[&](const string& what)
				{
				//aout(self) << "got result " << what << endl;
				}
		);

		self->send_exit(mirror_actor, caf::exit_reason::user_shutdown);
		self->wait_for(mirror_actor);
		}

	anon_send(core_actor, "shutdown");
	return 0;
	}
```

And at least on FreeBSD 11.1-RELEASE-p9, would find that to segfault in libcrypto around 1/1000 runs.  After the patch, I see no crashes after 20k runs.

Neverlord:
Thank you for hunting this bug down and fixing it! OpenSSL is really a pain to get right and work with.

PR #704, opened 2018-06-25
Fix get_mac_addresses() potential buffer over-read

jsiwek:
The macOS/BSD implementation of get_mac_addresses() can check the
link-layer address length early on to filter out interfaces that don't
have a MAC address.  e.g. loopback interfaces don't have one.

A concern is that the sdl_data array of sockaddr_dl contains both the
data for the interface name and link-layer address, but
get_mac_addresses() is always assuming enough space was allocated in the
array to read 6 bytes of the MAC address even when some interfaces
report an address length of 0.  e.g. long interface names without a MAC
address could be a problem.

### DIFF
--- a/libcaf_core/src/get_mac_addresses.cpp
+++ b/libcaf_core/src/get_mac_addresses.cpp
@@ -48,6 +48,9 @@ std::vector<iface_info> get_mac_addresses() {
     }
     auto ifm = reinterpret_cast<if_msghdr*>(buf.data());
     auto sdl = reinterpret_cast<sockaddr_dl*>(ifm + 1);
+    constexpr auto mac_addr_len = 6;
+    if (sdl->sdl_alen != mac_addr_len)
+      continue;
     auto ptr = reinterpret_cast<unsigned char*>(LLADDR(sdl));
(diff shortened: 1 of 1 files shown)


jsiwek:
I didn't actually get a test case that could trigger an over-read and only came upon this while looking into what became a false-positive from AddressSanitizer.

Maybe implementations would allocate 6+ bytes of slop such that the assumption of always being able to read 6-bytes beyond the interface name in `sdl_data` is ok, but the patch at least seems like a conservative thing to do.

jsiwek:
I amended the commit.

I'd agree it's maybe more readable to use an identifier vs. a literal, but beyond that I don't know what goal you're aiming at, so feel free to hunt around for other pre-defined macros that better suit you (IMO, it's more trouble than it's worth).  Also the change I suggest was at least not *introducing* a magic number -- the function already used '5' in the for loop, so kind of feels like a separate/pre-existing issue that's "not my problem" :)

Neverlord:
Sins of the past are no excuse for committing new ones.

Anyway, the current version LGTM. Thanks for the patch!

PR #715, opened 2018-06-29
Simplify UDP test, use dynamic ports

josephnoir:
This PR simplifies a UDP test that regularly failed on our Jenkins Windows VMs.


libcaf_io/test/remote_actor_udp.cpp
```diff
@@ -207,7 +207,12 @@ CAF_TEST(multiple_endpoints_udp) {
       }
     };
   });
-  server_sys.middleman().publish_udp(mirror, 12345);
+  // Publish server actor.
+  auto eport = server_sys.middleman().publish_udp(mirror, 0);
```
Neverlord: Simpler:

`auto port = unbox(server_sys.middleman().publish_udp(mirror, 0));`

libcaf_io/test/remote_actor_udp.cpp
```diff
@@ -220,32 +225,25 @@ CAF_TEST(multiple_endpoints_udp) {
...
-  // setup client a
-  CAF_MESSAGE("creating first client");
-  config client_cfg;
-  actor_system client_sys{client_cfg};
-  auto client = client_sys.spawn(client_fun);
-  // acquire remote actor from the server
-  auto client_srv = client_sys.middleman().remote_actor_udp("localhost", 12345);
-  CAF_REQUIRE(client_srv);
-  // setup other clients
-  for (int i = 0; i < 5; ++i) {
```
Neverlord: Seems like the unit test now reduced coverage by using a single client instead of 5. What's the motivation for this change? More is better, when it comes to testing. 😉

josephnoir: It looks like the tests take too long I wanted to check if this makes a difference. Then I remembered that our PR builder doesn't build on Windows. I'm not sure what the good solution there is as I can't reproduce the error manually.

libcaf_io/test/remote_actor_udp.cpp
josephnoir: It seems to be local to the [config_option test](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/test/config_option.cpp#L91).

libcaf_io/test/remote_actor_udp.cpp
josephnoir: Just saw your comment in Gitter. I'll update the PR.

libcaf_io/test/remote_actor_udp.cpp
Neverlord: Nope, just include `dsl.hpp`: https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/dsl.hpp#L627

Neverlord: Sorry, my bad. You're not using `test_coordinator_fixture`.

Issue #710, opened 2018-06-27
Clean up default_multiplexer header


Neverlord:
The `default_multiplexer` has many classes that we will touch during the I/O redesign. So the first step is separating all the classes into individual headers for a fresh and clean setup.

Issue #696, opened 2018-05-11
Move socket_guard to namespace detail and apply it to more use cases


Neverlord:
We should move `socket_guard` to namespace `detail` to re-use it in more places. See  #694 for the original discussion.

Neverlord:
#716 already took care of that.

PR #716, opened 2018-07-02
Split up default multiplexer file

josephnoir:
This PR addresses #710. It adds a lot of files. I'm not sure about the code organization. Does a policy namespace make sense, four namespaces seems a little "deep".


Neverlord:
Good to see this moving forward. 👍

> Does a policy namespace make sense

Good question. The problem with the current `policy` namespace is that it can be ambiguous. `policy::foo` can now either be `caf::policy::foo` or (if you happen to be in namespace `caf::io::network`) `caf::io::network::policy::foo`.

So, if we leave `caf::io::network::policy` then we should always qualify any policy from there with `network::policy::` (even when in namespace `caf::io::network`) to make sure statements are never ambiguous with `caf::policy`.

We also could flatten our namespaces and move all policies always to `caf::policy` instead. In this case we should also get rid of `caf::opencl::detail`.

@josephnoir, @mavam: what do you guys think? Should we put up with deep namespace nesting for more structure or flatten the namespaces and move every policy class regardless of the module it's defined in to `caf::policy`?

mavam:
> We also could flatten our namespaces and move all policies always to `caf::policy` instead. In this case we should also get rid of `caf::opencl::detail`

This would be my favorite.

josephnoir:
I'd prefer network related code to be placed in the `network` namespace. Granted, the policies in this PR just bundle a few functions in the policy class, but going forward policies might gain more functionality and implement part of the network functionality themselves.

Since having a simpler hierarchy is also a value I'm just slightly leaning towards the deeper namespace.

Neverlord:
> implement part of the network functionality themselves.

Absolutely. However, should we take this point into consideration? The policy only makes sense when plugging it into the right component. Just like a mixin, policies draw in as many dependencies as they need to fit their role.

If we would nest policies consequently, then we would need to move several classes. For example, `caf::policy::work_stealing` would need to go to `caf::scheduler::policy::work_stealing`. I'm more afraid of fragmenting namespaces and introducing complex structure than having dependencies between related components "across namespaces".

I think we should strive for simple rules. We already have quite a handful of namespaces:

- `caf` all non-optional, user-facing parts of the API (we *could* consider an inline namespace `caf::core` to add a bit more structure without forcing users to type more than necessary)
- `caf::detail` private implementation details, not safe for third parties to use
- `caf::intrusive` custom container types with intrusive node API
- `caf::scheduler` all parts of the cooperative scheduler module
- `caf::policy` policy classes for configuring other components
- `caf::mixin` utility classes for sharing code without common base types
- `caf::meta` optional annotations when implementing `inspect` functions
- `caf::io` user-facing parts of the I/O module
- `caf::io::basp` CAF's binary wire protocol
- `caf::io::network` asynchronous, actor-oriented network I/O
- `caf::io::filesystem` asynchronous, actor-oriented file I/O (not yet, but likely addition)
- `caf::opencl` user-facing API for spawning CL actors
- `caf::openssl` user-facing API of the OpenSSL module
- `caf::python` user-facing API for the Python binding
- `caf::test` unit test framework

Plus two under-developed namespaces that we either should get rid of or use consistently: `caf::tag`, `caf::decorator`.

All in all, I think we should consolidate `caf::opencl::detail` and `caf::io::network::policy`. @josephnoir I hope you're OK with that, since you're only slightly leaning towards the deeper namespace. Mind making the changes? 🙂

josephnoir:
Sure. Hope I didn't miss something.

josephnoir:
Something with our Jenkins conf is off, if you look at the [build](https://jenkins.inet.haw-hamburg.de/blue/organizations/jenkins/CAF%2Factor-framework/detail/PR-716/2/pipeline) all builds succeeded.

josephnoir:
Jenkins is finally happy. 😺

libcaf_io/src/default_multiplexer.cpp
```diff
@@ -990,341 +663,7 @@ int64_t default_multiplexer::next_endpoint_id() {
...
-    writing_ = false;
-    backend().del(operation::write, fd(), this);
-  } else {
-    wr_buf_.swap(wr_offline_buf_.front());
-    wr_offline_buf_.pop_front();
-  }
-}
+// -- Related helper functions -------------------------------------------------
 
 class socket_guard {
```
Neverlord: Since we're already touching this: this is a good opportunity to move the socket guard out of here. See https://github.com/actor-framework/actor-framework/issues/696.

libcaf_io/src/socket_utils.cpp
```diff
@@ -0,0 +1,369 @@
...
+    default:
+      break;
+  }
+  CAF_CRITICAL("invalid protocol family");
+}
+
+} // namespace network
+} // namespace io
+} // namespace caf
+
```
Neverlord: Extra newline

libcaf_io/src/stream.cpp
```diff
@@ -0,0 +1,207 @@
+
```
Neverlord: Extra newline.

libcaf_io/src/stream.cpp
```diff
@@ -0,0 +1,207 @@
...
+  close_read_channel();
+  passivate();
+}
+
+void stream::removed_from_loop(operation op) {
+  CAF_LOG_TRACE(CAF_ARG(op));
+  switch (op) {
+    case operation::read:  reader_.reset(); break;
+    case operation::write: writer_.reset(); break;
+    case operation::propagate_error: break;
```
Neverlord: The last `case` doesn't need a `break`.

libcaf_io/src/tcp.cpp
```diff
@@ -0,0 +1,76 @@
...
+using caf::io::network::native_socket;
+using caf::io::network::no_sigpipe_io_flag;
+
+namespace caf {
+namespace policy {
+
+rw_state tcp::read_some(size_t& result, native_socket fd, void* buf,
+                        size_t len) {
+  CAF_LOG_TRACE(CAF_ARG(fd) << CAF_ARG(len));
+  auto sres = ::recv(fd, reinterpret_cast<io::network::socket_recv_ptr>(buf),
```
Neverlord: `recv` is used without including the right header (we shouldn't rely on other headers to include them for us).

libcaf_io/src/tcp.cpp
```diff
@@ -0,0 +1,76 @@
...
+    return rw_state::failure;
+  }
+  result = (sres > 0) ? static_cast<size_t>(sres) : 0;
+  return rw_state::success;
+}
+
+rw_state tcp::write_some(size_t& result, native_socket fd, const void* buf,
+                         size_t len) {
+  CAF_LOG_TRACE(CAF_ARG(fd) << CAF_ARG(len));
+  auto sres = ::send(fd, reinterpret_cast<io::network::socket_send_ptr>(buf),
```
Neverlord: `send`: missing include.

libcaf_io/src/tcp.cpp
```diff
@@ -0,0 +1,76 @@
...
+    return rw_state::failure;
+  result = (sres > 0) ? static_cast<size_t>(sres) : 0;
+  return rw_state::success;
+}
+
+bool tcp::try_accept(native_socket& result, native_socket fd) {
+  using namespace io::network;
+  CAF_LOG_TRACE(CAF_ARG(fd));
+  sockaddr_storage addr;
+  memset(&addr, 0, sizeof(addr));
```
Neverlord: `memset`: missing include.

libcaf_io/src/tcp.cpp
```diff
@@ -0,0 +1,76 @@
...
+  socklen_t addrlen = sizeof(addr);
```
Neverlord: `socklen_t`: missing include, but we should use `socket_size_type` anyways.

libcaf_io/caf/io/network/socket_utils.hpp
```diff
@@ -0,0 +1,165 @@
...
+namespace io {
+namespace network {
+
+  // annoying platform-dependent bootstrapping
+#ifdef CAF_WINDOWS
+  using setsockopt_ptr = const char*;
+  using getsockopt_ptr = char*;
+  using socket_send_ptr = const char*;
+  using socket_recv_ptr = char*;
+  using socklen_t = int;
```
Neverlord: Defining UNIX types was OK in a `.cpp` file where the symbols wouldn't leak out. However, here we should not "overload" those names.

I propose `socklen_t => socket_size_type` and `ssize_t => signed_size_type`.

libcaf_io/src/event_handler.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // enable nonblocking IO, disable Nagle's algorithm, and suppress SIGPIPE
+  nonblocking(fd_, true);
+  tcp_nodelay(fd_, true);
+  allow_sigpipe(fd_, false);
+}
+
+} // namespace network
+} // namespace io
+} // namespace caf
+
```
Neverlord: Extra newline.

libcaf_io/src/pipe_reader.cpp
```diff
@@ -0,0 +1,68 @@
...
+    : event_handler(dm, invalid_native_socket) {
+  // nop
+}
+
+void pipe_reader::removed_from_loop(operation) {
+  // nop
+}
+
+resumable* pipe_reader::try_read_next() {
+  intptr_t ptrval;
```
Neverlord: Missing headers for `intptr_t`, `recv` and `read`.

libcaf_io/src/udp.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#include "caf/policy/udp.hpp"
+
+#include "caf/logger.hpp"
+#include "caf/io/network/socket_utils.hpp"
+
+#ifdef CAF_WINDOWS
+using caf::io::network::ssize_t;
```
Neverlord: See previous comment, should become superfluous after defining `signed_size_type`.

libcaf_io/caf/io/network/socket_utils.hpp
josephnoir: Do we have to defined these ourselves or which header do we have to include here?
A comment next to a win header reads `// socket_size_type, etc. (MSVC20xx)`, but I wouldn't know where to get them on a unix system.

josephnoir:
Jenkins is happy again. Mind taking another look?

Neverlord:
@josephnoir I did a bit more cleanup of the default multiplexer. There's still room for improvement, but it's a good-enough scaffold for now. If Jenkins is still happy after my commits you can go ahead and merge. 👍

PR #717, opened 2018-07-02
Fix "unused" warnings when using OpenSSL 1.1

jsiwek:
Per the OpenSSL CHANGES file:

* OpenSSL 1.1 no longer requires setting locking callbacks and
  previous functions are replaced with no-op compatibility macros.

* SSL_CTX_set_ecdh_auto() is removed (replaced with no-op macro) and
  ECDH support is always enabled by default.


Neverlord:
Thanks for the patch. 👍

Good to see that OpenSSL finally recognized that thread-safety is no longer optional. A bit late for my taste, but better late than never.

Issue #718, opened 2018-07-04
CMake whitespace error CMP0004 with CentOS 7 rpmbuild system LDFLAGS


antcodd:
CAF does not build using the LDFLAGS supplied by rpmbuild in CentOS 7 because of trailing whitespace and CMake policy CMP0004.

This is similar to #653, but instead with externally added flags. CMakelists.txt has also changed significantly since that issue.

After some investigation this is because of the way RPM (4.11.3) itself is constructing the LDFLAGS supplied with ``%configure``:
```
$ rpm --showrc | grep relro
-14: __global_ldflags   -Wl,-z,relro %{_hardened_ldflags}
```
where ``%{_hardened_ldflags}`` is not set.

For context I was attempting to build a local RPM of Bro git master which currently builds libcaf master as a 3rdparty submodule. 

I have since discovered the slightly outdated OBS CAF repository with spec file. That doesn't use %configure so the default system LDFLAGS aren't set, presumably because of a lack of supported arguments. A quick hack forward-port seems to build on CentOS 7.4 now (GCC 4.8.5), except for some packaged file issues. As a separate issue it might make sense to update that and pull in parts of the %configure macro. Bro requires caf 0.15.5 so I can't use the prebuilt versions.

```
********** Begin CAF External Project CMake Output ************

-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler: /bin/cc
-- Check for working C compiler: /bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /bin/c++
-- Check for working CXX compiler: /bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Found g++ version 4.8
-- Found OpenSSL: /usr/lib64/libssl.so;/usr/lib64/libcrypto.so (found version "1.0.2k") 
-- Performing Test OPENSSL_CORRECT_VERSION_NUMBER
-- Performing Test OPENSSL_CORRECT_VERSION_NUMBER - Success
-- Found Doxygen: /bin/doxygen (found version "1.8.5") 
-- 
====================|  Build Summary  |====================

CAF version:           0.15.7

Build type:            RelWithDebInfo
Build static:          no
Build static only:     no
Build static runtime:  no
Runtime checks:        no
Log level:             none
With mem. mgmt.:       yes
With exceptions:       yes

[...]

PR #720, opened 2018-07-04
Fix LDFLAG handling with whitespaces

josephnoir:
It seems cmake automatically populates the `CMAKE_{EXE,MODULE,STATIC,SHARED}_LINKER_FLAGS` using the LDFLAGS environment variable. Adding those a second time via [target_link_libraries](https://github.com/actor-framework/actor-framework/blob/master/CMakeLists.txt#L604) seems unnecessary.

The only way I found to print them is using one of the variables above. I picked one and added the "shared" hint to avoid possible confusion. Tested this with cmake 2.8.12 and 3.7.2. 

Should our Cmake requirement someday meet 3.12.something we could use [add_link_options](https://cmake.org/cmake/help/git-stage/command/add_link_options.html) instead of passing LDFLAGS ourselves.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,12 +129,6 @@ else()
 endif()
 
 ################################################################################
-#                        respect environment variables                         #
-################################################################################
-
-set(LDFLAGS $ENV{LDFLAGS})
-
-################################################################################
 #                              get version of CAF                              #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


antcodd:
What about the references to $LDFLAGS in the subdirectory CMakeLists.txts?

Seems to successfully configure now though. I now have issues with CXXFLAGS, will comment in #718 about it.

antcodd:
I think this PR was incomplete, sorry if it wasn't clear. Doesn't $LDFLAGS need to be renamed to $CAF_EXTRA_LDFLAGS in the subproject targets as well (such as libcaf_io/CMakeLists.txt) for the extra flags to be used? Compile on windows is probably broken.

josephnoir:
The subprojects also use the variable LD_FLAGS which, I guess, was named wrong previously. I'll take a look at that and the problem with CXX FLAGS you mentioned in #718.

PR #726, opened 2018-07-09
Add URI class, close #708

### LINKED ISSUE #708: Implement URI class
We need an URI support for technology-independent addressing in the middleman.

Neverlord:
An URI is modeled as immutable value type that is optimized for message passing, i.e., it is cheap to copy.

- [x] Add classes for representing IP, because an URI can contain an IP address as host
- [x] Implement URI class with an opaque PIMPL (and builder class)
- [x] Implement RFC 3986 conformant URI parser
- [x] Make URI serializable

@josephnoir I got only one TODO left (serialization), so you can go ahead and review all finished parts.

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -122,6 +122,9 @@ set(LIBCAF_CORE_SRCS
   src/type_erased_value.cpp
   src/uniform_type_info_map.cpp
   src/unprofiled.cpp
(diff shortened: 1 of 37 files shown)


Neverlord:
> I'm not sure about the compare of the string view.

You were spot on! Fixing the `compare` also revealed a lot of bugs that needed fixing, but the PR should be good to go now.

PR #731, opened 2018-07-16
Call remove_from_loop in manager::detach

Neverlord:
Relates #695.

@jsiwek I've consolidated all `io_failure` implementations into a single one (all implementations were identical) and call `remove_from_loop` in `detach` (whether or not detaching from parent actually succeeds). This *should* make sure the sockets get removed on all code paths. Can you still reproduce hangs with this patch?


jsiwek:
With this patch, I don't reproduce any hangs using the test case from #695, thanks.

PR #732, opened 2018-07-16
Remove CAF_EXP_THROW and simplify CAF_RAISE_ERROR

Neverlord:
The old `CAF_EXP_THROW` is obsolete because of the much nicer `unbox` utility in the testing DSL. Further, `CAF_RAISE_ERROR` can simply take a C string instead of doing the `std::string` conversions.

Getting rid of macros also feels really good. 😁


Neverlord:
@josephnoir Jenkins is finally happy

libcaf_opencl/src/manager.cpp
```diff
@@ -169,12 +161,12 @@ program_ptr manager::create_program(const char* kernel_source,
...
       // pfn_notify callback, but we can get
       // the build log
-#ifndef __APPLE__
+#ifndef CAF_MACOS
       CAF_LOG_ERROR(CAF_ARG(ss.str()));
 #endif
       oss << endl << ss.str();
     }
-    CAF_RAISE_ERROR(oss.str());
+    CAF_RAISE_ERROR("clBuildProgram failed");
```
josephnoir: If I'm not mistaken `oss` is never read now and should be removed.

Neverlord: Fixed.

PR #736, opened 2018-07-20
Fix data buffered in OpenSSL not being flushed

jsiwek:
Reading through the two commits should be straight-forward and provide better detail, but generally: one commit adds a test case that shows the undesired behavior (the restricted_ping_pong test case should hang/timeout) and the other is a proposed fix for the underlying problem of data not getting properly flushed from internal OpenSSL buffers.

### DIFF
--- a/libcaf_openssl/caf/openssl/session.hpp
+++ b/libcaf_openssl/caf/openssl/session.hpp
@@ -59,6 +59,7 @@ public:
   bool try_connect(native_socket fd);
   bool try_accept(native_socket fd);
   const char* openssl_passphrase();
+  int pending_read_fd();
 
 private:
   rw_state do_some(int (*f)(SSL*, void*, int), size_t& result, void* buf,
@@ -74,6 +75,8 @@ private:
   std::string openssl_passphrase_;
   bool connecting_;
   bool accepting_;
+  bool had_pending_;
+  std::pair<native_socket, native_socket> pipe_fds_;
 };
 
 /// @relates session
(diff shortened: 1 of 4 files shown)


Neverlord:
Thank you for digging into this! To summarize the issue: data can get lost if a broker fails to read all data that's buffered outside of the socket buffer.

Your patch essentially bypasses the `max-consecutive-reads` parameter by trigger the broker from another I/O handler again, but only for the OpenSSL scribe. I think this is a general problem that we need to solve once instead of relying on individual scribes to deploy workarounds. Otherwise, another scribe (e.g. a QUIC scribe) will have the same issue sooner or later. That's why I don't merge this PR.

I will solve this one layer above by allowing scribes to signal brokers when they must read more data even if `max-consecutive-reads` was reached.

jsiwek:
Thanks for the feedback -- I agree you'd likely run into this type of situation again anytime the data read off the socket buffer gets transformed in some way and doesn't fit inside the supplied `stream::rd_buf_` buffer.  If there's not any way to signal that the last read attempt in the `max-consecutive-reads` loop failed to fit the transformed data into the supplied buffer, then there's a problem because the event loop won't call us back again in the case the actual socket buffer has been emptied.

Let me know when you've added such a signaling mechanism or want any help testing (probably the test case from this PR is still a good enough check, though I don't know why Jenkins failed to build it -- seemed to work for me even on matching gcc 4.8.4).

Neverlord:
> Let me know when you've added such a signaling mechanism or want any help testing

I will get in touch, additional testing would be most welcome.

> probably the test case from this PR is still a good enough check

Yes, that's a very good starting point. I only need to port it to the latest testing API, since `CAF_EXP_THROW` was removed recently.

PR #740, opened 2018-07-28
Allow policies to signal internal buffering

Neverlord:
Followup to #736. This patch extends I/O policies to signal upstream when data must get read even after reaching the `max-consecutive-reads` threshold.

@jsiwek Setting `max-consecutive-reads` to 1 reliably reproduces the reported issue. This patch solves it for me, it'd be great if you could test this in your setup as well.


Neverlord:
Yes, I've seen it. Sorry for not waiting on the test results in the first place.

Neverlord:
@jsiwek Jenkins seems happy. Mind giving it another try?

jsiwek:
Thanks, our tests look stable w/ this now.

PR #744, opened 2018-08-07
Fix the "quiet" log level

jsiwek:
The "first line" and "last line" log messages were causing a log to be generated because they only checked whether CAF_LOG_LEVEL was defined, but it's now always defined.


Neverlord:
Thanks a lot! 👍

PR #745, opened 2018-08-08
Make container streambuffer more flexible

mavam:
This PR improves the interface of the container streambuffer. Most notably, we can now reassign the get area. We also hide the `sputn` implementation, because the default has an unnecessary overflow check, but the put area never overflows (because it does not exist). Also, we can now seek on the container streambuffer get area, which required factoring the seek functions.


Neverlord:
@mavam sorry, that was a bad timing to branch off---and a reminder to me that pushing into master is a bad habit of mine. Please rebase to current master to get this properly tested by Jenkins.

mavam:
Done. Let's wait for Jenkins.

PR #747, opened 2018-08-20
Remove <memory> include from expected.hpp

tobim:


tobim:
Oops, don't know how I missed that.

PR #751, opened 2018-09-01
Fix segfault on changing to empty behavior

### LINKED ISSUE #750: Segfault while receiving unhandled message after changing behaviour using become
Changing the behaviour of an actor results in segfault or in unhandled exception St9_bad_alloc. 
Caf version: 0.15.5
Minimal 'working' example:

```cpp
#include  <caf/all.hpp>

using namespace caf;

behavior actorTwo(event_based_actor* self) {
  return {};
}
behavior actorOne(event_based_actor* self) {
  return {[=](const std::string& msg) { self->become(actorTwo(self)); }};
}
void caf_main(actor_system& system, const actor_system_config& cfg) {
  auto one = system.spawn(actorOne);

  scoped_actor sc{system};
  sc->send(one, "Hey");
}
CAF_MAIN()
```

Hamdor:
With this commit, empty behaviors are not added to the behavior stack.
Closes #750

### DIFF
--- a/libcaf_core/src/scheduled_actor.cpp
+++ b/libcaf_core/src/scheduled_actor.cpp
@@ -768,7 +768,8 @@ void scheduled_actor::do_become(behavior bhvr, bool discard_old) {
   if (discard_old && !bhvr_stack_.empty())
     bhvr_stack_.pop_back();
   // request_timeout simply resets the timeout when it's invalid
-  bhvr_stack_.push_back(std::move(bhvr));
+  if (bhvr)
+    bhvr_stack_.push_back(std::move(bhvr));
   set_receive_timeout();
 }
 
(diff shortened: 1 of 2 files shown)


Neverlord:
Thanks! 👍

Neverlord:
Yes. We could discuss whether setting an empty behavior should have the same effect as calling `unbecome`. I think it's a weird behavior either way, but for now I prefer at least not segfaulting in this case.

josephnoir:
Fixing the segfault is the right move, no argument here. But I have a hard time figuring out what I would expect to happen. 

* Calling `unbecome` seems somewhat unexpected. (Maybe only at first?)
* Transforming the actor into a sink seems weird because you can't exit the state anymore. It would simply wait for all its references to go out of scope and drop all messages.

Either way, this sounds like something we should address and document.

PR #754, opened 2018-09-07
Make build_config.hpp actually part of the build

tobim:
This also moves build_config into the libcaf_core target.


.gitignore
```diff
@@ -12,7 +12,6 @@ manual
 *.swp
 bii/*
 .idea/
-libcaf_core/caf/detail/build_config.hpp
```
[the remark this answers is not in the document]
tobim: This commit moves the destination of that file into the build tree. This path doesn't exit any longer.

libcaf_core/CMakeLists.txt
```diff
@@ -131,33 +127,64 @@ set(LIBCAF_CORE_SRCS
...
                         PROPERTIES
                         SOVERSION ${CAF_VERSION}
                         VERSION ${CAF_VERSION}
                         OUTPUT_NAME caf_core)
-  if (CYGWIN)
-    install(TARGETS libcaf_core_shared RUNTIME DESTINATION bin)
-  else()
-    install(TARGETS libcaf_core_shared LIBRARY DESTINATION lib)
-  endif()
+  install(TARGETS libcaf_core_shared
```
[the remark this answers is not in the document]
tobim: Unfortunately I don't have a setup where I can test it atm, but [according to the documentation](https://cmake.org/cmake/help/v3.0/command/install.html) the `if-then-else` here was completely superfluous.

tobim:
The intention here is to fix the following bug in the build process:

``` sh
cmake -Bbuild_info -H. -DCAF_LOG_LEVEL=2
cmake -Bbuild_trace -H. -DCAF_LOG_LEVEL=4
cmake --build build_info  # <- the libraries will be built with log level TRACE
```

libcaf_core/CMakeLists.txt
```diff
@@ -131,33 +127,64 @@ set(LIBCAF_CORE_SRCS
...
+# configure build_config.hpp header
+macro(to_int_value name)
+  if(${name})
+    set(${name}_INT 1)
+  else()
+    set(${name}_INT -1)
+  endif()
+endmacro()
+
+to_int_value(CAF_NO_EXCEPTIONS)
```
Neverlord: Actually, this entire `to_int_value` stuff is obsolete since 3be2020441de126661cba825e485aa3c4e1c1258. Rather than moving it from one CMake file to the other, we can just drop it.

CMakeLists.txt
```diff
@@ -365,37 +364,18 @@ if (DEFINED CMAKE_LD_LIBS)
...
-               "${CMAKE_CURRENT_SOURCE_DIR}/libcaf_core/caf/detail/build_config.hpp"
-               IMMEDIATE @ONLY)
-
-
 ################################################################################
 #                           setup for install target                           #
 ################################################################################
 
 # install includes from core
-install(DIRECTORY libcaf_core/caf/
```
Neverlord: I could get behind moving the `install` statements to the submodule CMake files, but only if we're consistent with it. This moves the install for the core module, but leaves everything else in the root setup.

libcaf_core/CMakeLists.txt
```diff
@@ -131,33 +127,64 @@ set(LIBCAF_CORE_SRCS
...
+  )
   set_target_properties(libcaf_core_static PROPERTIES OUTPUT_NAME caf_core_static)
   install(TARGETS libcaf_core_static ARCHIVE DESTINATION lib)
   add_dependencies(libcaf_core_static libcaf_core)
 endif ()
 
-link_directories(${LD_DIRS})
-include_directories(. ${INCLUDE_DIRS})
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/caf"
+        DESTINATION include
```
Neverlord: Aside from the discussion above (either move it back or consistently place the installs in the submodule CMake files), this now installs all CAF headers into the wrong destination. Should read `DESTINATION include/caf`.

CMakeLists.txt
```diff
@@ -98,7 +98,6 @@ if(CAF_USE_CCACHE)
   if(CCACHE_PROGRAM)
     message(STATUS "Using ccache command: ${CCACHE_PROGRAM}")
     set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "${CCACHE_PROGRAM}")
```
Neverlord: Seems like an orthogonal concern to the purpose of this commit/PR.

libcaf_core/CMakeLists.txt
```diff
@@ -1,10 +1,6 @@
...
+cmake_minimum_required(VERSION 2.8.12)
 project(caf_core C CXX)
 
-set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR})
-
-include_directories(.)
-
 # get header files; only needed by CMake generators,
 # e.g., for creating proper Xcode projects
 file(GLOB_RECURSE LIBCAF_CORE_HDRS "caf/*.hpp")
```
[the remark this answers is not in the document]
tobim: it doesn't.

CMakeLists.txt
tobim: I was hoping for an answer like that :wink:.

CMakeLists.txt
tobim: Yeah, that made it in out of accident, I just had it fixed for my personal use. I can revert if you want to.

libcaf_core/CMakeLists.txt
tobim: Note the removed directory separator at the end of the `DIRECTORY` argument, this way the source directory itself gets copied into the destination. The old form of the call used `libcaf_core/caf/` as the argument, which is handled by copying every member inside that source into `DESTINATION`.

CMakeLists.txt
Neverlord: I'm fine with the change, as long as fixes an issue for you. I merely would've liked to see this in its own commit with a brief explanation what it solves. :)

libcaf_core/CMakeLists.txt
Neverlord: Gotya, thanks. Sometimes it isn't easy to think of all the subtleties in a review.

CMakeLists.txt
tobim: Not a real issue, but [ccache doesn't support linking and just passes the command back to the regular compiler driver](https://ccache.samba.org/manual/latest.html#_limitations). Enabling it for the linker is pointless.

Neverlord: 👍

PR #767, opened 2018-10-02
Build windows dll

uentity:
This patchset enables successful shared library (DLL) build on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.4.0)
 project(caf C CXX)
 
 include(CheckCSourceCompiles)
@@ -24,10 +24,9 @@ if(APPLE AND NOT DEFINED CMAKE_MACOSX_RPATH)
   set(CMAKE_MACOSX_RPATH true)
 endif()
 
-# Shared libs are currently not supported on Windows.
-if(WIN32 AND NOT CAF_BUILD_STATIC_ONLY)
-  message(STATUS "CAF currently only supports static-only builds on Windows")
-  set(CAF_BUILD_STATIC_ONLY yes)
+# Auto-export all symbols from Windows DLL
+if(WIN32)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
 if(CAF_BUILD_STATIC_RUNTIME)
--- a/libcaf_core/caf/binary_deserializer.hpp
+++ b/libcaf_core/caf/binary_deserializer.hpp
@@ -29,7 +29,7 @@
 namespace caf {
 
 /// Implements the deserializer interface with a binary serialization protocol.
-class binary_deserializer final : public deserializer {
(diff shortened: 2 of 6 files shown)


uentity:
The disadvantage is that `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` requires CMake v. 3.4 and higher.
Maybe it would be feasible to bump CMake version only for Windows build (if that's possible, I'm not a CMake expert).

Neverlord:
I find it odd, that we still have to do the export/import hack for global variables after setting `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`. However, the docs say so...

FYI, I merged this into an integration branch with a conditional version check.

uentity:
> I find it odd, that we still have to do the export/import hack for global variables after setting `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`. However, the docs say so...

Yes, that's sick.
Thanks for merging!

Neverlord:
@uentity doesn't work for me, I get tons of `unresolved external symbol` errors: https://jenkins.inet.haw-hamburg.de/blue/organizations/jenkins/CAF%2Factor-framework/detail/integration%2Fuentity-build-windows-dll/2/pipeline/

uentity:
@Neverlord I think that's because for some reason in the very beginnig  you pass `-DCAF_BUILD_STATIC_ONLY:BOOL=yes` flag to CMake.

Neverlord:
Does this means that static builds no longer work when using import/export annotations? What a mess. 😕

uentity:
Well, it really seem weird. I will test static/shared builds from scratch. Maybe, `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` should be enabled ONLY for shared builds...

uentity:
@Neverlord that's not so easy to fix problem as it seems at first glance.

When CAF if being used by third-party code there is no evidence whether static or shared CAF library will be used during linking. That means we can't define `CAF_API` macro to work properly with both static and shared builds _without any additional information from third-party code_ that uses CAF.

So without deeper modification of hardcoded defaults I see only two solutions.
1) State that we're switching to shared library ONLY build type on Windows and deprecate static builds.
2) Force CAF users to define additional preprocessor macro to provide information about what kind of CAF build they are using. We can assume, for example, that shared library is the default and users have to define `CAF_USE_STATIC_LIB` in case static library is being used. Or vice versa, make static build default and make define for shared build to keep compatibility.

Completely another approach is to switch defaults from raw variables to one or several functions that will return some PODs without static variables. This can be implemented using hidden globals (singletons). I know that globals are bad, etc, but this can be acceptable for hardcoded defaults.

So what do you think? Personally I'd go with solution 2 defaulting to static build, because it will require minimal additional modifications and won't break anything for current users.

uentity:
@Neverlord actually I use now only shared build. Release version of `caf-core` is only about 2.6 Mb in (`caf-io` is also around this size).

So, if there isn't any resistance for keeping static builds, I can only support that.

Neverlord:
@uentity thank you once more for your work (and patience) here! 🙏🏻

Closed via https://github.com/actor-framework/actor-framework/pull/992.

PR #772, opened 2018-10-02
Set CLOEXEC flag on socket/accept/pipe calls

jsiwek:
I did a cursory review of these calls to add close-on-exec flags as a generally accepted Good Idea.

The particular problem that precipitated this PR was that a long-lived fork/exec'd process ended up inheriting the FD from a port binding, and then restarting the original parent process failed to re-bind to the same port.

The Windows implementation for the new `child_process_inherit()` is an untested guess.


jsiwek:
Just forced pushed my branch to fix platform-dependent build issues, but looks like I still have the Windows function wrong, so would need help with that.  Or another idea would be to not bother with it on Windows (since it's default inheritance behavior is already what we want: child procs don't inherit) and instead just set CLOEXEC at the required call sites, wrapped in `#ifndef CAF_WINDOWS` blocks.

I generally don't mind if this PR gets closed and reimplemented differently, so long as the underlying problem gets fixed.  Just hoping it helps at least for demonstration purposes.

jsiwek:
Also worth calling out just in case, but this PR would change the default semantics of `create_pipe()` in case anyone out there was actually was relying on the default behavior of FDs inheriting across an exec.  (Maybe the argument is that it's just Better for such users to explicitly ask for that behavior when necessary anyway... )

jsiwek:
> I don't think it's ever sane to `fork` a CAF application. Whether or not you close your sockets, the bigger issue is that the actor system of the forked process now still assumes to have a running scheduler, a middleman with its multiplexer, a logger with its background thread, and so on.

Yep, understood.  The problem isn't strictly `fork` here, it's `exec`, hence the close-on-exec flag.

> So, the `fork`-ed process would have to never touch anything CAF. I assume the `fork` comes from some legacy part of your application that doesn't use CAF at all, correct?

There's code paths that may end up invoking something like `system` to run external shell commands:

http://pubs.opengroup.org/onlinepubs/9699919799/functions/system.html

As it says, it operates as if doing a `fork` followed by `exec` to overwrite the child process image.

> Sorry, we're not Windows experts either. However, I wouldn't mind making `child_process_inherit` a no-op on Windows (leaving a fixme/implement-me comment) to get this past Jenkins.

Ok, just force pushed to turn it into a no-op/fixme.

Neverlord:
Ok, if this fixes issues with `system(...)` then it definitely makes sense. Thanks for the patch!

PR #777, opened 2018-10-07
Add non-overriding receive timouts to actor clock

josephnoir:
This PR adds a  new function to the actor clock that allows adding a new receive timeout without overriding existing receive timeouts.

A use case is the new broker implementation that needs this feature for retransmit timeouts (among others) where multiple messages might need retransmission.


Neverlord:
I think sooner or later this would result in some *very* hard to debug problem when someone mixes `set_ordinary_timeout` and `add_ordinary_timeout`. I can understand the thoughts behind re-using existing mechanisms as much as possible, but we also need to follow the principle of least surprise.

josephnoir:
Yeah, I get it.

josephnoir:
How about this?

Neverlord:
I'm not quite happy with the name `multi-timeout`, but I couldn't come up with something better either. So let's merge it for now and maybe rename later.

PR #779, opened 2018-10-09
Use insert for faster appending in xsputn

tobim:
Unfortunately, the performance benefit from this change does not apply if compiled with clang and libc++. See https://bugs.llvm.org/show_bug.cgi?id=35637 for details.

### DIFF
--- a/libcaf_core/caf/streambuf.hpp
+++ b/libcaf_core/caf/streambuf.hpp
@@ -365,11 +365,7 @@ protected:
   }
 
   std::streamsize xsputn(const char_type* s, std::streamsize n) override {
-    // TODO: Do a performance analysis whether the current implementation based
-    // on copy_n is faster than these two statements:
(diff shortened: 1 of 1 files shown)


libcaf_core/caf/streambuf.hpp
```diff
@@ -365,11 +365,7 @@ class containerbuf
   }
 
   std::streamsize xsputn(const char_type* s, std::streamsize n) override {
-    // TODO: Do a performance analysis whether the current implementation based
-    // on copy_n is faster than these two statements:
-    // (1) container_.resize(container_.size() + n);
-    // (2) std::memcpy(this->pptr(), s, n * sizeof(char_type));
```
mavam: Have you compared the performance against what was written in the comment?

mavam: Awesome, I love how you're on top of it!

PR #780, opened 2018-10-15
Re-implement ip_endpoint

Neverlord:
Replace the old heap-allocating `ip_endpoint` with a new implementation that leverages CAF's new `ip_address` type.


josephnoir:
Only had a brief look at it and will do a review later. Quick question: why did the `ip_endpoint` move from io to core?

Neverlord:
For the same reason we moved `ip_address`: it's just a data type and a platform-independent representation is in itself useful to have, even without doing I/O.

However, I think we should move the `assign` functions back to the I/O module. Converting to/from `sockaddr` is only needed when talking to the socket API.

josephnoir:
The last Jenkins errors are unrelated. I'll do the review when I have a bit of time.

Neverlord:
How are they unrelated? `io_dynamic_remote_actor_udp` is actually the only unit test that is affected by the changes. I don't have the cycles to hunt this down on Windows, so I'll close this for now. 😒

PR #783, opened 2018-10-23
Fix exit reason propagation in monitorable_actor

tobim:


Neverlord:
Thanks a lot for the fixes! 👍

PR #787, opened 2018-10-30
Fix stringification inspector concept modeling

mavam:
Prior to this change, it wasn't possible write generic code that checks for `Inspector::writes_state`.


Neverlord:
Thanks! 👍

PR #789, opened 2018-11-07
Fix missing return

russpowers:
Hi, I'm getting warnings in Visual Studio for this line, and it looks like it's just missing a `return`.  I haven't tested the change, but it seems pretty straightforward.  I'm surprised it's not causing any bugs, though, maybe this function is not used?


Neverlord:
Welcome aboard and thanks for the fix! A quick grep shows that this function is indeed unused at the moment, so maybe we should consider removing it.

russpowers:
Yeah, that's what I figured, it would have shown up fairly quickly in tests if it was being used.  Well, now it's unused **and** bug free! 😁

PR #793, opened 2018-11-15
Add missing <string> include to use std::string

Cheney-W:
The latest source failed with error C2039 in is_primitive_config_value.hpp, so add \<string> to fix it.


Neverlord:
Welcome to CAF and thank you for the patch! 🙂

PR #795, opened 2018-11-17
Fix invalid class names in CAF log output

### LINKED ISSUE #792: Logger produces invalid class names
Here's an example from a VAST run:
```
9702 vast DEBUG actor13 0x7000042c2000 vast.system archive archive.cpp:59 archive got query for 8289 events in range [ 0 , 8462 )
9702 vast DEBUG actor13 0x7000042c2000 >%20vast.segment_store get segment_store.cpp:91 getting table slices with ids
```

The `>%20vast.segment_store` class has some extra characters in the beginning.

Neverlord:
Close #792.

### DIFF
--- a/libcaf_core/src/logger.cpp
+++ b/libcaf_core/src/logger.cpp
@@ -63,7 +63,12 @@ constexpr string_view fun_prefixes[] = {
   "unsigned ",
 };
 
-constexpr string_view anon_ns = "(anonymous namespace)";
+// Various spellings of the anonymous namespace as reported by CAF_PRETTY_FUN.
+constexpr string_view anon_ns[] = {
+  "(anonymous namespace)", // Clang
+  "{anonymous}", // GCC
+  "`anonymous-namespace'", // MSVC
+};
 
 /// Reduces symbol by printing all prefixes to `out` and returning the
 /// remainder. For example, "ns::foo::bar" prints "ns.foo" to `out` and returns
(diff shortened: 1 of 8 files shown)


josephnoir:
Is my assignment here a review request or how can I help?

libcaf_core/src/logger.cpp
```diff
@@ -47,14 +47,116 @@ namespace caf {
...
+    if (starts_with(symbol, str))
+      symbol.remove_prefix(str.size());
+  };
+  // MSVC adds `struct` to symbol names. For example:
+  // void __cdecl `anonymous-namespace'::foo::tpl<struct T>::run(void)
+  //                                              ^~~~~~
+  skip("struct ");
+  string_view last = "";
+  bool printed = false;
+  // Prints the content of `last` and then replaces it with `x`.
```
josephnoir: What's x here, it seems you replace `last` with `.` (line 91).

libcaf_core/src/logger.cpp
```diff
@@ -146,6 +249,13 @@ logger::line_builder& logger::line_builder::operator<<(const std::string& str) {
   return *this;
 }
 
+logger::line_builder& logger::line_builder::operator<<(string_view str) {
+  if (!str_.empty() && str_.back() != ' ')
```
josephnoir: Why does this check the last character while the overload below does not?

Neverlord: Good question. I fixed the other implementations with c083daa.

PR #798, opened 2018-11-20
Clean up string algorithms header and fix caf::split

Neverlord:
Fixes a bug in `caf::split` that was spotted by @tobim, adds unit tests for the string algorithms, and removes unused functions.

### DIFF
--- a/libcaf_core/test/string_algorithms.cpp
+++ b/libcaf_core/test/string_algorithms.cpp
@@ -1,0 +1,109 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2018 Dominik Charousset                                     *
+ *                                                                            *
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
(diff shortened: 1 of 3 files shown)


libcaf_core/test/string_algorithms.cpp
```diff
@@ -0,0 +1,109 @@
...
+}
+
+std::string join(str_list vec) {
+  return caf::join(vec, ",");
+}
+
+template <class... Ts>
+std::string splice(string_view glue, const Ts&... xs) {
+  std::string result;
+  caf::splice(result, glue, xs...);
```
josephnoir: Didn't you remove this function from the file `string_algorithms.hpp`?

libcaf_core/test/string_algorithms.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <string>
+#include <vector>
+
+using namespace caf;
+
+namespace {
+
+struct fixture {
+
+};
```
josephnoir: Probably unnecessary?

libcaf_core/test/string_algorithms.cpp
```diff
@@ -0,0 +1,109 @@
...
+  CAF_CHECK_EQUAL(join({"a", "b"}), "a,b");
+  CAF_CHECK_EQUAL(join({"a", "b", "c"}), "a,b,c");
+}
+
+CAF_TEST(starts with) {
+  CAF_CHECK(starts_with("foobar", "f"));
+  CAF_CHECK(starts_with("foobar", "fo"));
+  CAF_CHECK(starts_with("foobar", "fooba"));
+  CAF_CHECK(starts_with("foobar", "foobar"));
+  CAF_CHECK(!starts_with("foobar", "o"));
```
josephnoir: Cases where I can easily imagine errors would be something like `starts_with("foobar", "fa")` or `starts_with("foobar", "foobaro")` that include a partial match (same for `ends_with`). Not that I'm expecting this here.

Neverlord: Good point. I've added a couple of extra tests.

PR #799, opened 2018-11-21
Fix comments of 2 UDP functions, where port still referred to TCP

chhinze:
`TCP` -> `UDP` in doc comments for functions `publish_udp` and `remote_actor_udp`


Neverlord:
Welcome to CAF and thanks for the PR! 👍

PR #801, opened 2018-11-22
Allow copy of config_option

tobim:
This adds a copy constructor and copy assignment operator for `config_option`. It is implemented as a deep copy. The alternative solution with a shared_ptr was not used because `make_shared` does not support dynamic arrays until c++20.


libcaf_core/caf/config_option.hpp
```diff
@@ -54,10 +54,18 @@ class config_option {
   config_option(string_view category, string_view name, string_view description,
                 const meta_state* meta, void* value = nullptr);
 
+  config_option(const config_option&);
+
   config_option(config_option&&) = default;
 
+  config_option& operator=(config_option);
```
Neverlord: Should read `operator=(const config_option&)`. For some types it makes sense to provide only this signature, but since we want to have a separate move-assign operator we should use the standard signatures.

libcaf_core/src/config_option.cpp
```diff
@@ -77,6 +77,33 @@ config_option::config_option(string_view category, string_view name,
...
+    short_names_separator_{other.short_names_separator_},
+    buf_size_{other.buf_size_},
+    meta_{other.meta_},
+    value_{other.value_} {
+  buf_.reset(new char[buf_size_]);
+  std::copy_n(other.buf_.get(), buf_size_, buf_.get());
+}
+
+config_option& config_option::operator=(config_option other) {
+  swap(*this, other);
```
Neverlord: With the signature change this becomes the classic two-liner:
```
config_option tmp{other};
swap(*this, tmp);
```

libcaf_core/test/config_option_set.cpp
```diff
@@ -44,8 +44,9 @@ CAF_TEST_FIXTURE_SCOPE(config_option_set_tests, fixture)
 
 CAF_TEST(lookup) {
   opts.add<int>("opt1,1", "test option 1")
-    .add<float>("test", "opt2,2", "test option 2")
-    .add<bool>("test", "flag,fl3", "test flag");
+    .add<float>("test", "opt2,2", "test option 2");
+  auto flag = make_config_option<bool>("test", "flag,fl3", "test flag");
```
Neverlord: I would've preferred new unit tests for the copying. This mixes copy-semantic testing into the lookup test, making this test harder to read and reason about.

libcaf_core/test/config_option.cpp
```diff
@@ -87,8 +87,29 @@ void check_integer_options() {
   check_integer_options<T>(tk);
 }
 
+void compare(const config_option& lhs, const config_option& rhs) {
+  CAF_CHECK_EQUAL(lhs.category(), rhs.category());
+  CAF_CHECK_EQUAL(lhs.long_name(), rhs.long_name());
+  CAF_CHECK_EQUAL(lhs.short_names(), rhs.short_names());
+  CAF_CHECK_EQUAL(lhs.short_names(), rhs.short_names());
```
Neverlord: `short_names` gets compared twice.

PR #804, opened 2018-12-06
Conditionally print logger framing messages

mavam:
The logger uncondionally prints a begin and end message that frames the log messages. This PR changes this behavior such that the logger now only prints the framing when at least one event is logged.

### DIFF
--- a/libcaf_core/src/logger.cpp
+++ b/libcaf_core/src/logger.cpp
@@ -547,13 +547,18 @@ string_view logger::skip_path(string_view path) {
 }
 
 void logger::run() {
-  log_first_line();
+  auto at_least_one_event = false;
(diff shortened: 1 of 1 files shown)


mavam:
Not sure what Jenkins is complaining about:
```

- remote_link
WATCHDOG: unit test did not finish within 300s, abort

        Start 112: openssl_authentication
112/112 Test #112: openssl_authentication ...........   Passed    0.05 sec
```

Seems unrelated to this change, no?

Neverlord:
Thanks for the patch. However, I prefer a solution without extra state and branching (see 2cc4377).

mavam:
Certainly nicer that way.

PR #811, opened 2019-01-23
Cap logger level to not exceed CAF_LOG_LEVEL

jsiwek:
If the build system isn't configured at a high enough level, then a user setting a higher log level in their config shouldn't do anything. For example, if the configured level is QUIET, then a log file
should never be produced (the original logic would actually result in producing an empty log file by default).

To follow along with that example, see the logger initialize a "quiet" verbosity level from `CAF_LOG_LEVEL` here:

https://github.com/actor-framework/actor-framework/blob/14c187b81c77c7c76734adc1daf414d109ac489b/libcaf_core/src/logger.cpp#L249

This logic overwrite's the logger's level with "trace":

https://github.com/actor-framework/actor-framework/blob/14c187b81c77c7c76734adc1daf414d109ac489b/libcaf_core/src/logger.cpp#L381-L389

This condition is now false:

https://github.com/actor-framework/actor-framework/blob/14c187b81c77c7c76734adc1daf414d109ac489b/libcaf_core/src/logger.cpp#L636

A log file gets opened:

https://github.com/actor-framework/actor-framework/blob/14c187b81c77c7c76734adc1daf414d109ac489b/libcaf_core/src/logger.cpp#L666

But nothing will be written to it as technically it's log level "quiet" and logging should be pre-processed out.

### DIFF
--- a/libcaf_core/src/logger.cpp
+++ b/libcaf_core/src/logger.cpp
@@ -384,8 +384,10 @@ void logger::init(actor_system_config& cfg) {
   file_verbosity = get_or(cfg, "logger.file-verbosity", file_verbosity);
   console_verbosity =
     get_or(cfg, "logger.console-verbosity", console_verbosity);
-  cfg_.file_verbosity = to_level_int(file_verbosity);
-  cfg_.console_verbosity = to_level_int(console_verbosity);
+  cfg_.file_verbosity = std::min(static_cast<unsigned>(CAF_LOG_LEVEL),
+                                 to_level_int(file_verbosity));
+  cfg_.console_verbosity = std::min(static_cast<unsigned>(CAF_LOG_LEVEL),
+                                    to_level_int(console_verbosity));
   cfg_.verbosity = std::max(cfg_.file_verbosity, cfg_.console_verbosity);
   // Parse the format string.
   file_format_ =


jsiwek:
Also just wanted to note I don't think this logging behavior/issue was in 0.16.2, but I saw it after upgrading to 0.16.3.

Neverlord:
This was a deliberate change with 0.16.3. We want to allow users to make use of CAFs logger, even if CAF itself was compiled without logging. This allows projects to use the logging infrastructure and configure a log level independently of what CAF was build with.

For example, the new logging macros for VAST are independent of `CAF_LOG_LEVEL`. Combined with the new component blacklists, this gives users a high degree of flexibility for configuring what goes into a log file.

I think the actual bug here is the default value in `src/defaults.cpp`. It's always set to `trace` for both console and file. However, the default should reflect `CAF_LOG_LEVEL` instead. Also, creating the file and logging the first line should happen lazily on the first logged event to avoid empty files.

jsiwek:
> We want to allow users to make use of CAFs logger, even if CAF itself was compiled without logging. This allows projects to use the logging infrastructure and configure a log level independently of what CAF was build with.

Ack.  Feel free to close the PR and re-open an issue for whenever someone gets around to a more appropriate fix.

Neverlord:
I tackled this in `topic/logger` by creating the file lazily when printing the first event. Also, users can now wrap the `CAF_LOG_IMPL` macro to get logging regardless of the setting in CAF's build config.

PR #813, opened 2019-01-29
Port auto-connect changes to cleanup branch

josephnoir:


libcaf_io/src/instance.cpp
```diff
@@ -349,7 +359,9 @@ bool instance::handle(execution_unit* ctx, connection_handle hdl, header& hdr,
       binary_deserializer bd{ctx, *payload};
       node_id source_node;
       std::string remote_appid;
-      if (bd(source_node, remote_appid))
+      basp::routing_table::endpoint autoconn_addr;
+      // TODO: Read addrs separately.
```
Neverlord: How and why?

libcaf_io/test/automatic_connection.cpp
```diff
@@ -65,223 +75,282 @@ using pong_atom = atom_constant<atom("pong")>;
...
+public:
+  simple_config() {
+    load<caf::io::middleman>();
+    set("middleman.enable-automatic-connections", true);
+    set("middleman.enable-tcp", true);
+    set("middleman.enable-udp", false);
+  }
+};
+
+class fixture {
```
Neverlord: Is there a particular reason not to use the `belt_fixture` form the `io_dsl.hpp` here?

Neverlord:
As discussed, let's postpone discussion on this. Since we are in the process of evaluating alternative options I'll merge this for now and maybe we revisit this later.

PR #815, opened 2019-02-06
Clarify bit representation of ipv4 address factory

josephnoir:
I think this is what the function does. Is this expected and should we actually store the bits in network byte order instead of host byte order?

### DIFF
--- a/libcaf_core/caf/ipv4_address.hpp
+++ b/libcaf_core/caf/ipv4_address.hpp
@@ -46,6 +46,7 @@ public:
 
   explicit ipv4_address(array_type bytes);
 
+  /// Constructs an IPv4 address from bits in network byte order.
   static ipv4_address from_bits(uint32_t bits) {
     ipv4_address result;
     result.bits(bits);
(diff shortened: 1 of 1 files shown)


Neverlord:
This function is not meant for public consumption. Hence, it should be declared `@private`. Also, this simply sets all bits. It's a pure convenience function for default-constructing and then calling `.bits(x)`. Hence, the argument naturally must have the *host network byte order* (little endian, most likely).

Neverlord:
Ah, I'm sorry. The bits in `ipv4_address` are in fact network byte order.

libcaf_core/caf/ipv4_address.hpp
```diff
@@ -46,6 +46,8 @@ class ipv4_address : public byte_address<ipv4_address> {
 
   explicit ipv4_address(array_type bytes);
 
+  /// Construct an ipv4 address from the network representation of the address
```
Neverlord: To me, `network representation` is very vague and an unusual wording. I'd prefer the much more common term "network byte order". Also, this function is not meant for public use. Hence, please add `@private`.

josephnoir: Why is this a private API? I'm using it to parse addresses from pcap files. (I'm sure those are only v4.) What would be the alternative here? Deconstructing the int into four octets or the string representation seems like an unnecessary work-around.

Neverlord: It's private simply by extension, because the `bits(uint32_t)` setter is private. I don't have strong feeling towards making both public, though. Im finde with removing that restriction as long as you update the documentation of the `bits` getter and setter with mentioning "network byte order".

josephnoir: :+1: I'll update the comments.

PR #821, opened 2019-02-20
Parallelize deserialization of incoming BASP messages

Neverlord:
# Summary

This set of changes addresses performance issues when distributing CAF applications.

The BASP broker is currently bottlenecked by deserializing incoming messages. For evaluating the performance, I've used the `simple_streaming` benchmark, which runs a trivial streaming source that produces short strings endlessly and a sink that drops everything it receives.

# Baseline

The baseline deploys both actors in the same process:

```
$ simple_streaming -m both
975586 messages/s
924538 messages/s
1088718 messages/s
976075 messages/s
994920 messages/s
1095617 messages/s
789944 messages/s
782430 messages/s
825262 messages/s
848951 messages/s
```

# Distributed

Deploying each actor in its own process (connected via localhost) drops the performance to roughly half on current `master` (with the source already running in the background):

```
$ simple_streaming -m sink -p 4242
538069 messages/s
557683 messages/s
566474 messages/s
556036 messages/s
537674 messages/s
568108 messages/s
559711 messages/s
547152 messages/s
550078 messages/s
542460 messages/s
```

Looking at the bandwidth via `iftop -i lo` reveals that CAF only utilizes 46 mbits.

# Distributed after the Patch

With this patch, performance doubles:

```
$ ./build/release/bin/simple_streaming --middleman.workers=2 -m sink -p 4242
1125932 messages/s
1254350 messages/s
1258735 messages/s
1285260 messages/s
1289829 messages/s
1215599 messages/s
1326506 messages/s
1274993 messages/s
1239853 messages/s
1388835 messages/s
```

Looking at the bandwidth via `iftop -i lo` again shows that CAF utilizes about 100 mbits now.

And yeah, this is now faster than the baseline. I don't know how that's possible. 🤷‍♂️

The single-process setup is a topic for another day, though.


mavam:
> And yeah, this is now faster than the baseline. I don't know how that's possible.

@tobim had a hunch that the scheduler might be a bottleneck in the baseline. Maybe there are also other central components in CAF that were fully saturated. Duplicating (by distribution) them gives room for more throughput.

Neverlord:
Hm, there is no central component source or sink need to speak to. The only thing I can imagine is this:
- the source sends a batch to the sink, wich is currently awaiting messages
- the sink gets pulled to the current worker
- the source still runs, so the sink gets no CPU time
- the sink now has to wait until it gets stolen by another worker or the source gives the CPU free

The performance increase could simply be purely by the fact the the BASP broker and the source can never share the same worker, because the BASP broker lives in the MM multiplexer.

Neverlord:
Interesting, I've ran a small scheduling experiment. Actively forcing sink and source to not share a worker makes performance drop by about 20% in the baseline. Detaching the source also has a negative effect, though not quite as much.

mavam:
> Actively forcing sink and source to not share a worker makes performance drop by about 20% in the baseline.

Weird. There's a mismatch between our mental model and what the code does. This seems worth investigating.

josephnoir:
Nice. Not what I expected when we talked about separating (de)serialization form the BASP broker, but it definitely seems to pay off! Got a few comments:

* As far as I can tell there is no minimum amount of workers and on a system with less than four cores there won't be any? Might be worth it to add a comment here.
* Did you measure if two workers are sufficient to handle all deserialization or do you ever hit the [else](https://github.com/actor-framework/actor-framework/blob/15a0b49f9f9b748d79b9ea928a61f586a6553709/libcaf_io/src/instance.cpp#L397) path?
* You set the workers to 2 in your measurement. How many cores do you have and how does the number of workers affect the measurements? (I can imagine that on a system with many cores we don't actually need cores/4 workers, i.e., 16 for 64 cores.)
* It might be interesting to run an measurement for a scenario that features multiple proxies instead of just one. Maybe saturate all of the scheduler threads?

Neverlord:
> Not what I expected when we talked about separating (de)serialization form the BASP broker

For `caf_net` we are going to get rid of the BASP broker entirely. I think this is as far as we're going to push the current design.

> As far as I can tell there is no minimum amount of workers and on a system with less than four cores there won't be any? Might be worth it to add a comment here.

This formula is by no means set in stone. Maybe it's a good idea to go with something like `max(3, hardware_concurrency() / 4) + 1`. That would give us 1-4 workers.

> You set the workers to 2 in your measurement.

I've tested on mobi8, i.e., 64 cores. I simply played with the parameter a bit and it made no difference to put more than 2. Honestly, I simply forgot to remote the setting for the benchmarking.

> It might be interesting to run an measurement for a scenario that features multiple proxies

Sure, multiple remote actors / connections would be interesting to see as part of a follow-up performance analysis.

josephnoir:
> This formula is by no means set in stone. Maybe it's a good idea to go with something like `max(3, hardware_concurrency() / 4) + 1`. That would give us 1-4 workers.

I think that sounds reasonable! I'm not sure how much it would pay off to do more measurements here. Similar question for the runtime behavior with a fully loaded scheduler. If we don't expect any surprises we can go ahead with this PR.

@mavam How much of a block is explaining the performance difference for you?

mavam:
> @mavam How much of a block is explaining the performance difference for you?

I'm mostly concerned with introducing an inadvertent performance regression. Before merging, I'd like to use that branch for a bit in VAST, so that we can see how it affects throughput. For example, because we're adding more workers, I'd like to get a feel for what that means in an oversubscribed setting.

josephnoir:
Fair enough. That should be helpful!

@Neverlord Running the distributed case with the patch and 0 workers should give an estimate of the patch impact aside from parallelization. Could you post numbers for that? I can run it myself if you give me the code for the `simple_streaming` benchmark.

Neverlord:
> I can run it myself if you give me the code for the `simple_streaming` benchmark.

https://github.com/actor-framework/benchmarks 😉

josephnoir:
Hmmm, I played around with the benchmark a bit. I got widely different results for different commits. My initial results for master differed from yours and I noticed that I wasn't on the up to date master and using https://github.com/actor-framework/actor-framework/commit/792f39c297ecac8367585fe712eeb7a317b13c45 instead. Here are measurements on my MacBook for three hashes: master, topic/basp-worker, and 792f39c2. (CAF and benchmarks are build with `--build-type=release`.)

# Baseline

Sink and source in the same process. The measurements for 792f39c2 are similar to topic/basp-worker and it looks more like something in the master branch changed in-between that leads to problems?

In fact, topic/basp-worker "[...] is 19 commits ahead, 7 commits behind master." Maybe its one of those 7 commits that leads to the degradation of the baseline performance.

## master

```
$ ./build/bin/simple_streaming -m both
9179592 messages/s
9333368 messages/s
9380246 messages/s
9406093 messages/s
9417797 messages/s
9365075 messages/s
9369084 messages/s
9354812 messages/s
9288539 messages/s
9301096 messages/s
9330199 messages/s
```

## hash 792f39c2

```
$ ./build/bin/simple_streaming -m both
36395155 messages/s
38402593 messages/s
37280274 messages/s
37320577 messages/s
38405536 messages/s
38076313 messages/s
37870813 messages/s
37861896 messages/s
37373127 messages/s
38135564 messages/s
38393431 messages/s
37324010 messages/s
37647258 messages/s
38035939 messages/s
37947184 messages/s
```

## topic/basp-worker

```
$ ./build/bin/simple_streaming -m both
36344469 messages/s
37190880 messages/s
37256645 messages/s
37751733 messages/s
37712177 messages/s
36856158 messages/s
37815160 messages/s
37959410 messages/s
37151576 messages/s
37185461 messages/s
36523857 messages/s
```

## topic/basp-worker rebased onto master

```
$ ./build/bin/simple_streaming -m both
9488023 messages/s
9437534 messages/s
9434740 messages/s
9568037 messages/s
9536930 messages/s
9623891 messages/s
9586408 messages/s
9520629 messages/s
```

# Distributed

## master

```
$ ./build/bin/simple_streaming -m sink -p 4242
86138 messages/s
1464058 messages/s
2058408 messages/s
2478780 messages/s
2154069 messages/s
2256937 messages/s
2324106 messages/s
2478440 messages/s
1904654 messages/s
2411618 messages/s
1930842 messages/s
2036666 messages/s
2082396 messages/s
1706818 messages/s
2129211 messages/s
```

Here is 792f39c2 as a comparison to the measurements below.

```
$ ./build/bin/simple_streaming -m sink -p 4242
3169258 messages/s
6664097 messages/s
7403145 messages/s
8165313 messages/s
8839700 messages/s
8591621 messages/s
8644764 messages/s
8091869 messages/s
8770711 messages/s
9183423 messages/s
8823838 messages/s
8013785 messages/s
8029027 messages/s
8518952 messages/s
8182198 messages/s
```

## topic/basp-workers

Event without workers this branch performs better.

```
$ ./build/bin/simple_streaming -m sink -p 4242 --middleman.workers=0
4543170 messages/s
10804201 messages/s
10627355 messages/s
11487112 messages/s
11695092 messages/s
11443263 messages/s
11009361 messages/s
12500169 messages/s
11283109 messages/s
10896291 messages/s
10400493 messages/s
10583239 messages/s
10997706 messages/s
12327173 messages/s
10655649 messages/s
11401075 messages/s
11378095 messages/s
10892043 messages/s
11676990 messages/s
10655247 messages/s
11558498 messages/s
10370257 messages/s
10809343 messages/s
11732321 messages/s
11381980 messages/s
```

Adding one or two workers improves performance, although less than I would have expected. Maybe this change is more evident if there are more actors communicating via the broker?

```
$ ./build/bin/simple_streaming -m sink -p 4242 --middleman.workers=1
4502770 messages/s
12366598 messages/s
14167966 messages/s
13147887 messages/s
15017593 messages/s
13020927 messages/s
13278203 messages/s
13316992 messages/s
14703513 messages/s
13289149 messages/s
12291563 messages/s
12575136 messages/s
13084065 messages/s
12856170 messages/s
14210567 messages/s
12904425 messages/s
```

```
$ ./build/bin/simple_streaming -m sink -p 4242 --middleman.workers=2
6223490 messages/s
14557784 messages/s
14572587 messages/s
15125580 messages/s
17592252 messages/s
14966868 messages/s
15085509 messages/s
16797462 messages/s
16633251 messages/s
17015461 messages/s
14870052 messages/s
16154902 messages/s
```

At least for this benchmark, adding more workers doesn't seem to pay off much.

```
$ ./build/bin/simple_streaming -m sink -p 4242 --middleman.workers=3
4701346 messages/s
13377680 messages/s
15295182 messages/s
14863476 messages/s
16382346 messages/s
14675376 messages/s
15581456 messages/s
16408251 messages/s
14390780 messages/s
16394753 messages/s
17613908 messages/s
15828813 messages/s
14403740 messages/s
15506960 messages/s
```

```
$ ./build/bin/simple_streaming -m sink -p 4242 --middleman.workers=4
5391728 messages/s
14835998 messages/s
15606474 messages/s
15081277 messages/s
16500765 messages/s
15575883 messages/s
14091946 messages/s
15998640 messages/s
16348271 messages/s
16388614 messages/s
14506458 messages/s
17783201 messages/s
14885644 messages/s
13637856 messages/s
```

Neverlord:
Thanks for this performance study! Looking at the delta between `basp-worker` and `master`, I think the most likely culprit is: c736d97883b8f64830c7c345bf4dbaef39a3b7cf. I'll look into it.

Neverlord:
@josephnoir Turns out c736d97883b8f64830c7c345bf4dbaef39a3b7cf sneaked `caf_flow` logging into every build. Can you double-check by retrying with current master?

josephnoir:
Yes and it looks better. Here are the new measurements for master:

```
$ Baseline
$ ./build/bin/simple_streaming -m both
34760548 messages/s
35513571 messages/s
35499853 messages/s
35320678 messages/s
35224665 messages/s
$ # Distributed
$ ./build/bin/simple_streaming -m sink -p 4242
4596296 messages/s
9773151 messages/s
10751395 messages/s
11327629 messages/s
11187282 messages/s
```

Now we no longer have a distributed case that outperforms the baseline.

Neverlord:
@mavam I was testing the branch today in VAST with the ICTF log as well as looking into Broker (see referenced Broker PR). For Broker, I get quite the boost in performance (70k msgs/s up to 100k msgs/s).

For VAST, this branch achieves ~5% speedup on my quad core iMac. This iMac is as over-subscribed as it gets. IMPORTER, ARCHIVE, INDEX and depending on the layout ~20 INDEXER actors all share four cores and the performance *still* improves then pushing deserialization work from the I/O thread to the thread pool of the scheduler.

```
# Baseline

time vast -d importer-master import -b zeek -r ictf.log

real    3m15.721s
user    1m59.285s
sys     0m36.449s

# Topic Branch

$ time vast -d importer import -b zeek -r ictf.log

real    3m5.265s
user    1m50.442s
sys     0m20.348s
```

Any reservations left against merging this?

mavam:
> Any reservations left against merging this?

Nope.

These are quite promising results. Good work!

Neverlord:
@josephnoir b628ea4 tweaks the formula as discussed. No more blockers from my side. :)

libcaf_io/src/routing_table.cpp
```diff
@@ -34,122 +34,99 @@ routing_table::~routing_table() {
...
-    ++res;
-  }
-  return res;
+  std::unique_lock<std::mutex> guard{mtx_};
+  // Never add indirect entries if we already have direct connection.
+  if (direct_by_nid_.count(dest) != 0)
+    return false;
+  // Never add indirect entries if we don't have a connection to the hop.
+  if (direct_by_nid_.count(hop) == 0)
+    return false;
```
josephnoir: I have conflicting views on the return value of this function. Just looking at the name, sure, it returns true if the entry was added. But this is not helpful in telling us what what the state is now. Even if it returns `false` we might be able to reach the node (line 121) or not (line 124).

libcaf_io/caf/io/basp/message_queue.hpp
```diff
@@ -0,0 +1,81 @@
...
+    uint64_t id;
+    strong_actor_ptr receiver;
+    mailbox_element_ptr content;
+  };
+
+  // -- constructors, destructors, and assignment operators --------------------
+
+  message_queue();
+
+  // -- member variables -------------------------------------------------------
```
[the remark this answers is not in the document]
Neverlord: Fixed.

libcaf_io/src/routing_table.cpp
Neverlord: I agree, the semantics of the routing table is convoluted. However, since we already voted to rip out the entire routing stuff in `caf_net`, I'd say we don't invest further cycles here. Ok?

PR #824, opened 2019-03-04
Allow IPO on gcc/clang

tobim:
This allows `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` to work.


CMakeLists.txt
```diff
@@ -315,7 +327,7 @@ if(MINGW)
   build_string("EXTRA_FLAGS" "-static")
 elseif(CYGWIN)
   build_string("EXTRA_FLAGS" "-U__STRICT_ANSI__")
-else()
+elseif(NOT CAF_BUILD_STATIC_ONLY)
   build_string("EXTRA_FLAGS" "-fPIC")
```
Neverlord: Wouldn't this result in a broken `.a` file when building both shared in static? I think this dual build has to go away anyway. Users that wish to get both simply have to build both versions. We eventually have to make this change anyway if we wish to get CAF build proper static/shared libraries on Windows (cf. #767).

Not a change request, btw. Just thinking loudly.

examples/CMakeLists.txt
```diff
@@ -1,4 +1,3 @@
-cmake_minimum_required(VERSION 2.8)
```
Neverlord: Are these changes on purpose or did they sneak in?

tobim: That is on purpose. `cmake_minimum_required(VERSION)` has side-effects (sigh) such as setting all `CMP00XX` policies to `NEW` if they were defined in cmake version `VERSION` or to unset the variable otherwise. It should only be called once.

CMakeLists.txt
tobim: `CAF_BUILD_STATIC_ONY` should not be on if it also builds shared.

> I think this dual build has to go away anyway. Users that wish to get both simply have to build both versions.

Full ACK. But out of scope for this PR?

examples/CMakeLists.txt
Neverlord: 🤦‍♂️

PR #827, opened 2019-03-13
 Add marker to make categories optional on the CLI

Neverlord:
Categories are great at organizing program options. However, on the CLI
they get in the way quickly. This change allows developers to prefix
category names with '?' to make it optional on the CLI. For example, an
option added via

```
opt_group{custom_options_, "?foo"}
.add<std::string>("bar,b", "some string");
```

allows users to use any of these on the CLI:

- `-b text`
- `--bar=text`
- `--foo.bar=text`

All options in the `global` category already allowed this abbreviated
form implicitly.

### DIFF
--- a/libcaf_core/caf/config_option.hpp
+++ b/libcaf_core/caf/config_option.hpp
@@ -100,6 +100,9 @@ public:
   /// Returns whether this config option stores a boolean flag.
   bool is_flag() const noexcept;
 
+  /// Returns whether the category is optional for CLI options.
+  bool has_flat_cli_name() const noexcept;
+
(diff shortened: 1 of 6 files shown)


Neverlord:
> would it be possible to have a config option without a category?

Currently we kinda treat `global` like no category. Maybe we want to migrate away from this "special" name eventually.

> Just asking because the code uses `buf_[0]`.

Currently there's always a category. However, even if we relax this restriction: This is guaranteed to be safe, because the buffer always contains a formatted string in the form `<category>.<name>,<short names>,<description>`. So an empty category would mean `buf_[0] == '.'`.

> feel free to hit merge.

Ok. 🙂

Issue #839, opened 2019-05-13
--config-file option mentioned in section 11.2 of manual is not recognized


ksachdeva:
Section 11.2 of the manual that talks about command line and ini support mentions the possibility to provide the path to the ini file using --config-file.

When I provide this option, the application terminates with an error that it is not a recognized option. 

-h option on the command line does not have any entry for 'config-file' either.

As long as my caf-application .ini file is in the same directory as the executable it does work.

My environment is OSX 10.13.6 / Clang 10.0.0

PR #842, opened 2019-05-23
Add config file option from manual and help text

josephnoir:
Addresses #839. The global option listed in the [latest manual](https://actor-framework.readthedocs.io/en/latest/ConfiguringActorApplications.html#command-line-options-and-ini-configuration-files) was not checked before. `--help` lists the option now.

### DIFF
--- a/libcaf_core/caf/sec.hpp
+++ b/libcaf_core/caf/sec.hpp
@@ -114,6 +114,8 @@ enum class sec : uint8_t {
   bad_function_call = 40,
   /// Feature is disabled in the actor system config.
   feature_disabled,
+  /// Failed to open file.
+  cannot_open_file,
 };
 
 /// @relates sec
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -67,7 +67,8 @@ actor_system_config::actor_system_config()
   opt_group{custom_options_, "global"}
     .add<bool>("help,h?", "print help text to STDERR and exit")
     .add<bool>("long-help", "print long help text to STDERR and exit")
-    .add<bool>("dump-config", "print configuration to STDERR and exit");
+    .add<bool>("dump-config", "print configuration to STDERR and exit")
+    .add<string>("config-file", "set config file (default: caf-application.ini)");
   opt_group{custom_options_, "stream"}
     .add<timespan>(stream_desired_batch_complexity, "desired-batch-complexity",
                    "processing time per batch")
@@ -367,13 +368,26 @@ std::string actor_system_config::render_pec(uint8_t x, atom_value,
 
 void actor_system_config::extract_config_file_path(string_list& args) {
   static constexpr const char needle[] = "--caf#config-file=";
-  auto last = args.end();
-  auto i = std::find_if(args.begin(), last, [](const std::string& arg) {
-    return arg.compare(0, sizeof(needle) - 1, needle) == 0;
-  });
-  if (i == last)
-    return;
-  auto arg_begin = i->begin() + strlen(needle);
+  static constexpr const char needle_global[] = "--config-file=";
+  const auto last = args.end();
+  size_t len = 0;
+  auto i = args.begin();
+  // Look for match and set len and i.
+  auto check = [&](const char* str) {
+    i = std::find_if(args.begin(), last, [&](const std::string& arg) {
+      return arg.compare(0, strlen(str), str) == 0;
+    });
+    len = strlen(str);
+  };
+  // Check for global name.
+  check(needle_global);
+  if (i == last) {
+    // Check for full name.
+    check(needle);
+    if (i == last)
+      return;
+  }
+  auto arg_begin = i->begin() + len;
   auto arg_end = i->end();
   if (arg_begin == arg_end) {
     // Missing value.
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -35,6 +35,12 @@
 
 namespace caf {
 
+namespace {
+
+constexpr const char* default_config_file = "caf-application.ini";
+
+} // namespace anonymous
+
 actor_system_config::~actor_system_config() {
   // nop
 }
@@ -45,7 +51,7 @@ actor_system_config::~actor_system_config() {
 actor_system_config::actor_system_config()
     : cli_helptext_printed(false),
       slave_mode(false),
-      config_file_path("caf-application.ini"),
+      config_file_path(default_config_file),
       slave_mode_fun(nullptr) {
   // add `vector<T>` and `stream<T>` for each statically known type
   add_message_type_impl<stream<actor>>("stream<@actor>");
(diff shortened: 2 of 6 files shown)


libcaf_core/src/actor_system_config.cpp
```diff
@@ -367,13 +368,26 @@ std::string actor_system_config::render_pec(uint8_t x, atom_value,
...
+  static constexpr const char needle_global[] = "--config-file=";
+  const auto last = args.end();
+  size_t len = 0;
+  auto i = args.begin();
+  // Look for match and set len and i.
+  auto check = [&](const char* str) {
+    i = std::find_if(args.begin(), last, [&](const std::string& arg) {
+      return arg.compare(0, strlen(str), str) == 0;
+    });
+    len = strlen(str);
```
Neverlord: I find this code hard to follow because of the side effects going on and the two different needle variables. Why is one called `needle` and the other one `needle_global`? Both of them are global constants. Also, we're adding `config-file` as global option but don't do anything with it. This function still invokes the parser by hand rather than using `config_option::parse` and `config_option::store`.

First of all, I think this code would greatly benefit from turning the side effects into a proper algorithm (untested).

```c++
/// Finds ... in range ...
template <class ForwardIterator, class Sentinel>
std::pair<ForwardIterator, string_view>
find_by_long_name(const config_option& x, ForwardIterator first, Sentinel last) {
  auto long_name = x.long_name();
  for (; first != last; ++first) {
    string_view str{*first};
    // Make sure this is a long option starting with "--".
    if (str.compare(0, 2, "--") != 0)
      continue;
    str.remove_prefix(2);
    // Skip optional "caf#" prefix.
    if (str.compare(0, 4, "caf#") != 0)
      str.remove_prefix(4);
    // Make sure we are dealing with the right key.
    if (str.substr(0, long_name.size()) != long_name)
      continue;
    // Make sure the key is followed by an assignment.
    str.remove_prefix(long_name.size());
    if (str.compare(0, 1, "=") != 0)
      continue;
    // Remove leading '=' and return the value.
    str.remove_prefix(1);
    return {first, str};
  }
  return {first, string_view{}};
}
```

This algorithm is general enough to find its way into `config_option.hpp` as a free function. Bonus points for implementing and using [`string_view::starts_with`](https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with) to further improve readability by getting rid of the awkward `compare(...)`. ;)

After separating the algorithm, the remainder of this function becomes very straightforward.

```c++
  auto ptr = custom_options_.qualified_name_lookup("global.config-file");
  CAF_ASSERT(ptr != nullptr);
  string_list::iterator i;
  string_view path;
  std::tie(i, path) = find_by_long_name(*ptr, args.begin(), args.end());
  if (i == args.end())
    return;
  if (path.empty()) {
    args.erase(i);
    return;
  }
  // ... ptr->parse(...) ... ptr->store(...) ...
```

josephnoir:
Got it running, but unsure about the escaped string handling that happened here before. It looks like `parse` will [handle it](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/src/config_value.cpp#L78)?

josephnoir:
Should we add some error handling while we are on it. Maybe something like this?

```diff
diff --git a/libcaf_core/caf/actor_system_config.hpp b/libcaf_core/caf/actor_system_config.hpp
index 976a398b..358794d9 100644
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -342,7 +342,7 @@ private:

   actor_system_config& set_impl(string_view name, config_value value);

-  void extract_config_file_path(string_list& args);
+  expected<void> extract_config_file_path(string_list& args);

   /// Adjusts the content of the configuration, e.g., for ensuring backwards
   /// compatibility with older options.
diff --git a/libcaf_core/src/actor_system_config.cpp b/libcaf_core/src/actor_system_config.cpp
index e13a21ae..cf8c0658 100644
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -287,7 +287,9 @@ error actor_system_config::parse(string_list args, const char* ini_file_cstr) {
   if (ini_file_cstr != nullptr)
     config_file_path = ini_file_cstr;
   // CLI arguments always win.
-  extract_config_file_path(args);
+  auto success = extract_config_file_path(args);
+  if (!success)
+    return success.error();
   std::ifstream ini{config_file_path};
   return parse(std::move(args), ini);
 }
@@ -367,22 +369,23 @@ std::string actor_system_config::render_pec(uint8_t x, atom_value,
                         meta::omittable_if_empty(), xs);
 }

-void actor_system_config::extract_config_file_path(string_list& args) {
+expected<void> actor_system_config::extract_config_file_path(string_list& args) {
   auto ptr = custom_options_.qualified_name_lookup("global.config-file");
   CAF_ASSERT(ptr != nullptr);
   string_list::iterator i;
   string_view path;
   std::tie(i, path) = find_by_long_name(*ptr, args.begin(), args.end());
   if (i == args.end())
-    return;
+    return unit;
   if (path.empty()) {
     args.erase(i);
-    return;
+    return make_error(pec::missing_argument, std::string{*i});
   }
   auto evalue = ptr->parse(path);
   if (!evalue)
-    return;
+    return std::move(evalue.error());
   ptr->store(*evalue);
+  return unit;
 }

 error actor_system_config::adjust_content() {
```

libcaf_core/caf/sec.hpp
```diff
@@ -114,6 +114,8 @@ enum class sec : uint8_t {
   bad_function_call = 40,
   /// Feature is disabled in the actor system config.
   feature_disabled,
+  /// Failed to open file to read actor system config.
```
Neverlord: That's overly specific. I think the error code should apply to any file, not just the config.

libcaf_core/src/actor_system_config.cpp
```diff
@@ -287,8 +293,12 @@ error actor_system_config::parse(string_list args, const char* ini_file_cstr) {
   if (ini_file_cstr != nullptr)
     config_file_path = ini_file_cstr;
   // CLI arguments always win.
-  extract_config_file_path(args);
+  auto err = extract_config_file_path(args);
```
Neverlord: That's more verbose than it needs to be:

```c++
if (auto err = extract_config_file_path(args))
  return err;
```

libcaf_core/src/actor_system_config.cpp
```diff
@@ -367,22 +377,23 @@ std::string actor_system_config::render_pec(uint8_t x, atom_value,
...
-void actor_system_config::extract_config_file_path(string_list& args) {
+error actor_system_config::extract_config_file_path(string_list& args) {
   auto ptr = custom_options_.qualified_name_lookup("global.config-file");
   CAF_ASSERT(ptr != nullptr);
   string_list::iterator i;
   string_view path;
   std::tie(i, path) = find_by_long_name(*ptr, args.begin(), args.end());
   if (i == args.end())
-    return;
+    return {};
```
Neverlord: It's more idiomatic (and easier to comprehend IMO) to write `return none` instead.

libcaf_core/src/actor_system_config.cpp
```diff
@@ -367,22 +377,23 @@ std::string actor_system_config::render_pec(uint8_t x, atom_value,
...
     args.erase(i);
-    return;
+    return make_error(pec::missing_argument, std::string{*i});
   }
   auto evalue = ptr->parse(path);
   if (!evalue)
-    return;
+    return std::move(evalue.error());
   ptr->store(*evalue);
+  return {};
```
Neverlord:
```c++
return none;
```

Issue #836, opened 2019-05-02
Warning from static analyzer


vyivanov:
[PVS-Studio](https://www.viva64.com/en/w/v1023/) issues this warning on version [0.16.3](https://github.com/actor-framework/actor-framework/blob/ec6a89271472aa7e45cfb72d483fd8cfce0210ac/libcaf_core/caf/actor_system_config.hpp#L240):

```
caf/actor_system_config.hpp:240:1: warning: V1023 A pointer without owner is added to the 'thread_hooks_' container by the 'emplace_back' method. A memory leak will occur in case of an exception.
```

josephnoir:
Should by addressed by #843.

PR #843, opened 2019-05-23
Fix potential memory leak

josephnoir:
Addresses #836. If reallocation throws the pointer is not cleaned up correctly.

### DIFF
--- a/libcaf_core/caf/actor_system_config.hpp
+++ b/libcaf_core/caf/actor_system_config.hpp
@@ -231,7 +231,8 @@ public:
   /// Adds a hook type to the scheduler.
   template <class Hook, class... Ts>
   actor_system_config& add_thread_hook(Ts&&... ts) {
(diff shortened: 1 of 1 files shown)


libcaf_core/caf/actor_system_config.hpp
```diff
@@ -231,7 +231,8 @@ class actor_system_config {
   /// Adds a hook type to the scheduler.
   template <class Hook, class... Ts>
   actor_system_config& add_thread_hook(Ts&&... ts) {
-    thread_hooks_.emplace_back(new Hook(std::forward<Ts>(ts)...));
+    std::unique_ptr<thread_hook> hook(new Hook(std::forward<Ts>(ts)...));
```
Neverlord:
```suggestion
    std::unique_ptr<thread_hook> hook{new Hook(std::forward<Ts>(ts)...)};
```

PR #850, opened 2019-06-20
Set SO_KEEPALIVE on TCP connections

josephnoir:
This PR enables keep alive socket option on TCP sockets. The socket timeout depends on the system settings. On macOS this is apparently 2 hours (keepidle is in msec according to the [Internet](https://stackoverflow.com/questions/15860127/how-to-configure-tcp-keepalive-under-mac-os-x)):

```
$ sysctl -A | grep "net.inet.tcp.*keep"
net.inet.tcp.keepidle: 7200000
net.inet.tcp.keepintvl: 75000
net.inet.tcp.keepinit: 75000
net.inet.tcp.keepcnt: 8
net.inet.tcp.always_keepalive: 0
```


josephnoir:
Oh, and I'm not entirely sure we need a separate function for Windows. The difference is the type of the flag but that could probably solved otherwise.

PR #852, opened 2019-06-27
Fix integer overflow check in number parser

### LINKED ISSUE #851: Parsing IPv4 address from a string sometimes works although it should fail
The parser behaves weirdly if an octet has 4 digits starting with a 1 in the range `1137` to `1279`.

Here's a few observations:
* `"1136.12.12.12"` fails to parse
* `"1137.12.12.12"` parses to `113.12.12.12`
*  `"1138.12.12.12"` parses to `114.12.12.12`
*  continues up to `"1279.12.12.12"`  which parses to `255.12.12.12`
*  `"1280.12.12.12"`  and above fails again

And a test case for `test/ipv4_address.hpp`

```cpp
CAF_TEST(bug) {
  auto should_fail = [](std::string addr) {
    ipv4_address x;
    auto err = parse(addr, x);
    CAF_CHECK(err);
  };
  should_fail("1136.12.12.12");
  should_fail("1137.12.12.12");
  should_fail("1279.12.12.12");
  should_fail("1280.12.12.12");
}
```

Neverlord:
Fixes #851.

### DIFF
--- a/libcaf_core/caf/detail/parser/add_ascii.hpp
+++ b/libcaf_core/caf/detail/parser/add_ascii.hpp
@@ -31,6 +31,7 @@ namespace parser {
 // Sum up integers when parsing positive integers.
 // @returns `false` on an overflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
+// @warning can leave `x` in an intermediate state when retuning `false`
 template <int Base, class T>
 bool add_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {
   CAF_IGNORE_UNUSED(u);
--- a/libcaf_core/caf/detail/parser/add_ascii.hpp
+++ b/libcaf_core/caf/detail/parser/add_ascii.hpp
@@ -19,7 +19,10 @@
 
 #pragma once
 
+#include <limits>
+
 #include "caf/detail/parser/ascii_to_int.hpp"
+#include "caf/detail/type_traits.hpp"
 
 namespace caf {
 namespace detail {
@@ -29,11 +32,26 @@ namespace parser {
 // @returns `false` on an overflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
 template <int Base, class T>
-bool add_ascii(T& x, char c) {
+bool add_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {
+  CAF_IGNORE_UNUSED(u);
+  if (x > (std::numeric_limits<T>::max() / Base))
+    return false;
+  x *= Base;
   ascii_to_int<Base, T> f;
-  auto before = x;
+  auto y = f(c);
+  if (x > (std::numeric_limits<T>::max() - y))
+    return false;
+  x += y;
+  return true;
+}
+
+template <int Base, class T>
+bool add_ascii(T& x, char c,
+               enable_if_tt<std::is_floating_point<T>, int> u = 0) {
+  CAF_IGNORE_UNUSED(u);
+  ascii_to_int<Base, T> f;
   x = (x * Base) + f(c);
-  return before <= x;
+  return true;
 }
 
 } // namespace parser
--- a/libcaf_core/caf/detail/parser/sub_ascii.hpp
+++ b/libcaf_core/caf/detail/parser/sub_ascii.hpp
@@ -31,6 +31,7 @@ namespace parser {
 // Subtracs integers when parsing negative integers.
 // @returns `false` on an underflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
+// @warning can leave `x` in an intermediate state when retuning `false`
 template <int Base, class T>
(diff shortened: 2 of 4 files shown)


libcaf_core/caf/detail/parser/add_ascii.hpp
```diff
@@ -29,11 +33,24 @@ namespace parser {
 // @returns `false` on an overflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
 template <int Base, class T>
-bool add_ascii(T& x, char c) {
+bool add_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {
+  CAF_IGNORE_UNUSED(u);
+  ascii_to_int<Base, T> f;
+  auto y = f(c);
+  if (x > ((std::numeric_limits<T>::max() / Base) - y)
+            + (std::numeric_limits<T>::max() % Base))
```
josephnoir: This does not work in all cases. Here are a few test cases that should fail, but don't and give a wrong address as a result:

```
  should_fail("260.12.12.12");
  should_fail("261.12.12.12"); 
  should_fail("262.12.12.12");
  should_fail("263.12.12.12");
  should_fail("264.12.12.12");
  should_fail("270.12.12.12");
  should_fail("271.12.12.12"); 
  should_fail("272.12.12.12"); 
  should_fail("273.12.12.12"); 
  should_fail("280.12.12.12"); 
  should_fail("281.12.12.12");  
  should_fail("282.12.12.12"); 
  should_fail("291.12.12.12"); 
  should_fail("300.12.12.12");
```

libcaf_core/caf/detail/parser/read_ipv4_address.hpp
```diff
@@ -53,7 +53,8 @@ void read_ipv4_octet(state<Iterator, Sentinel>& ps, Consumer& consumer) {
   uint8_t res = 0;
   // Reads the a decimal place.
   auto rd_decimal = [&](char c) {
-    return add_ascii<10>(res, c);
+    auto b = add_ascii<10>(res, c) && res <= 255;
```
josephnoir: Shouldn't `res <= 255` always be true for `uint8_t`? (And if the check is necessary, shouldn't it use `std::numeric_limits<uint8_t>::max()`.)

libcaf_core/caf/detail/parser/add_ascii.hpp
josephnoir: Oh, and why is the `BASE` part of the check. It is only important for converting the ascii value to an integer.

josephnoir: Ok, I see why the BASE is necessary here. Sorry for the spam. Still doesn't work though.

josephnoir: Adding a check to see if we can even add another digit would address this:
```cpp  
template <int Base, class T>      
bool add_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {      
  CAF_IGNORE_UNUSED(u);      
  // Check if we can add another digit.      
  T res = (x * Base);                                          
  if (res / Base != x)                                         
    return false;                                           
  // Check if we can add this specific digit.              
  ascii_to_int<Base, T> f;                                                                
  auto y = f(c);                                                                                  
  if (x > ((std::numeric_limits<T>::max() / Base) - y)                                                           
            + (std::numeric_limits<T>::max() % Base))                                                            
    return false;                                                                                                
  x = res + y;                                                                                                   
  return true;                                                                                                   
}
```

Neverlord:
Thanks for double-checking my bad code! The new bound checks in `add_ascii` and `sub_ascii` should do the trick - and are reasonably tested now.

libcaf_core/caf/detail/parser/add_ascii.hpp
```diff
@@ -29,11 +32,26 @@ namespace parser {
 // @returns `false` on an overflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
 template <int Base, class T>
-bool add_ascii(T& x, char c) {
+bool add_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {
+  CAF_IGNORE_UNUSED(u);
+  if (x > (std::numeric_limits<T>::max() / Base))
+    return false;
+  x *= Base;
```
josephnoir: If parsing fails this mutates the input value `x`.

libcaf_core/caf/detail/parser/sub_ascii.hpp
```diff
@@ -29,11 +32,26 @@ namespace parser {
 // @returns `false` on an underflow, otherwise `true`.
 // @pre `isdigit(c) || (Base == 16 && isxdigit(c))`
 template <int Base, class T>
-bool sub_ascii(T& x, char c) {
+bool sub_ascii(T& x, char c, enable_if_tt<std::is_integral<T>, int> u = 0) {
+  CAF_IGNORE_UNUSED(u);
+  if (x < (std::numeric_limits<T>::min() / Base))
+    return false;
+  x *= Base;
```
josephnoir: Same as above.

Neverlord:
I think it's never safe to rely on parser output variables in case of an error. However, documenting this fact does not hurt either.

PR #854, opened 2019-07-10
Parallelize deserialization of incoming BASP messages

Neverlord:
This is a followup to https://github.com/actor-framework/actor-framework/pull/821 with fixes for the issues that required to revert the changes via https://github.com/actor-framework/actor-framework/pull/828.


libcaf_io/src/basp_broker.cpp
```diff
@@ -121,6 +121,7 @@ behavior basp_broker::make_behavior() {
     send(this, tick_atom::value, heartbeat_interval);
   }
   return {
+
```
josephnoir: O'rly?

libcaf_io/src/basp_broker.cpp
```diff
@@ -144,17 +145,17 @@ behavior basp_broker::make_behavior() {
...
         message_id mid, const message& msg) {
-      CAF_LOG_TRACE(CAF_ARG(src) << CAF_ARG(dest)
-                    << CAF_ARG(mid) << CAF_ARG(msg));
+      CAF_LOG_TRACE(CAF_ARG(src)
+                    << CAF_ARG(dest) << CAF_ARG(mid) << CAF_ARG(msg));
       if (!dest || system().node() == dest->node()) {
-        CAF_LOG_WARNING("cannot forward to invalid or local actor:"
-                        << CAF_ARG(dest));
+        CAF_LOG_WARNING(
+          "cannot forward to invalid or local actor:" << CAF_ARG(dest));
```
josephnoir: I'm not sure about this one either. Doesn't really matter though.

libcaf_io/src/worker.cpp
```diff
@@ -31,11 +31,11 @@ namespace basp {
 // -- constructors, destructors, and assignment operators ----------------------
 
 worker::worker(worker_hub& hub, message_queue& queue, proxy_registry& proxies)
-    : next_(nullptr),
-      hub_(&hub),
-      queue_(&queue),
-      proxies_(&proxies),
-      system_(&proxies.system()) {
+  : next_(nullptr),
```
josephnoir: Our [contributing](https://github.com/actor-framework/actor-framework/blob/master/CONTRIBUTING.md#breaking-statements) document says something else. There are a few other constructors that were formatted this way in this commit.

libcaf_io/src/basp_broker.cpp
Neverlord: This newline only kept `clang-format` from formatting the lambdas below aligned after the `{`. I did find a better solution though: `behavior{` is a tiny bit more verbose but it gets `clang-format` to correctly align the lambdas two-spaces from the left.

libcaf_io/src/basp_broker.cpp
Neverlord: I've broken up the string literal to get a nicer formatting.

PR #858, opened 2019-07-30
Don't consider uint8_t and bool as same 

tobim:
This extra check is necessary for instantiations of `caf::variant`s containing `bool` and `uint8_t` members in that order. If one tries to construct the variant with `uint8_t`, the `type_id` calculation will place the constructor argument at the `T` argument and the type list entry at `U`. Therefore this additional condition is needed.


tobim:
`clang-format` wants the line above to be
```
bool Enable = std::is_integral<T>::value&& std::is_integral<U>::value
```
A logical and should not attach to a type, that should be reserved to references. I opted to turn clang-format off for the template expression.

Neverlord:
Thanks for the patch! 👍

PR #860, opened 2019-07-30
Move worker_hub to the core library

Neverlord:
We need to re-use the `worker_hub` in `caf::net`, so it makes sense to move the implementation of `worker_hub` to `caf::detail` instead of having it in `caf::io`.


josephnoir:
What is `libcaf_core/caf/detail/worker.hpp` there for? maybe I'm missing something, but it doesn't seem to be use anywhere.

Neverlord:
You are correct, this one simply slipped through earlier refactoring.

PR #865, opened 2019-08-01
Add byte and span replacements

Neverlord:
`std::byte` and `std::span` are essential additions to the STL. Unfortunately, they're C++17 and C++20 respectively. This PR adds standard-compatible types that we can start using right away and work in C++11.


mavam:
We might switch to your implementation and ditch our hand-rolled version in VAST. 🙂 

@tobim, any objections?

Neverlord:
As long as you're not using the `Extent` parameter of `vast::span` for doing spawn manipulation in `constexpr` functions, there shouldn't be showstoppers.

tobim:
> @tobim, any objections?

Sounds good to me.

PR #879, opened 2019-08-14
Fix potential segfault with logging enabled

### LINKED ISSUE #878: Potential segfault when using streams with logging enabled
When streaming a type that wasn't added to the uniform type info list (e.g. when using whitelisted types), CAF segfaults in the constructor of `inbound_path` if the log level is DEBUG or lower.

Neverlord:
Close #878.

### DIFF
--- a/libcaf_core/caf/uniform_type_info_map.hpp
+++ b/libcaf_core/caf/uniform_type_info_map.hpp
@@ -75,32 +75,45 @@ public:
 
   /// Returns the portable name for given type information or `nullptr`
   /// if no mapping was found.
-  const std::string* portable_name(uint16_t nr, const std::type_info* ti) const;
+  const std::string& portable_name(uint16_t nr, const std::type_info* ti) const;
 
   /// Returns the portable name for given type information or `nullptr`
   /// if no mapping was found.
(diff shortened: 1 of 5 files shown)


libcaf_core/caf/uniform_type_info_map.hpp
```diff
@@ -75,32 +75,45 @@ class uniform_type_info_map {
...
+
+  /// Lock for accessing `ad_hoc_`.`
   mutable detail::shared_spinlock ad_hoc_mtx_;
 
-  // message type names
+  /// Names of builtin types.
   std::array<std::string, type_nrs - 1> builtin_names_;
+
+  /// Displayed name for unknown types.
+  std::string default_type_name_;
```
josephnoir: This could be const?

Neverlord: Making member variables `const` comes with many strings attached (makes a type unmovable, for example). Generally, I wouldn't put `const` to member variables except for a few niche applications.

josephnoir: I see.

PR #883, opened 2019-08-21
Fix cmake output dir checks

jakobod:
This PR fixes the check for output dirs in CMakeLists.txt


jakobod:
I believe, we still have to think about the "release-type" directory in MSVC builds. We probably have to fix the `FindCAF.cmake` script along this PR.

Neverlord:
> We probably have to fix the FindCAF.cmake script along this PR.

You're right, we need to update that as well.

PR #884, opened 2019-08-21
Add to_string for authority type of URI class

josephnoir:


libcaf_core/caf/detail/append_uri.hpp
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include "caf/detail/append_hex.hpp"
+
+namespace caf {
+namespace detail {
+
+// Escapes all reserved characters according to RFC 3986 in `x` and
+// adds the encoded string to `str`.
+void append_uri(std::string& str, string_view x, bool is_path = false);
```
Neverlord: `append_uri` does not make a lot of sense to me, since it takes a `string_view` instead of an URI. Either we should make the context clear via an extra suffix or (my vote) we use the actual name of encoding: [`append_percent_encoded`](https://en.wikipedia.org/wiki/Percent-encoding).

libcaf_core/caf/detail/append_uri.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * License 1.0. See accompanying files LICENSE and LICENSE_ALTERNATIVE.       *
+ *                                                                            *
+ * If you did not receive a copy of the license files, see                    *
+ * http://opensource.org/licenses/BSD-3-Clause and                            *
+ * http://www.boost.org/LICENSE_1_0.txt.                                      *
+ ******************************************************************************/
+
+#pragma once
+
+#include "caf/detail/append_hex.hpp"
```
Neverlord: Missing include: `caf/fwd.hpp` (for the forward declaration of `string_view`).

libcaf_core/src/uri.cpp
```diff
@@ -101,6 +102,27 @@ std::string to_string(const uri& x) {
...
 }
 
+std::string to_string(const uri::authority_type& x) {
+  std::string str;
+  if (!x.userinfo.empty()) {
+    detail::append_uri(str, x.userinfo);
+    str += '@';
+  }
+  auto addr = get_if<ip_address>(&x.host);
+  if (addr == nullptr) {
```
Neverlord:
```suggestion
  if (auto addr = get_if<ip_address>(&x.host)) {
```

(note: switches the `if` and `else` branch)

libcaf_core/src/uri.cpp
```diff
@@ -101,6 +102,27 @@ std::string to_string(const uri& x) {
...
+    detail::append_uri(str, get<std::string>(x.host));
+  } else {
+    str += '[';
```
Neverlord: IPv4 addresses should not use the `[]` notation. It's "technically valid", because any IPv4 address also is an IPv6 address, but `to_string` output is meant for humans. 🙂

```c++
if (addr->embeds_v4()) // omit parentheses
else // print parentheses
```

libcaf_core/src/uri.cpp
```diff
@@ -101,6 +102,27 @@ std::string to_string(const uri& x) {
...
 
+std::string to_string(const uri::authority_type& x) {
+  std::string str;
+  if (!x.userinfo.empty()) {
+    detail::append_uri(str, x.userinfo);
+    str += '@';
+  }
+  auto addr = get_if<ip_address>(&x.host);
+  if (addr == nullptr) {
+    detail::append_uri(str, get<std::string>(x.host));
```
Neverlord: It's unlikely that the `host` variant receives further types, but I'd still prefer either using `caf::visit` for the dispatching instead to get a compiler error if the definition of `host` ever changes or at least put a `CAF_ASSERT(holds_alternative<std::string>(x.host))` here.

libcaf_core/caf/detail/append_uri.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace caf {
+namespace detail {
+
+// Escapes all reserved characters according to RFC 3986 in `x` and
+// adds the encoded string to `str`.
+void append_uri(std::string& str, string_view x, bool is_path = false);
+
+} // namespacvec detail
```
Neverlord:
```suggestion
} // namespace detail
```

Issue #866, opened 2019-08-03
Conflicting 'uninstall' targets with CMake.


ChristofferGreen:
If you add actor-framework together with another library via for example the CMake FetchContent_Declare functionality and they both declare an uninstall target cmake will complain with: "add_custom_target cannot create target "uninstall" because another target with the same name already exists." since CMake targets must be unique. 

Possible solution: First check if the uninstall target already exists before adding it.

Here is another library that had the same issue (I worked around this issue with the same solution): https://github.com/leethomason/tinyxml2/issues/516

Neverlord:
> Possible solution: First check if the uninstall target already exists before adding it.

Sounds reasonable. However, looking at the patch in `tinyxml`, I wonder whether dropping the `uninstall` instructions entirely if the target already exists is the right approach. Shouldn't the `uninstall` target get expanded in this case? Would this be the correct CMake patch?

```cmake
# process cmake_uninstall.cmake.in
configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
               "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
               IMMEDIATE @ONLY)
# add uninstall target
if(NOT TARGET uninstall)
  add_custom_target(uninstall
                    COMMAND "${CMAKE_COMMAND}" -P
                    "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake")
else()
  add_custom_command(TARGET uninstall
                     PRE_BUILD
                     COMMAND "${CMAKE_COMMAND}" -P
                     "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake")
endif()
```

tobim:
how about:
``` cmake
# process cmake_uninstall.cmake.in
configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
               "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake"
               IMMEDIATE @ONLY)
if(NOT TARGET uninstall)
  add_custom_target(uninstall)
endif()
add_custom_target(uninstall_caf
                  COMMAND "${CMAKE_COMMAND}" -P
                  "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake")
add_dependencies(uninstall uninstall_caf)
```
?

Neverlord:
[...]

PR #886, opened 2019-08-21
Make it easier to build CAF as a CMake subproject

tobim:
Closes https://github.com/actor-framework/actor-framework/issues/866.


Neverlord:
Thanks for the fix! 👍

ChristofferGreen:
Great to see it fixed, thank you.

PR #888, opened 2019-08-22
Add overloads to detail::parse for all IP types

Neverlord:


libcaf_core/caf/ipv6_endpoint.hpp
```diff
@@ -18,19 +18,26 @@
...
+#include "caf/fwd.hpp"
 #include "caf/ipv6_address.hpp"
 #include "caf/meta/type_name.hpp"
 
 namespace caf {
 
 /// An IP endpoint that contains an ::ipv6_address and a port.
-class ipv6_endpoint : detail::comparable<ipv6_endpoint> {
+class ipv6_endpoint : detail::comparable<ipv6_endpoint>,
+                      detail::comparable<ipv6_endpoint, ipv4_endpoint> {
```
jakobod: Wouldn't it be useful to add this to the ipv4_endpoint too? Now one would have to compare `ipv6_endpoint`s with `ipv4_endpoint`s like this: `ipv6 > ipv4` but comparing like `ipv4 < ipv6` would not be possible, wouldn't it?

Neverlord: That would actually produce ambiguous overloads. `comparable` injects operators for both directions.

jakobod: Nevermind. I just figured this out. 👍

libcaf_core/test/ipv6_endpoint.cpp
```diff
@@ -138,6 +104,12 @@ CAF_TEST(constructing assigning and hash_code) {
   CAF_CHECK_EQUAL(ep1.hash_code(), ep2.hash_code());
 }
 
+CAF_TEST(comparison to IPv4) {
+  ipv4_endpoint v4{ipv4_address({127, 0, 0, 1}), 8080};
+  ipv6_endpoint v6{v4.address(), v4.port()};
+  CAF_CHECK_EQUAL(v4, v6);
```
jakobod: Maybe add a test that checks the other way around too. (`v6 == v4`)

PR #891, opened 2019-08-25
Add function overloads for URIs to the middleman

Neverlord:


josephnoir:
Just glanced over this so far. I'll give it a closer look (hopefully tomorrow).

I think we talked about this before, but do we want to use `tcp` as a scheme here? Actors use BASP over TCP which would be different to TCP or UDP used by broker. This leads to the second question: How does this affect brokers? Will this remain a high-level API specific to publishing and "remote-actoring" actors or do we want to to use this everywhere eventually.

Neverlord:
> How does this affect brokers? 

I don't think we need to refine the Broker API further. They're going to get deprecated anyways.

> Will this remain a high-level API specific to publishing and "remote-actoring" actors (...)

Yes. The one thing I want to address is to make it easier to setup CAF connection and also simplify switching between modules. For example, switching between `caf_io` and `caf_openssl` should be as simple as replacing `tcp://hostname:8080` with `ssl://hostname:8080` in the config file.

> I think we talked about this before, but do we want to use `tcp` as a scheme here?

I don't think BASP is something we need or should expose to users. If we really want to avoid using URIs with plain `tcp`, I think we could follow Akka's example and use `caf.tcp`, `caf.ssl`, `caf.udp`, `caf.quic`, etc. as scheme.

josephnoir:
> I don't think BASP is something we need or should expose to users. If we really want to avoid using URIs with plain tcp, I think we could follow Akka's example and use caf.tcp, caf.ssl, caf.udp, caf.quic, etc. as scheme.

Well, if we only offer this API for the "actor" functions I don't think we will run into a problems.

Looking at the `net` module using a prefix might make the transition easier since the module will separate actors and BASP more cleanly. I'm not sure if we might run into ambiguities then. Think of a potential new broker implementation where we would want `tcp` to mean TCP and not BASP over TCP.

Neverlord:
>Looking at the net module using a prefix might make the transition easier

How so? Ideally this is transparent to the user. We could simply always prefer `caf_net`, throw an error when users load both `caf_io` and `caf_net`, or something along those lines. But I think when we provide a URI-based API the thing we don't want to do is changing our scheme.

josephnoir:
> How so?

If use the prefix from the beginning we don't have to address a conflict later on and have a different API for `net` and `io`. 

> But I think when we provide a URI-based API the thing we don't want to do is changing our scheme.

Yeah, I think that's what I wanted to say. So, an alternative to just using `tcp` as a scheme and have it mean BASP over TCP for actors and raw TCP for brokers (not yet, but potentially later when we introduce a related API in the net module) is the caf prefix you suggested. That would make `caf.$TRANSPORT` mean `$TRANSPORT` *for actors* and a broker would simply use `$TRANSPORT` for *raw* `$TRANSPORT`.

Neverlord:
Ok. One thing, though: the URIs are going to change anyways when transitioning to `caf_net`, because we'll use names instead of ports.

For example:

```ini
; three different actors on earth, all published to a different port
remotes = [
  <caf.tcp://earth:8001>,
  <caf.tcp://earth:8002>,
  <caf.tcp://earth:8003>
]
```

would become

```ini
; three different actors on earth, all registered under a different name
remotes = [
  <caf.tcp://earth:8000/actors/actor-a>,
  <caf.tcp://earth:8000/actors/actor-b>,
  <caf.tcp://earth:8000/actors/actor-c>
]
```

It's probably a bad idea to introduce an URI-based interface for the old middleman when we want to switch everyone to the new `caf_net` API soon(-ish). Sorry for the noise.

PR #893, opened 2019-08-27
Add unit tests for settings

Neverlord:
I noticed that the `settings` header had no unit tests. This test suite covers all functions defined in `settings.hpp`.

### DIFF
--- a/libcaf_core/test/settings.cpp
+++ b/libcaf_core/test/settings.cpp
@@ -1,0 +1,162 @@
+/******************************************************************************
+ *                       ____    _    _____                                   *
+ *                      / ___|  / \  |  ___|    C++                           *
+ *                     | |     / _ \ | |_       Actor                         *
+ *                     | |___ / ___ \|  _|      Framework                     *
+ *                      \____/_/   \_|_|                                      *
+ *                                                                            *
+ * Copyright 2011-2019 Dominik Charousset                                     *
+ *                                                                            *
+ * Distributed under the terms and conditions of the BSD 3-Clause License or  *
+ * (at your option) under the terms and conditions of the Boost Software      *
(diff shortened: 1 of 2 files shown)


libcaf_core/test/settings.cpp
```diff
@@ -0,0 +1,162 @@
...
+  CAF_CHECK(x.contains("logger"));
+  CAF_CHECK(x.contains("one"));
+  CAF_CHECK_EQUAL(unpack(x, "foo"), "bar"_s);
+  CAF_CHECK_EQUAL(unpack(x, "logger", "console"), atom("none"));
+  CAF_CHECK_EQUAL(unpack(x, "one", "two", "three"), "four"_s);
+  put(x, "logger.console", atom("trace"));
+  CAF_CHECK_EQUAL(unpack(x, "logger", "console"), atom("trace"));
+}
+
+CAF_TEST(put missing) {
```
jakobod: You are not checking that an already contained value isn't added again! Maybe try to add some contained value with a different key and check that the size stays the same and the old key + value is still contained afterwards.

Neverlord: The last `put_missing` checks that, does it not?

```c++
  put_missing(x, "logger.console", atom("trace"));
  CAF_CHECK_EQUAL(unpack(x, "logger", "console"), atom("none"));
```

jakobod: True. I misread `atom("trace")` for `atom("none")`.
I'm still confused by the description of the function regarding this test:
```
/// Converts `value` to a `config_value` and assigns it to `key` if `value` is
/// currently missing in `xs`.
```
Shouldn't you try to add the same value with a different key? The description doesn't say anything about handling same keys.

Neverlord: Documentation fail. Will fix.

PR #895, opened 2019-08-28
Fix infinite recursion in read_ini

### LINKED ISSUE #894: endless loop while parsing a config file if line starts with a non-alphanumeric character
Let's take a minimal program minapp.cpp:
`#include "caf/exec_main.hpp"`
``void caf_main(caf::actor_system& system)``
`{`
`}`
`CAF_MAIN()`

and a minimal config file caf-application.ini with 1 line and 1 character:
`#`
Running it as
`minapp --config-file=caf-application.ini`
goes into endless loop, without error diagnostics.

Looks like it is endlessly bouncing in libcaf_core/caf/detail/parser/read_ini.hpp, in read_ini() between term_state(init) and unstable_state(return_to_global).

The same effect if '#" in config file is replaced with '/', '\', '.', ',', '!', '@', '$', '%', , '^', '&', '('.

Neverlord:
Close #894.

### DIFF
--- a/libcaf_core/test/read_ini.cpp
+++ b/libcaf_core/test/read_ini.cpp
@@ -106,7 +106,7 @@ log_type make_log(Ts&&... xs) {
 }
 
 // Tests basic functionality.
-const char* ini0 = R"(
+const auto ini0 = R"(
 [1group]
 1value=321
 [_foo]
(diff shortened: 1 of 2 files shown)


libcaf_core/test/read_ini.cpp
```diff
@@ -242,6 +242,14 @@ const auto ini1_log = make_log(
 );
 // clang-format on
 
+const char* ini2 = "#";
```
josephnoir: This  and `ini3` below could be `auto` as well, right?

PR #897, opened 2019-08-28
Streamline CMake scaffold

tobim:
This PR replaces the `libcaf_*_shared` and `libcaf_*_static` targets with a consolidated `libcaf_*`. Switching between `SHARED` and `STATIC` builds can be done with the builtin mechanism of `BUILD_SHARED_LIBS` which is defaulted to `ON`. This means that shared and static can not be built in the same tree any more. In reality this is not a loss, as having both variants in the same environment is nothing but asking for trouble.
In addition, the new targets are now properly exported to the build tree and installed to `$prefix/lib/cmake/caf`. This allows cmake to find CAF without the need to add `FindCAF.cmake`.


Neverlord:
I'm very reluctant to merge big CMake changes like this. I see a lot of potential for breaking things at various user setups and I want to avoid spending time fiddling with CMake as much as possible.

That being said, I agree that removing the option to build both static and shared libraries in the same tree is the right thing to do. That's not the only thing this overhaul does, though. So before going into details, I got a couple of questions:

- Does this affect downstream projects such as our [incubator](https://github.com/actor-framework/incubator)?
- Does it change the behavior of CAF for third-parties that integrate CAF as CMake subfolder?
- Are our old `FindCAF` scripts still properly detecting CAF installations generated after the chagnes?
- You say one benefit of the changes is that we no longer need `FindCAF.cmake`, but why did you then make changes to that file?

tobim:
I understand your worries and you're raising important points. 

> That's not the only thing this overhaul does

But it is the only "breaking" thing in this PR. Regarding that, there is an option to implement a workaround that makes this not breaking:
 - Users that integrate CAF via `find_package()` _should_ not be affected at all.
 - Users that use `add_subdirectory(caf)` and try to link to `libcaf_*_shared` or `libcaf_*_static` would have their setup break. This could be worked around by declaring those targets as aliases to `libcaf_*` depending on `BUILD_SHARED_LIBS`, and setting `BUILD_SHARED_LIBS=OFF` in the CAF project if `CAF_BUILD_STATIC` is on.


>  Does this affect downstream projects such as our incubator?

Since that project uses `find_package(caf)` it should not be affected.

>  Does it change the behavior of CAF for third-parties that integrate CAF as CMake subfolder?

See above.

>  Are our old FindCAF scripts still properly detecting CAF installations generated after the chagnes?

Yes, I did not make any substantial change to `FindCAF.cmake`.

> You say one benefit of the changes is that we no longer need FindCAF.cmake, but why did you then make changes to that file?

I added `imported` targets, to make it possible to use all methods of linking to CAF in the same way:
`target_link_libraries(myTarget PUBLIC caf::core caf::io ...)` regardless of using `find_package(CAF)` with  the `FindCAF.cmake` module, with the `CAFConfig` (new) style, or with `add_subdirectory()`.
This allows for a smoother upgrade Path from the module style to the config style: First update `FindCAF.cmake` and change to linking in the way I described before, this will also work with earlier versions of CAF. When you are confident everyone updated to a CAF version that provides the `CAFConfig` support, just remove `FindCAF.cmake` and everything should continue to work.

jsiwek:
Since the concern of breaking downstream user setups came up, just thought I'd point out our own usage for consideration.  Our default build embeds CAF via a CMake ExternalProject and does reference options regarding CAF static libs:

https://github.com/zeek/broker/blob/6be21839f112057a4f20b62d847ab0a272b78773/CMakeLists.txt#L147-L148
https://github.com/zeek/broker/blob/6be21839f112057a4f20b62d847ab0a272b78773/CMakeLists.txt#L176-L184

This has been stable for us for a while now, so it would be a bit of a nuisance to have to change things up and potentially introduce new bugs while doing so.  (Not that I've looked at all into what types of changes would be needed).

Other initial impressions/opinions in case it helps:

* I didn't completely understand what the benefits of this PR are for CAF (maybe my fault for not looking more closely at the changes).  There's maintainability issues with the current approach?  Is it preventing some use-case or does the PR enable a new use-case?

* I don't completely buy the claim that it's "wrong" to build both shared and static libs within the same tree -- that's the user's prerogative.  Maybe it gets them into trouble, or maybe they really do know what they are doing and this ability makes their life a bit simpler.  Since users are currently trusted to be in the later camp, this move seems less like an improvement and more like a regression.

tobim:
Thanks for bringing your use case to my attention: I tried to build `broker` against the branch in this PR and had to make a few modifications to the ExternalProject path:
1. The files of the static libraries don't have to `_static` suffix any more, so the `IMPORTED_LOCATION` properties change.
 -.> The output paths could be changed back, but I don't know of any other library that does this and it requires extra work for users without any benefit, so I'd prefer to keep the change.
2. The flag `CMAKE_POSITION_INDEPENDENT_CODE` has to be set to `${ENABLE_SHARED}` in the call to build CAF. This is because if one builds broker with `./configure --enable-static` caf is linked statically even to the shared version of `libbroker` (This behavior is the same with the current master), and cmake doesn't build static libs with `-fpic` by default.

The full diff is:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index acdc61a..7d9cb08 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,13 @@ if ( ENABLE_CCACHE )
   set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_PROGRAM})
 endif ()

+set(ENABLE_SHARED true)
+
+if (ENABLE_STATIC_ONLY)
+  set(ENABLE_STATIC true)
+  set(ENABLE_SHARED false)
+endif ()
+
 # -- Dependencies -------------------------------------------------------------

 if ( CAF_ROOT_DIR )
@@ -146,6 +153,7 @@ else ()
         -DCAF_NO_OPENCL:bool=yes
         -DCAF_BUILD_STATIC:bool=${ENABLE_STATIC}
         -DCAF_BUILD_STATIC_ONLY:bool=${ENABLE_STATIC_ONLY}
+        -DCMAKE_POSITION_INDEPENDENT_CODE:bool=${ENABLE_SHARED}
         # Sticking CAF libs in the same output dir as libbroker is a bit of a
         # hack until CMake 3.8, which has BUILD_RPATH that can instead be set
         # on libbroker.  The problem here is that broker unit tests will fail
@@ -177,11 +185,11 @@ else ()
     add_library(libcaf_core_static STATIC IMPORTED)
     add_library(libcaf_io_static STATIC IMPORTED)
     set_property(TARGET libcaf_core_static PROPERTY IMPORTED_LOCATION
-                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_core_static${static_ext})
+                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_core${static_ext})
     set_property(TARGET libcaf_io_static PROPERTY IMPORTED_LOCATION
-                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_io_static${static_ext})
+                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_io${static_ext})
     set_property(TARGET libcaf_openssl_static PROPERTY IMPORTED_LOCATION
-                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_openssl_static${static_ext})
+                 ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libcaf_openssl${static_ext})
   else ()
     add_library(libcaf_core_shared SHARED IMPORTED)
     add_library(libcaf_io_shared SHARED IMPORTED)
@@ -275,12 +283,6 @@ list(GET _version_numbers 1 BROKER_VERSION_MINOR)

 # The SO number shall increase only if binary interface changes.
 set(BROKER_SOVERSION 2)
-set(ENABLE_SHARED true)
-
-if (ENABLE_STATIC_ONLY)
-  set(ENABLE_STATIC true)
-  set(ENABLE_SHARED false)
-endif ()

 # -- compiler -----------------------------------------------------------------
```

> I didn't completely understand what the benefits of this PR are for CAF (maybe my fault for not looking more closely at the changes). There's maintainability issues with the current approach? Is it preventing some use-case or does the PR enable a new use-case?

I have the same use case as `broker`: I want to be able to use CAF from another cmake based project. But instead of invoking `ExternalProject`, [I combine the build instruction with `add_subdirectory()`](https://github.com/tenzir/vast/blob/master/CMakeLists.txt#L429-L467). I opened this PR for the most part to get rid of the boilerplate that is in the second half of the highlighted code.

The other part of this is the addition of the generated `CAFConfig.cmake` and `CAFTargets.cmake` files. These enable the `CONFIG` variant of `find_package()` and should be seen as the successor of the `FindCAF.cmake` method. See https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages for a simple example of that method.

> I don't completely buy the claim that it's "wrong" to build both shared and static libs within the same tree -- that's the user's prerogative. Maybe it gets them into trouble, or maybe they really do know what they are doing and this ability makes their life a bit simpler. Since users are currently trusted to be in the later camp, this move seems less like an improvement and more like a regression.

I'm not able to come up with a legitimate use case of having both static and shared libraries in the build tree. It's just an avenue to get your final build results into an inconsistent state, and therefore a bad API.

Neverlord:
I think @jsiwek raises some valid concerns and this PR can introduce a regression for users without providing a serious upside at the moment.

Regarding compiling both static and shared objects in the same build: putting flavor discussions aside, there is nothing wrong with giving users the option to do this... on UNIX. However, on Windows we actually need to generate different build config headers for static and shared library builds. Hence, "dual builds" will get the axe anyway. @uentity already did a lot of the heavy lifting for proper shared library support on Windows in https://github.com/actor-framework/actor-framework/pull/767. I only need some more time to really think this through, put in some testing on Windows and come to a proper solution that works in the long term.

Since we have to introduce this regression anyway, I'm also in favor of "picking up" the improvements to the CMake setup in this PR. @tobim consequently, I am going to coordinate the CMake changes here and the support for shared libraries on Windows, so this PR might remain open for a while (until maybe 0.18 or 0.19) and I'm probably going to merge it into an integration branch first.

tobim:
Do you want me to merge @uentity's work into this PR, or shall we do that in you proposed integration branch?

PR #904, opened 2019-09-02
Add scheduled_send for sending with absolute timeout

### LINKED ISSUE #901: Add scheduled_send for delayed sends with absolute timeout
Currently, our API only allows sending messages with relative timeout via `delayed_send`.

Neverlord:
Close #901.

### DIFF
--- a/libcaf_core/caf/send.hpp
+++ b/libcaf_core/caf/send.hpp
@@ -71,6 +71,7 @@ void send_as(const Source& src, const Dest& dest, Ts&&... xs) {
 template <message_priority P = message_priority::normal, class Source,
           class Dest, class... Ts>
 void unsafe_send_as(Source* src, const Dest& dest, Ts&&... xs) {
+  static_assert(sizeof...(Ts) > 0, "no message to send");
   if (dest)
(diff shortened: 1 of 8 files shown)


libcaf_core/caf/send.hpp
```diff
@@ -108,6 +110,38 @@ void anon_send(const Dest& dest, Ts&&... xs) {
                   std::forward<Ts>(xs)...);
 }
 
+template <message_priority P = message_priority::normal, class Dest = actor,
+          class Rep = int, class Period = std::ratio<1>, class... Ts>
+detail::enable_if_t<!std::is_same<Dest, group>::value>
+delayed_anon_send(const Dest& dest, std::chrono::duration<Rep, Period> rtime,
+                  Ts&&... xs) {
+  static_assert(sizeof...(Ts) > 0, "no message to send");
```
josephnoir: This check are no longer part of the sender mixin. Does one of the other templates implicitly perform it or why did you add it here but remove it in those functions?

Neverlord: Looks like a simple oversight. Good catch!

PR #913, opened 2019-09-10
Reduce stack usage and indirections in serializers

### LINKED ISSUE #912: High stack usage of serializers
Passing a struct with many members (≥100) to a serializer or deserializer can cause a stack overflow. One of the culprits is `apply_derived` in `data_processor` that essentially doubles the stack depth (we assumed compilers to inline these calls, but that appears to not work as expected).

Neverlord:
Optimize `data_processor` in order to reduce the risk of stack overflows when using classes with many (≥100) data members.

Close #912.

### DIFF
--- a/libcaf_core/src/error.cpp
+++ b/libcaf_core/src/error.cpp
@@ -19,10 +19,11 @@
 #include "caf/error.hpp"
 
 #include "caf/config.hpp"
+#include "caf/deep_to_string.hpp"
+#include "caf/deserializer.hpp"
+#include "caf/make_message.hpp"
 #include "caf/message.hpp"
 #include "caf/serializer.hpp"
-#include "caf/deserializer.hpp"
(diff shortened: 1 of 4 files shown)


libcaf_core/src/error.cpp
```diff
@@ -19,10 +19,11 @@
 #include "caf/error.hpp"
 
 #include "caf/config.hpp"
+#include "caf/deep_to_string.hpp"
+#include "caf/deserializer.hpp"
```
jakobod: Is this include necessary?

libcaf_core/src/error.cpp
```diff
@@ -19,10 +19,11 @@
...
+#include "caf/make_message.hpp"
 #include "caf/message.hpp"
 #include "caf/serializer.hpp"
```
jakobod: This seems unecessary too.

PR #917, opened 2019-09-16
Use CMAKE_INSTALL_* vars for install paths

dcode:
Updates each `CMakeLists.txt` to use `CMAKE_INSTALL_*` from GNUInstallDirs. This makes it easier to package for package repositories while maintaining user overrides via CMake cache and defines.


dcode:
I made this patch against 0.17.1, but looks like it still applies to master

Neverlord:
Welcome to CAF and thank you for submitting the patch! 🎉

PR #927, opened 2019-10-08
Auto-generate to_string implementations for enums

Neverlord:
The generator only works for enums that live in their own header. Hence, I've excluded enums that live as implementation detail within classes etc. The "big ones" like SEC and PEC are the important ones anyway, because the frequently change.


cmake/caf-generate-enum-strings.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return EXIT_FAILURE;
+  }
+  vector<string> namespaces;
+  string enum_name;
+  string line;
+  bool is_enum_class = false;
+  // Locate the beginning of the enum.
+  for (;;) {
+    if (!getline(in, line)) {
+      cerr << "unable to locate enum in file file: " << argv[1] << '\n';
```
josephnoir:
```suggestion
      cerr << "unable to locate enum in file: " << argv[1] << '\n';
```

PR #943, opened 2019-10-23
Fix compilation of the python binding and apply clang format

mroethke:
What the title says.


Neverlord:
Thanks for the fix! 🎉

PR #945, opened 2019-10-23
correct selecting actor_system for actor when sending message

d-chugunov:
For example, we have two actor system. When an actor from the system 1 calls method send of an actor from the system 2 the latter actor will be executed in a worker of the system 1 instead of its system 2. My patch fixes this behaviour. Besides there are another places in the framework which should be fixed as well.

### DIFF
--- a/libcaf_core/caf/mixin/sender.hpp
+++ b/libcaf_core/caf/mixin/sender.hpp
@@ -65,7 +65,7 @@ public:
     type_check(dest, args_token);
     if (dest)
       dest->eq_impl(make_message_id(P), dptr()->ctrl(),
-                    dptr()->context(), std::forward<Ts>(xs)...);
+                    nullptr, std::forward<Ts>(xs)...);
   }
 
   /// Sends `{xs...}` as an asynchronous message to `dest` with priority `mp`.
@@ -79,7 +79,7 @@ public:
                   "communicating with dynamically typed actors");
     if (dest)
       dest->get()->eq_impl(make_message_id(P), dptr()->ctrl(),
-                           dptr()->context(), std::forward<Ts>(xs)...);
+                           nullptr, std::forward<Ts>(xs)...);
   }
 
   template <message_priority P = message_priority::normal,


Neverlord:
Welcome to CAF! I'm always excited to see more people contributing! 🙂

Unfortunately, though, I have to reject this PR. Passing actor handles between actor systems is *not* supported and fundamentally unsafe. Actors should never "leak" out of their enclosing actor system, because this breaks fundamental assumptions regarding object lifetimes, actor identities, and so on.

By not propagating the context, this patch also pessimizes performance of the work stealing scheduler by forcing actors to go through the central dispatching and also by reducing cache locality.

What use case do you have for using two actor systems in one process? If you can share your use case and some of your requirements, maybe we can find a safe way to achieve what you are looking for by adding necessary API pieces if necessary.

The only safe way to interconnect two actor systems at the moment is to use `publish`/`remote_actor` over localhost connection. This is awkward, because you have to open a local port. Our redeisgned I/O layer [`caf_net`](https://github.com/actor-framework/incubator) will allow you to use connected [socket pair](https://linux.die.net/man/2/socketpair), which is more efficient and does not require opening ports. The new module still needs some time in the oven, though.

d-chugunov:
> Passing actor handles between actor systems is not supported and fundamentally unsafe. 

So I speak not about passing actor handles between actor systems but about executing actor in a worker in its actor system. In current implementation actor is executed in the sender actor system. For example, if actor1 is from system1 and actor2 is from system2 and the behaviourcode for actor1 executed from worker of system1 calls ` actor2.send(...)` the behaviour code for actor2 will be executed in worker of system1 (not in worker of system2).

Neverlord:
Nice to see that @woelke's thesis makes some waves. 🙂

One goal in the thesis was to keep the NUMA HW transparent. This does not seem relevant to you, since you already designed your system around NUMA.

Ok, so pinning actors to threads seems to be the problem. Let's see if we can find a solution that does not involve multiple actor systems (and potential UB as a result).

Do you have a fixed number of workers per NUMA node? That would make it rather simple: use blocking / detached actors and then pin their threads to the right NUMA node. You can define thread hooks that CAF will call whenever it spins up a new thread.

If you do need a variable set of actors: would extending `spawn` with an option to pin an actor to a particular execution unit solve your problem? Detached actors are pinned to their private thread, so a patch to generalize this facility is pretty straight forward.

> So I speak not about passing actor handles between actor systems but about executing actor in a worker in its actor system.

Just to clarify: if you send a message from actor A living in system 1 to actor B living in system 2, you are passing an actor handle from A to B (because the message contains a handle to the sender). Also, actor A has to get access to B in the first place. You could send a message without sender information via `anon_send`, but you are still holding some reference to B in actor A (which is fundamentally unsafe). Let's find a solution that's not involving two actor systems in the first place. 😉

d-chugunov:
> Do you have a fixed number of workers per NUMA node? That would make it rather simple: use blocking / detached actors and then pin their threads to the right NUMA node. You can define thread hooks that CAF will call whenever it spins up a new thread.

Yes, we have fixed number of workers per NUMA node (one for each processor unit). Currently we use event_based_actor. So it would be better to find a solution for this type of actor. Besides if we put all actors in one actor system the work stealing policy may be used and tasks may be processed on wrong processor unit, isn't it? Moreover it would be better if work of actors for connections didn't depend on actors for tasks processing.

Neverlord:
> Yes, we have fixed number of workers per NUMA node (one for each processor unit)

If you have exactly one actor per CPU, than it seems like you use a simpler design:
- you can give each event-based actor (worker) its own thread by spawning it with the `detached` flag
- you can add custom code that runs whenever CAF spins up a thread by calling `add_thread_hook` on the [`actor_system_config`](http://actor-framework.org/doc/classcaf_1_1actor__system__config.html)
- in your thread hook, you can use the native threading API (e.g. POSIX) to pin the thread to specific core (unfortunately there's no API for that in `std::thread`)

Is there anything that speaks against using this simple design in your case?

d-chugunov:
> you can add custom code that runs whenever CAF spins up a thread by calling add_thread_hook on the actor_system_config
This hook will be executed for every threads and even in actor_system contructor. But in my case the hook should be executed for detached actors only. Besides the hook should have some information (like NUMA node index or actor index) about the actor which have created c new thread.

PR #946, opened 2019-10-23
Add experimental API for actor profiling

Neverlord:
Currently, we ship the `profiled_coordinator` for collecting some data about actors at runtime. However, doing this at the scheduler level is inflexible, operates only at the actor level rather than on the message level, and is blind to detached and blocking actors.

This PR adds a new interface `actor_profiler` that gives users callbacks for actor creation and message processing. Currently, this feature is off by default and requires a CMake option. Mainly because we need some benchmarking to see if there are performance implications. I don't expect impact on performance when not setting a profiler, but we need to measure it regardless.


libcaf_core/caf/actor_profiler.hpp
```diff
@@ -0,0 +1,76 @@
...
+  /// member function after the constructor of `self` has completed but before
+  /// constructing the behavior .
+  /// @param self The new actor.
+  /// @param parent Points to the parent actor unless `self` is a top-level
+  ///               actor (in this case, `parent` has the value `nullptr`).
+  /// @thread-safe
+  virtual void add_actor(const local_actor& self, const local_actor* parent)
+    = 0;
+
+  /// Called whenever the actor system spawns a new actor.
```
joakim-brannstrom: "spawns a new actor"? copy/paste error from the add_actor description? :)

Neverlord: Oops, definitely a copy & paste error. Thanks! 🙂

libcaf_core/test/actor_profiler.cpp
```diff
@@ -0,0 +1,163 @@
...
+        CAF_CHECK_EQUAL(str, "hello bar");
+        return "hello foo";
+      },
+    };
+  };
+  auto foo = [bar](stateful_actor<foo_state>* self) -> behavior {
+    auto b = self->spawn(bar);
+    self->send(b, "hello bar");
+    return {
+      [](const std::string& str) mutable { CAF_CHECK_EQUAL(str, "hello foo"); },
```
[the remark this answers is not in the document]
Neverlord: To test your vigilance. Good job, you've passed the test! 😆

Neverlord:
> Are we assuming those work as well because the general actor message processing works?

Good question. The unit test definitely doesn't trigger all code paths. Triggering a reasonable number of these paths is going to be tricky, though. We need `request().then()` and well as `request().await()` with and without error / timeout as well as half a dozen corner cases inside `consume` with and without streaming... and then we have only covered `scheduled_actor`.

I think we should go ahead and start play with the feature while thinking about how to either (1) increase code coverage on this or to (2) restructure the code. Probably option (2) is better. In theory, all we need is a decorator for the current function so that the profiler gets the invocation result in a single place.

josephnoir:
I approve.

PR #949, opened 2019-10-25
OpenBSD compilation support

douglas-carmichael:
I was able to get the CAF framework to compile in OpenBSD with the attached patches.
[zeek-3.0.0-caf-openbsd.txt](https://github.com/actor-framework/actor-framework/files/3772949/zeek-3.0.0-caf-openbsd.txt)


Neverlord:
Actual PR is #950.

PR #950, opened 2019-10-25
OpenBSD compilation support (real fixes.)

douglas-carmichael:
These are the fixes from my fork of the actor-framework repository to compile on OpenBSD.
(Tested on OpenBSD 6.6.)

### DIFF
--- a/libcaf_io/src/io/network/native_socket.cpp
+++ b/libcaf_io/src/io/network/native_socket.cpp
@@ -96,7 +96,12 @@ const int ec_interrupted_syscall = EINTR;
 // platform-dependent SIGPIPE setup
 #if defined(CAF_MACOS) || defined(CAF_IOS) || defined(CAF_BSD)
 // Use the socket option but no flags to recv/send on macOS/iOS/BSD.
+// (OpenBSD doesn't have SO_NOSIGPIPE)
+#ifndef __OpenBSD__
 const int no_sigpipe_socket_flag = SO_NOSIGPIPE;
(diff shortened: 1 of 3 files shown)


libcaf_io/src/io/network/native_socket.cpp
```diff
@@ -96,7 +96,12 @@ const int ec_interrupted_syscall = EINTR;
 // platform-dependent SIGPIPE setup
 #if defined(CAF_MACOS) || defined(CAF_IOS) || defined(CAF_BSD)
 // Use the socket option but no flags to recv/send on macOS/iOS/BSD.
+// (OpenBSD doesn't have SO_NOSIGPIPE)
```
Neverlord: This patch seems incorrect to me. It may get past the compiler, but I question whether it achieves the desired behavior of blocking SIGPIPE. Looking into the man pages, it looks like OpenBSD provides the same API as Linux (i.e., `MSG_NOSIGNAL`): `https://man.openbsd.org/sendmsg.2`.

I think the correct patch is to replace `defined(CAF_BSD)` with `defined(__FreeBSD__)` in line 97 and then mention OpenBSD along Linux/Android in the `#else` branch.

I have no experience on OpenBSD, though. Do I miss something related to `MSG_NOSIGNAL` on OpenBSD?

Neverlord:
@douglas-carmichael thanks again for you contribution! I merged this PR into `topic/openbsd` and opened #955 with additional changes.

PR #954, opened 2019-10-30
Avoid repeated calls to stream finalizers

### LINKED ISSUE #937: Finalize handlers called twice on streams
A stream manager that receives an error seems to call the finalize handler twice: once with the actual error and then again with default-constructed error.

Neverlord:
Fixes #937.

### DIFF
--- a/libcaf_core/caf/stream_manager.hpp
+++ b/libcaf_core/caf/stream_manager.hpp
@@ -47,12 +47,12 @@ public:
   /// outbound paths exist.
   static constexpr int is_continuous_flag = 0x0001;
 
-  /// Denotes whether the stream is about to stop, only sending already
-  /// buffered elements.
+  /// Denotes whether the stream is about to stop, only sending buffered
+  /// elements.
   static constexpr int is_shutting_down_flag = 0x0002;
 
-  /// Denotes whether the manager has already stopped. Calling member functions
-  /// such as stop() or abort() on it no longer has any effect.
+  /// Denotes whether the manager has stopped. Calling member functions such as
+  /// stop() or abort() on it no longer has any effect.
   static constexpr int is_stopped_flag = 0x0004;
 
   // -- member types -----------------------------------------------------------
@@ -147,14 +147,9 @@ public:
 
   // -- properties -------------------------------------------------------------
 
-  /// Returns whether this stream is shutting down.
-  bool shutting_down() const noexcept {
-    return getf(is_shutting_down_flag);
-  }
-
-  /// Returns whether this manager has already stopped.
-  bool stopped() const noexcept {
-    return getf(is_stopped_flag);
+  /// Returns whether this stream is neither shutting down nor has stopped.
+  bool running() const noexcept {
+    return getf(is_shutting_down_flag | is_stopped_flag) == 0;
   }
 
   /// Returns whether this stream remains open even if no in- or outbound paths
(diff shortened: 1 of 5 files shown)


Neverlord:
@josephnoir sorry for the noise in this PR, I accidentally pushed commits into the branch for other stuff I'm working on. This is done and ready for review as of d52371c. :)

libcaf_core/src/stream_manager.cpp
```diff
@@ -116,15 +113,21 @@ void stream_manager::handle(stream_slots slots, upstream_msg::ack_batch& x) {
...
+  CAF_LOG_TRACE(CAF_ARG(slots) << CAF_ARG(x));
   if (out().remove_path(slots.receiver, x.reason, true))
     stop(std::move(x.reason));
 }
 
 void stream_manager::stop(error reason) {
+  CAF_LOG_TRACE(CAF_ARG(reason));
+  if (stopped())
+    return;
+  flags_ = is_stopped_flag;
```
josephnoir: `flags_` seems to store bit flags. I'm assuming overwriting them is intended here?

libcaf_core/src/stream_manager.cpp
```diff
@@ -135,12 +138,12 @@ void stream_manager::stop(error reason) {
 
 void stream_manager::shutdown() {
   CAF_LOG_TRACE("");
-  // Mark as shutting down and reset other flags.
-  if (shutting_down())
```
josephnoir: The header has a call to [`shutting_down()`](https://github.com/actor-framework/actor-framework/blob/d52371cff540c11d968b960a7643633371371902/libcaf_core/caf/stream_manager.hpp#L170) as well. I was wondering if that should test for `is_stopped_flag`. And if so, maybe `shutting_down()` should just check both flags?

libcaf_core/caf/stream_manager.hpp
```diff
@@ -51,6 +51,10 @@ class stream_manager : public ref_counted {
   /// buffered elements.
   static constexpr int is_shutting_down_flag = 0x0002;
 
+  /// Denotes whether the manager has already stopped. Calling member functions
```
josephnoir:
```suggestion
  /// Denotes whether the manager has stopped. Calling member functions
```

libcaf_core/caf/stream_manager.hpp
```diff
@@ -148,16 +152,21 @@ class stream_manager : public ref_counted {
     return getf(is_shutting_down_flag);
   }
 
+  /// Returns whether this manager has already stopped.
```
josephnoir:
```suggestion
  /// Returns whether this manager has stopped.
```

libcaf_test/caf/test/dsl.hpp
```diff
@@ -632,12 +632,12 @@ class test_coordinator_fixture {
   // -- constructors, destructors, and assignment operators --------------------
 
   static Config& init_config(Config& cfg) {
+    cfg.set("logger.file-verbosity", caf::atom("quiet"));
```
josephnoir: Fixing #959 as a side effect here?

libcaf_core/src/stream_manager.cpp
Neverlord: Yes, it is. Setting stop clears all other flags.

libcaf_test/caf/test/dsl.hpp
Neverlord: Apparently yes, I needed a way to look at the log output while working on this.

libcaf_core/src/stream_manager.cpp
Neverlord: Good catch. Will fix.

PR #961, opened 2019-11-03
Make messages visible to the actor profiler

Neverlord:
Tapping into actor messages is crucial for understanding system behavior and performance. With this addition, an `actor_profiler` sees not only incoming messages but also outgoing messages.

### DIFF
--- a/libcaf_core/src/response_promise.cpp
+++ b/libcaf_core/src/response_promise.cpp
@@ -16,13 +16,15 @@
  * http://www.boost.org/LICENSE_1_0.txt.                                      *
  ******************************************************************************/
 
-#include <utility>
 #include <algorithm>
+#include <utility>
 
 #include "caf/response_promise.hpp"
 
-#include "caf/logger.hpp"
+#include "caf/detail/profiled_send.hpp"
 #include "caf/local_actor.hpp"
+#include "caf/logger.hpp"
+#include "caf/no_stages.hpp"
 
 namespace caf {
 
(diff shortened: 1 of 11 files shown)


libcaf_core/src/response_promise.cpp
```diff
@@ -67,32 +68,60 @@ bool response_promise::async() const {
...
-                  context());
+    detail::profiled_send(self, std::move(source_), next, id_,
+                          std::move(stages_), self->context(), std::move(msg));
+    self_.reset();
     return;
   }
   if (source_) {
-    source_->enqueue(std::move(self_), id_.response_id(),
-                     std::move(msg), context());
+    detail::profiled_send(self, self_, source_, id_.response_id(), no_stages,
```
josephnoir: I think having `self` and `self_` here is a bit confusing. What differentiates them? (In `profiled_send.hpp` the second parameter is `sender`.)

libcaf_core/src/response_promise.cpp
```diff
@@ -67,32 +68,60 @@ bool response_promise::async() const {
...
+                          self->context(), std::move(msg));
```
Neverlord: > I think having `self` and `self_` here is a bit confusing. What differentiates them? (In `profiled_send.hpp` the second parameter is `sender`.)

`self_` has the actual actor type, `self` is of type `actor_control_block*`. They point to the same thing, but with different types. Ideas for a better name then `self`? `ctrl`?


(@josephnoir sorry, I accidentally deleted your original comment)

josephnoir: Wait, isn't it the other way around? `self` is a `local_actor` as returned by `self_ptr()` and the `actor_control_block*` is the member `self_`, a `strong_actor_ptr`. Doesn't help with the naming as it contradicts renaming `self` to `ctrl`.

Neverlord: You're right, I mixed up the two. :)

josephnoir: Would the member `self_` be the sender of the message? If so we could rename it to `sender_` (or `origin_` because it might not be the next sender in the presence of stages)?

Neverlord: I think we should stick to `self_` for the member. We use this consistently and it makes it very clear that this is not *some* sender but actually *this* actor, i.e., the owner of the promise. I instead renamed the local variable to `dptr`, following our naming convention for downcasted pointers. I hope we can agree on that. 🙂

josephnoir: Sounds good.

PR #963, opened 2019-11-06
Raise error when opening user-defined config fails

### LINKED ISSUE #947: Config file could be silently ignored, application can run with unintended parameters
Attachment [demo_files.tar.gz](https://github.com/actor-framework/actor-framework/files/3769372/demo_files.tar.gz) contains a minimal CAF application which uses CAF built-in feature for parsing and getting a user-defined parameter from the config file.

It works as expected when the config file exists and is accessible via absolute pathname or a simple relative pathname:
```
$ ./myapp --config-file=/home/user/my_app.ini
DebugLevel = 4
$ ./myapp --config-file=my_app.ini
DebugLevel = 4
$ ./myapp --config-file=./my_app.ini
DebugLevel = 4
$ ./myapp --config-file=../user/my_app.ini
DebugLevel = 4
```
However, it silently ignores a non-existing config file or one with '~':
```
$ ./myapp --config-file=zzzzzz
DebugLevel = 65535
$ ./myapp --config-file=~/my_app.ini
DebugLevel = 65535
```
I expected that the application stops with error message about a problem with config file.

Neverlord:
Closes #947.

### DIFF
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -342,9 +342,14 @@ void print(const config_value::dictionary& xs, indentation indent) {
 
 error actor_system_config::parse(string_list args, std::istream& ini) {
   // Content of the INI file overrides hard-coded defaults.
-  if (ini.good())
+  if (ini.good()) {
     if (auto err = parse_config(ini, custom_options_, content))
       return err;
+  } else {
+    // Not finding an explicitly defined config file is an error.
+    if (auto fname = get_if<std::string>(&content, "config-file"))
+      return make_error(sec::cannot_open_file, *fname);
+  }
   // CLI options override the content of the INI file.
   using std::make_move_iterator;
   auto res = custom_options_.parse(content, args);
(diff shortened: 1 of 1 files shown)


libcaf_core/src/actor_system_config.cpp
```diff
@@ -516,6 +521,7 @@ error actor_system_config::extract_config_file_path(string_list& args) {
   auto evalue = ptr->parse(path);
   if (!evalue)
     return std::move(evalue.error());
+  put(content, "config-file", *evalue);
```
josephnoir: How does this relate to the addresses issue? Parse already failed here if the config file is not available.

Neverlord: `extract_config_file_path` gets called before parsing any other CLI option. The work flow is basically:
- check if there's a `--config-file=` in the CLI arguments
- if yes: parse that argument (but not any other)
- read the config file
- read the CLI arguments

By putting `config-file` into our `content` map, we have access to the information whether the user has requested to read that file or if we simply try opening the default.

josephnoir: Ok.

PR #965, opened 2019-11-08
Remove localhost connections from I/O unit tests

Neverlord:
This was motivated by #955, because the I/O unit tests fail on OpenBSD due to differences in how each OS handles localhost connections. Since this isn't the first time we're having trouble with our I/O unit tests because of this, I've ported all unit tests to the `io_dsl.hpp` header. Some of the unit tests were also quite dated, had large overlaps, or were flat out weird (like the `triggering` unit test). The changes are substantial enough that I think discussing this separately from the actual OpenBSD changes makes more sense.

I ended up consolidated a bunch of stuff such as the remote actor suite and its typed counterpart and dropped several tests that don't have clear, concise goal.


libcaf_io/test/io/remote_spawn.cpp
```diff
@@ -72,57 +73,56 @@ calculator::behavior_type typed_calculator_fun() {
...
-  config cfg{argc, argv};
-  actor_system system{cfg};
-  auto port = unbox(system.middleman().open(0));
-  std::thread child{[=] { run_client(argc, argv, port); }};
-  child.join();
-}
+struct fixture : point_to_point_fixture<test_coordinator_fixture<config>> {
+  fixture() {
+    prepare_connection(mars, earth, "mars", 8080);
+    // ssp = std::make_shared<suite_state>();
```
josephnoir: Can probably be removed.

Issue #966, opened 2019-11-09
Minor issue with caf::callback_impl::operator()


Crzyrndm:
The line [here](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/callback.hpp#L57) should be forwarding the arguments. As it is now, the callback function will not work correctly with r-value reference parameters

I have pulled out the neccessary components to demonstrate the issue [here](https://godbolt.org/z/Cs4UMu). The corrected version (using std::forward) is [here](https://godbolt.org/z/eVc4EX).

Even though the function call has no universal reference types, forward is the correct operation (the *-value'ness of the Ts type is unknown).

Neverlord:
Thank you for tracking this down and taking the time to writing down your analysis in such great detail!

Would you mind opening a PR for this? This way, your contribution remains visible in the git history. 🙂

Crzyrndm:
Done

PR #967, opened 2019-11-10
forward arguments as *-value'ness is unknown

Crzyrndm:
Resolves https://github.com/actor-framework/actor-framework/issues/966

Willing to add a few tests, but unfamiliar with the project and didn't see any obvious existing location for callback tests


Neverlord:
Thanks a lot! 🎉

PR #969, opened 2019-11-14
Revert breaking change

Neverlord:
Removing `is_blocking_requester` outright breaks existing code. It's not much of a hassle to add a workaround for now and go through a deprecation cycle.

### DIFF
--- a/libcaf_core/caf/actor_traits.hpp
+++ b/libcaf_core/caf/actor_traits.hpp
@@ -22,15 +22,13 @@
 
 #include "caf/fwd.hpp"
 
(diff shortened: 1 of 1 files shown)


libcaf_core/caf/actor_traits.hpp
```diff
@@ -22,6 +22,16 @@
 
 #include "caf/fwd.hpp"
 
+namespace caf {
+namespace mixin {
```
jakobod: It would probably be a good idea to use concatenated namespaces here already.

PR #970, opened 2019-11-15
Generate unique node ID for each actor system

Neverlord:
Currently, CAF generates the same node ID when running on the same machine and only differentiates actors systems by their process ID. When running CAF instances in a container, this process ID is most likely the same for each run. This means two containers can produce the same node ID and thus equivalent actor IDs.

### DIFF
--- a/libcaf_core/src/node_id.cpp
+++ b/libcaf_core/src/node_id.cpp
@@ -73,9 +73,9 @@ node_id node_id::default_data::local(const actor_system_config&) {
   // process ID is most likely the same.
   std::random_device rd;
   std::minstd_rand gen{rd()};
-  std::uniform_int_distribution<char> dis(33, 126);
+  std::uniform_int_distribution<> dis(33, 126);
   for (int i = 0; i < 8; ++i)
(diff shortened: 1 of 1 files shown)


libcaf_core/src/node_id.cpp
```diff
@@ -64,12 +65,24 @@ node_id node_id::default_data::local(const actor_system_config&) {
   macs.reserve(ifs.size());
   for (auto& i : ifs)
     macs.emplace_back(std::move(i.second));
-  auto hd_serial_and_mac_addr = join(macs, "") + detail::get_root_uuid();
+  auto seeded_hd_serial_and_mac_addr = join(macs, "") + detail::get_root_uuid();
```
grahamreeds: Why not use RFC 4122 to generate a UUID?

Neverlord: At this point, we might as well. Originally, producing the same `node_id` for each run was intended as feature. 😉

However, switching to UUIDs in node IDs would potentially break downstream code. CAF uses 160-bit numbers, while UUIDs are 128-bit numbers. Since `caf_io` is already on its way out, I rather not change data formats or APIs at this point.

PR #971, opened 2019-11-15
Reformat nested namespaces

jakobod:
This PR reformats previously nested namespaces to concatenated namespaces if possible.


Neverlord:
Thanks for taking care of this. 👍

PR #974, opened 2019-11-18
Multiple typos fix

andrismednis:

PR #977, opened 2019-11-20
Fix pointed out typos

jakobod:
This PR fixes two typos that @andrismednis pointed out in #974 .

### DIFF
--- a/libcaf_core/caf/actor_storage.hpp
+++ b/libcaf_core/caf/actor_storage.hpp
@@ -132,7 +132,7 @@ void intrusive_ptr_release(actor_storage<T>* x) {
   // and destroy the data block
   if (x->ctrl.strong_refs.fetch_sub(1, std::memory_order_acq_rel) == 1) {
     x->destroy_data();
-    intrusive_ptr_relase_weak(x);
+    intrusive_ptr_release_weak(x);
   }
 }
 
--- a/libcaf_core/caf/actor_storage.hpp
+++ b/libcaf_core/caf/actor_storage.hpp
@@ -92,50 +92,6 @@ private:
   }
 };
 
-/// @relates actor_storage
-template <class T>
-bool intrusive_ptr_upgrade_weak(actor_storage<T>* x) {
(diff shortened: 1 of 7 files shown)


andrismednis:
Here is another case with "relase":
In libcaf_core/caf/actor_storage.hpp there is a template
```
template <class T>
void intrusive_ptr_release(actor_storage<T>* x) {
  // release implicit weak pointer if the last strong ref expires
  // and destroy the data block
  if (x->ctrl.strong_refs.fetch_sub(1, std::memory_order_acq_rel) == 1) {
    x->destroy_data();
    intrusive_ptr_relase_weak(x);    <---- relase !?
  }
}
```
In 6 other places it is 'intrusive_ptr_release_weak'.

libcaf_core/caf/actor_storage.hpp
```diff
@@ -132,7 +132,7 @@ void intrusive_ptr_release(actor_storage<T>* x) {
   // and destroy the data block
   if (x->ctrl.strong_refs.fetch_sub(1, std::memory_order_acq_rel) == 1) {
     x->destroy_data();
-    intrusive_ptr_relase_weak(x);
+    intrusive_ptr_release_weak(x);
```
Neverlord: This is the only place where we've misspelled `intrusive_ptr_release_weak`, which means this should've resulted in a compiler error. But it didn't. Dead code?

Seems like it when looking at the [coverage report](https://jenkins.inet.haw-hamburg.de/job/CAF/job/actor-framework/job/master/378/cobertura/caf_test/libcaf_core_caf_actor_storage_hpp/). Can you check whether we can just remove all the `intrusive_ptr_` functions from `actor_storage.hpp`? As far as I can tell, we never form an `intrusive_ptr` to an actor storage. Only to an actor control block.

jakobod: I will check that tomorrow.

jakobod: Done.

Issue #978, opened 2019-11-21
Remote address lookup does not work for datagram handles


josephnoir:
Calling `abstract_broker::remote_addr(datagram_handle)` does not work as expected. Instead of looking up the address in the known endpoints of the underlying datagram handler, it tries to retrieve the address from the socket. Since the datagram sockets in use are not connected this cannot work.

PR #979, opened 2019-11-21
Fix remote address lookup for datagram handles

josephnoir:
Addresses #978. Still needs testing.


stashlukj:
This works for me. I am able to reply to the sender via unconnected UDP.

Neverlord:
@stashlukj thank you for double-checking the patch! 👍

PR #980, opened 2019-11-22
Add thead to core affinity

lucarin91:
This pull request adds the ability to use the thread to core affinity in CAF. 
Pinning threads to specific cores can, in general, improve performances and can also permit the isolation of different threads in order to avoid performance degradation (e.g., worker threads and blocking actors)

The patch adds the *Affinity Manager* which is called by the *Actor System* every time a new thread is created and manages the affinity setting for each of them.
There is the possibility to set the affinity for the different types of threads with the following configurations:
- `affinity.worker-cores` for the Run-time threads
- `affinity.blocking-cores` for the blocking actors
- `affinity.detached-cores` for the detached actors 
- `affinity.other-cores` for other threads

Each configuration supports a string that specifies a set of groups of cores in which each thread must be pinned.
For instance, the string `"<0,3><4-7>"` specifies two groups of cores, the first is composed by the cores 0, 3 and the second by the cores 4, 5, 6, 7. Thus the first thread spawned will be restricted to the threads 0 and 3 and the second thread to the cores 4, 5, 6, and 7. Then the third thread will be pinned on the first group again and so on.

The patch supports both Linux and Windows, macOS, instead, does not support thread to core affinity.
Right now I only test it with g++ on Linux and with the Microsoft C++ compiler on Windows 10.

I am available for any clarification.


Neverlord:
Thank you for submitting this! Adding an API for core affinities looks like a good addition.

> Pinning threads to specific cores can, in general, improve performances

Did you measure performance gains for CAF applications? What are the metrics one should look for when tuning CAF applications with affinities? Can you share some experience with this feature?

The code itself does need some work in my eyes. In particular:

- I don't see what value `examples/affinity_test.cpp` provides to a CAF user. It looks more like a half-baked unit test to me. As it stands, the output of this program depends on the configuration. And the example does not tell the user which configuration parameters to use and what output to expect.
- Modules in CAF are either optional background services with separate threads like a logger or extensions shipped in a separate library. The affinity manager only adds a single function, `set_affinity`, to the actor system and gets always loaded unconditionally. I think we can integrate this with much less boilerplate code and without the extra heap indirection.
- Configuring the affinity groups as strings and then have a special parser for them seems very odd to me. Our parsers already support nested lists. Without any extra code, we can configure `[[0,3], [4,5,6,7]]` and read this as `vector<set<int>>` via `get_if`. If we want to support the `4-7` notation, then I think adding this one feature to the CAF parsers is the better option.
- The code still seems rough in places. For example, there are commented-out `printf` statements and TODOs. The code claims that macOS ignores thread affinities. Looking a bit around, it seems like there's an API available: https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/. Looks like this API does not force a thread-to-processor mapping, but it still seems to give users *some* control over thread placement.
- Regarding the configuration parameters, I think we should consider alternative names. A `affinity.blocking-cores` parameter would not give me a good intuition as a user. A "blocking core" requires some explanation. In any case, the parameter is about *blocking actors*. Maybe simply `affinity.scheduled-actors`, `affinity.blocking-actors`, `affinity.detached-actors`, `affinity.non-actors`?

lucarin91:
Thank for considering the proposed feature.
Sorry if the code does not respect some development principles of the CAF framework, I will try to update it and in order to better feat with it.

I will respond with multiple comments to have a more clear discussion.

lucarin91:
> Did you measure performance gains for CAF applications? What are the metrics one should look for when tuning CAF applications with affinities? Can you share some experience with this feature?

The affinity feature was used in the paper [Accelerating  Actor-based Applications with Parallel  Patterns](http://pages.di.unipi.it/mencagli/publications/preprint-pdp-2019.pdf). In the paper you can find some performance analysis.

lucarin91:
> I don't see what value examples/affinity_test.cpp provides to a CAF user. It looks more like a half-baked unit test to me. As it stands, the output of this program depends on the configuration. And the example does not tell the user which configuration parameters to use and what output to expect.

Yes, I agree with you. The example was created to verify that the affinity works as expected. Maybe it is possible to add the configuration directly in the code.

lucarin91:
> Modules in CAF are either optional background services with separate threads like a logger or extensions shipped in a separate library. The affinity manager only adds a single function, set_affinity, to the actor system and gets always loaded unconditionally. I think we can integrate this with much less boilerplate code and without the extra heap indirection.

The problem is that the `affinity_manager` has four atomics one for each thread type to remember in which group the next thread has to be pinned.
I think that in some ways we need to maintain a "singleton" pattern to achieve this.
Another possibility is to remove the affinity_manager and put all the affinity logic directly in the `actor_system`, maybe in the function `thread_started`.

lucarin91:
> Configuring the affinity groups as strings and then have a special parser for them seems very odd to me. Our parsers already support nested lists. Without any extra code, we can configure [[0,3], [4,5,6,7]] and read this as vector<set<int>> via get_if. If we want to support the 4-7 notation, then I think adding this one feature to the CAF parsers is the better option.

Yes, this will integrate more the affinity feature in the framework. I can work on this.

lucarin91:
> The code still seems rough in places. For example, there are commented-out printf statements and TODOs. The code claims that macOS ignores thread affinities. Looking a bit around, it seems like there's an API available: https://developer.apple.com/library/archive/releasenotes/Performance/RN-AffinityAPI/. Looks like this API does not force a thread-to-processor mapping, but it still seems to give users some control over thread placement.

Yes, sorry I will clean up the code.
For what concerning macOS, I was honestly thinking that it was not supported at all. 
Maybe it is possible to add support in the future, but unfortunately, I do not have a mac to implement it and test it.

lucarin91:
> Regarding the configuration parameters, I think we should consider alternative names. A affinity.blocking-cores parameter would not give me a good intuition as a user. A "blocking core" requires some explanation. In any case, the parameter is about blocking actors. Maybe simply affinity.scheduled-actors, affinity.blocking-actors, affinity.detached-actors, affinity.non-actors

Yes, this is a better naming, I will update it.

lucarin91:
I refactoring the affinity feature following your suggestion. In particular:

- remove the _Affinity Manager_ and move the affinity logic directly inside the _Actor System_.
- use the CAF internal parser for the affinity configuration and use a `vector<vector<int>>` instead of a custom string.
- implement a macOS version that I have tested only on a virtual machine.
- implement a FreeBSD version that I do not have tested yet.
- update the `affinity_test` example.

I am here for further problems or suggestions :)

libcaf_core/caf/actor_system.hpp
```diff
@@ -524,7 +534,9 @@ class actor_system {
 
   /// Calls all thread started hooks
   /// @warning must be called by thread which is about to start
-  void thread_started();
+  void thread_started(const thread_type tt=other_thread);
+
+  void set_affinity(const thread_type tt);
```
Neverlord: This entire PR essentially boils down to adding this function. Could you please add documentation to it? From browsing through the code, I think this does the following:
- Calling this function affects the current thread only.
 -It picks the next core for that affinity group in round-robin order. This means calling `set_affinity` N+1 times will bind the the thread to the first core again, given group size N. Thus, two threads can share one core but only within the affinity group.
- This does nothing if not configuring the corresponding type-to-cores parameters.

That's my understanding after reading the code, but a user should receive this information upfront. 🙂

libcaf_core/src/detail/affinity.cpp
```diff
@@ -0,0 +1,95 @@
...
+#ifdef CAF_LINUX
+#  ifndef _GNU_SOURCE
+#    define _GNU_SOURCE
+#  endif // _GNU_SOURCE
+#  include <sched.h>
+#endif // CAF_LINUX
+
+#ifdef CAF_WINDOWS
+#  ifndef WIN32_LEAN_AND_MEAN
+#    define _GNU_SOURCE
```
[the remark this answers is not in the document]
lucarin91: Yes, sorry. It was a copy-paste error. I will change it in the next commit.

libcaf_core/caf/actor_system.hpp
lucarin91: Yes, you are correct the code does exactly that. I will add some documentation on the code in the next commit.

Neverlord:
Just a quick heads-up: I didn't forget this PR. I'm prioritizing #1006 right now and once that's done I'll integrating this PR for 0.18.

Neverlord:
@lucarin91 I appreciate your efforts to stay on par with `master`, but #1057 is going to introduce a bunch more conflicts anyways. Can you please reset this PR to 7e94303? I'm going to take care of all merge conflicts when integrating your changes. 🙂

Thanks! 🙏🏻

Neverlord:
Thanks! Sorry that this is open for such a long time. This PR had an unfortunate timing. Not your fault of course. 🙂

mavam:
We also have a new use case in [VAST](https://github.com/tenzir/vast) for CPU pinning of detached actors that read packets from the NIC. We're looking forward to see progress on this issue.

(cc @dominiklohmann)

Neverlord:
This is going to be an 0.18 feature, so it's not going to be available in an official release until then. If you have immediate need for this, maybe you can use the thread hooks in the meantime to do the pinning yourself.

dominiklohmann:
It's not too urgent, we can wait for 0.18. And if we need it quicker than that, we can always use the patch from this branch on 0.17.4 with slight modifications.

norg:
Hi,

is there any chance that we might seen this being integrated? On Linux systems with other applications it is very helpful to set affinity directly. For example a setup with dual socket CPUs we would like to pin the CAF part to one CPU/NUMA node with the SSDs attached while something else (suricata for example) is pinned to the second NUMA Node where the NICs are.
This would ensure that there is no unnecessary cross talk and also no sharing of ressources that should be isolated.

Neverlord:
> is there any chance that we might seen this being integrated?

Good question. After reviewing the code and checking the documentation on the various OS APIs in use, I'm not so sure at the moment what's the best way moving forward. As a CAF core feature, it would need to be consistent across all platforms. However, on macOS you have to pick a single core per thread. With this PR, CAF would put all threads (silently) on the first CPU in a group on macOS. Apple also introduced affinity tags with 10.5, which may be an alternative but they still seem to work differently. Looks like Linux/Windows/BSD (when opting into the non-portable GNU extension) have a similar notion of affinity masks, but I didn't check for iOS/Android yet (seems like CAF gets used on those as well).

So one option could be to only pick the improvement to the thread hooks here. If users know they only ever deploy on Linux, they can then install a hook to call `sched_setaffinity` themselves and we can provide a recipe in some article or manual section. We could also make it a platform-specific configuration option that only works on some of the platforms. But then users at least should get an actual error when trying to configure this on macOS (and other platforms where we can't implement it properly) instead of silently getting unexpected system behavior.

> would like to pin the CAF part to one CPU/NUMA node

If we go down this rabbit hole: isn't there another problem with the current threading in general? CAF uses `std::thread`, which has no notion of a three-step create, configure, launch. We would need to set an affinity before actually launching the thread, though. The OS migrates a thread to allowed cores when setting the mask later, but they could still end up running briefly on the wrong NUMA node. When starting on the wrong NUMA node, wouldn't they also receive a stack from the NUMA node they first run on? Meaning that after pinning them at a later time, you are now guaranteed to pay for remote memory access on each stack access.

norg:
I agree with your thoughts. I didn't know about the issue with the timing when to assign the affinity. So if that makes it more complicated, that's difficult to deal with. I don't have a proper solution.

Coming from Suricata where I have this option to set exact CPU affinity I was hoping to see it in other areas as well as it helps a lot with performance optimization on high load systems as described above (SSDs on Node 0 where CAF should work, NICs on Node 1 where Suricata works).

Neverlord:
> So one option could be to only pick the improvement to the thread hooks here. If users know they only ever deploy on Linux, they can then install a hook to call `sched_setaffinity` themselves and we can provide a recipe in some article or manual section.

That's what we're going with. I really would've liked to see this as a builtin feature, but since it pretty much only makes sense on Linux as-is and pins threads only after they've been started, I'm going to pick up the thread hook improvements for 0.19 and then we'll have some article or manual entry with an example on how to call `sched_setaffinity`.

PR #982, opened 2019-11-25
Fix undefined behavior, build with UBSan in Jenkins

Neverlord:
Big thanks to @alessandrogario for bringing a couple of UBSan errors to our attention. I think none of the findings could spell any trouble, but of course fixing is the right thing to do so regardless.

UBSan found mostly unsafe downcasts. The only actual problematic piece was `before_begin` in the intrusive task list. However, this function was unused outside of unit tests. So I've just removed it.

I also tweaked our CMake scaffold a bit to make it easier to enable more sanitizers and I've enabled ASAN and UBSan by default now in Jenkins.

### DIFF
--- a/configure
+++ b/configure
@@ -79,6 +79,12 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --with-sanitizers=LIST      build with this list of sanitizers enabled
     --with-actor-profiler       enables the (experimental) actor_profiler API
 
+  Convenience options:
+    --dev-mode                  sets --build-type=debug, --no-examples,
+                                --no-tools, --with-runtime-checks,
+                                --log-level=trace, and
+                                --sanitizers=address,undefined
+
   Influential Environment Variables (only on first invocation):
     CXX                         C++ compiler command
     CXXFLAGS                    C++ compiler flags (overrides defaults)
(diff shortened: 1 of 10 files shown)


configure
```diff
@@ -76,17 +76,9 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
...
-    --enable-asan               alias for --with-address-sanitier
-    --with-gcov                 build with gcov coverage enabled
+    --with-sanitizers=LIST      build with this list of sanitizers enabled
     --with-actor-profiler       enables the (experimental) actor_profiler API
 
-  Convenience options:
-    --dev-mode                  sets --build-type=debug, --no-examples,
-                                --no-tools, --with-runtime-checks,
-                                --log-level=trace, and --enable-asan
-
```
josephnoir: Why remove this one? I use it a lot!

configure
```diff
@@ -76,17 +76,9 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
                                   - INFO
                                   - DEBUG
                                   - TRACE
-    --with-address-sanitizer    build with address sanitizer if available
-    --with-asan                 alias for --with-address-sanitier
-    --enable-asan               alias for --with-address-sanitier
-    --with-gcov                 build with gcov coverage enabled
+    --with-sanitizers=LIST      build with this list of sanitizers enabled
```
josephnoir: I think it would be nice  to have a convenience options such as `--enable-sanitizers` that simply sets whatever we usually use or alternatively list the possible options for `LIST` in the description.

Neverlord: I'm not a fan of introducing too many ways for the same thing. I'll give you `--dev-mode` back, ok? 😇

PR #984, opened 2019-11-26
Fix conversion constructor for caf::span

Neverlord:
@jakobod can you double-check whether this fixes the issues you're having in the incubator?


jakobod:
I have just checked and the problem still persists..

jakobod:
Nevermind. This actually fixed the problem. I just checked the wrong call... :no_mouth:

PR #987, opened 2019-11-30
Resurrect serialized_size function

Neverlord:


libcaf_core/src/detail/serialized_size.cpp
```diff
@@ -0,0 +1,153 @@
...
+  return end_sequence();
+}
+
+error serialized_size_inspector::apply(const std::u32string& x) {
+  begin_sequence(x.size());
+  result_ += x.size() * sizeof(uint32_t);
+  return end_sequence();
+}
+
+error serialized_size_inspector::apply(span<const byte> x) {
```
josephnoir: The implementations for other types of variable length start with `begin_sequence(x.size())` but this one doesn't. Is that intentional? (And `span<const byte>` does not appear in the test cases.)

Neverlord: > Is that intentional?

Yes, because this isn't variable length. It's a BLOB for all CAF cares about.

> And `span<const byte>` does not appear in the test cases.

Everything else is implemented using byte spans (see `apply_int`).

josephnoir: Oh. Well. Ack.

Issue #951, opened 2019-10-27
Minimal possible example blocks forever in caf::actor_system destructor


ChristofferGreen:
The smallest possible code example I can do:

`#include <caf/all.hpp>

int main(int argc, char *argv[]) {
  caf::actor_system_config cfg;
  caf::actor_system system(cfg);

  return 0;
}
`

Blocks forever, never exists, in the destructor of system. I am using (mingw/msys2) GCC 9.2.0  on windows 10. Using CAF 0.17.2.

ChristofferGreen:
From printf debugging it appears to be these lines at the end of the destructor that hangs forever:

"  while (!logger_dtor_done_)
    logger_dtor_cv_.wait(guard);
"

ChristofferGreen:
It looks like the logger constructor gets called but the destructor (via reset on it's pointer) does not ever get called.

ChristofferGreen:
Here is the ref_counted::deref() function in ref_counted.cpp

"
void ref_counted::deref() const noexcept {
  if (unique()) {
    request_deletion(false);
    return;
  }
  if (rc_.fetch_sub(1, std::memory_order_acq_rel) == 1) {
    request_deletion(true);
  }
}
"

Neither if statement is entered. I don't know the code very well but I suspect this causes the logger to not be deleted.

ChristofferGreen:
Putting: 
    "std::cout << "handle count: " << logger_.get()->get_reference_count() << std::endl;" 
Before the "logger_.reset();" line in actor_system::~actor_system() results in: "handle count: 27"

Figuring out what is going on here is a bit above my head.

Neverlord:
Thank you for narrowing this down already! We'll try to reproduce this next week and see who's holding onto the references. :)

Neverlord:
@ChristofferGreen I had no luck reproducing this on my end so far.

> I am using (mingw/msys2) GCC 9.2.0 on windows 10. Using CAF 0.17.2.

Please note that we do *not* support MinGW officially. We're open to patches from the community for MinGW, but we don't maintain or test the MinGW code paths. Maintainers for that are welcome, btw. 😉

However, we *do* support Visual Studio for regular Windows builds. To make sure we're not dealing with some MingGW issue here: can you reproduce this behavior when using MSVC or when running on Linux?

ChristofferGreen:
[...]

PR #988, opened 2019-11-30
Do not use thread_local for MinGW

ch1973:
MinGW's support for thread_local seems to be broken. When used, many unit tests stay in actor_system destructor forever. See also #951 . Without thread_local everything works fine.
This pull request fixes it by disabling thread_local for MinGW.


Neverlord:
Thanks for the patch! 🎉

PR #999, opened 2019-12-19
Enable compact list notation in config files

Neverlord:
Adds support for `from..to..step` syntax (borrowed form `bash` & friends) in config files (steps are optional). Particularly useful for defining affinity groups (see https://github.com/actor-framework/actor-framework/pull/980).

Two motivating examples:
- `[1..3]` expands to `[1, 2, 3]`
- `[4..-4..-2]` expands to `[4, 2, 0,- 2, -4]`

### DIFF
--- a/libcaf_core/caf/detail/parser/read_ini.hpp
+++ b/libcaf_core/caf/detail/parser/read_ini.hpp
@@ -71,8 +71,9 @@ void read_ini_comment(State& ps, Consumer&&) {
   // clang-format on
 }
 
-template <class State, class Consumer>
-void read_ini_value(State& ps, Consumer&& consumer);
+template <class State, class Consumer, class InsideList = std::false_type>
+void read_ini_value(State& ps, Consumer&& consumer,
+                    InsideList inside_list = {});
 
 template <class State, class Consumer>
 void read_ini_list(State& ps, Consumer&& consumer) {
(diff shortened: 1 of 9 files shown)


libcaf_core/test/detail/parser/read_number.cpp
```diff
@@ -269,4 +301,44 @@ CAF_TEST(fractional mantissa with negative exponent) {
...
+CAF_TEST(ranges can use negative step values) {
+  CHECK_RANGE("6..2..-2", 6, 4, 2);
+  CHECK_RANGE("8..3..-3", 8, 5);
+}
+
+CAF_TEST(ranges can use signed integers) {
+  CHECK_RANGE("+2..+6..+2", 2, 4, 6);
+  CHECK_RANGE("+6..+2..-2", 6, 4, 2);
+  CHECK_RANGE("+2..-2..-2", 2, 0, -2);
+  CHECK_RANGE("-2..+2..+2", -2, 0, 2);
```
josephnoir: What happens for a range like this: `"-2..+2..-2"`?

libcaf_core/caf/detail/parser/read_ini.hpp
```diff
@@ -183,16 +184,17 @@ void read_ini_uri(State& ps, Consumer&& consumer) {
   // clang-format on
 }
 
-template <class State, class Consumer>
-void read_ini_value(State& ps, Consumer&& consumer) {
+template <class State, class Consumer, class InsideList>
+void read_ini_value(State& ps, Consumer&& consumer, InsideList inside_list) {
```
josephnoir: The type `InsideList` is passed to `read_number_or_timespan` as the type for `EnableRange`, does the tag have a different meaning in both cases?

Neverlord: That' correct. We enable the range syntax when inside a list. 🙂

libcaf_core/test/detail/parser/read_number.cpp
Neverlord: You'll get an error. I've added a new test case for that.

PR #1000, opened 2019-12-20
Fix issues with XCode CMake Generator

jokalode:
The CMake Xcode generator does not like targets that consist solely of `TARGET_OBJECTS` from an object library.

This pull request introduces a utility function `xcode_assert_target_not_empty` to create a simple default source file when using the Xcode generator with "objects only" targets.

Also, ignores deprecated build variables `EXECUTABLE_OUTPUT_PATH` and `LIBRARY_OUTPUT_PATH` when using the Xcode generator.


Neverlord:
Thanks for the fix! 🙏🏻

PR #1005, opened 2019-12-27
netinet/ip.h needs sys/types.h on FreeBSD gcc architectures

leres:
Changes from https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=242908


leres:
(I should have mentioned that Piotr Kubaj submitted the PR and provided the patches.)

Neverlord:
Thanks for the fix! 🙏🏻

PR #1023, opened 2020-01-20
Report errors while connecting two CAF nodes

Neverlord:
Connecting two CAF nodes can fail for several reasons:
- unexpected TCP disconnects
- an application ID mismatch
- a CAF version mismatch

Currently, CAF reports all of the above errors as `disconnect_during_handshake`. Masking the true error makes it impossible for users to troubleshoot CAF applications or to respond to errors programmatically. For example, an application may try to reconnect on actual TCP disconnects but there's no point in disconnecting to an incompatible node.

For original context, see https://github.com/zeek/broker/issues/80.

(Note: the branch `topic/basp-error-codes` contains the same fix but branched off from 0.17.3 to be included in CAF 0.17.4.)

### DIFF
--- a/libcaf_core/caf/sec.hpp
+++ b/libcaf_core/caf/sec.hpp
@@ -126,6 +126,17 @@ enum class sec : uint8_t {
   socket_operation_failed = 45,
   /// A resource is temporarily unavailable or would block.
   unavailable_or_would_block,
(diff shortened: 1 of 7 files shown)


Neverlord:
Agree, I wouldn't mind cleaning those up as well.

PR #1033, opened 2020-01-24
Remove the global `to_string` overload

Neverlord:
The global `to_string` overload may have seemed like a good idea at the
time (4 years ago), but it interacts very badly with types such as
`variant` that have implicit conversion operators. The global
`to_string` also has a very weird interaction with the
`stringification_inspector`: we enable the `to_string` if a type is
inspectable but we use this as a fallback anyway. However, a custom
`to_string` that gets picked up via ADL can become ambiguous due to the
`caf::to_string` for no good reason.

@josephnoir this patch applies to the 0.17.4 release branch. I'll also do a similar patch to `master` as a followup.


Neverlord:
After more consideration: It's ultimately still an API change, so let's drop this with 0.18 but not in a bugfix release such as 0.17.4.

PR #1034, opened 2020-01-30
Add typed view types for messages

Neverlord:
The typed message view types are a missing puzzle piece for reimplementing the matching based on `type_id_list`.


Neverlord:
> Since these are public headers we might want to add some documentation?

Yes, I'll add some more documentation in a followup.

> And the license in the header stats 2019 not 20. Not sure if you want to address that.

The dates are all over the place TBH. I'm thinking of minimizing the header boilerplate altogether to only a link to the original repo plus license and add a `COPYING` file to the repo instead.

PR #1035, opened 2020-01-31
profiled_coordinator.hpp: check for RUSAGE_THREAD

ffontaine:
RUSAGE_THREAD is not defined on uclibc, so use RUSAGE_SELF if
RUSAGE_THREAD is undefined

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


Neverlord:
Thanks for the fix! 🙏🏻

PR #1045, opened 2020-02-07
Remove extra semicolon to quiet -Wpedantic

stashlukj:


Neverlord:
Oh, these pesky semicolons... Thanks for the fix! 🙂

PR #1052, opened 2020-02-17
 Convert manual to reStructuredText

Neverlord:
Our HTML version of the manual is more or less broken ever since our `pandoc`-based conversion broke a while back. Since our HTML version of the manual is definitely more important than a polished PDF version, it's time to make the jump and switch over to reStructuredText. In my experience, working with `.rst` files is also easier / more productive than using LaTeX.


josephnoir:
I had trouble building this on macOS. What worked for me was creating a venv and installing the necessary tools inside. Think that was `sphinx`, `sphinx_rtd_theme`, and `gitpython`. I still had to reactivate the venv. Don't ask me why ...

The output looks mostly good. I did find a problem in ...
* The `Scheduler` section. Near the end of the first paragraph.
* Same in the section on `ConfiguringActorApplications.html`, subsection `command-line-options-and-ini-configuration-files`, third paragraph.
* `Messages` in the subsection `extracting`, are the `$` signs supposed to be there?
* `Group Communication`, subsection `Remote Groups`, last paragraph.
* Second to last paragraph in `ManagingGroupsOfWorkers`
* In `Network I/O with Brokers` after the first paragraph, the `experimental` has a visible `:sup:` in front of it.

PR #1053, opened 2020-02-17
Minor refactoring

dosuperuser:
This pull request introduces: 

- Check of emptiness using` empty() `instead of size compared to 0. According to standard `empty()` is guaranteed to have O(1) complexity.
- Removal of unnecessary` <iostream>` include and duplicate `<utility>`
- Adding const qualifier for 2 member functions *REVERTED*
- Using` clear()` instead of assignment of empty C str

I would love to get your feedback and opinions.

### DIFF
--- a/libcaf_core/caf/local_actor.hpp
+++ b/libcaf_core/caf/local_actor.hpp
@@ -340,12 +340,12 @@ public:
   /// Serializes the state of this actor to `sink`. This function is
   /// only called if this actor has set the `is_serializable` flag.
   /// The default implementation throws a `std::logic_error`.
-  virtual error save_state(serializer& sink, unsigned int version) const;
+  virtual error save_state(serializer& sink, unsigned int version);
 
   /// Deserializes the state of this actor from `source`. This function is
   /// only called if this actor has set the `is_serializable` flag.
   /// The default implementation throws a `std::logic_error`.
-  virtual error load_state(deserializer& source, unsigned int version) const;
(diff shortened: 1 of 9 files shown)


libcaf_core/caf/local_actor.hpp
```diff
@@ -340,12 +340,12 @@ class CAF_CORE_EXPORT local_actor : public monitorable_actor {
...
   /// only called if this actor has set the `is_serializable` flag.
   /// The default implementation throws a `std::logic_error`.
-  virtual error save_state(serializer& sink, unsigned int version);
+  virtual error save_state(serializer& sink, unsigned int version) const;
 
   /// Deserializes the state of this actor from `source`. This function is
   /// only called if this actor has set the `is_serializable` flag.
   /// The default implementation throws a `std::logic_error`.
-  virtual error load_state(deserializer& source, unsigned int version);
+  virtual error load_state(deserializer& source, unsigned int version) const;
```
Neverlord:
```suggestion
  virtual error load_state(deserializer& source, unsigned int version);
```

A const load function makes no sense, since it's purpose is to override the state of an actor.

dosuperuser: Thank you I oversee this as a quick refactor. Thank you I reverted back this changes.

Neverlord:
FYI, I've fixed the formatting issue during merge. Thank you for helping to improve the code quality! 🎉

PR #1059, opened 2020-03-04
Add CAF_CORE_EXPORT to tracing_data_factory

CppPhil:
This set of changes has tracing_data_factory be declared with CAF_CORE_EXPORT.
Additionally include directives for tracing_data.hpp and tracing_data_factory.hpp are added to all.hpp.

This fixes linker errors regarding tracing_data_factory when using the actor_profiler API.


Neverlord:
Thanks for the fix! 🙏🏻

PR #1067, opened 2020-03-13
Fix unexpected responses to anon_send, close #1046

### LINKED ISSUE #1046: Unexpected message after anon_send
Getting some odd behaviour, I'm using anon_send(this, get_buff_atom_v)  from my actor to it's self, the handler will respond with a caf::result<mybuffer>, which if fulfilled via a response promise.

I'm expecting this to basically trigger the message handler and throw away the reply as I'm not using request.

What appears to be happening is that the reply is sent back to my message queue, as if I'd done send(this, mybuffer) 

This then kills the actor with unexpected message.

*** unexpected message [id: 20, name: PlayheadActor]: (<unprintable>)

Am I doing something wrong, or have I found a nasty bug ?

Neverlord:


libcaf_core/caf/detail/profiled_send.hpp
```diff
@@ -46,16 +46,21 @@ void profiled_send(Self* self, Sender&& sender, const Handle& receiver,
 }
 
 template <class Self, class Sender, class Handle, class... Ts>
-void profiled_send(Self* self, Sender&& sender, const Handle& receiver,
+void profiled_send(Self* self, Sender&& sender, const Handle& dst,
```
jakobod: Consistent naming would be cool.
```suggestion
void profiled_send(Self* self, Sender&& src, const Handle& dst,
```
This fits line 35 too.

libcaf_core/test/mixin/sender.cpp
```diff
@@ -55,28 +55,50 @@ struct fixture : test_coordinator_fixture<> {
 
 CAF_TEST_FIXTURE_SCOPE(sender_tests, fixture)
 
-CAF_TEST(delayed actor message) {
+CAF_TEST(delayed actor messages receive responses) {
   self->delayed_send(testee, seconds(1), "hello world");
   sched.trigger_timeout();
   expect((std::string), from(self).to(testee).with("hello world"));
```
jakobod: Maybe use a variable in the fixture for the teststring to not write it out every time.
```suggestion
  expect((std::string), from(self).to(testee).with(test_string));
```

libcaf_core/caf/detail/profiled_send.hpp
```diff
@@ -31,31 +31,36 @@
 
 namespace caf::detail {
 
-template <class Self, class Sender, class Handle, class... Ts>
-void profiled_send(Self* self, Sender&& sender, const Handle& receiver,
+template <class Self, class SelfHandle, class Handle, class... Ts>
+void profiled_send(Self* self, SelfHandle&& src, const Handle& dst,
                    message_id msg_id, std::vector<strong_actor_ptr> stages,
                    execution_unit* context, Ts&&... xs) {
   CAF_IGNORE_UNUSED(self);
```
jakobod: Dont we use `[[maybe_unused]]` to silence unused warnings?

```suggestion
void profiled_send([[maybe_unused]] Self* self, SelfHandle&& src, const Handle& dst,
                   message_id msg_id, std::vector<strong_actor_ptr> stages,
                   execution_unit* context, Ts&&... xs) {
```

libcaf_core/caf/detail/profiled_send.hpp
```diff
@@ -31,31 +31,36 @@
...
   }
 }
 
-template <class Self, class Sender, class Handle, class... Ts>
-void profiled_send(Self* self, Sender&& sender, const Handle& receiver,
+template <class Self, class SelfHandle, class Handle, class... Ts>
+void profiled_send(Self* self, SelfHandle&& src, const Handle& dst,
                    actor_clock& clock, actor_clock::time_point timeout,
                    message_id msg_id, Ts&&... xs) {
   CAF_IGNORE_UNUSED(self);
```
jakobod: Same here

```suggestion
void profiled_send([[maybe_unused]] Self* self, SelfHandle&& src, const Handle& dst,
                   actor_clock& clock, actor_clock::time_point timeout,
                   message_id msg_id, Ts&&... xs) {
```

Neverlord:
> Dont we use `[[maybe_unused]]` compiler directives to silence unused warnings?

Yes. There are currently 40 references to `CAF_IGNORE_UNUSED`. I'll clean those up in a followup once the current PRs got merged.

PR #1071, opened 2020-03-17
Allow actors to monitor CAF nodes

### LINKED ISSUE #1042: Allow monitoring of CAF nodes
Currently, actors can only monitor other actors but not entire nodes. The use case for monitoring nodes is primarily when [remotely spawning](https://actor-framework.readthedocs.io/en/stable/RemoteSpawn.html) actors to have a way of detecting when the entire nodes fails and becomes unavailable for spawning more actors.

Neverlord:
Closes #1042.

### DIFF
--- a/libcaf_io/src/io/basp_broker.cpp
+++ b/libcaf_io/src/io/basp_broker.cpp
@@ -77,6 +77,13 @@ void basp_broker::on_exit() {
   //       that the middleman calls this in its stop() function. However,
   //       ultimately we should find a nonblocking solution here.
   instance.hub().await_workers();
+  // All nodes are offline now. We use a default-constructed error code to
+  // signal ordinary shutdown.
+  for (const auto& [node, observer_list] : node_observers)
+    for (const auto& observer : observer_list)
+      if (auto hdl = actor_cast<actor>(observer))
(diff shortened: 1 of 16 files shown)


libcaf_io/src/io/basp_broker.cpp
```diff
@@ -567,8 +625,10 @@ void basp_broker::connection_cleanup(connection_handle hdl, sec code) {
   CAF_LOG_TRACE(CAF_ARG(hdl) << CAF_ARG(code));
   // Remove handle from the routing table and clean up any node-specific state
```
jakobod:
```suggestion
  // Remove handle from the routing table, notify all observers, and clean up any node-specific state
```

Neverlord:
Thanks for the quick review! 🙏🏻

I've added your suggestions and merge when Jenkins gives green light.

> Where would such a behavior be useful?

Event-based actors don't really have a use case for multiple monitors. However, it's useful in blocking actors, where you monitor an actor (or node) inside a scope and then wait for the matching down message. Since `down_msg` behaves in this way, breaking the symmetry for nodes and actors here would IMO make for a bad API.

PR #1074, opened 2020-03-24
Add new UUID class

Neverlord:
Going forward, we should replace our custom actor system ID based on a 160bit hash + PID with standard UUIDs. Also, UUIDs are widely used as transaction or resource IDs. Giving our users access to a builtin UUID implementation is a good quality-of-life addition.


libcaf_core/test/uuid.cpp
```diff
@@ -0,0 +1,146 @@
...
+                  pec::invalid_argument);
+  CAF_CHECK_EQUAL(make_uuid("cbba369a-6ceb-F1ea-bc55-0242ac130003"),
+                  pec::invalid_argument);
+  CAF_CHECK_EQUAL(make_uuid("cbba38fc-6ceb-01ea-bc55-0242ac130003"),
+                  pec::invalid_argument);
+}
+
+#define WITH(uuid_str)                                                         \
+  x = uuid_str##_uuid;                                                         \
+  for (int dummy = 0; dummy < 1; ++dummy)
```
jakobod: Just asking myself: wouldn't something like
```
if (true)
```
be sufficient for this case? Why use a `for` loop?

PR #1080, opened 2020-04-04
Add missing cast

josephnoir:
As the issue stated, 0.17.4 has the same problem. Do we want as well and would this be the way to do it?


Neverlord:
I've already cherry-picket your patch (e83e5f84fb018716738c319400e992565b77413b).

josephnoir:
Oh, then I forgot to build the pull that branch before checking. Sorry for the noise.

Neverlord:
No worries. I much rather have a couple extra PRs than getting things overlooked. 🙂

PR #1085, opened 2020-04-08
Re-implement the message builder

Neverlord:
Implement remaining TODOs in the `message_builder`.

### DIFF
--- a/libcaf_core/caf/message_builder.hpp
+++ b/libcaf_core/caf/message_builder.hpp
@@ -62,7 +62,7 @@ public:
     using namespace detail;
     using value_type = strip_and_convert_t<T>;
     static_assert(sendable<value_type>);
-    storage_size_ += padded_size_v<strip_and_convert_t<value_type>>;
+    storage_size_ += padded_size_v<value_type>;
     types_.push_back(type_id_v<value_type>);
     elements_.emplace_back(make_message_builder_element(std::forward<T>(x)));
     return *this;
--- a/libcaf_core/caf/message_builder.hpp
+++ b/libcaf_core/caf/message_builder.hpp
@@ -23,6 +23,7 @@
 #include "caf/detail/core_export.hpp"
 #include "caf/detail/implicit_conversions.hpp"
 #include "caf/detail/message_builder_element.hpp"
+#include "caf/detail/padded_size.hpp"
 #include "caf/detail/type_id_list_builder.hpp"
 #include "caf/fwd.hpp"
 #include "caf/message.hpp"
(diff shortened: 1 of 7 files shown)


libcaf_core/caf/message_builder.hpp
```diff
@@ -61,6 +62,7 @@ class CAF_CORE_EXPORT message_builder {
     using namespace detail;
     using value_type = strip_and_convert_t<T>;
     static_assert(sendable<value_type>);
+    storage_size_ += padded_size_v<strip_and_convert_t<value_type>>;
```
josephnoir, line 65: `value_type` is already "stripped and converted" two lines up.

libcaf_core/caf/message_builder.hpp
```diff
@@ -91,7 +93,7 @@ class CAF_CORE_EXPORT message_builder {
   /// Converts the buffer to an actual message object and transfers
   /// ownership of the data to it, leaving this object in an invalid state.
   /// @warning Calling *any*  member function on this object afterwards
-  ///          is undefined behavior (dereferencing a `nullptr`)
+  ///          is undefined behavior.
```
josephnoir, line 96: Wouldn't it make sense to allow calling `clear()`?

Neverlord: I don't think we should support any operation on a moved-from object in general. What use case would require clearing moved-from objects instead of simply creating new builders?

Supporting `clear` would require the `type_id_list_builder` to reset its members in `move_to_list`. That would be dead code 99% of the time and only add a maintenance headache: one (weird) self-incurred requirement we need to track, test, and continue to support.

libcaf_core/src/detail/type_id_list_builder.cpp
```diff
@@ -135,14 +139,15 @@ type_id_list type_id_list_builder::move_to_list() {
...
+  if (list_size == 0)
     return make_type_id_list();
-  storage_[0] = static_cast<type_id_t>(size());
   auto vptr = malloc(size_ * sizeof(type_id_t));
   if (vptr == nullptr)
     throw std::bad_alloc();
   auto copy = reinterpret_cast<type_id_t*>(vptr);
-  memcpy(copy, storage_, size_ * sizeof(type_id_t));
+  copy[0] = static_cast<type_id_t>(list_size);
+  memcpy(copy + 1, storage_ + 1, list_size * sizeof(type_id_t));
```
josephnoir, line 150: Why did you change this?

Neverlord: Because `storage_[0]` doesn't have the final size at this point.

josephnoir: Ok. That is what line 140 ensured before and since you removed that you now have to set the size of the copy explicitly?

Neverlord: Line 140 before was flat out wrong. A `const` member function should not modify the content of the object. 😉

josephnoir: 🤭

PR #1087, opened 2020-04-11
Modernize CMake setup

Neverlord:
Our CMake scaffold goes all the way back to a time when CAF was still named `libcppa`. Back then, CMake 2.8 and GCC 4.7 were the hot stuff. 😉

A rewrite of CAF's CMake scripts is long overdue. I left some minor ToDos in places where we can improve things further when bumping the minimum CMake version.

With the new scaffold, we leave CXX flags alone. All targets properly pull in their dependencies (via PUBLIC / INTERFACE link libraries) and all CAF options are now properly sorted into CACHE variables and listed at the beginning of the file. We consequently use the `CAF::` namespace, e.g. `CAF::core` is the `libcaf_core` library.

I've also got rid of all the guess-and-disable code. By default, all modules are enabled. If the build fails, users are encouraged to *opt out* of features they don't care about such as the OpenSSL module.

Remaining ToDo items:
- [x] Update `Jenkinsfile`
- [x] Update `configure` script


CMakeLists.txt
```diff
@@ -514,74 +282,12 @@ write_basic_package_version_file(
...
-        "\n"
-        "\n"
-        "\nSource directory:      ${CMAKE_CURRENT_SOURCE_DIR}"
-        "\nBuild directory:       ${CMAKE_CURRENT_BINARY_DIR}"
-        "\nExecutable path:       ${EXECUTABLE_OUTPUT_PATH}"
-        "\nLibrary path:          ${LIBRARY_OUTPUT_PATH}"
-        "\nInstall prefix:        ${CMAKE_INSTALL_PREFIX}"
-        "\nGenerator:             ${CMAKE_GENERATOR}"
-        "\n"
-        "\n===========================================================\n")
```
josephnoir, line 586: I liked the summary, just to check flags and stuff.

configure
```diff
@@ -1,387 +1,241 @@
...
+  export-compile-commands   write JSON compile commands database [ON]
+  prefer-pthread-flag       prefer -pthread flag if available  [ON]
+  curl-examples             build examples with libcurl [OFF]
+  protobuf-examples         build examples with Google Protobuf [OFF]
+  qt5-examples              build examples with the Qt5 framework [OFF]
+  runtime-checks            build CAF with extra runtime assertions [OFF]
+  utility-targets           include targets like consistency-check [OFF]
+  actor-profiler            enable experimental proiler API [OFF]
+  examples                  build small programs showcasing CAF features [ON]
+  io-module                 build networking I/O module [ON]
```
josephnoir, line 58: What about the openssl module?

configure
```diff
@@ -1,387 +1,241 @@
...
+#   $2: CMake type
+#   $3: value
+append_cache_entry() {
+  case "$3" in
+    *\ * )
+      # string contains whitespace
+      CMakeCacheEntries="$CMakeCacheEntries -D \"$1:$2=$3\""
+      ;;
+    *)
+      # string contains whitespace
```
josephnoir, line 82:
```suggestion
      # string contains no whitespace
```

configure
```diff
@@ -1,387 +1,241 @@
...
+  protobuf-examples         build examples with Google Protobuf [OFF]
+  qt5-examples              build examples with the Qt5 framework [OFF]
+  runtime-checks            build CAF with extra runtime assertions [OFF]
+  utility-targets           include targets like consistency-check [OFF]
+  actor-profiler            enable experimental proiler API [OFF]
+  examples                  build small programs showcasing CAF features [ON]
+  io-module                 build networking I/O module [ON]
+  testing                   build unit test suites [ON]
+  tools                     build utility programs such as caf-run [ON]
+  with-exceptions           build CAF with support for exceptions [ON]
```
josephnoir, line 61: Shouldn't this just be `exceptions`?

libcaf_core/CMakeLists.txt
```diff
@@ -321,21 +223,139 @@ install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/caf"
...
+target_compile_definitions(caf-core-test PRIVATE libcaf_core_EXPORTS)
 
-set(CAF_LIBRARIES ${CAF_LIBRARIES} PARENT_SCOPE)
+target_link_libraries(caf-core-test PUBLIC CAF::test)
+
+target_link_libraries(caf-core-test PRIVATE
+  $<TARGET_PROPERTY:libcaf_core,INTERFACE_LINK_LIBRARIES>)
+
+target_include_directories(caf-core-test PRIVATE
+                           "${CMAKE_BINARY_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}")
```
josephnoir, line 245: Why is `CMAKE_BINARY_DIR` included here?

tools/CMakeLists.txt
```diff
@@ -4,10 +4,7 @@ include_directories(${LIBCAF_INCLUDE_DIRS})
 
 macro(add name)
   add_executable(${name} ${name}.cpp ${ARGN})
-  target_link_libraries(${name}
-                        ${CAF_EXTRA_LDFLAGS}
-                        ${CAF_LIBRARIES}
-                        ${PTHREAD_LIBRARIES})
+  target_link_libraries(${name} CAF::core CAF::io)
```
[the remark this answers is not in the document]
Neverlord: They are propagated through the interface properties of the targets, e.g., https://cmake.org/cmake/help/v3.2/module/FindThreads.html.

CMakeLists.txt
Neverlord: Those are available in the CMake cache and/or compile database. Including flags here is borderline useless, because they depend on the configuration when using a multi-configuration generator.

configure
Neverlord: Yep!

configure
Neverlord: Good question. Looks like I've missed it.

libcaf_core/CMakeLists.txt
Neverlord: For finding `build_config.hpp`.

PR #1079, opened 2020-04-04
Add missing cast

josephnoir:
Fixes #1077. Just tested with the master and clang10 (llvm via brew, clang version 10.0.0).

Issue #1077, opened 2020-04-01
caf-core-test target compilation failure with clang 10


dominiklohmann:
The target `caf-core-test` currently fails to compile with clang 10, tested on macOS.

#### Steps to reproduce

1. `brew install llvm` (this installs clang 10)
2. Configure environment to use the bundled clang and libc++:
```
❯ typeset -pm {{LD,CPP,CXX}FLAGS,PATH,CMAKE_\*}
export -UT LDFLAGS ldflags=( -Wl,-rpath,/usr/local/opt/llvm/lib -L/usr/local/opt/llvm/lib ) ' '
export -UT CPPFLAGS cppflags=( '-isystem /usr/local/opt/llvm/include' ) ' '
export -UT CXXFLAGS cxxflags=( '-isystem /usr/local/opt/llvm/include/c++/v1' ) ' '
export -UT PATH path=( /usr/local/opt/ccache/libexec /usr/local/opt/llvm/bin /usr/local/bin /usr/local/sbin /usr/bin /bin /usr/sbin /sbin )
export -r CMAKE_EXPORT_COMPILE_COMMANDS=ON
export -r CMAKE_GENERATOR=Ninja
export -r CMAKE_BUILD_PARALLEL_LEVEL=12
```
3. Try to build current CAF master or even the latest release (0.17.4)
```
❯ ./configure
❯ cmake --build build --target caf-core-test
```

#### Error message

<details>
<summary>Click to expand</summary>

```
❯ cmake --build build --target caf-core-test --clean-first --parallel 1
[1/1 0.031s] Cleaning all built files...
Cleaning... 136 files.
[133/247 5.660s] Building CXX object libcaf_core/CMakeFiles/caf-core-test.dir/test/behavior.cpp.o
FAILED: libcaf_core/CMakeFiles/caf-core-test.dir/test/behavior.cpp.o
[...]

PR #1089, opened 2020-04-18
Fix build on Clang 10

Neverlord:
This patch should fix the testing DSL on Clang 10. #1079 didn't take care of all conversions from `scoped_actor` to `strong_actor_ptr`. I've run into a couple more errors today.

Relates #1077.


josephnoir:
Oh, I'm sorry I didn't catch those.

Neverlord:
No worries. The CAF tests don't trigger the other paths. :)

PR #1090, opened 2020-04-18
Use only constexpr values as defaults

Neverlord:
In a few places where constexpr couldn't get used (e.g. app IDs), I fell back to fetching config values via `get_if` and then computing a fallback as necessary.


libcaf_core/caf/defaults.hpp
```diff
@@ -20,77 +20,89 @@
...
-extern CAF_CORE_EXPORT const string_view console_verbosity;
-extern CAF_CORE_EXPORT string_view file_format;
-extern CAF_CORE_EXPORT string_view file_name;
-extern CAF_CORE_EXPORT const string_view file_verbosity;
+constexpr auto component_filter = string_view{""};
+constexpr auto console = string_view{"none"};
+constexpr auto console_format = string_view{"%m"};
+constexpr auto console_verbosity = default_log_level;
+constexpr auto file_format = string_view{"%r %c %p %a %t %C %M %F:%L %m%n"};
+constexpr auto file_name = string_view{"actor_log_[TIMESTAMP]_[PID].log"};
```
josephnoir, line 94: Removing the node id from the file name was an intentional change?

libcaf_io/src/io/basp/instance.cpp
```diff
@@ -312,8 +322,13 @@ connection_state instance::handle(execution_unit* ctx, connection_handle hdl,
...
       // Check the application ID.
-      auto whitelist = get_or(config(), "middleman.app-identifiers",
-                              defaults::middleman::app_identifiers);
+      string_list whitelist;
+      if (auto lst = get_if<string_list>(&config(),
+                                         "middleman.app-identifiers")) {
+        whitelist = std::move(*lst);
+      } else {
+        whitelist.emplace_back(to_string(defaults::middleman::app_identifier));
+      }
```
josephnoir, line 331: The curly braces are here because the if is broken into two lines?

libcaf_core/caf/defaults.hpp
Neverlord: Yes, timestamp + PID are more then enough to avoid name clashes on disk. The NODE ID only blows up the file names.

josephnoir: K. We might want to add this to the patch notes? Someone might have written a script to handle / parse the logs.

Neverlord: Actually... let me revert that change. I just realized the PID + time would not disambiguate the logs of two actor systems running in the same process. We do that in a couple of unit tests.

josephnoir: Damn, you are right 😲

libcaf_io/src/io/basp/instance.cpp
Neverlord: No particular reason. I've removed the braces.

libcaf_core/caf/defaults.hpp
```diff
@@ -91,8 +91,9 @@ constexpr auto console = string_view{"none"};
 constexpr auto console_format = string_view{"%m"};
 constexpr auto console_verbosity = default_log_level;
 constexpr auto file_format = string_view{"%r %c %p %a %t %C %M %F:%L %m%n"};
-constexpr auto file_name = string_view{"actor_log_[TIMESTAMP]_[PID].log"};
 constexpr auto file_verbosity = default_log_level;
+constexpr auto file_name
+  = string_view{"actor_log_[TIMESTAMP]_[PID]_[NODE].log"};
```
josephnoir, line 96:
```suggestion
  = string_view{"actor_log_[PID]_[TIMESTAMP]_[NODE].log"};
```

I think that's how it was previously.

Neverlord: Indeed.

dominiklohmann:
Don't these all need to be `inline` as well when they're not static member variables?

Neverlord:
No, `constexpr` implies internal linkage.

dominiklohmann:
> No, `constexpr` implies internal linkage.

That's not always true, and it is false in the case of this PR.

From [cppreference](https://en.cppreference.com/w/cpp/language/constexpr) (because it is easier to quote just the relevant bits than the standard), added emphasis mine:

> The constexpr specifier declares that it is possible to evaluate the value of the function or variable at compile time. Such variables and functions can then be used where only compile time constant expressions are allowed (provided that appropriate function arguments are given). A constexpr specifier used in an object declaration or non-static member function (until C++14) implies const. **A constexpr specifier used in a function or static member variable (since C++17) declaration implies inline.** If any declaration of a function or function template has a constexpr specifier, then every declaration must contain that specifier.

Neverlord:
Please read that again:

> A constexpr specifier (...) implies inline.

Besides, the build would have failed with tons of duplicate symbol errors if `constexpr` would not imply internal linkage.

dominiklohmann:
>  used in a function or static member variable (since C++17)

The part you omitted is restricting it to functions and static member variables, which notably excludes global variables.

From the page on the [inline specifier](https://en.cppreference.com/w/cpp/language/inline):

> A static member variable **(but not a namespace-scope variable)** declared constexpr is implicitly an inline variable.

> Besides, the build would have failed with tons of duplicate symbol errors

Works in practice does not necessarily mean conforms to the standard. This only violates the ODR if the compiler decides not to inline the value for the variable, e.g., if you take its address.

Neverlord:
> `constexpr` implies const and const on global/namespace scope implies static (internal linkage), which means that every translation unit including this header gets its own copy of PI. The memory for that static is only going to be allocated if an address or reference to it is taken, and the address is going to be different in each translation unit.

via https://isocpp.org/blog/2018/05/quick-q-use-of-constexpr-in-header-file

Internal linkage is exactly the behavior I was looking for. If you put an `inline` here, you allow the compiler to produce a single address for the constants (which is also semi-true across .dll/.so boundaries due to visibility=hidden - to really guarantee unique addresses, we would have to use `extern`).

dominiklohmann:
I stand corrected in that case. I was thinking about inline, not about just internal linkage. Thanks!

Issue #1095, opened 2020-04-22
Type mismatch in the heartbeat interval parameter


Neverlord:
Our actor system config has the heartbeat parameter as `timespan`:

```c++
    .add<timespan>("heartbeat-interval", "interval of heartbeat messages")
```

The BASP broker reads this config parameter when spinning up:

```c++
  auto heartbeat_interval = get_or(config(), "middleman.heartbeat-interval",
                                   defaults::middleman::heartbeat_interval);
```

However, it is impossible as  a user to effectively provide a valid interval. Because:

```c++
constexpr auto heartbeat_interval = size_t{0};
```

When converting the config parameter to `timespan`, we forgot to update the BASP broker. It still believes the config parameter is an integer with millisecond resolution.

On the command line, users cannot set this parameter. CAF expects a `timespan`, so the parser complains when finding an integer without time suffix (`s`, `ms`, etc.).

Setting the parameter via configuration file as an integer works. Maybe we also should consider this a bug and reject config files with type mismatches for known parameters.

Neverlord:
I found this bug while working on #1094.

Neverlord:
Fixed via #1096.

Issue #1094, opened 2020-04-22
Frozen BASP nodes never disconnect


Neverlord:
We currently rely on TCP for detecting disconnect. However, (very rarely) we've observed frozen nodes where the TCP connection despite heartbeats appeared to remain intact while the remote side never responded to anything.

Rather than relying on TCP, we should actively monitor remote nodes and sever the connection after a (configurable) certain number of heartbeats were skipped.

josephnoir:
There are other approaches to failure detection aside from just counting missing heartbeats, such as the [Phi Accrual Failure Detector](https://pdfs.semanticscholar.org/11ae/4c0c0d0c36dc177c1fff5eb84fa49aa3e1a8.pdf) which is available in [Akka](https://doc.akka.io/docs/akka/current/typed/failure-detector.html) and [Cassandra](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/gms/FailureDetector.java). Not sure if we just want to wing it or maybe give this some thought.

Neverlord:
Gossip protocols certainly make sense in a cluster setup to agree on whether a node failed or not. I think the connection monitoring discussed here is complementary to that.

Neverlord:
Sorry, next time I should click the link first. :)

This is not about the gossip protocol they're using in their cluster framework. I think a more sophisticated error detector would be nice to have, but I think it should go into `caf_net` instead. Connection timeout + heartbeat should do the trick for now.

Neverlord:
Fixed via #1096.

PR #1096, opened 2020-04-22
Disconnect BASP nodes when reaching connection timeouts

Neverlord:
Note: this PR takes care of #1094 and #1095 for CAF 0.17.5. Once this got merged, I'll port the changes to `master` for CAF 0.18.

To enable the frozen-node-detection by default, I've turned heartbeats to *on* by default. The default timeouts are very conservative: heartbeats every 10s, disconnects after not hearing anything from another node after 30s.

### DIFF
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -115,8 +115,6 @@ actor_system_config::actor_system_config()
                                   "excluded components for logging")
     .add<bool>("inline-output", "disable logger thread (for testing only!)");
   opt_group{custom_options_, "middleman"}
-    .add<atom_value>("network-backend",
-                     "either 'default' or 'asio' (if available)")
     .add<std::vector<string>>("app-identifiers",
                               "valid application identifiers of this node")
     .add<string>("app-identifier", "DEPRECATED: use app-identifiers instead")
--- a/libcaf_core/src/actor_system_config.cpp
+++ b/libcaf_core/src/actor_system_config.cpp
@@ -127,7 +127,7 @@ actor_system_config::actor_system_config()
     .add<timespan>("heartbeat-interval", "interval of heartbeat messages")
     .add<timespan>("connection-timeout",
                    "max. time between messages before declaring a node dead "
-                   "(ignored if heartbeats are disabled)")
+                   "(disabled if 0, ignored if heartbeats are disabled)")
     .add<bool>("attach-utility-actors",
                "schedule utility actors instead of dedicating threads")
     .add<bool>("manual-multiplexing",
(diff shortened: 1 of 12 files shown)


libcaf_io/src/io/basp_broker.cpp
```diff
@@ -329,10 +334,46 @@ behavior basp_broker::make_behavior() {
...
+          auto hdl = i->second.hdl;
+          // connection_cleanup below calls ctx.erase, so we need to increase
+          // the iterator now, before it gets invalidated.
+          ++i;
+          connection_cleanup(hdl, sec::connection_timeout);
+          close(hdl);
+        } else {
+          ++i;
+        }
+      }
```
josephnoir, line 373: We should only evaluate disconnects if the `connection_timeout  > 0`.

libcaf_io/src/io/basp_broker.cpp
```diff
@@ -329,10 +334,46 @@ behavior basp_broker::make_behavior() {
...
+                      "Other nodes may disconnect.");
+        while (now >= next_tick)
+          next_tick += heartbeat_interval;
+
+      } else if (now >= scheduled + (heartbeat_interval / 2)) {
+        CAF_LOG_WARNING("Lagging more than 50% of a heartbeat interval behind! "
+                        "Interval too low or BASP actor overloaded!");
+      }
+      // Send out heartbeats.
       instance.handle_heartbeat(context());
```
josephnoir, line 359: I think `handle_heartbeat` is a misleading name. The function only sends out heartbeat messages. How about renaming it to `send_heartbeats` or `write_heartbeats`?

libcaf_core/src/actor_system_config.cpp
```diff
@@ -125,6 +125,9 @@ actor_system_config::actor_system_config()
     .add<size_t>("max-consecutive-reads",
                  "max. number of consecutive reads per broker")
     .add<timespan>("heartbeat-interval", "interval of heartbeat messages")
```
josephnoir, line 127: A bit of a conceptual weird point is that the `heartbeat-interval` of one node is related to the `connection-timeout` of its peers. Since heartbeats are uni-directional and don't provoke a response you have to be careful to keep that in mind. E.g., you can't have a configuration where some nodes are more interested in the liveliness status of their peers than others. Well you can, but the `heartbeat-interval` must fit the lowest timeout of all nodes.

I think we should make that more explicit in the description.

Neverlord: Agreed. Ideally, BASP nodes would agree on a heartbeat interval during the handshake. However, for CAF 0.17.5, I'd like to avoid (breaking) changes to the network protocol. We should consider changing the handshake for 0.18, though.

Neverlord: Hm, by the same reasoning: I think we should turn this off by default. If you are running CAF 0.17.4 and 0.17.5 side-by-side (e.g. during a migration phase), your 0.17.5 nodes otherwise sever a connection after 30s of no activity (0.17.4 has heartbeats *off* by default). We should recommend setting this for now and implement a better handshake with 0.18.

PR #1098, opened 2020-04-23
Update FindCAF.cmake

jakobod:
The `FindCAF.cmake` script has not been updated when the cmake scaffold has been reworked.
I copied the updated version from the incubator to keep it updated for users.

### DIFF
--- a/cmake/FindCAF.cmake
+++ b/cmake/FindCAF.cmake
@@ -13,9 +13,10 @@
 # Variables defined by this module:
 #
 #  CAF_FOUND              System has CAF headers and library
-#  CAF_VERSION            Found CAF release number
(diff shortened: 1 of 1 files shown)


Neverlord:
Thanks for paying attention to details and filing the PR!

I took my time to think a bit about the script. Modern CMake generally favors the `CONFIG` mode of [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) over the old `MODULE` approach using `FindX` scripts. Our setup already includes a rudimentary CMake config file and we install the CMake config alongside CAF in our `install` target.

Since we use this release to overhaul the entire CMake setup, I think we should get rid of the script and only support the `CONFIG` mode going forward. We should also extend our build pipeline to check whether a CAF installation works as expected for downstream projects.

Issue #1097, opened 2020-04-23
Use nodiscard attribute where appropriate


Neverlord:
At least, we should add `[[nodiscard]]` to (almost?) all function that return an `error`, e.g., `actor_system_config::parse`.

CppPhil:
Good idea.
Maybe we can add the nodiscard attribute to the error type itself to cut down on the typing a bit.
IIRC you can apply the attribute to types as well.

Neverlord:
Yes, I like that suggestion. We can probably add the attribute to `error` and `error_code`.

CppPhil:
Adding the `nodiscard` attribute to `error` and `error_code` turns out not to be very easy to do because of `CAF_CORE_EXPORT`.
See: https://stackoverflow.com/questions/57993818/gcc-how-to-combine-attribute-dllexport-and-nodiscard-in-a-struct-de
`class [[gnu::visibility("default")]] [[nodiscard]] error` seems to work and does generate additional warnings.
We should be able to use the `[[gnu::]]` attributes since we don't support compilers without C++17 support anymore. (Or do we?)
CMake's `generate_export_header` (https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html) doesn't seem to have an option to get it to use something akin to `[[gnu::visibility("default")]]`.
Inspecting https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/GenerateExportHeader.cmake we can see that the `__attribute__((visibility("default")))` attributes are just hard coded as such.
How do we want to proceed?
Do we want to maintain a custom `generate_export_header` implementation?

Neverlord:
AFAIK, we can place the attribute to the declaration. I haven't tried whether all compilers are fine with having standard attributes on the declaration and the visibility flag on the implementation, but it's worth a shot.

```C++
// fwd.hpp
...
class [[nodiscard]] error; 
...

// error.hpp
...
class CAF_CORE_EXPORT error : detail::comparable<error> { 
...
```

CppPhil:
> AFAIK, we can place the attribute to the declaration. I haven't tried whether all compilers are fine with having standard attributes on the declaration and the visibility flag on the implementation, but it's worth a shot.
> 
> ```c++
> // fwd.hpp
> ...
> class [[nodiscard]] error; 
> ...
> 
> // error.hpp
> ...
> class CAF_CORE_EXPORT error : detail::comparable<error> { 
> ...
> ```

Sounds like a clever idea.

PR #1109, opened 2020-06-02
Use nodiscard attribute where appropriate

CppPhil:
Declares `error_code` and `error` in `fwd.hpp` as `[[nodiscard]]`.
This does generate some additional warnings that may or may not indicate bugs.
I've taken the liberty to silence the warning in `libcaf_core/caf/detail/type_traits.hpp` since `x->apply(std::declval<Arg>())` is only used for the sake of expression SFINAE and I don't think the return value matters here.
Closes: #1097


Neverlord:
Thanks for taking care of this! I'll check this out locally to see whether the patch introduces warnings. Otherwise LGTM.

PR #1115, opened 2020-06-23
Switch to new, cleaner config syntax

### LINKED ISSUE #941: Accept deep_to_string output for settings in the config parser
Currently, the config parser does not accept the output that `caf::deep_to_strings` generates when passing a `caf::settings`. Tweaking the parser to accept that format would enable "round trips" and makes it easier to work with settings (e.g., by allowing users to just print the output of `deep_to_string` to a config file).

Neverlord:
At long last, it's time to say goodbye to the INI syntax. Initially, we've picked INI for its simplicity and because we could parse it without an actual parser backend. Now that we do have a parser backend that can cope with nested configuration blocks, it's time to move on. The new format is essentially the syntax we've been supporting for quite a while minus the `[category]` notation. There are basically two additions:

- Keys can be quoted.
- Strings can use either `'` or `"`. They behave like they would in JavaScript (i.e., only affect escaping).

By supporting quoted keys, the parser can now finally read the `to_string` output of `settings`.

Closes #941.


libcaf_core/src/detail/config_consumer.cpp
```diff
@@ -0,0 +1,201 @@
...
+  : options_(other.options_), parent_(other.parent_), cfg_(other.cfg_) {
+  other.parent_ = none;
+}
+
+config_consumer& config_consumer::operator=(config_consumer&& other) {
+  destroy();
+  options_ = other.options_;
+  parent_ = other.parent_;
+  cfg_ = other.cfg_;
+  other.parent_ = none;
```
josephnoir, line 107: Same.
```suggestion
  other.parent_ = none;
  other.cfg_ = nullptr;
```

libcaf_core/src/detail/config_consumer.cpp
```diff
@@ -0,0 +1,201 @@
...
+    } else {
+      dst.insert_or_assign(key, std::move(value));
+    }
+  }
+}
+
+} // namespace
+
+pec config_consumer::value_impl(config_value&& x) {
+  // See whether there's an config_option associated to this key and perform a
```
josephnoir, line 159:
```suggestion
  // See whether there's a config_option associated to this key and perform a
```

libcaf_core/test/detail/config_consumer.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+CAF_TEST_FIXTURE_SCOPE(config_consumer_tests, fixture)
+
+CAF_TEST(config_consumer) {
+  string_view str = test_config1;
+  detail::config_consumer consumer{options, config};
+  string_parser_state res{str.begin(), str.end()};
+  detail::parser::read_config(res, consumer);
+  CAF_CHECK_EQUAL(res.code, pec::success);
+  CAF_CHECK_EQUAL(string_view(res.i, res.e), string_view());
```
josephnoir, line 84:
```suggestion
  CAF_CHECK_EQUAL(res.i, res.e);
```

libcaf_core/src/detail/config_consumer.cpp
```diff
@@ -0,0 +1,201 @@
...
+  // nop
+}
+
+config_consumer::config_consumer(settings& cfg) : cfg_(&cfg) {
+  // nop
+}
+
+config_consumer::config_consumer(config_consumer&& other)
+  : options_(other.options_), parent_(other.parent_), cfg_(other.cfg_) {
+  other.parent_ = none;
```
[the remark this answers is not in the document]
Neverlord: That would be a dead write. The only legal things to do on moved-from objects is calling the destructor or move something new into it. We must set the `parent_` to `none` to stop the destructor from deleting `cfg_`, but everything else is irrelevant.

libcaf_core/test/detail/config_consumer.cpp
Neverlord: These two checks give the same yes/no answer, but there's a reason I've written it like this: writing it this way will print what's left in the buffer. This means, if the check fails, it also gives you the information you need to hunt for the bug right away. 🙂

josephnoir: Neat.

PR #1118, opened 2020-07-13
Capture message text of exceptions if actors die

### LINKED ISSUE #1117: Capture message text of exceptions if actors die
If a worker dies due to an unhandled exception, the default exception handler creates an `error` with `sec::runtime_error`. For exceptions that derive from `std::exception`, CAF could at least store the exception message in the error as well.

Neverlord:
Closes #1117.


libcaf_core/src/scheduled_actor.cpp
```diff
@@ -96,22 +96,24 @@ void scheduled_actor::default_exit_handler(scheduled_actor* ptr, exit_msg& x) {
...
 
 #ifdef CAF_ENABLE_EXCEPTIONS
-error scheduled_actor::default_exception_handler(pointer ptr,
+error scheduled_actor::default_exception_handler(local_actor* ptr,
                                                  std::exception_ptr& x) {
   CAF_ASSERT(x != nullptr);
   try {
     std::rethrow_exception(x);
-  } catch (const std::exception& e) {
+  } catch (std::exception& e) {
```
josephnoir, line 104: Why loose the `const` here?

Neverlord: It just makes little sense to capture exceptions by const reference. Capturing by reference is the idiomatic way and there's always going to be a single reference to the exception anyway.

josephnoir: k

PR #1122, opened 2020-07-24
Update CAF_VERSION in config.hpp

tobim:


Neverlord:
Noticed it as well, thanks.

PR #1131, opened 2020-07-30
Fix missing header file in install directory

### LINKED ISSUE #1130: Missing header file (caf/detail/build_config.hpp) in install location
Building from `master`, using a minimal setup, I tried integrating the framework into a simple application. `make install` produced a nice CAF installation package using prefix. Here are my source files:
```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.17 FATAL_ERROR)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)

project(test)
    find_package(CAF REQUIRED)
    add_executable(test main.cpp)

    target_link_libraries(test PRIVATE CAF::core)
```

```cpp
// main.cpp
#include <caf/all.hpp>
#include <iostream>

int main()
{
    std::cout << "Hello, world!" << std::endl;

    caf::actor{};

    return 0;
}
```
I received the following error: 
```
[ 50%] Building CXX object CMakeFiles/test.dir/main.cpp.o
In file included from /home/<USER>/CAF/include/caf/all.hpp:21:0,
                 from /home/<USER>/test/main.cpp:1:
/home/<USER>/CAF/include/caf/config.hpp:22:10: fatal error: caf/detail/build_config.hpp: No such file or directory
 #include "caf/detail/build_config.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
As I understand, this stems from the fact the file is only placed into binary output directory, but not installed.

Bedzior:
Fixes #1130 

This is a draft, because component-wise `install()` is typically called from within `caf_export_and_install_lib`. Is this fine?

Secondly, this could also call for an immediate staging phase on project's CI, where the framework would be compiled into a simple executable to verify validity of the installation.

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -187,6 +187,8 @@ add_library(libcaf_core "${PROJECT_SOURCE_DIR}/cmake/dummy.cpp"
 caf_core_set_default_properties(libcaf_core_obj libcaf_core)
 
 caf_export_and_install_lib(core)
+install(FILES "${CMAKE_BINARY_DIR}/caf/detail/build_config.hpp"
+          DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/caf/detail")
 
 # -- build unit tests ----------------------------------------------------------
 


Neverlord:
> Is this fine?

Yes, totally fine. The build header isn't part of the source tree, so it needs special attention. Thanks a lot for the fix! 🙂

Neverlord:
> verify validity of the installation

That's a good point as well, but we should track it separately. The examples compile binaries, but they don't use `find_package(CAF)` since they can just use the CMake targets.

Issue #1121, opened 2020-07-16
Build failure when building with Bazel: error: 'type_id' is not a class, namespace, or enumeration


yesudeep:
Hi!

I'm attempting to build the CAF using [Bazel](bazel.build). However, 
I get an error:

## Clang 10.0.0 on Mac OS X 10.15.5:

```shell
❯ bazel build --sandbox_debug "@org_actor_framework_caf//:core"
... elided ...
external/org_actor_framework_caf/libcaf_core/src/type_id.cpp:23:1: error: 'type_id' is not a class, namespace, or enumeration
type_id::type_id() {
^
external/org_actor_framework_caf/libcaf_core/caf/type_id.hpp:42:8: note: 'type_id' declared here
struct type_id;
       ^
external/org_actor_framework_caf/libcaf_core/src/type_id.cpp:27:1: error: 'type_id' is not a class, namespace, or enumeration
type_id::~type_id() {
^
external/org_actor_framework_caf/libcaf_core/caf/type_id.hpp:42:8: note: 'type_id' declared here
struct type_id;
       ^
2 errors generated.
Target @org_actor_framework_caf//:core failed to build
INFO: Elapsed time: 6.398s, Critical Path: 0.89s
INFO: 1 process: 1 darwin-sandbox.
FAILED: Build did NOT complete successfully
```

## GCC 10.1.1 on Linux 5.7.8 x86_64:

```shell
❯ bazel build "@org_actor_framework_caf//:core"
... elided ...
Use --sandbox_debug to see verbose messages from the sandbox
external/org_actor_framework_caf/libcaf_core/src/type_id.cpp:23:8: error: deduced class type 'type_id' in function return type
   23 | type_id::type_id() {
      |        ^~
In file included from external/org_actor_framework_caf/libcaf_core/caf/type_id.hpp:30,
                 from external/org_actor_framework_caf/libcaf_core/src/type_id.cpp:19:
external/org_actor_framework_caf/libcaf_core/caf/fwd.hpp:49:25: note: 'template<class> struct caf::type_id' declared here
   49 | template <class> struct type_id;
      |                         ^~~~~~~
external/org_actor_framework_caf/libcaf_core/src/type_id.cpp:27:10: error: expected id-expression before '~' token
   27 | type_id::~type_id() {
      |          ^
Target @org_actor_framework_caf//:core failed to build
INFO: Elapsed time: 1.295s, Critical Path: 1.22s
INFO: 0 processes.
[...]

PR #1132, opened 2020-08-01
Remove obsolete files, fix serial_reply unit test suite

Neverlord:
Relates #1121.


libcaf_core/test/serial_reply.cpp
```diff
@@ -29,48 +29,50 @@ CAF_TEST(test_serial_reply) {
...
-    self->become([=](int) mutable {
-      auto rp = self->make_response_promise();
-      CAF_MESSAGE("received 'hi there'");
-      self->request(c0, infinite, sub0_atom::value)
-        .then([=](sub0_atom) mutable {
+    return {
+      [=](hi_atom) mutable {
+        auto rp = self->make_response_promise();
+        CAF_MESSAGE("received 'hi there'");
+        self->request(c0, infinite, sub0_atom_v).then([=](sub0_atom) mutable {
```
jakobod, line 50: Just for my understanding: Why do these handlers have to be marked `mutable`? Wouldn't referencing the context be sufficient?

Neverlord: We call `deliver` on the captured response promise, which requires non-const access.

PR #1133, opened 2020-08-04
Fix datagram servant shutdown

josephnoir:
In cases where a datagram server did not participate in any communication it did not get added to its broker and wasn't shut down correctly. As a result, the multiplexer would hang in its poll loop and prevent the actor system from shutting down.


Neverlord:
@josephnoir Please also make a followup for cherry-picking this to `master` (`git cherry-pick -x`) and adding a changelog entry. 🙂

PR #1135, opened 2020-08-17
Handle overflows in stream slot id selection

lava:
Previously, the next stream slot would always come out as `0` as soon as one stream slot with id `65535` was created.

NOTE: I did try for a while, but I did not see a good way to exercise the new code path in a unit test. If you have an idea how to do it, i'd be happy to add another commit with the test to this PR.

### DIFF
--- a/libcaf_core/src/scheduled_actor.cpp
+++ b/libcaf_core/src/scheduled_actor.cpp
@@ -1023,7 +1023,17 @@ uint64_t scheduled_actor::set_timeout(std::string type,
 stream_slot scheduled_actor::next_slot() {
   stream_slot result = 1;
   auto nslot = [](const stream_manager_map& x) -> stream_slot {
-    return x.rbegin()->first + 1;
+    auto highest = x.rbegin()->first;
+    if (highest < std::numeric_limits<decltype(highest)>::max())
+      return highest + 1;
+    // Back-up algorithm in the case of an overflow: Take the minimum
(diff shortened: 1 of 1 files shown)


libcaf_core/src/scheduled_actor.cpp
```diff
@@ -1023,7 +1023,23 @@ uint64_t scheduled_actor::set_timeout(std::string type,
...
+    if (highest < std::numeric_limits<decltype(highest)>::max())
+      return highest + 1;
+    // Back-up algorithm in the case of an overflow: Take the minimum
+    // id `1` if its still free, otherwise take an id from the first
+    // gap between two consecutive keys.
+    if (1 < x.begin()->first)
+      return 1;
+    auto current = x.begin();
+    auto next = ++x.begin();
+    while (next != x.end()) {
```
Neverlord, line 1036: I feel like the hand-rolled loop makes this more complicated than it ought ought to be. Why not use `adjacent_find`? Something along those lines:

```c++
  auto has_gap = [](auto& a, auto& b) { return b.first - a.first > 1; };
  auto i = std::adjacent_find(x.begin(), x.end(), has_gap);
  // ... return i->second + 1 if i != x.end() ...
```

Neverlord (approved):
Thanks for submitting this bugfix! I think the code could be improved, though.

Neverlord:
Thanks for the fix! 🙏

PR #1138, opened 2020-08-27
Add logging statements when adjusting actor registry count

lava:
Add a debug statements printing the actor id whenever the running actor count is adjusted.

This speeds up debugging for errors where the running count does not reach zero at program shutdown but it is not clear which actor is still alive.

NOTE: The compatibility header is based on similar code I wrote at https://mesos.apache.org/api/latest/c++/3rdparty_2stout_2include_2stout_2attributes_8hpp_source.html . If CAF already requires a compiler with support for standard attributes this can also be removed to reduce cruft.

### DIFF
--- a/libcaf_core/caf/actor_registry.hpp
+++ b/libcaf_core/caf/actor_registry.hpp
@@ -65,11 +65,11 @@ public:
   /// leaving `reason` for future reference.
   void erase(actor_id key);
 
-  /// Increases running-actors-count by one.
-  void inc_running();
+  /// Increases running-actors-count by one. Returns the increased count.
+  size_t inc_running();
 
-  /// Decreases running-actors-count by one.
-  void dec_running();
+  /// Decreases running-actors-count by one. Returns the decreased count.
+  size_t dec_running();
 
   /// Returns the number of currently running actors.
   size_t running() const;
--- a/libcaf_core/src/abstract_actor.cpp
+++ b/libcaf_core/src/abstract_actor.cpp
@@ -95,14 +95,16 @@ void abstract_actor::register_at_system() {
   if (getf(is_registered_flag))
     return;
   setf(is_registered_flag);
-  home_system().registry().inc_running();
+  [[maybe_unused]] auto count = home_system().registry().inc_running();
+  CAF_LOG_DEBUG("actor " << id() << " increased running count to " << count);
 }
 
 void abstract_actor::unregister_from_system() {
   if (!getf(is_registered_flag))
     return;
   unsetf(is_registered_flag);
-  home_system().registry().dec_running();
+  [[maybe_unused]] auto count = home_system().registry().dec_running();
+  CAF_LOG_DEBUG("actor " << id() << " decreased running count to " << count);
 }
 
 } // namespace caf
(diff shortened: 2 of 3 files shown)


Neverlord:
1. Before adding an unnecessary macro workaround, you could have easily checked whether we use the attribute: `fgrep -R '[[maybe_unused]]' libcaf_*`
2. How is the new logging speeding up debugging over `grep 'TERMINATE\|SPAWN' my.log`?

lava:
Thanks for the reply. I actually did (1), but on the 0.17 branch since that's what we're currently using internally, so I did not get any results, and I didn't think to check again after rebasing.

For (2), it was not obvious to me what guarantees there are on the relationship between spawn()/terminate() and the registry. Not all spawned actors are registered, and from the docs it seemed like usage of the registry is intended to be somewhat dynamic, e.g. with middleman adding and removing actors as needed. Since shutdown is usually blocking on `await_running_count_equal(0)`, adding print statements directly in the registry strips away all abstraction and directly gives the offending actor id by subtracting the removed ids from the added ids.

sourcedelica:
I'm all in favor of making this easier to debug.  It's one of the things that trips up new CAF developers on our team pretty regularly.

Neverlord:
> I'm all in favor of making this easier to debug.

Me too. 😉

I'm just wondering whether we can advance things further. So far, this revealed two things for me:

> the relationship between spawn()/terminate() and the registry.

The counter is actually a system metric and has nothing to do with the registry. The counter is part of the registry for historical reasons only. So, I think it's about time we get that right in the API and move the counter up to the `actor_system`. The SPAWN/TERMINATE events are the right things to look at. It seems like there's only one information missing: whether the actor is hidden or not. Hidden actors simply don't increase the `running` count. I feel like properly documenting that and maybe providing a small Python script that spits out unterminated actors from a log would help more in the long run.

> ...  strips away all abstraction and directly gives the offending actor id by subtracting the removed ids from the added ids.

So, here's another idea. We already have the metric `caf.running-actors`. We could rename that to `caf.system.running-actors` (total number of actors in the system) and add a new `caf.actor.running` metric (number of running actor for one kind of actor). With that, once your actor system hangs, you simply enable metrics for your actors, fire up your application again and get to the offending actor(s) with one shot using `curl` (plus `grep`, maybe) rather than combing through the debug logs. Because not only do you need CAF debug logs otherwise, you also need to have a CAF version with debug logging builtin. Metrics are always available to you.

Once you know what you are looking for, you can dive into the post-mortem analysis with log files (which you may generate with other means that CAF logging).

sourcedelica:
Not having to build logging variant to debug this issue would be very nice.

lava:
Adding the information to the metrics sounds like a great plan. It might also be worth considering doing both, since having the information in the logs would probably make it easier to get it in non-interactive contexts (eg. a CI run, or when post-mortem debugging from a customer issue)

Neverlord (changes requested):
Regardless of the other ideas mentioned to improve debugging, I think this is generally a useful addition. However, I do have a request regarding the code. Passing the actor ID in semantically makes no sense, since the function does nothing with it (other than logging). Please have `inc_running` and `dec_running` return the new value and log at the calling actor.

Both functions are called only in `libcaf_core/src/abstract_actor.cpp`, so it's just shifting code around a bit.

lava:
@Neverlord I just pushed an updated version of this PR that includes the changes you suggested, thanks for taking the time.

Neverlord:
Thanks for the PR! 👍

PR #1141, opened 2020-09-24
Fix segfault when setting an invalid credit policy

### LINKED ISSUE #1140: Setting caf.stream.credit-policy to an invalid value causes segfault
When running the integer-stream example with `--caf.stream.credit-policy=xyz`, the application segfaults because it tries to dereference a nullptr instead of exiting gracefully with an error message. Ideally this would list the available credit policies.

This was tested with the current rc for 0.18.

dominiklohmann:
Fixes #1140


Neverlord:
Thanks for the fix! 🙏

PR #1142, opened 2020-09-24
Fix warnings with clang-10

dominiklohmann:
I've had these fixed for a while locally, so why not push them. This allows for building CAF without warnings using clang-10, thus removing all the unnecessary noise in the compiler output.

@Neverlord did you consider running clang-10 in CI because of the new warnings in `-Wall`?

### DIFF
--- a/libcaf_core/caf/telemetry/metric_registry.hpp
+++ b/libcaf_core/caf/telemetry/metric_registry.hpp
@@ -407,8 +407,8 @@ public:
   ///               as well as prefixes starting with an underscore are
   ///               reserved.
   /// @param name The human-readable name of the metric, e.g., `requests`.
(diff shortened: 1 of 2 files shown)


Neverlord:
> I've had these fixed for a while locally, so why not push them

Indeed, thanks for sharing!

> did you consider running clang-10 in CI

We don't pick individual compiler versions. Rather, we try following the main Linux distributions plus MSVC and XCode. Speaking of, our selection currently isn't up-to-date, though: at least Fedora 32 is missing. @jakobod mind adding the new Docker image and check whether we miss something else?

Btw, not sure why merge status is red here. It's green in the CI, but it tries to set the commit status for the merge-SHA instead of 44b091c.

jakobod:
@Neverlord Noted.

PR #1151, opened 2020-10-17
Allow aborting nested downstream managers

### LINKED ISSUE #1150: Stream managers may access destroyed path objects
The segfault I was investigating in #1148 actually seems to be a heap-use-after-free bug related to the `downstream_manager`. This may be specific to the `fused_downstream_manager` implementation. In this concrete instance, the actor called a function on an `outbound_path` after `caf::downstream_manager_base::clear_paths()` destroyed the object.

Neverlord:
So, the segfault I've investigated in #1150 turned out to be caused by user code that called `abort` on a nested downstream manager of a `fused_downstream_manager`. The fused manager holds non-owning pointers to the paths, so users must make sure that these pointers are cleaned up as well. The new function adds a safe way to abort a nested manager.

The fused managers are difficult to use correctly in general, so we probably should revisit this API sooner rather than later.

Close #1150.

### DIFF
--- a/libcaf_core/caf/fused_downstream_manager.hpp
+++ b/libcaf_core/caf/fused_downstream_manager.hpp
@@ -99,6 +99,11 @@ struct init_ptr_array<I, I> {
 template <class T, class... Ts>
 class fused_downstream_manager : public downstream_manager {
 public:
+  // -- sanity checks ----------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


Neverlord:
I've actually found a way to this without the new API. If we do end up revisiting this API, maybe it's better to not add new stuff.

PR #1153, opened 2020-10-30
Link against libatomic on 32-bit systems

### LINKED ISSUE #1129: CAF fails to compile on 32-bit Linux with GCC
The master branch can not be compiled in Raspbian. 
I use raspberry pi 3 and my GCC version is gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1). It gave me these errors:

```
[ 35%] Linking CXX executable caf-core-test
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/actor_clock.cpp.o: in function `std::__atomic_base<long long>::operator++()':
actor_clock.cpp:(.text._ZNSt13__atomic_baseIxEppEv[_ZNSt13__atomic_baseIxEppEv]+0x28): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/collector/prometheus.cpp.o: in function `caf::telemetry::int_gauge::inc(long long)':
prometheus.cpp:(.text._ZN3caf9telemetry9int_gauge3incEx[_ZN3caf9telemetry9int_gauge3incEx]+0x3c): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/collector/prometheus.cpp.o: in function `caf::telemetry::int_gauge::value(long long)':
prometheus.cpp:(.text._ZN3caf9telemetry9int_gauge5valueEx[_ZN3caf9telemetry9int_gauge5valueEx]+0x4c): undefined reference to `__atomic_store_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/counter.cpp.o: in function `caf::telemetry::int_gauge::value() const':
counter.cpp:(.text._ZNK3caf9telemetry9int_gauge5valueEv[_ZNK3caf9telemetry9int_gauge5valueEv]+0x40): undefined reference to `__atomic_load_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/counter.cpp.o: in function `std::atomic<double>::load(std::memory_order) const':
counter.cpp:(.text._ZNKSt6atomicIdE4loadESt12memory_order[_ZNKSt6atomicIdE4loadESt12memory_order]+0x38): undefined reference to `__atomic_load_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/counter.cpp.o: in function `std::atomic<double>::compare_exchange_weak(double&, double, std::memory_order, std::memory_order)':
counter.cpp:(.text._ZNSt6atomicIdE21compare_exchange_weakERddSt12memory_orderS2_[_ZNSt6atomicIdE21compare_exchange_weakERddSt12memory_orderS2_]+0x78): undefined reference to `__atomic_compare_exchange_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/gauge.cpp.o: in function `caf::telemetry::int_gauge::dec(long long)':
gauge.cpp:(.text._ZN3caf9telemetry9int_gauge3decEx[_ZN3caf9telemetry9int_gauge3decEx]+0x3c): undefined reference to `__atomic_fetch_sub_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/gauge.cpp.o: in function `std::atomic<double>::store(double, std::memory_order)':
gauge.cpp:(.text._ZNSt6atomicIdE5storeEdSt12memory_order[_ZNSt6atomicIdE5storeEdSt12memory_order]+0x58): undefined reference to `__atomic_store_8'
/usr/bin/ld: CMakeFiles/caf-core-test.dir/test/telemetry/gauge.cpp.o: in function `std::__atomic_base<long long>::operator--()':
gauge.cpp:(.text._ZNSt13__atomic_baseIxEmmEv[_ZNSt13__atomic_baseIxEmmEv]+0x28): undefined reference to `__atomic_fetch_sub_8'
collect2: error: ld returned 1 exit status
```

But the actor framework 0.17.6 can be compiled in my Raspbian without any problem.

Neverlord:
Closes #1129.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 - Building executables that link to CAF on 32-bit Linux versions using GCC
   failed due to undefined references to `__atomic_fetch` symbols. Adding a CMake
   dependency for `caf_core` to libatomic gets executables to compile and link as
-  expected.
+  expected (#1153).
 
 ## [0.18.0-rc.1] - 2020-09-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   case or do unexpected things in the latter case. The stringification inspector
   now matches precisely on pointer types to stop the compiler from doing
   implicit conversions in the first place.
+- Building executables that link to CAF on 32-bit Linux versions using GCC
+  failed due to undefined references to `__atomic_fetch` symbols. Adding a CMake
+  dependency for `caf_core` to libatomic gets executables to compile and link as
+  expected.
 
 ## [0.18.0-rc.1] - 2020-09-09
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(CAF CXX)
 
 include(CMakeDependentOption)      # Conditional default values
 include(CMakePackageConfigHelpers) # For creating .cmake files
+include(CheckCXXSourceCompiles)    # For checking whether some C++ features work
 include(GNUInstallDirs)            # Sets default install paths
 include(GenerateExportHeader)      # Auto-generates dllexport macros
 
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -194,7 +194,26 @@ caf_core_set_default_properties(libcaf_core_obj libcaf_core)
 
 caf_export_and_install_lib(core)
 install(FILES "${CMAKE_BINARY_DIR}/caf/detail/build_config.hpp"
-          DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/caf/detail")
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/caf/detail")
+
+# Check whether we need to link against libatomic.
+if(NOT CMAKE_CROSSCOMPILING)
+  set(snippet "#include <cstdint>
+               #include <atomic>
+               std::atomic<uint64_t> x;
+               int main(int, char**) { return static_cast<int>(x.load()); }")
+  check_cxx_source_compiles("${snippet}" has_64bit_atomic)
+  if(NOT has_64bit_atomic)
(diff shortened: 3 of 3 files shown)


josephnoir (commented):
Looks good! The only comment I have is that the referenced issue only mentions Linux + GCC, does it make sense to do this on all platforms nonetheless?

Neverlord:
> Looks good! The only comment I have is that the referenced issue only mentions Linux + GCC, does it make sense to do this on all platforms nonetheless?

The only other platform we support is MSVC, which doesn't need extra dependencies for standard library features. So it just isn't an issue there.

josephnoir:
I was more thinking of Linux + clang. But now I feel like I missed something.

Neverlord:
Please note that the CMake codes does not check for GCC. If you're using Clang but with GCC's `libstdc++`, you'll have the same issue. The reason is that GCC's standard library implementation does not come with the necessary symbols. It's the same for `std::filesystem`, btw. If you're using MSVC or Clang with `libc++`, you're good. If you're using GCC or Clang with `libstdc++`, you'll have to link against `libatomic` in some cases.

In practice, CAF users run into this on Linux with GCC on 32-bit platforms. There are a couple more edge cases (like using Clang with GCC's standard library implementation), but please let's not overcomplicate the changelog with too much detail.

josephnoir:
Ah, so `libc++` includes the 64 bit atomics and the [check here](https://github.com/actor-framework/actor-framework/pull/1153/files#diff-e273b7c9e9c4e95f44167a54bdc4e0712e714bef8f345dabd1caaf22045b96baR206) will fail?

Neverlord:
> Ah, so `libc++` includes the 64 bit atomics and the [check here](https://github.com/actor-framework/actor-framework/pull/1153/files#diff-e273b7c9e9c4e95f44167a54bdc4e0712e714bef8f345dabd1caaf22045b96baR206) will fail?

Yep, the first `check_cxx_source_compiles` should succeed and set `has_64bit_atomic` to `true` unless symbols for `std::atomic` are missing. AFAIK, only GCC's `libstdc++` requires the extra dependency.

PR #1161, opened 2020-11-13
Fix handling of excluded-components settings

Neverlord:
See CHANGELOG entry:

> CAF 0.18 introduced the option to set different `excluded-components` filters
> for file and console log output. However, CAF rejected all events that matched
> either filter. The new implementation uses the *intersection* of both filters
> to reject log messages immediately (before enqueueing it to the logger's
> queue) and then applies the filters individually when generating file or
> console output.


josephnoir (approved):
Seems straight forwards.

PR #1165, opened 2020-11-19
Fix qt and protobuf examples

### LINKED ISSUE #1156: Qt examples doesn't build in master and 0.18.0-rc.1 tag
# Steps to reproduce
1. git clone https://github.com/actor-framework/actor-framework
2. cd actor-framework
3. git checkout master
4. ./configure --prefix=/usr --enable-qt5-examples
5. cd build
6. make

# Expected behavior
Build should finish with no errors

# Observed behavior
Build fails with error messages:
```

actor-framework/examples/qtsupport/chatwidget.hpp:25:30: error: ‘atom_constant’ in namespace ‘caf’ does not name a template type
   using set_name_atom = caf::atom_constant<caf::atom("setName")>;

actor-framework/examples/qtsupport/qt_group_chat.cpp:77:29: error: ‘atom’ was not declared in this scope
     send_as(client, client, atom("setName"), move(name));
```

Maybe there are other examples which doesn't build now, didn't check yet.
Updates:
- protobuf example have this issue too.
- In Qt example core issues are:
   * usage of old caf::(atom("something")) syntax
   * usage of dynamic dispatch on string as atoms `mb.append(atom_from_string(words.front()))`
   * inconsisten behaviour of split function, due to inconsistent behaviour of caf::span (gcc version has constructor overload for const char * and clang - doesn't - inspect #ifdef CAF_GCC)

RG-707:
Closes #1156

### DIFF
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -81,10 +81,10 @@ void ChatWidget::sendChatMessage() {
           print("*** error: "
                 + QString::fromUtf8(to_string(x.error()).c_str()));
         else
-          self()->send(self(), atom("join"), std::move(*x));
+          self()->send(self(), join_atom_v, std::move(*x));
       },
-      [=](set_name_atom atm, string& name) {
-        send_as(as_actor(), as_actor(), atm, std::move(name));
+      [=](set_name_atom, string& name) {
+        send_as(as_actor(), as_actor(), set_name_atom_v, std::move(name));
       }
     });
     if (! res)
@@ -130,12 +130,12 @@ void ChatWidget::joinGroup() {
   if (! x)
     QMessageBox::critical(this, "Error", to_string(x.error()).c_str());
   else
-    self()->send(self(), join_atom::value, std::move(*x));
+    self()->send(self(), join_atom_v, std::move(*x));
 }
 
 void ChatWidget::changeName() {
   auto name = QInputDialog::getText(this, "Change Name",
                                     "Please enter a new name");
   if (! name.isEmpty())
-    send_as(as_actor(), as_actor(), atom("setName"), name.toUtf8().constData());
+    send_as(as_actor(), as_actor(), set_name_atom_v, name.toUtf8().constData());
 }
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -68,31 +68,22 @@ void ChatWidget::sendChatMessage() {
   if (line.startsWith('/')) {
     vector<string> words;
     split(words, line.midRef(1).toUtf8().constData(), is_any_of(" "));
-    message_builder mb;
-    if (words.size() > 1) {
-      // convert first word to an atom
-      mb.append(atom_from_string(words.front()))
-        .append(words.begin() + 1, words.end());
-    };
-    auto res = mb.apply({
-      [=](join_atom, const string& mod, const string& g) {
-        auto x = system().groups().get(mod, g);
-        if (! x)
-          print("*** error: "
-                + QString::fromUtf8(to_string(x.error()).c_str()));
-        else
-          self()->send(self(), join_atom_v, std::move(*x));
-      },
-      [=](set_name_atom, string& name) {
-        send_as(as_actor(), as_actor(), set_name_atom_v, std::move(name));
+    if (words.size() > 1)
+      if (words.front() == "join" && words.size() == 3) {
+          auto x = system().groups().get(words[1], words[2]);
+          if (!x)
+            print("*** error: "
+                  + QString::fromUtf8(to_string(x.error()).c_str()));
+          else
+            self()->send(self(), join_atom_v, std::move(*x));
+      } else if (words.front() == "setName" && words.size() == 2)
+          send_as(as_actor(), as_actor(), set_name_atom_v, std::move(words[1]));
+      } else {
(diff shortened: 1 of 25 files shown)


examples/qtsupport/chatwidget.cpp
```diff
@@ -68,30 +74,22 @@ void ChatWidget::sendChatMessage() {
...
-      mb.append(atom_from_string(words.front()))
-        .append(words.begin() + 1, words.end());
-    };
-    auto res = mb.apply({
-      [=](join_atom, const string& mod, const string& g) {
-        auto x = system().groups().get(mod, g);
-        if (! x)
+      if (words.front() == "join" && words.size() == 3) {
+        auto x = system().groups().get(words[1], words[2]);
+        if (!x)
```
Neverlord, line 80:
```suggestion
        if (auto x = system().groups().get(words[1], words[2]))
```

And don't forget to switch the bodies for `if` and `else`. Generally, I think putting the "good" case first makes the code easier to follow.

examples/qtsupport/chatwidget.cpp
```diff
@@ -68,30 +74,22 @@ void ChatWidget::sendChatMessage() {
...
         else
-          self()->send(self(), atom("join"), std::move(*x));
-      },
-      [=](set_name_atom atm, string& name) {
-        send_as(as_actor(), as_actor(), atm, std::move(name));
-      }
-    });
-    if (! res)
+          self()->send(self(), join_atom_v, std::move(*x));
+      } else if (words.front() == "setName" && words.size() == 2)
```
Neverlord, line 85:
```suggestion
      } else if (words[0] == "setName" && words.size() == 2)
```

examples/qtsupport/chatwidget.hpp
```diff
@@ -11,6 +11,13 @@ CAF_PUSH_WARNINGS
 #include <QTextEdit>
 CAF_POP_WARNINGS
 
+CAF_BEGIN_TYPE_ID_BLOCK(qt_support, first_custom_type_id)
```
Neverlord, line 14:
```suggestion
CAF_BEGIN_TYPE_ID_BLOCK(qtsupport, first_custom_type_id)
```

I find it confusing to introduce an underscore here if the example directory does not.

examples/qtsupport/qt_group_chat.cpp
```diff
@@ -4,9 +4,9 @@
  * terminal version in remote_actors/group_chat.cpp.                          *
  *                                                                            *
  * Setup for a minimal chat between "alice" and "bob":                        *
- * - ./build/bin/group_server -p 4242                                         *
- * - ./build/bin/qt_group_chat -g remote:chatroom@localhost:4242 -n alice     *
- * - ./build/bin/qt_group_chat -g remote:chatroom@localhost:4242 -n bob       *
+ * - ./build/examples/group_server -p 4242                                    *
+ * - ./build/examples/qt_group_chat -g remote:chatroom@localhost:4242 -n alice*
+ * - ./build/examples/qt_group_chat -g remote:chatroom@localhost:4242 -n bob  *
```
Neverlord, line 9: Please drop the paths here as well.

examples/qtsupport/qt_group_chat.cpp
```diff
@@ -26,57 +26,62 @@ CAF_PUSH_WARNINGS
...
-  auto name = cfg.name;
+  init_global_meta_objects<id_block::qt_support>();
+  core::init_global_meta_objects();
+  io::middleman::init_global_meta_objects();
+  config cfg{};
+  auto err = cfg.parse(argc, argv);
+  if (cfg.cli_helptext_printed)
+    return 0;
+  cfg.load<io::middleman>();
+  actor_system sys{cfg};
```
Neverlord, line 52: Please use the `CAF_MAIN` scaffold here instead of making all of these steps by hand. This example simple hasn't been ported to it yet.

examples/qtsupport/qt_group_chat.cpp
```diff
@@ -26,57 +26,62 @@ CAF_PUSH_WARNINGS
...
+  }
   group grp;
   // evaluate group parameters
-  if (! cfg.group_id.empty()) {
-    auto p = cfg.group_id.find(':');
-    if (p == std::string::npos) {
-      cerr << "*** error parsing argument " << cfg.group_id
-         << ", expected format: <module_name>:<group_id>";
+  if (auto locator = get_if<std::string>(&cfg, "group")) {
+    if (auto grp_p = sys.groups().get(*locator)) {
```
Neverlord, line 66:
```suggestion
    if (auto grp_p = sys.groups().get(*locator)) {
```

What is the `_p` suffix standing for? In other places, we usually use a `maybe_` prefix for catching the result of something that returns `expected<T>` or `optional<T>`.

Neverlord (changes requested):
Thanks for taking care of this! 🙏

Currently, this does not compile for me (see [fail.log](https://github.com/actor-framework/actor-framework/files/5569421/fail.log)). Also, this isn't covered by our CI. Please make sure that at least one build in our matrix includes the Qt5 examples.

I've also found a couple minor things in the code, mostly stylistically.

examples/qtsupport/qt_group_chat.cpp
RG-707: I will port it to `CAF_MAIN`.

RG-707:
I will look into the build problems and add it to the CI.

PR #1166, opened 2020-11-19
Fix unsafe use of strncmp

### LINKED ISSUE #1164: Undefined behavior in string_view::compare on strncmp call
# Problem
In file function  [string_view::compare](https://github.com/actor-framework/actor-framework/blob/f84a26cf1597995f846cd277dc3e6b9238999d50/libcaf_core/src/string_view.cpp#L104)
When string_view::compare is called and
1. s0 == s1
2. s0 == 0

Then strncmp is called with nullptr as values:
```return strncmp(data(), str.data(), s0);```

And calling [strncmp](https://en.cppreference.com/w/cpp/string/byte/strncmp) with nullptr arguments triggers undefined behaviour:
"The behavior is undefined when either lhs or rhs is the null pointer."

# Steps to reproduce
In order to separate this problem from other memory issues we will use branch with fixed memory leaks: issue/1160 or master if issue/1160 branch will be merged.
- Os: Docker image of Ubuntu 16
- Compiler: gcc 8.4.0

1. git clone https://github.com/actor-framework/actor-framework
2. cd actor-framework/
3. git checkout issue/1160
4. CXXFLAGS="-fuse-ld=gold -fno-sanitize-recover=undefined" ./configure --dev-mode
5. cd build/
6. cmake --build . 
7. ctest

# Expected behavior
All test should pass with no errors

# Observed behavior
4 tests failing:
```
The following tests FAILED:
	 30 - detail.config_consumer (Failed)
	 36 - detail.meta_object (Failed)
	107 - string_view (Failed)
	124 - uri (Failed)
```

Each of them points to the same root of problem, for example
```
ctest -VV -R "uri"
```
gives output
```
UpdateCTestConfiguration  from :/tmp/actor-framework/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/tmp/actor-framework/build/DartConfiguration.tcl
Test project /tmp/actor-framework/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 124
    Start 124: uri

124: Test command: /tmp/actor-framework/build/libcaf_core/caf-core-test "-r300" "-n" "-v5" "-s"^uri$""
124: Test timeout computed to be: 10000000
124: +----------------------------------------------------------------------+
124:                                   uri
124: +----------------------------------------------------------------------+
124: 
124: - default URIs are empty
124: ** /tmp/actor-framework/libcaf_core/test/uri.cpp:223  ::caf::test::equal_to(x.empty(), true)
124: /tmp/actor-framework/libcaf_core/src/string_view.cpp:111:19: runtime error: null pointer passed as argument 1, which is declared to never be null
1/1 Test #124: uri ..............................***Failed    0.06 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.07 sec

The following tests FAILED:
	124 - uri (Failed)
Errors while running CTest
```
So [actor-framework/libcaf_core/src/string_view.cpp:111:19](https://github.com/actor-framework/actor-framework/blob/f84a26cf1597995f846cd277dc3e6b9238999d50/libcaf_core/src/string_view.cpp#L111)

# Thoughts
I think it would be great if -fno-sanitize-recover=undefined flag and ubsan in general would be added at least for some compiler/targes into CI system, to prevent regression in the future.

### PR DESCRIPTION
**Pull request by Neverlord**
Closes #1164 and also adds an UBSan check to our CI pipeline.

### DIFF
--- a/.ci/run.sh
+++ b/.ci/run.sh
@@ -28,8 +28,9 @@ elif [ $# = 2 ]; then
   if [ "$1" = 'test' ] && [ -d "$2" ]; then
     Mode='test'
     BuildDir="$2"
-  elif [ "$1" = 'assert' ] && [ "$2" = 'LeakSanitizer' ]; then
+  elif [ "$1" = 'assert' ]; then
     Mode='assert'
+    What="$2"
   else
     usage
   fi
@@ -85,10 +86,37 @@ runLeakSanitizerCheck() {
   fi
 }
 
+runUBSanitizerCheck() {
+  UBSanCheckStr="
+    int main(int argc, char**) {
+      int k = 0x7fffffff;
+      k += argc;
+      return 0;
+    }
+  "
+  echo "${UBSanCheckStr}" > UBSanCheck.cpp
+  c++ UBSanCheck.cpp -o UBSanCheck -fsanitize=undefined -fno-omit-frame-pointer
+  out=`./UBSanCheck 2>&1 | grep -o 'signed integer overflow'`
+  if [ -z "$out" ]; then
+    echo "unable to detected undefined behavior on this platform!"
+    return 1
+  fi
+}
+
 if [ "$Mode" = 'build' ]; then
   runBuild
 elif [ "$Mode" = 'test' ]; then
   runTest
 else
-  runLeakSanitizerCheck
+  case "$What" in
+    LeakSanitizer)
+      runLeakSanitizerCheck
+      ;;
+    UBSanitizer)
+      runUBSanitizerCheck
+      ;;
+    *)
+    echo "unrecognized tag: $What!"
+    return 1
+  esac
 fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   console output.
 - Fix memory leaks when deserializing URIs and when detaching the content of
   messages (#1160).
+- Fix undefined behavior in `string_view::compare` (#1164).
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by josephnoir — COMMENTED** [review]
Looks good. I checked the `string_view` test and it looks like we don't have a test case that validates the return values for strings of unequal length. I think we should add one.

I added this to my local test to check. It succeeds.
```
string_view a = "cdef";              
CAF_CHECK(z.compare(a) < 0);          
CAF_CHECK(a.compare(z) > 0); 
```

However, while reading the test I noticed the following line. It checks *substrings* of unequal length and and finds them equal. Is that supposed to happen?
```
CAF_CHECK(x.compare(1, 2, y, 0, 2) == 0);
```

And the check below that one appears twice.
```
CAF_CHECK(x.compare(2, 1, z, 0, 1) == 0);    
CAF_CHECK(x.compare(2, 1, z, 0, 1) == 0); 
```

Neverlord:
Thanks for the review! I'll make sure to add the new test before merge (and remove the duplicate).

> However, while reading the test I noticed the following line. It checks _substrings_ of unequal length and and finds them equal. Is that supposed to happen?
> 
> ```
> CAF_CHECK(x.compare(1, 2, y, 0, 2) == 0);
> ```

This checks equal lengths (substrings of size 2). The API is `compare(size_type pos1, size_type count1, basic_string_view v, size_type pos2, size_type count2)`.

josephnoir:
Ah, sorry. The second argument is a length, not an index. 👍

PR #1179, opened 2020-11-23
Add new capabilities for BDD-style unit tests

Neverlord:
We've recently added a couple BDD-style unit tests to the code base. While I find this style generally working well for structuring tests, our current unit test macros add a lot of noise to these tests.

The new `SCENARIO`, `GIVEN`, `WHEN`, `THEN` macros cut down that noise and make the tests much more readable. I've also added short versions for our check and require macros. Since these additions could break client code, the new BDD-style macros go into a new header. There is of course more we could do (like nicely indenting unit test output based on scenario/given/when/then blocks) but this lets us reap 80% of the benefits with 20% effort right away. 🙂

### DIFF
--- a/libcaf_core/test/broadcast_downstream_manager.cpp
+++ b/libcaf_core/test/broadcast_downstream_manager.cpp
@@ -342,7 +342,7 @@ receive_checker<F> operator<<(receive_checker<F> xs, not_empty_t) {
 
 #define BATCH(first, last) make_batch(first, last)
 
-#define AND <<
+#define AND_RECEIVED <<
 
 // -- unit tests ---------------------------------------------------------------
 
@@ -367,7 +367,7 @@ CAF_TEST(one_path_force) {
   }
   // Give 11 credit (more than 10).
   AFTER ENTITY alice TRIED FORCE_SENDING 11 ELEMENTS {
-    ENTITY bob RECEIVED BATCH(14, 23) AND BATCH(24, 24);
+    ENTITY bob RECEIVED BATCH(14, 23) AND_RECEIVED BATCH(24, 24);
     ENTITY alice HAS 0u CREDIT FOR bob;
   }
   // Drain all elements except the last 5.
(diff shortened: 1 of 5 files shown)


josephnoir (commented):
Looks good. I think this makes the tests easier to read. I'm wondering if we should do something about the output formatting to visually separate the test cases in a scenario. At the moment, everything in a scenario has the same indentation:

```
- SCENARIO tunnels automatically unsubscribe from their origin
  -> GIVEN a group with two local subscribers locally and tunneled [line 197]
  -> WHEN the first actor leaves the tunnel [line 203]
  -> THEN no messaging occurs [line 205]
  -> WHEN an actors sends to the group after the unsubscribe [line 209]
  -> THEN the unsubscribed actor no longer receives the message [line 211]
  -> expect(put_atom, int).from(self).to(t1).with(_, 42) [line 212]
  -> expect(put_atom, int).from(self).to(t2).with(_, 42) [line 213]
  -> expect(put_atom, int).from(self).to(worker).with(_, 42) [line 214]
  -> disallow(put_atom, int).from(self).to(t3).with(_, 42) [line 215]
  -> expect(put_atom, int).from(self).to(t4).with(_, 42) [line 216]
  -> WHEN the second actor also unsubscribes from the tunnel [line 220]
  -> THEN the tunnel unsubscribes from its origin [line 222]
  -> expect(sys_atom, leave_atom).to(worker) [line 223]
  -> expect(leave_atom, strong_actor_ptr).from(worker).to(intermediary).with(_, worker) [line 225]
  -> WHEN an actors sends to the group after the tunnel left [line 228]
  -> THEN no message arrives at the tunnel [line 230]
  -> expect(put_atom, int).from(self).to(t1).with(_, 42) [line 231]
  -> expect(put_atom, int).from(self).to(t2).with(_, 42) [line 232]
  -> disallow(put_atom, int).from(self).to(worker).with(_, 42) [line 233]
  -> 40 checks took 2 ms
```

Neverlord:
> I'm wondering if we should do something about the output formatting to visually separate the test cases in a scenario.

I think we definitely should go there eventually. However, the current unit testing framework really shows its age by now and many features were bolted on. Let's stick to the 20% effort solution for now and then we'll either revisit the unit testing framework as a whole at some point or (what I'm leaning towards) make a clean from-scratch design and deprecate the old stuff in some future CAF.

PR #1180, opened 2020-11-23
Tidy up CMake setup, propagate flags via targets

Neverlord:
- Make defining CAF components easier with the new `caf_add_component` function.
- Use a central `caf_interal` target for propagating flags.
- Be a good citizen and don't globally set CMake policies (use push/pop).


CMakeLists.txt
```diff
@@ -77,67 +80,26 @@ if(MSVC AND CAF_SANITIZERS)
   message(FATAL_ERROR "Sanitizer builds are currently not supported on MSVC")
 endif()
 
-# -- compiler setup ------------------------------------------------------------
+# -- base target setup ---------------------------------------------------------
+
+# This target propagates compiler flags, extra dependencies, etc. All other CAF
+# targets pull this target in as PRIVATE dependency. Users that embed CAF into
+# their own CMake scaffold (e.g., via FetchContent) may pass this target in with
+# some properties predefined in order to force compiler flags or dependencies.
```
josephnoir, line 88: If this is a recommendation we should document it somewhere outside of this comment.

CMakeLists.txt
```diff
@@ -1,13 +1,16 @@
-cmake_minimum_required(VERSION 3.4 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.5...3.18 FATAL_ERROR)
```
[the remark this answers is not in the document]
Neverlord: This is the same version range we use for the policies. Basically, I've tested that setting `NEW` for all policies introduced up until 3.18 work and are anticipated by our CMake scaffold but going back as far as 3.5 still gives a valid build.

CMakeLists.txt
Neverlord: Good point. I think this should go into a cohesive CMake/build guide, though. The wiki might be a good place for that.

josephnoir: This does not stand in the way of merging this PR right?

josephnoir (approved):

CMakeLists.txt
Neverlord: Don't think so. 😊

PR #1183, opened 2020-12-08
Align typed_response_promise with response_promise

dominiklohmann:
This aligns the functionality of `typed_response_promise` with the functionality of `response_promise`.

I am in the process of migrating some code to use typed actors and response promises, and found `static_cast<response_promise&>(typed_rp)` quite inconvenient to use.

### DIFF
--- a/libcaf_core/caf/response_promise.hpp
+++ b/libcaf_core/caf/response_promise.hpp
@@ -53,6 +53,8 @@ public:
   response_promise& operator=(response_promise&&) = default;
   response_promise& operator=(const response_promise&) = default;
 
+  virtual ~response_promise() = default;
+
   /// Satisfies the promise by sending a non-error response message.
   template <class T, class... Ts>
   detail::enable_if_t<((sizeof...(Ts) > 0)
@@ -73,12 +75,20 @@ public:
         make_message(std::forward<T>(x), std::forward<Ts>(xs)...));
   }
 
+  /// Satisfies the promise by sending an error or non-error response message.
   template <class T>
   void deliver(expected<T> x) {
     if (x)
       return deliver(std::move(*x));
     return deliver(std::move(x.error()));
   }
+
+  /// Satisfies the promise by sending an error response message.
+  void deliver(error x);
+
+  /// Satisfies the promise by sending an empty message if this promise has a
+  /// valid message ID, i.e., `async() == false`.
+  void deliver(unit_t x);
 
   /// Satisfies the promise by delegating to another actor.
   template <message_priority P = message_priority::normal, class Handle = actor,
@@ -103,13 +113,6 @@ public:
                     make_message(std::forward<Ts>(xs)...));
     return {};
   }
-
-  /// Satisfies the promise by sending an error response message.
-  void deliver(error x);
-
-  /// Satisfies the promise by sending an empty message if this promise has a
-  /// valid message ID, i.e., `async() == false`.
-  void deliver(unit_t x);
 
   /// Returns whether this response promise replies to an asynchronous message.
   bool async() const;
--- a/libcaf_core/caf/typed_response_promise.hpp
+++ b/libcaf_core/caf/typed_response_promise.hpp
@@ -28,13 +28,15 @@ namespace caf {
 /// response message from the server (i.e. receiver of the request)
 /// to the client (i.e. the sender of the request).
 template <class... Ts>
-class typed_response_promise {
+class typed_response_promise : public response_promise {
 public:
+  using super = response_promise;
+
   /// Constructs an invalid response promise.
   typed_response_promise() = default;
 
   typed_response_promise(strong_actor_ptr self, mailbox_element& src)
-    : promise_(std::move(self), src) {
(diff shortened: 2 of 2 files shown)


Neverlord:
This causes a is-a relation between the two promise types I'd rather not introduce. I also don't recall why the implicit conversion operator exists, but it looks fishy (like a hack we've forgotten to clean up) TBH.

Aside from syntax, what is the actual use case you have where this gets in the way? Would an `untyped` member function for the conversion be an agreeable alternative?

dominiklohmann:
If you compare the interfaces of `typed_response_promise<Ts...>` and  `response_promise`, you'll see that the latter has a few utilities that the former does not provide. Most notably, `source()`, `stages()`, `next()`, `id()`, `deliver(unit_t)`, and `deliver(expected<T>)`.

Unless you want to keep the APIs in sync manually, or it is by design that a typed response promise cannot be used like a response promise without hassle on the user side, inheritance surely is the way to go. What did you have in mind?

Side note: I noticed only after creating this PR that `typed_response_promise<T>::delegate(...)` does not work when sending to a `typed_actor<Ts...>::pointer` or `typed_actor<Ts...>::stateful_pointer<State>`, which is a common case when delegating to self. If you want to go a different route with fixing these inconsistencies, maybe you can fix that as well while you're at it.

Neverlord:
Ok, duly noted that there are inconsistencies between the promise types.

> If you want to go a different route with fixing these inconsistencies, maybe you can fix that as well while you're at it.

If find it a tad sad that your approach to contributing to open source projects seems to be "take it or leave it". But that is of course your choice!

Contributions are welcome, but the bar is high. Changes have to fit into the overall design of the framework and should move the code into the right direction. CAF evolved over a long time span (since 2011) and initially looked quite differenlty. I believe discussing problems and mapping out the solution space should be part of the contribution and vetting process to make sure new code additions not eventually become a burden by themselves. In particular when touching central pieces of the messaging layer.

dominiklohmann:
> If find it a tad sad that your approach to contributing to open source projects seems to be "take it or leave it". But that is of course your choice!

Sorry if I came across that way. Let's drop the heat and get back to discussing the problem space. Generally, I am always open to adapting this as you deem okay—it's your code base after all—, and what I wanted to bring across was that I think that my solution is a good solution for this problem.

Neverlord:
> Sorry if I came across that way. Let's drop the heat and get back to discussing the problem space.

No worries, I can understand it if you had something locally that seemed to work and simply wanted to "push it upstream". That works fine for small tweaks and straightforward fixes, but things have more complexity to them than it may seem at a first glance more often that not.

The typed version should not inherit from the untyped version for the same reason `typed_actor` does not inherit from `actor`: one models dynamically typed communication and the other models statically typed communication. You can't be both and they should not convert between one another without explicit "I-know-what-I'm-doing" input from the developer, e.g., via `actor_cast` (which is why I'm quite irritated by the implicit conversion operator, but that one gets deprecated with 0.18 no matter what now).

I see three options here that I would happily merge if you're up to it:
1. Only add a member function to give you the wrapped promise. That let's you at least replace the `static_cast` with something less verbose. Haven't fully made up my mind yet whether this should be a member function or a free function like `promise_cast` (which then could convert both ways).
2. Add the missing utility functions to the typed promise, always delegating to the wrapped promise. Won't result in particular pretty code, but at least gets the job done and plays nice in templatized code. Doesn't change anything fundamental about the code, so it just boils down to some grunt work. The benefit here is the very low risk additional code introduces, so it's also quick to review.
3. Factor out state and all functions except the non-error `deliver` from `response_promise` to a new class `response_promise_base`. Then, both the typed and the untyped version can extend that type and add their `deliver` functions. This is the cleanest solution, but may also have some repercussions while implementing it. I didn't scan the code base for call sites that would require cleanup (or dig into why the implicit conversion operator exists). The conversion operator also can no longer return a reference, so it must return a value instead which may potentially break code (still prefer adding the conversion back in as a deprecated function rather than removing it outright). I'd be willing to go that route, but it certainly represents the most work.

dominiklohmann:
I'd probably go for option (2) if I were to implement this. For (1) I honestly do not know enough of the internals to do the proper type checking, and (3) I think would be implemented faster by yourself.

The deprecation sounds like a very good idea.

I cannot re-open this PR since I did not close it myself. It also might take me a few days to find the time to work on this.

Neverlord:
> I'd probably go for option (2) if I were to implement this.

Yeah, probably the safest short-term solution.

> I cannot re-open this PR since I did not close it myself. It also might take me a few days to find the time to work on this.

Whatever your time / priorities allow. If you're going to tackle this, just push to your branch and give me a ping when it's ready for reopen/review. 🙂

dominiklohmann:
I implemented it accordingly @Neverlord. CAF master compiles without triggering the added deprecation warning.

Neverlord:
Thanks!

Apparently, I can't re-open PRs if the branch has been force-pushed, though. Sorry for the inconvenience, but seems like you need to open another PR.

PR #1187, opened 2020-12-12
Use CMAKE_INSTALL_FULL_LIBDIR to install CMake files

jsiwek:
CMAKE_INSTALL_LIBDIR is underspecified and may be set to either a
relative or absolute path by an upstream project or user.  In the case
it's an absolute path, the install path for CAF's CMake files
incorrectly prepend CMAKE_INSTALL_PREFIX.

CMAKE_INSTALL_FULL_LIBDIR is specified to be an absolute path, with
CMAKE_INSTALL_PREFIX prepended if needed, so using it is preferable.


Neverlord:
Thanks a lot for the fix! 🙏

PR #1190, opened 2020-12-13
Align typed_response_promise with response_promise

dominiklohmann:
This aligns the functionality of `typed_response_promise` with the functionality of `response_promise`.

### DIFF
--- a/libcaf_core/caf/response_promise.hpp
+++ b/libcaf_core/caf/response_promise.hpp
@@ -73,7 +73,7 @@ public:
         make_message(std::forward<T>(x), std::forward<Ts>(xs)...));
   }
 
-  /// Satisfies the promise by sending either an error or a non-erorr response
+  /// Satisfies the promise by sending either an error or a non-error response
   /// message.
   template <class T>
   void deliver(expected<T> x) {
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/response_promise.hpp
```diff
@@ -105,6 +106,7 @@ class CAF_CORE_EXPORT response_promise {
   }
 
   /// Satisfies the promise by sending an error response message.
+  /// For non-requests, nothing is done.
```
Neverlord, line 109: This isn't true. It would mean that CAF silently drops errors instead of propagating them for asynchronous messages.

```suggestion
```

libcaf_core/caf/typed_response_promise.hpp
```diff
@@ -77,11 +100,43 @@ class typed_response_promise {
     return *this;
   }
 
+  /// Satisfies the promise by sending an empty message if this promise has a
+  /// valid message ID, i.e., `async() == false`.
+  void deliver(unit_t x) {
+    promise_.deliver(x);
+  }
+
```
Neverlord, line 108:
```suggestion
```

An actor that promises a tuple of type `Ts...` must not send an empty message instead. For "void" promises, `typed_actor_promise<unit_t>` should do the trick IIRC.

Neverlord (changes requested):
Thank you for tackling this!

LGTM, except a couple lines that should get removed.

Neverlord (approved):
Thanks! 👍

PR #1192, opened 2020-12-17
Add new get_as and get_or utility functions

Neverlord:
@josephnoir as discussed, here's the first chunk for the new config-value API overhaul. 🙂

After integrating that, I'm also going to base the `get_or` in `settings.hpp` on this new code as well as replacing the custom parsing that's currently going on in the config options.


libcaf_core/test/config_value.cpp
```diff
@@ -81,13 +86,739 @@ config_value cfg_lst(Ts&&... xs) {
...
+  }
+  GIVEN("config_values of null, URI, boolean, list or dictionary") {
+    WHEN("using get_as with floating point types") {
+      THEN("conversion fails") {
+        CHECK_EQ(get_as<int64_t>(cv_null), sec::conversion_failed);
+        CHECK_EQ(get_as<int64_t>(cv_true), sec::conversion_failed);
+        CHECK_EQ(get_as<int64_t>(cv_false), sec::conversion_failed);
+        CHECK_EQ(get_as<int64_t>(cv_empty_uri), sec::conversion_failed);
+        CHECK_EQ(get_as<int64_t>(cv_empty_list), sec::conversion_failed);
+        CHECK_EQ(get_as<int64_t>(cv_empty_dict), sec::conversion_failed);
```
josephnoir, line 306: Not a floating point type.

libcaf_core/test/config_value.cpp
```diff
@@ -81,13 +86,739 @@ config_value cfg_lst(Ts&&... xs) {
...
+      }
+    }
+  }
+}
+
+SCENARIO("get_as can convert config values to timespans") {
+  using namespace std::chrono_literals;
+  GIVEN("a config value with value 42s") {
```
josephnoir, line 314:
```suggestion
  GIVEN("a config value of 42s") {
```

libcaf_core/test/config_value.cpp
```diff
@@ -81,13 +86,739 @@ config_value cfg_lst(Ts&&... xs) {
...
+      THEN("conversion succeeds") {
+        CHECK_EQ(get_as<timespan>(x), timespan{42s});
+        CHECK_EQ(get_as<std::string>(x), "42s");
+      }
+    }
+    WHEN("using get_as with type other than timespan or string") {
+      THEN("conversion fails") {
+        CHECK_EQ(get_as<int64_t>(x), sec::conversion_failed);
+        CHECK_EQ(get_as<double>(x), sec::conversion_failed);
+        // CHECK_EQ(get_as<uri>(x), sec::conversion_failed);
```
josephnoir, line 326: Hm?

libcaf_core/test/config_value.cpp
```diff
@@ -81,13 +86,739 @@ config_value cfg_lst(Ts&&... xs) {
...
+    WHEN("using get_as with string") {
+      THEN("CAF renders the value as string") {
+        CHECK_EQ(get_as<string>(cv_null), "null");
+        CHECK_EQ(get_as<string>(cv_true), "true");
+        CHECK_EQ(get_as<string>(cv_false), "false");
+        CHECK_EQ(get_as<string>(cv_empty_list), "[]");
+        CHECK_EQ(get_as<string>(cv_empty_dict), "{}");
+        CHECK_EQ(get_as<string>(config_value{42}), "42");
+        CHECK_EQ(get_as<string>(config_value{4.2}), "4.2");
+        CHECK_EQ(get_as<string>(config_value{timespan{4}}), "4ns");
```
josephnoir, line 346: Since there is no other test case, just curious: Would `timespan{4s}` (assuming `using namespace std::chrono_literals;`) return a nanosecond resolution as well?

libcaf_core/src/config_value.cpp
```diff
@@ -126,6 +153,312 @@ const char* config_value::type_name_at_index(size_t index) noexcept {
...
+
+expected<timespan> config_value::to_timespan() const {
+  using result_type = expected<timespan>;
+  auto f = detail::make_overload(
+    no_conversions<timespan, none_t, bool, integer, real, uri,
+                   config_value::list, config_value::dictionary>(),
+    [](timespan x) {
+      // This cast may lose precision on the value. We could try and check that,
+      // but refusing to convert on loss of precision could also be unexpected
+      // behavior. So we rather always convert, even if it costs precision.
```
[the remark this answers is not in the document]
Neverlord: Oops. Copy & paste error. 🙈

libcaf_core/test/config_value.cpp
Neverlord: No, that would be silly. Our timestamp has nanosecond resolution, so using this constructor gives you the default resolution.

libcaf_core/src/config_value.cpp
```diff
@@ -126,6 +153,312 @@ const char* config_value::type_name_at_index(size_t index) noexcept {
...
+      if (detail::parse(x, dict, detail::require_opening_char) == none) {
+        tmp.clear();
+        for (const auto& [key, val] : dict) {
+          list kvp;
+          kvp.reserve(2);
+          kvp.emplace_back(key);
+          kvp.emplace_back(val);
+          tmp.emplace_back(std::move(kvp));
+        }
+        return result_type{std::move(tmp)};
```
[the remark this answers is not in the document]
Neverlord: Calling the function recursively only to avoid some code repetition feels a bit heavy handed to me. I moved the repeated code into a lambda instead.

libcaf_core/src/config_value.cpp
```diff
@@ -126,6 +153,312 @@ const char* config_value::type_name_at_index(size_t index) noexcept {
...
+        ptr->inc_constructed_elements();
+        if (!meta->load(reader, pos))
+          return false;
+        pos += meta->padded_size;
+      }
+      result.reset(ptr.release(), false);
+      return reader.end_sequence();
+    };
+    if (std::any_of(allowed_types.begin(), allowed_types.end(), converts))
+      return {std::move(result)};
```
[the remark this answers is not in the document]
Neverlord: I think warnings should stay reserved for functions with non-obvious pitfalls (like requiring a certain mutex locked for example). FWIW, I think it's obvious that this message needs to perform several memory allocations, otherwise it couldn't return a `message` and the documentation already mentions that it attempts to parse a config-value list.

libcaf_core/src/config_value_reader.cpp
```diff
@@ -563,15 +593,14 @@ bool config_value_reader::value(span<byte> bytes) {
...
   if (auto str = get_if<std::string>(obj, "@type"); str == nullptr) {
-    emplace_error(sec::runtime_error,
-                  "cannot fetch object type: no '@type' entry found");
-    return false;
-  } else if (auto id = query_type_id(*str); id == invalid_type_id) {
-    emplace_error(sec::runtime_error, "no such type: " + *str);
-    return false;
-  } else {
+    type = type_id_v<config_value::dictionary>;
+    return true;
```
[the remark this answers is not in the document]
Neverlord: What are you referring to as "this"? The fact that we return the type ID for `dictionary` here? That's simply the type ID of the argument `obj`.

libcaf_core/test/config_value.cpp
Neverlord: Correct. 🧐

libcaf_core/caf/binary_deserializer.hpp
```diff
@@ -35,7 +35,8 @@
 
 namespace caf {
 
-/// Deserializes objects from sequence of bytes.
+/// Deserializes objects from sequence of bytes. Does not perform run-time type
```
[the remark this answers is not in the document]
Neverlord: Yep, good catch! 🙂

libcaf_core/test/config_value.cpp
Neverlord: Probably some leftover from the prototyping phase. I've added it back in.

libcaf_core/src/config_value_reader.cpp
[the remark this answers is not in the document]
josephnoir: Yes. I don't find it obvious that the check in the `if` statement leads to a `dictionary` type.

libcaf_core/test/config_value.cpp
josephnoir: I think the we check twice now that the conversion to `int64_t` fails. I should have been more clear in my first comment on that, sorry. The first block that checks this is in lines 253 to 265. The comment was probably correct, but the conversion type should be different?

Neverlord: Yeah, looks like it. Thanks for the extra context, this makes it clear now.

libcaf_core/src/config_value_reader.cpp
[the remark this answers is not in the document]
Neverlord: Ok, I've added a comment to `config_value_reader.hpp`.

PR #1193, opened 2020-12-20
Check whether CMAKE_INSTALL_PREFIX is absolute

dominiklohmann:
CAF uses `CMAKE_INSTALL_FULL_<GNUInstallDir>` instead of `CMAKE_INSTALL_<GNUInstallDir>`, which by default are the latter prefixed with `CMAKE_INSTALL_PREFIX`. However, there is no further checking in place whether the resulting `CMAKE_INSTALL_FULL_<GNUInstallDir>` is absolute, and if it is not, then using it as a destination in `install(... DESTINATION ${CMAKE_INSTALL_FULL_<GNUInstallDir>})` will cause it to be prefixed with `CMAKE_INSTALL_PREFIX` again.

This PR adds a sanity check whether `CMAKE_INSTALL_PREFIX` is absolute.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,10 @@ if(MSVC AND CAF_SANITIZERS)
   message(FATAL_ERROR "Sanitizer builds are currently not supported on MSVC")
 endif()
 
+if (NOT IS_ABSOLUTE "${CMAKE_INSTALL_PREFIX}")
+  message(FATAL_ERROR "CMAKE_INSTALL_PREFIX must be absolute")
+endif ()
+
 # -- base target setup ---------------------------------------------------------
 
 # This target propagates compiler flags, extra dependencies, etc. All other CAF


Neverlord:
I'm not quite sure what to make of this. The [docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#special-cases) have this to say (emphasis mine):

> `CMAKE_INSTALL_FULL_<dir>`
>
>    **The absolute path** generated from the corresponding `CMAKE_INSTALL_<dir>` value. If the value is not already an absolute path, an absolute path is constructed typically by prepending the value of the `CMAKE_INSTALL_PREFIX` variable. However, there are some special cases as documented below.

The special cases boil down to `/`, `/usr` and `/opt`. So if `CMAKE_INSTALL_FULL_<dir>` ever results in a relative path for you, wouldn't that be a CMake bug?

Restricting `CMAKE_INSTALL_PREFIX` does not feel like a good thing to do even if `CMAKE_INSTALL_FULL_<dir>` is buggy. `CAF_INSTALL_CMAKEDIR` does not specify how (and if) it treats relative paths yet. So we could just specify that it treats relative paths relative to `CMAKE_INSTALL_PREFIX` and always convert it to an absolute path before passing it to `install` using this simple rule. That would mean reverting the previous patch (using `CMAKE_INSTALL_LIBDIR` again) and creating an absolute path (e.g., `CAF_INSTALL_CMAKEDIR_ABSOLUTE`) that we pass to `install` exclusively. Does this makes sense or am I missing something?

dominiklohmann:
> That would mean reverting the previous patch (using `CMAKE_INSTALL_LIBDIR` again) and creating an absolute path (e.g., `CAF_INSTALL_CMAKEDIR_ABSOLUTE`) that we pass to `install` exclusively. Does this makes sense or am I missing something?

I think the calls to `install` should just use `CMAKE_INSTALL_LIBDIR` over `CMAKE_INSTALL_FULL_LIBDIR`, then this wouldn't be an issue at all. There is no need to specify these as absolute.

Additionally, CPack does not package files that are installed using absolute paths, so said previous patch broke the `package` target silently.

Neverlord:
> > That would mean reverting the previous patch (using `CMAKE_INSTALL_LIBDIR` again) and creating an absolute path (e.g., `CAF_INSTALL_CMAKEDIR_ABSOLUTE`) that we pass to `install` exclusively. Does this makes sense or am I missing something?
> 
> I think the calls to `install` should just use `CMAKE_INSTALL_LIBDIR` over `CMAKE_INSTALL_FULL_LIBDIR`, then this wouldn't be an issue at all. There is no need to specify these as absolute.
> 
> Additionally, CPack does not package files that are installed using absolute paths, so said previous patch broke the `package` target silently.

Sounds like the solution should be stripping `CAF_INSTALL_CMAKEDIR` altogether then? At least as a configurable cache variable. Passing `${CMAKE_INSTALL_LIBDIR}/cmake/CAF` to `install` would mean it's no longer customizable, but if CPack doesn't support this anyway then there's no point in doing it.

dominiklohmann:
I agree with using a fixed `${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}`. You might also want to take a look at how [PackageProject.cmake](https://github.com/TheLartians/PackageProject.cmake/blob/master/CMakeLists.txt) does this, which I found to be excellent for simple toy libraries.

jsiwek:
Just want to point out a potential pitfall: as already mentioned, the CMake docs do specify `CMAKE_INSTALL_FULL_<dir>` as (supposedly) being an absolute path, but the docs under-specify whether `CMAKE_INSTALL_<dir>` is absolute or relative and this ambiguity was at least a historical issue for others.  Example reference, https://bugzilla.redhat.com/show_bug.cgi?id=795542, has this suggestion:

> And if someone really needs a relative path, the only portable way I know to achieve it until all parties are convinced that CMAKE_INSTALL_LIBDIR should be always relative, is:
>
> file (RELATIVE_PATH CMAKE_INSTALL_RELATIVE_LIBDIR "${CMAKE_INSTALL_PREFIX}" "${CMAKE_INSTALL_FULL_LIBDIR}")

But that's still assuming `CMAKE_INSTALL_FULL_<dir>`is correctly set as absolute-path, so not sure what to ultimately suggest.  Maybe:

* It's understandable how, for compat reasons, the underspecification of `CMAKE_INSTALL_<dir>` still exists
* I'm skeptical that CMake wouldn't be receptive to making changes/fixes for bugs that contradict what they have specified for `CMAKE_INSTALL_FULL_<dir>`
* If there are actually bugs that make `CMAKE_INSTALL_FULL_<dir>` ambiguous and won't be resolved in reasonable timeframe, then consumers of this CMake API that seek perfection likely have to implement the absolute vs. relative path sanity checks themselves everywhere.

Neverlord:
@jsiwek thanks for pointing this out. I feel like this entire installation / packaging process really is more complicated than it ought to be. 🙄

In any case, we currently export our targets like this:

```
  install(TARGETS libcaf_${component}
          EXPORT CAFTargets
          ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT ${component}
          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT ${component}
          LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT ${component})
```

So installing the package files to `${CMAKE_INSTALL_LIBDIR}/cmake/CAF` at least makes it consistent with the rest. I guess we'll wait and see if all users / package maintainers can live with that or if we see complaints at some point. 🤷🏻‍♂️

@dominiklohmann thanks for driving the discussion. Closing here, in favor of dropping `CAF_INSTALL_CMAKEDIR` as discussed.

dominiklohmann:
Thanks for taking care of this!

PR #1198, opened 2021-01-05
Fix handling of expected<T> in response promises

### LINKED ISSUE #1197: Compiler error when resolving a promise with an expected<T>
Compiling the code below results in the following error:
/usr/local/include/caf/typed_response_promise.hpp:84:21: error: return-statement with a value, in function returning ‘void’ [-fpermissive]

```cpp
#include <caf/all.hpp>
struct whatever : caf::event_based_actor {
  whatever(caf::actor_config &cfg): caf::event_based_actor(cfg) {
    auto prom = make_response_promise<int>();
    prom.deliver(caf::expected<int>{42});
  }
};
void caf_main(caf::actor_system as) {}
CAF_MAIN()
```

The cause of the problem is that deliver overload that takes caf::expected returns void but delegates and returns to other "deliver" overloads which however return *this:

```cpp
  template <class T>
  void deliver(expected<T> x) {
    if (x)
      return deliver(std::move(*x));
    return deliver(std::move(x.error()));
  }
```

Neverlord:
Fixes #1197.

I've discovered (and fixed) another bug with the response promises: they must not send empty messages in response to asynchronous messages.

Also, the previous unit tests for the promises were a mess. Or rather: quite dated. So I've ended up implementing a new set of unit tests. The tests for the typed and untyped version are a bit repetitive due to duplicated code, but I think trying to "unify" them would only make the test code worse and less comprehensible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   trying to parse the input of `x` if it contains a string. The function
   `get_or` already existed for `settings`, but we have added new overloads for
   generalizing the function to `config_value` as well.
+- The `typed_response_promise` received additional member functions to mirror
+  the interface of the untyped `response_promise`.
 
 ### Deprecated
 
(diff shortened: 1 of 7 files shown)


josephnoir (commented):
Looks good. Left a comment, but if you don't think it needs addressing, feel free to merge the PR.

Neverlord:
> don't think it needs addressing

Nah, makes sense to dispatch to the regular `deliver` functions in all cases. Also makes the code more robust. 🙂

josephnoir:
Nice :)

Issue #1051, opened 2020-02-14
Unhandled messages sent through the middleman lead to the sender hanging


Serbis:
Hi, I found incorrect behavior when the middleman module was working. Suppose there are two processes A and B. A has two actors A1 and A2. B also has two actors B1 and B2. A1 sends a message to A2, which delegates the received message to B2 through middleman. B2 receiving the message delegates the message to B1. If actor B1 does not have a handler for this message and it falls into the deadletter (in akka terminology), the original sender actor in the chain on side A (A1) hangs and is no longer able to receive any messages.

Neverlord:
Thanks for reporting this issue! From your description, this may also happen without involving the middleman.

tariksmajic:
Hello. This is my understanding concerning this problem. Actor communicate using messages, and when something goes wrong, actors will send error messages to notify their senders that something went wrong. One of those error messages is meant to notify the original sender that the receiving actor didn't handle the message. This happens when a default handler triggers, more specifically caf::print_and_drop or caf::drop handlers. Those handlers return error messages. And the default default handler is caf::print_and_drop.

Upon receiving an error message, actor will trigger its error message handler, which by default calls quit(). Error handle will trigger on the actor that last send the message (called send()), while skipping those who delegated it (called delegate()). 

I tested these situations, recreated your problem and resolved it by implementing a default handler on the receiving actor, or an error handler that doesn't call quit, on the sending actor.

PR #1200, opened 2021-01-12
Detect and report broken response promises

Neverlord:
Relates #1051.

Consider the following scenario: a client sends a request to the server. When receiving the request, the server creates a response promise and sends request to one or more workers. After receiving all results, the server calls `deliver` on its promise. In the current implementation, the client never receives anything if the server dies before receiving all results. With this change, the client would at least receive a `broken_promise` error instead of nothing. The gist of the changes boils down to keeping track of whether a user called `deliver` on a promise before destroying it.

There is a slight complication, though, because we use promises like this:

```c++
adder_actor::behavior_type server_impl(adder_actor::pointer self,
                                       adder_actor worker) {
  return {
    [=](add_atom, int32_t y, int32_t z) {
      auto rp = self->make_response_promise<int32_t>();
      self->request(worker, infinite, add_atom_v, y, z)
        .then([rp](int32_t result) mutable { rp.deliver(result); },
              [rp](error& err) mutable { rp.deliver(std::move(err)); });
      return rp;
    },
  };
}
```

This snippet creates three copies of the promise: one for the success handler, one for the error handler and one for returning from the message handler to tell CAF that we produce a response message at some later point.

These three promises must stay connected in some way, because we need to check whether a user called `deliver` (or `dispatch`) on *any* of them. Hence the implicit sharing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   string input on the command-line that starts with quotes in the same way it
   would parse strings from a config file, leading to very unintuitive results in
   some cases (#1113).
+- Response promises now implicitly share their state when copied. Once the
+  reference count for the state reaches zero, CAF now produces a
+  `broken_promise` error if the actor failed to fulfill the promise by calling
+  either `dispatch` or `delegate`.
 
 ### Fixed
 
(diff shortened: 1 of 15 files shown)


josephnoir (commented):
The `deliver` functions all have a `@post` that says `pending() == false` and then check the condition, silently no-op'ing. I'm wondering if that should be an error because it's a flaw in the code logic?

josephnoir:
Yeah, let's pretend I didn't say that.

PR #1207, opened 2021-01-22
Add {in,out}bound_stream_slot forward declarations

dominiklohmann:
This PR adds two missing forward declarations to `caf/fwd.hpp`.


Neverlord:
Thanks for the patch! 🙏

PR #1224, opened 2021-02-26
Fix creating typed_actor from typed_actor_pointer

### LINKED ISSUE #1218: Cannot construct typed_actor from typed_actor_pointer
This snippet is supposed to work:

```cpp
using my_actor = typed_actor<...>;
my_actor foo(my_actor::pointer_view self) {
  return my_actor{self};
}
```

However, trying to compile it gives:

```
(...)/caf/typed_actor.hpp|132 col 33| error: no matching constructor for initialization of 'caf::strong_actor_ptr' (aka 'intrusive_ptr<caf::actor_control_block>')
||   explicit typed_actor(T ptr) : ptr_(ptr.internal_ptr()) {
```

Neverlord:
Closes #1218.

### DIFF
--- a/libcaf_core/test/typed_spawn.cpp
+++ b/libcaf_core/test/typed_spawn.cpp
@@ -372,6 +372,34 @@ CAF_TEST(check_signature) {
   run();
 }
 
+SCENARIO("state-classes may use typed pointers") {
+  GIVEN("a state class for a statically typed actor type") {
+    using foo_type = typed_actor<result<int32_t>(get_atom)>;
+    struct foo_state {
+      foo_state(foo_type::pointer_view selfptr) : self(selfptr) {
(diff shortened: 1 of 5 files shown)


libcaf_core/test/typed_spawn.cpp
```diff
@@ -372,6 +372,34 @@ CAF_TEST(check_signature) {
   run();
 }
 
+SCENARIO("state-classes may use typed pointers") {
+  GIVEN("a state class for a statically typed actor type") {
```
josephnoir, line 376: Why is "state class" separated with a hyphen once and another time with a space?

Neverlord: No reason, just a typo. Fixed. :)

PR #1225, opened 2021-02-26
Allow setter to return bool, error or void

### LINKED ISSUE #1216: Inspectors only support primitive types with get/set API
A common idiom with the new inspector API is converting from an internal representation to an external representation during inspection. For example, an `enum` may convert between its constants and strings for human-readable formats and integers otherwise. This only boils down to writing a getter/setter pair and pass it to `apply`. However, this is not generalized at the moment. Consider this somewhat generic snippet:

```c++
  template <class Inspector>
  bool inspect(Inspector& f, MyType& x) {
    using value_type = /* ... */;
    auto get = [&x] { return to_value_type(x); };
    auto set = [&x](value_type val) { return from_value_type(val, x); };
    return f.apply(get, set);
  }
```

The problem is that `f.apply(get, set)` dispatches to `f.value` internally, which only accepts primitive types. The workaround is to dispatch on `Inspector::is_loading`, but the getter/setter API is simpler and should work with any type.

Neverlord:
Closes #1216.

### DIFF
--- a/libcaf_core/caf/load_inspector_base.hpp
+++ b/libcaf_core/caf/load_inspector_base.hpp
@@ -10,6 +10,7 @@
 
 #include "caf/inspector_access.hpp"
 #include "caf/load_inspector.hpp"
+#include "caf/sec.hpp"
 
 namespace caf {
 
@@ -61,7 +62,7 @@ public:
       auto emplace_result = xs.emplace(std::move(key), std::move(val));
       if constexpr (detail::is_pair<decltype(emplace_result)>::value) {
         if (!emplace_result.second) {
-          dref().emplace_error(sec::runtime_error, "multiple key definitions");
+          super::emplace_error(sec::runtime_error, "multiple key definitions");
           return false;
         }
       }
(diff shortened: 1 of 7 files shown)


libcaf_core/caf/load_inspector_base.hpp
```diff
@@ -106,17 +107,31 @@ class load_inspector_base : public load_inspector {
...
+        set(std::move(tmp));
+        return true;
+      } else {
+        return false;
+      }
     } else {
-      static_assert(std::is_convertible<setter_result, error>::value);
-      if (dref().value(tmp)) {
+      static_assert(std::is_convertible<setter_result, error>::value,
+                    "a setter must return caf::error, bool or void");
```
josephnoir, line 129: Could be aligned with the message in `load_inspector.hpp` (lines: 348, 370). Granted, this is a nitpick.
```suggestion
                    "setter must return caf::error, bool or void");
```

libcaf_core/test/load_inspector.cpp
```diff
@@ -567,4 +573,177 @@ CAF_TEST(load inspectors support messages) {
...
+        line default_line{{0, 0, 0}, {0, 0, 0}};
+        CHECK_EQ(x, default_line);
+      }
+    }
+    WHEN("passing a setter returning an error") {
+      f.reset();
+      auto x = line{{10, 10, 10}, {20, 20, 20}};
+      auto get = [&x] { return x; };
+      auto set = [](line) { return error{sec::runtime_error}; };
+      THEN("the inspector overrides the state using the setter") {
```
josephnoir, line 664:
```suggestion
      THEN("the inspection fails") {
```

libcaf_core/caf/load_inspector_base.hpp
Neverlord: Yep, I like the idea of being consistent. I've updated the other places, though, because I like an actual sentence starting with "a setter" better. 🙂

libcaf_core/test/load_inspector.cpp
Neverlord: Fixed.

Neverlord:
@josephnoir I've integrated your feedback and also rebased the branch to resolve conflicts.

PR #1228, opened 2021-03-03
Add metric_registry::merge function

Neverlord:
Add a new function that allows users to create some metrics even before the actor system gets created and then move them into the actor system after construction.

### DIFF
--- a/libcaf_core/caf/telemetry/metric_registry.hpp
+++ b/libcaf_core/caf/telemetry/metric_registry.hpp
@@ -471,6 +471,12 @@ public:
       visit_family(f, ptr.get());
   }
 
+  // -- modifiers --------------------------------------------------------------
+
+  /// Takes ownership of all metric families in `other`.
+  /// @pre `other` *must not* contain any duplicated metric family
+  void merge(metric_registry& other);
+
 private:
   /// @pre `families_mx_` is locked.
   metric_family* fetch(const string_view& prefix, const string_view& name);
--- a/libcaf_core/src/telemetry/metric_registry.cpp
+++ b/libcaf_core/src/telemetry/metric_registry.cpp
@@ -6,6 +6,7 @@
 
 #include "caf/actor_system_config.hpp"
 #include "caf/config.hpp"
+#include "caf/raise_error.hpp"
 #include "caf/telemetry/dbl_gauge.hpp"
 #include "caf/telemetry/int_gauge.hpp"
 #include "caf/telemetry/metric_family_impl.hpp"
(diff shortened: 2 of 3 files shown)


libcaf_core/src/telemetry/metric_registry.cpp
```diff
@@ -35,6 +36,21 @@ metric_registry::~metric_registry() {
   // nop
 }
 
+void metric_registry::merge(metric_registry& other) {
+  if (this == &other)
+    return;
+  std::unique_lock<std::mutex> guard1{families_mx_};
+  std::unique_lock<std::mutex> guard2{other.families_mx_};
```
dominiklohmann, line 43: Shouldn't this be `auto lock = std::scoped_lock{families_mx_, other.families_mx_};`?

Neverlord: No, we always use this notation for scope guards.

dominiklohmann: More in the sense of you need to lock both at the same time using either `std::lock(...)` with `std::{defer,adopt}_lock` or use the C++17 replacement for `std::unique_lock`, which is `std::scoped_lock`.

Neverlord: Switching to the "C++17 style" lock guards certainly is a good idea. I just wouldn't like them as a random one off in the code base via this PR.

dominiklohmann: Wit C++11 you were still able to do this:

```suggestion
  std::unique_lock<std::mutex> guard1{families_mx_, std::defer_lock};
  std::unique_lock<std::mutex> guard2{other.families_mx_, std::defer_lock};
  std::lock(guard1, guard2);
```

Neverlord: Are you suggesting this for the deadlock avoidance by `std::lock`? This is the only piece of code that touches both registries at once and there aren't any other algorithms or use cases that operate on two metric registries at once (this just isn't a thing), so I don't see a need for try-and-back-off or other avoidance strategies that `std::lock` may apply.

Switching to the new `scoped_lock` for cleaner code is a good idea, but rather as a refactoring step on the entire code base.

libcaf_core/caf/telemetry/metric_registry.hpp
```diff
@@ -471,6 +471,12 @@ class CAF_CORE_EXPORT metric_registry {
       visit_family(f, ptr.get());
   }
 
+  // -- modifiers --------------------------------------------------------------
+
+  /// Takes ownership of all metric families in `other`.
+  /// @pre `other` *must not* contain any duplicated metric family
```
josephnoir, line 477: What does "duplicated" mean here; as in `other` cannot contain the same metric twice or any metric that is present in `this` should not be in `other` (i.e., distinct sets of metric families)?

Neverlord: The latter. A registry cannot contain the same metric twice.

josephnoir: I can't make up my mind if this could be formulated more precisely or if that is even ambiguous in this context. This is the only nitpick I have. If you are ok with this, feel free to merge it.

PR #1242, opened 2021-03-22
Tie lifetime of meta objects table to CAF threads

### LINKED ISSUE #1241: Heap-use-after-free on meta objects if an app leaves main early
We currently use a static helper object to clean up meta object state. This works as long as applications destroy the actor system before leaving `main`. However, if an application only spins up threads in `main` and then carries on to use CAF in other threads, then CAF accesses the meta object table after it has been destroyed.

Original context see https://github.com/zeek/zeek/issues/1444#issuecomment-802800703.

Neverlord:
Fixes undefined behavior if an application leaves `main` while the threads started by CAF are still running. Usually, an application destroys the `actor_system` before leaving `main`. However, some applications may use the `main` function only for spinning up threads. In this case, the meta objects table must exceed the lifetime of the static helper object.

Closes #1241.

### DIFF
--- a/libcaf_core/src/logger.cpp
+++ b/libcaf_core/src/logger.cpp
@@ -648,9 +648,9 @@ void logger::start() {
     auto f = [this](auto guard) {
       CAF_IGNORE_UNUSED(guard);
       detail::set_thread_name("caf.logger");
-      this->system_.thread_started();
-      this->run();
-      this->system_.thread_terminates();
+      system_.thread_started();
+      run();
+      system_.thread_terminates();
     };
     thread_ = std::thread{f, detail::global_meta_objects_guard()};
   }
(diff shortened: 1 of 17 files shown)


Neverlord:
Closes #1241.

libcaf_core/src/logger.cpp
```diff
@@ -642,12 +643,16 @@ void logger::start() {
...
   } else {
-    thread_ = std::thread{[this] {
+    // Note: we don't call system_->launch_thread here since we don't want to
+    //       set a logger context in the logger thread.
+    auto f = [this](auto guard) {
+      CAF_IGNORE_UNUSED(guard);
       detail::set_thread_name("caf.logger");
       this->system_.thread_started();
       this->run();
       this->system_.thread_terminates();
```
josephnoir, line 653:
```suggestion
      system_.thread_started();
      run();
      system_.thread_terminates();
```

Those seem unnecessary.

Neverlord: Yep. Fixed.

PR #1244, opened 2021-03-24
Properly handle request.await in test DSL

### LINKED ISSUE #1232: expect() does not check awaited_responses
When using `request(…).await(…)` in tests, there's a mismatch between the checks by the `expect` clause and the actual behavior of the scheduler. The root cause is twofold:

- the `expect` clause does not check the `awaited_responses_` field in the `scheduled_actor`, while the `resume` call does
- the actor needs to be resumed once to skip (cache?) any messages at the front of the mailbox before processing the awaited response

Consider the following test case:

<details>
  <summary>Click to exand!</summary>

```c++
#define CAF_SUITE ping_pong

#include "caf/test/dsl.hpp"
#include "caf/test/unit_test_impl.hpp"

#include "caf/all.hpp"

using namespace caf;

// --(rst-ping-pong-begin)--
namespace {

struct empty_state {};
using actor_a = caf::typed_actor<reacts_to<pong_atom>>;
using actor_b = caf::typed_actor<reacts_to<ping_atom>>;
using actor_c = caf::typed_actor<result<pong_atom>(ping_atom)>;

actor_a::behavior_type runA(actor_a::stateful_pointer<empty_state> self, actor_b b) {
  self->send(b, ping_atom_v);
  aout(self) << "[A] -> ping" << std::endl;
  self->send(b, ping_atom_v);
  aout(self) << "[A] -> ping" << std::endl;
  return { [=](pong_atom) {
    aout(self) << "[A] <- pong" << std::endl;
  } };
}

actor_b::behavior_type runB(actor_b::stateful_pointer<empty_state> self, actor_c c) {
  aout(self) << "[B] init" << std::endl;
  return {
    [=](ping_atom) {
      aout(self) << "[B] <- ping" << std::endl;
      auto r = self->request(c, caf::infinite, ping_atom_v);
      aout(self) << "[B] -> ping" << std::endl;
      r.await([=](pong_atom) {
        aout(self) << "[B] <- pong" << std::endl;
      });
    }
  };
}

actor_c::behavior_type runC(actor_c::stateful_pointer<empty_state> self) {
  aout(self) << "[C] init" << std::endl;
  return {
    [=](ping_atom) {
      aout(self) << "[C] <- ping" << std::endl;
      aout(self) << "[C] -> pong" << std::endl;
      return pong_atom_v;
    }
  };
}

struct ping_pong_fixture : test_coordinator_fixture<> {
  actor_b b;
  actor_c c;

  ping_pong_fixture() {
    // Spawn the Pong actor.
    c = sys.spawn(runC);
    // Spawn the Pong actor.
    b = sys.spawn(runB, c);
    // Run initialization code for Pong.
    run();
  }
};

} // namespace

CAF_TEST_FIXTURE_SCOPE(ping_pong_tests, ping_pong_fixture)

CAF_TEST(three pings) {
  // Spawn the calling actor
  actor_a a = sys.spawn(runA, b);
  sched.run_once();

  // Test communication between A, B and C
  expect((ping_atom), from(a).to(b).with(_));
  // » Resume B to process the initial PING
  
  //expect((ping_atom), from(a).to(b).with(_));
  // » This resumes B but does not consume the message, only cache it (!)
  //assert(sched.prioritize(b)); sched.run_once();
  // » Alternative: Manually resuming B has the same effect
  
  expect((ping_atom), from(b).to(c).with(_));
  // » Resume C and put PONG into B's mailbox
  
  //expect((pong_atom), from(c).to(b).with(_));
  // » Note: If we don't resume B before this call,
  //   it fails because the message from A is at the top of the queue
  //assert(sched.prioritize(b)); sched.run_once();
  // » Alternative: Manually resuming B consumes and handles the message from C
  expect((ping_atom), from(a).to(b).with(_));
  // » Problem: This has the same effect as the previous line,
  //   but fails if we resumed B before the response arrived because
  //   the next message from A is cached.
  
  expect((ping_atom), from(a).to(b).with(_));
  // » Resume B again, this time consuming the second PING

  // Handle the second ping
  expect((ping_atom), from(b).to(c).with(_));
  expect((pong_atom), from(c).to(b).with(_));
  
  // No further messages allowed.
  disallow((ping_atom), from(a).to(b));
  disallow((ping_atom), from(b).to(c));
  disallow((pong_atom), from(c).to(b));
}

CAF_TEST_FIXTURE_SCOPE_END()
// --(rst-ping-pong-end)--
```

</details>

At some point, the second PING is in the mailbox of B and its PING to C has already been sent and is `await`ing a response. If we resume B now, the PING from A is temporarily set aside and a subsequent `expect` call can see and process the response from C just fine.

The problem is that we either need to use `sched.run_once()` manually, or `expect` the next message from A *without actually processing it*. The first option is only reliable if we make sure to `sched.prioritize(b)` (we need to schedule C once anyway). The second option produces a "phantom" message in the test log.

On the other hand, if we just `expect` the response from C, that fails because the PING from A is at the front of the mailbox. Instead, this should find that we're indeed awaiting a response from C, check whether it's in the inbox already and schedule the actor as needed.

Neverlord:
Closes #1232.

### DIFF
--- a/libcaf_core/src/scheduled_actor.cpp
+++ b/libcaf_core/src/scheduled_actor.cpp
@@ -175,6 +175,7 @@ void scheduled_actor::enqueue(mailbox_element_ptr ptr, execution_unit* eu) {
       break;
   }
 }
+
 mailbox_element* scheduled_actor::peek_at_next_mailbox_element() {
   if (mailbox().closed() || mailbox().blocked()) {
     return nullptr;
--- a/libcaf_core/src/scheduled_actor.cpp
+++ b/libcaf_core/src/scheduled_actor.cpp
@@ -176,7 +176,15 @@ void scheduled_actor::enqueue(mailbox_element_ptr ptr, execution_unit* eu) {
   }
 }
 mailbox_element* scheduled_actor::peek_at_next_mailbox_element() {
-  return mailbox().closed() || mailbox().blocked() ? nullptr : mailbox().peek();
+  if (mailbox().closed() || mailbox().blocked()) {
+    return nullptr;
+  } else if (awaited_responses_.empty()) {
+    return mailbox().peek();
+  } else {
+    auto mid = awaited_responses_.begin()->first;
+    auto pred = [mid](mailbox_element& x) { return x.mid == mid; };
+    return mailbox().find_if(pred);
+  }
 }
 
 // -- overridden functions of local_actor --------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 - Fix heap-use-after-free when accessing the meta objects table in applications
   that leave the `main` function while the actor system and its worker threads
   are still running (#1241).
+- The testing DSL now properly accounts for the message prioritization of actors
+  (suspending regular behavior until receiving the response) when using
+  `request.await` (#1232).
 
 ## [0.18.1] - 2021-03-19
 
--- a/libcaf_core/caf/intrusive/drr_cached_queue.hpp
+++ b/libcaf_core/caf/intrusive/drr_cached_queue.hpp
@@ -102,6 +102,13 @@ public:
     list_.peek_all(f);
   }
 
+  /// Tries to find the next element in the queue (excluding cached elements)
+  /// that matches the given predicate.
+  template <class Predicate>
+  pointer find_if(Predicate pred) {
+    return list_.find_if(pred);
+  }
+
   // -- modifiers -------------------------------------------------------------
 
   /// Removes all elements from the queue.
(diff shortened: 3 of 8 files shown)


libcaf_core/src/scheduled_actor.cpp
```diff
@@ -176,7 +176,15 @@ void scheduled_actor::enqueue(mailbox_element_ptr ptr, execution_unit* eu) {
   }
 }
 mailbox_element* scheduled_actor::peek_at_next_mailbox_element() {
```
josephnoir, line 178: Looks like there is a newline missing here.

josephnoir (commented):
Aside from that nitpick, I think the `find_if` function could use a short comment in all the files (`caf/intrusive/fifo_inbox.hpp`, `caf/intrusive/task_queue.hpp`, `caf/intrusive/wdrr_dynamic_multiplexed_queue.hpp`, `caf/intrusive/wdrr_fixed_multiplexed_queue.hpp`). Didn't make suggestions because I don't know whether the "excluding cached elements" part applies to all of them.

In any case, it looks like most of the functions in those files are documented, why leave this one out?

PR #1246, opened 2021-04-05
Fix `result<T>` constructors resolution during implicit conversions

### LINKED ISSUE #1245: `caf::result` constructor ambiguity
Here's simple code that trigger compile error in CAF 0.18.2:
```C++
#include <caf/result.hpp>

struct X {
	X(bool) {}
};

auto g() {
	return caf::result<X>{true};
}
```
GCC 11 output:
```
call of overloaded ‘result(<brace-enclosed initializer list>)’ is ambiguous
||    62 |         return caf::result<X>{true};
||       |                                   ^
|| In file included from kernel/src/property.cpp:14:
third_party/actor-framework/libcaf_core/caf/result.hpp|191 col 3| note: candidate: ‘caf::result<T>::result(caf::expected<T>) [with T = blue_sky::prop::X]’
||   191 |   result(expected<T> x) {
||       |   ^~~~~~
third_party/actor-framework/libcaf_core/caf/result.hpp|187 col 3| note: candidate: ‘caf::result<T>::result(T) [with T = blue_sky::prop::X]’
||   187 |   result(T x) : super(detail::result_base_message_init{}, std::move(x)) {
||       |   ^~~~~~
```
Porting my code to CAF 0.18.2 and constantly hitting this error to the point that no easy workaround exists.

uentity:
Convert existing value initializing constructor to templated one enabled
for all types convertible to T.

This change fixes #1245 and similar cases.

### DIFF
--- a/libcaf_core/caf/result.hpp
+++ b/libcaf_core/caf/result.hpp
@@ -184,7 +184,9 @@ public:
 
   using super::super;
 
-  result(T x) : super(detail::result_base_message_init{}, std::move(x)) {
+  template <class U, class = std::enable_if_t<std::is_constructible_v<T, U>>>
+  result(U&& x)
+    : super(detail::result_base_message_init{}, T{std::forward<U>(x)}) {
     // nop
   }
 


libcaf_core/caf/result.hpp
```diff
@@ -184,7 +184,8 @@ class result<T> : public result_base<T> {
 
   using super::super;
 
-  result(T x) : super(detail::result_base_message_init{}, std::move(x)) {
+  template <class U, class = std::enable_if_t<std::is_convertible_v<U, T>>>
+  result(U&& x) : super(detail::result_base_message_init{}, static_cast<T>(std::forward<U>(x))) {
```
Neverlord, line 188: Isn't this silently performing potentially unsafe integer conversions?

I probably would've gone with `is_constructible` instead for this reason:

```suggestion
  template <class U, class = std::enable_if_t<std::is_constructible_v<T, U>>>
  result(U&& x) : super(detail::result_base_message_init{}, T{std::forward<U>(x)}) {
```

Is this overlooking something that `is_constructible` gets right?

uentity: Yeah. I was deciding between `is_constructible` and `is_convertible`and left the latter. I checked your suggestion and seems that it works fine at lest in my code. I agree it's a better solution.

One edge case I was thinking about is the following -- what if somebody declare error enum as 'old' C-style `enum Error {}` with implicit conversion to `int`. Then let's consider the following code (based on CAF sources):
```C++
[=](int, int) -> result<int> { return Error::Happened; }
```
Will we get here wrong templated constructor being selected over overload that accepts error enums?

Neverlord:
Thanks for the fix! 🙏

libcaf_core/caf/result.hpp
Neverlord: Using C-style enums for error codes is a red flag in itself. For enum types, it actually doesn't matter which way we pick, since C-style enums are broken in both cases and enum classes do the opposite.

At least to my understanding of the language and recent Clang, GCC and MSVC all agree on this snippet:

```cpp
#include <type_traits>

enum class my_enum { a };

enum my_c_enum { argh };

static_assert(!std::is_convertible_v<my_enum, int>);

static_assert(!std::is_constructible_v<int, my_enum>);

static_assert(std::is_convertible_v<my_c_enum, int>);

static_assert(std::is_constructible_v<int, my_c_enum>);

int main() {
    return 0;
}
```

Issue #989, opened 2019-12-02
Too gready `caf::result` constructor


uentity:
After commit #9053a3afe5b86186621a2e80db546f2c0c0aa1b7 that is related to #908 and #902 `caf::result` now have the following constructor:
```C++
  // clang-format off
  template <class... Us,
            class = detail::enable_if_tt<
                     detail::all_constructible<
                       detail::type_list<Ts...>,
                       detail::type_list<detail::decay_t<Us>...>>>>
  // clang-format on
  result(Us&&... xs) : flag(rt_value) {
    value = make_message(Ts{std::forward<Us>(xs)}...);
  }
```

Now, let's consider that I have a *non-serializable* `tape` with similar gready perfect-forwarding constructor (that later dispatches to a set of private constructors) and a *serializable* `tape_box` that can be constructed from `tape` instance:
```C++
struct tape {
  template<typename... Ts> tape(Ts&&... ts) { ... }
};

struct tape_box {
  tape_box(const tape& t) { ... }
};
```

Then, returning `response_promise<tape_box>` from message handler will fail:
```C++
...
[=](int value) -> result<tape_box> {
  auto res = make_response_promise<tape_box>(...);
  return res; // <- compile error here
},
...
```
because compiler tries to check if `result<tape_box>` can be constructed from `response_promise&`, to do that it tries to evaluate constructor `tape(response_promise&)` and that check will succeed, because `tape` has perfect-forwarding constructor.
After that a compile error is produced on line `value = make_message(Ts{std::forward<Us>(xs)}...);` because `tape` doesn't expect `caf::response_promise`.

Maybe it's better to whitelist predefined types like `response_promise`, `typed_response_promise<...>`, `delegated<...>` in greedy `result` constructor?

PR #1249, opened 2021-04-16
Prevent utility types from matching to result<T> templated constructor

uentity:
Issue #989 has been reintroduced with recent commit 3fe3d5e33f8d985f532789f8dfb9170ca2427f16

Narrow `result<T>` templated constructor by one more condition: disable it for utility types that base class `result_base<T>` can be constructed from.

### DIFF
--- a/libcaf_core/caf/result.hpp
+++ b/libcaf_core/caf/result.hpp
@@ -184,7 +184,9 @@ public:
 
   using super::super;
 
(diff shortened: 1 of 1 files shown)


Neverlord:
Thanks once more! I hope the additional restriction to `enable_if` covers all the edge cases this time. 🤞

Btw, I've fixed the formatting during merge.

PR #1250, opened 2021-04-21
Fix typo in work-stealing config params reading

uentity:
Enclosing `caf.` section was missing.


Neverlord:
Thanks for the fix! I think we really ought to provide a better API than just passing strings around to avoid such mistakes in the future. 😕

PR #1252, opened 2021-04-23
Enable structured binding for cow_tuple

Neverlord:
Currently, using structured bindings for `cow_tuple` requires users to use the `.data()` member, e.g.:

```cpp
auto xs = make_cow_tuple(1, 2, 3);
auto [a, b, c] = xs.data();
```

By providing specializations for `std::tuple_size` and `std::tuple_element`, this simplifies to what users probably expected from a tuple-like type:

```cpp
auto xs = make_cow_tuple(1, 2, 3);
auto [a, b, c] = xs;
```

Of course this only grants const access (mutable access requires `xs.unshared()`).


Neverlord:
Actually, let me think about this some more. It may be convenient, but it always deconstructs into const references, which may lead to some very unexpected behavior / bugs.

PR #1259, opened 2021-05-21
Add missing namespace qualifications in caf unit test macros

lava:
The `CAF_CHECK_PASSED/_FAILED` macros would refer to the `term` identifier without namespace qualification, leading to a compilation error if invoked from outside the `caf` namespace.

### DIFF
--- a/libcaf_test/caf/test/unit_test.hpp
+++ b/libcaf_test/caf/test/unit_test.hpp
@@ -517,9 +517,10 @@ using caf_test_case_auto_fixture = caf::test::dummy_fixture;
 #define CAF_CHECK_PASSED(msg)                                                  \
   do {                                                                         \
     auto out = ::caf::test::logger::instance().massive();                      \
-    out << term::green << "** " << term::blue << __FILE__ << term::yellow      \
-        << ":" << term::blue << __LINE__                                       \
-        << ::caf::test::detail::fill(__LINE__) << term::reset << msg << '\n';  \
+    out << ::caf::term::green << "** " << ::caf::term::blue << __FILE__        \
+        << ::caf::term::yellow << ":" << ::caf::term::blue << __LINE__         \
+        << ::caf::test::detail::fill(__LINE__) << ::caf::term::reset << msg    \
+        << '\n';                                                               \
     ::caf::test::engine::current_test()->pass();                               \
     ::caf::test::engine::last_check_file(__FILE__);                            \
     ::caf::test::engine::last_check_line(__LINE__);                            \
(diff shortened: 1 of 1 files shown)


Neverlord:
How did you run into the compiler errors in the first place? Did you use them for custom macros on top of the CAF unit testing API? In this case I'd suggest implementing them differently. 🙂

After scanning the code base, it looks like these macros are basically legacy API that we forgot to clean up. I could find only one single leftover use of `CAF_CHECK_FAILED`. So my first reaction would be removing them TBH to avoid further bit rot.

lava:
Essentially yes: We did have some issues in the past with the automatic unboxing that is performed by the  which led to some broken tests being shown as green in CI, so we ditched the high-level test function and are using the lower-level unit testing macros directly.

Here is the code that triggered this: https://github.com/tenzir/vast/blob/3b41c8ef62df19f7e950d27e87ca9267c261706f/libvast_test/vast/test/test.hpp#L122-L129

If you see a better way to implement this, I'm open for suggestions ;)

Neverlord:
All CAF macros dispatch to one of the `::caf::test::detail::check*` functions. So I'd recommend wrapping them as well.

dominiklohmann:
I'm not quite sure why this was rejected. Isn't it a strict improvement, regardless of whether this is a good thing to do or not? Good macro hygiene doesn't hurt anyone.

Neverlord:
I don't think there are good arguments for leaving dead code in the source tree. Improved dead code is still dead code.

There were actually two unused macros for failing a check: `CAF_CHECK_FAILED` (which this PR wanted to fix since it was broken) and `CAF_CHECK_FAIL` (which also looked broken since it's passing `true` for the `should_fail` parameter). After looking at the sources, I believe the only correct macro to use is `CAF_ERROR`. Hence, I'm dropping the other two.

If you for some reason really don't want to use any of the check macros and roll your own, I'd recommend this for passing a test:

```cpp
#define MY_CHECK_PASS(msg)                                                         \
   ::caf::test::detail::check_un(true, __FILE__, __LINE__, msg)
```

Sorry for the inconvenience! Unfortunately, the unit testing part of CAF acquired its fair share of inconsistencies and quirks over the years. That's why we're [thinking to replace it eventually](https://www.actor-framework.org/blog/2021-05/roadmap-and-community-update). Also, I'm aware that I'm recommending a `detail` function. Only the `caf::test::detail` functions, though, since I'm very certain that we're not going to touch that (but migrate away to a new unit testing scaffold entirely). Please don't use `caf::detail` functions! 🙂

PR #1264, opened 2021-05-29
Fix typo

raphaelzulliger:
I'm not native English, but I'd say it was wrong before.


Neverlord:
Thanks for the fix! 🙏

Also, seems to be your first contribution. So, welcome to CAF. 👋

PR #1267, opened 2021-06-01
Fix CMakeLists.txt for installing caf-tools

Jasper-Ben:
This change fixes the outcome of `make install`, which previously would
install the tools' .cpp files to `/usr/local/share/caf/tools`. Now, we
install the built tool binaries instead.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -2,7 +2,7 @@ add_custom_target(all_tools)
 
 macro(add name)
   add_executable(${name} ${name}.cpp ${ARGN})
-  install(FILES ${name}.cpp DESTINATION ${CMAKE_INSTALL_DATADIR}/caf/tools)
+  install(TARGETS ${name} DESTINATION ${CMAKE_INSTALL_DATADIR}/caf/tools)
   add_dependencies(${name} all_tools)
 endmacro()
 


tools/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_custom_target(all_tools)
 
 macro(add name)
   add_executable(${name} ${name}.cpp ${ARGN})
-  install(FILES ${name}.cpp DESTINATION ${CMAKE_INSTALL_DATADIR}/caf/tools)
+  install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${name} DESTINATION ${CMAKE_INSTALL_DATADIR}/caf/tools)
```
dominiklohmann, line 5: This should either use `install(PROGRAMS $<TARGET_FILE:${name}> ...)` or `install(TARGETS ${name} ...)` so the installed target has the correct permissions. The latter is usually preferred.

Jasper-Ben: > This should either use `install(PROGRAMS $<TARGET_FILE:${name}> ...)` or `install(TARGETS ${name} ...)` so the installed target has the correct permissions. The latter is usually preferred.

Thanks for the hint. I'm not too familiar with cmake :slightly_smiling_face:

Neverlord:
@Jasper-Ben welcome to CAF! 👋

Thanks for the PR! Installing the binaries definitely makes more sense for the tools. 🙏

PR #1268, opened 2021-06-05
Fix typo

raphaelzulliger:
My default comment applies here as well :-).

(Btw: I'm not sure whether `that all actor should` handle is proper English or whether it should it be `that every actor should...` (or `each`). I'm really very unsure about that one!)

### DIFF
--- a/manual/MessagePassing.rst
+++ b/manual/MessagePassing.rst
@@ -70,7 +70,7 @@ Default and System Message Handlers
 -----------------------------------
 
 CAF has three system-level message types (``down_msg``, ``exit_msg``, and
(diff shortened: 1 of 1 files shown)


Neverlord:
Thank you once more for the proofreading! 🙏

> (Btw: I'm not sure whether `that all actor should` handle is proper English or whether it should it be `that every actor should...` (or `each`). I'm really very unsure about that one!)

I think it should be either "that all actors should ... regardless of their" or "that each actor should ... regardless of its". Since this section mostly uses the plural form, I've picked the former. Thanks for pointing it out!

PR #1272, opened 2021-06-11
Modernize functor_attachable and allow non-const default handlers

Neverlord:
I did some refactoring to use more C++17 constructs and also found an issue with the default handlers: passing a lambda declared as `mutable` to any of the `set_..._handler` functions fails to compile because the captured `fun` argument is `const` unless we also declare the lambdas that we use to construct the `std::function` object as `mutable`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   properly back off, causing high CPU load due to spinning and in some scenarios
   never recovering. This issue has been resolved by properly handling
   `SSL_ERROR_WANT_READ` on the transport (#1060).
+- Scheduled actors now accept default handlers for down messages etc. with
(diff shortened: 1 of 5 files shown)


josephnoir (commented):
Should we add a test to check the claim of the change log? Just in case we temper with it again in the future.

Neverlord:
> Should we add a test to check the claim of the change log? Just in case we temper with it again in the future.

Yeah, you're right of course. Sorry for filing the PR without adding a test for it. Will do.

Neverlord:
@josephnoir the handlers are checked in several test suites. Duplicating them only for the sake of using a mutable lambda did not strike me as a good approach. So I've added a new test suite that only does compile-time checks to make sure the setters accept both mutable and const function objects. Also rebased while I'm at it.

josephnoir (approved):

PR #1274, opened 2021-06-12
Fix shutdown issue when using blocking actors

### LINKED ISSUE #1266: Possible problem with private thread pool destruction
Is seems there is some problem here:
https://github.com/actor-framework/actor-framework/blob/9a09b5de89980a0fadb93b5eb13e8a280ea45f51/libcaf_core/src/actor_system.cpp#L399

Inside stop method of private_thread_pool, if we have some detached actors, private_thread_pool::running_ variable will be more than zero at this moment of time. Adding shutdown_helper node and processing it inside run_loop will do nothing since condition
https://github.com/actor-framework/actor-framework/blob/9a09b5de89980a0fadb93b5eb13e8a280ea45f51/libcaf_core/src/detail/private_thread_pool.cpp#L60
is not true - ( remaining > 0 )
Here shutdown_helper will be destroyed and lost. If some detached actor will be released after that - thread, performing stop will hang forever waiting loop thread to finish.
https://github.com/actor-framework/actor-framework/blob/9a09b5de89980a0fadb93b5eb13e8a280ea45f51/libcaf_core/src/detail/private_thread_pool.cpp#L49

Probably, some code could be added in actor_system destructor to wait until all detached actors return their threads back and stop private_thread_pool after this.

Neverlord:
Closes #1266.

@dfedin could you give this a try to see if it resolves the shutdown issues you've reported in #1266? I leave this in draft mode until then.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   when setting a `timespan` parameter such as `caf.middleman.heartbeat-interval`
   via config file or CLI to `0s` and then printing the config parameter, e.g.,
   via `--dump-config`.
+- Blocking actors now release their private thread before decrementing the
+  running-actors count to resolve a race condition during system shutdown that
+  could result in the system hanging (#1266).
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


dfedin:
> Closes #1266.
> 
> @dfedin could you give this a try to see if it resolves the shutdown issues you've reported in #1266? I leave this in draft mode until then.

Hi!
Yes, it works for me, no shutdown issues.

josephnoir (approved):

PR #1275, opened 2021-06-13
Fix some typos

Hamdor:
I recently found some tool that is able to find typos (https://github.com/crate-ci/typos) in source code. It can be simply used with github-actions, so if you are intrested I can add a workflow for that. Nevertheless the tool is not perfect and reports some false-positives.


Neverlord:
Thanks!

> if you are intrested I can add a workflow for that. Nevertheless the tool is not perfect and reports some false-positives.

Definitely interested! Whether I'd actually integrate it depends a on the false-positive, though. How many false positives did a pass over CAF produce?

Hamdor:
> How many false positives did a pass over CAF produce?

Actually, it isn't that bad. On current master, this would lead to 2 false-positives:

> error: `existend` should be `existed`
  --> .\libcaf_openssl\test\openssl\authentication.cpp:127:47
    |
127 |     // return if any file is unreadable or non-existend
    |                                                ^^^^^^^^
    |
error: `Whn` should be `When`
  --> .\libcaf_openssl\test\key.1.pem:8:3
  |
8 | modWhnI215kMqd9/atdT5EEVx8/f/MQMj0vII8GJSm6H6/duLIVFksMjTM+gCBtQ
  |    ^^^
  |

Neverlord:
> But the first isnt really a false-positive, the suggestion is just wrong.

Yeah, should be `non-existent`. Seems like all we need is to exclude `*.pem` files from spellchecking, then. 🙂

PR #1278, opened 2021-06-25
Add process metrics importer

Neverlord:
Move the logic for importing process metrics from the Prometheus broker to public API. This allows users to get access to the process metrics without having to load the HTTP server for Prometheus scraping.

### DIFF
--- a/libcaf_core/src/telemetry/importer/process.cpp
+++ b/libcaf_core/src/telemetry/importer/process.cpp
@@ -1,0 +1,229 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/telemetry/importer/process.hpp"
+
+#include "caf/logger.hpp"
+#include "caf/telemetry/gauge.hpp"
+#include "caf/telemetry/metric_registry.hpp"
+
+// -- detect supported platforms -----------------------------------------------
+
+#if defined(CAF_MACOS) || defined(CAF_LINUX)
+#  define CAF_HAS_PROCESS_METRICS
+#endif
+
+// -- gauge setup and boilerplate ----------------------------------------------
+
+#ifdef CAF_HAS_PROCESS_METRICS
+
+namespace {
+
+struct sys_stats {
+  int64_t rss;
+  int64_t vms;
+  double cpu_time;
+};
+
+sys_stats read_sys_stats();
+
+static constexpr bool platform_supported_v = true;
+
+template <class CpuPtr, class RssPtr, class VmsPtr>
+void sys_stats_init(caf::telemetry::metric_registry& reg, RssPtr& rss,
+                    VmsPtr& vms, CpuPtr& cpu) {
+  rss = reg.gauge_singleton("process", "resident_memory",
+                            "Resident memory size.", "bytes");
+  vms = reg.gauge_singleton("process", "virtual_memory", "Virtual memory size.",
+                            "bytes");
+  cpu = reg.gauge_singleton<double>("process", "cpu",
+                                    "Total user and system CPU time spent.",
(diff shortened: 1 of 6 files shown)


libcaf_core/src/telemetry/importer/process.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+namespace {
+
+sys_stats read_sys_stats() {
+  sys_stats result{0, 0, 0};
+  // Fetch memory usage.
+  {
+    mach_task_basic_info info;
+    mach_msg_type_number_t count = MACH_TASK_BASIC_INFO_COUNT;
+    if (task_info(mach_task_self(), MACH_TASK_BASIC_INFO, (task_info_t) &info,
```
josephnoir, line 92: A C-style cast.

libcaf_core/src/telemetry/importer/process.cpp
```diff
@@ -0,0 +1,229 @@
...
+        == KERN_SUCCESS) {
+      result.rss = info.resident_size;
+      result.vms = info.virtual_size;
+    }
+  }
+  // Fetch CPU time.
+  {
+    task_thread_times_info info;
+    mach_msg_type_number_t count = TASK_THREAD_TIMES_INFO_COUNT;
+    if (task_info(mach_task_self(), TASK_THREAD_TIMES_INFO, (task_info_t) &info,
```
josephnoir, line 103: A C-style cast.

libcaf_core/src/telemetry/importer/process.cpp
```diff
@@ -0,0 +1,229 @@
...
+#ifdef CAF_LINUX
+
+#  include <cstdio>
+#  include <unistd.h>
+
+namespace {
+
+std::atomic<long> global_ticks_per_second;
+
+std::atomic<long> global_page_size;
```
josephnoir, line 131: What are these here for? They are not used anywhere ... at least that's what my grep says.

libcaf_io/src/detail/prometheus_broker.cpp
```diff
@@ -267,17 +110,12 @@ behavior prometheus_broker::make_behavior() {
 }
 
 void prometheus_broker::scrape() {
-#ifdef HAS_PROCESS_METRICS
-  // Collect system metrics at most once per second.
+  // Scrape system metrics at most once per second.
   auto now = time(NULL);
```
josephnoir, line 114:
```suggestion
  auto now = time(nullptr);
```

josephnoir: Actually, is `nullptr == NULL` in these cases?

libcaf_core/src/telemetry/importer/process.cpp
Neverlord: They are used in `TRY_LOAD` to cache the result of syscalls.

libcaf_io/src/detail/prometheus_broker.cpp
Neverlord: Sure, but it's preferable since `nullptr` is never convertible to integers whereas `NULL` may be defined as just `0`.

libcaf_core/src/telemetry/importer/process.cpp
josephnoir: Ah, thanks.

libcaf_io/src/detail/prometheus_broker.cpp
josephnoir: I'm not sure in which direction you are arguing. The time [man page](https://linux.die.net/man/2/time) says "non-NULL" value. Could an implementation of time compare to 0 to check that?

Neverlord: Just saying we should avoid old C macros if there's a better alternative available with `nullptr`. The function takes a `time_t*` argument, so here's no ambiguity either way.

Issue #1236, opened 2021-03-11
CAF 0.19 API cleanup


Neverlord:
This is a meta issue for deprecating/removing obsolete API parts with 0.19.

We can replace some CAF classes with standard types now. TODOs are:

- [x] make `caf::byte` an alias for `std::byte` and deprecate the type alias
- [x] make `caf::string_view` an alias for `std::string_view` and deprecate the type alias
- [x] deprecate `caf::optional` and use `std::optional` in CAF instead, note that we can't switch to a type alias here because `caf::optional` has features that `std::optional` does not implement such as optional references
- [x] deprecate `caf::variant` and use `std::variant` in CAF instead, a simple type alias isn't enough since we'd also have to "import" `std::visit` and so on plus there might be API incompatibilities; thus it's probably better to leave `caf::variant` as it is and just deprecate it

Obsolete utilities that we can deprecate:

- [x] `caf::replies_to` and `caf::reacts_to`
- [x] the entire `SumType` API in `sum_type.hpp`

APIs that can leave the experimental phase:

- [x] connecting CAF nodes via `io::middleman::open`/`io::middleman::connect`
- [x] remote spawning of actors

Classes that are already obsolete and that we can remove:

- [x] `memory_managed`
- [x] `static_visitor`
- [x] `type_erased_value` (technically not deprecated, but simply has been overlooked and is now unused)
- [x] `caf::meta` API for legacy inspectors

Headers that are already obsolete and that we can remove:

- [x] `atom.hpp`

jakobod:
I will start working on this now. Any objections?

Neverlord:
> I will start working on this now. Any objections?

No, just recommendations: `byte` and `string_view` should be very straightforward. 🙂

Neverlord:
Slight change of plans. The `sum_type` API and `variant` had some nasty templates that weren't properly guarded with `enable_if`. It would have been nice to give them a deprecation cycle as well, but the effort to 'sanitize' the implementation to behave nicely when mixing with `std::variant` isn't worth the time investment. So get the axe right away.

PR #1287, opened 2021-08-03
Deprecate caf::byte

jakobod:
This PR deprecates the custom byte implementation in CAF. All internal references and usages have been refactored to use std::byte.

Relates #1236


libcaf_core/caf/byte.hpp
```diff
@@ -1,73 +1 @@
...
-namespace caf {
-
-/// A C++11/14 drop-in replacement for C++17's `std::byte`.
-enum class byte : uint8_t {};
-
-template <class IntegerType,
-          class = std::enable_if_t<std::is_integral<IntegerType>::value>>
-constexpr IntegerType to_integer(byte x) noexcept {
-  return static_cast<IntegerType>(x);
-}
```
Neverlord, line 19: Please leave this function in the header (but it should take a `std::byte`) and deprecate it.

Neverlord (changes requested):
Great to see this moving forward! 👏

It seems like you've used text-replacement over the entire repository. Nothing wrong with that, but please be careful to not replace the generic word "byte" with "std::byte" in comments. I hope I got all occurrences tagged, but please double-check. 🙂

Plus, please continue to use `to_integer` instead of `static_cast`.

libcaf_core/caf/byte.hpp
jakobod: I'm confused about this. Why leave this function and change it to take `std::byte` when the standard-lib provides it already?

Neverlord: Because removing the function outright may break client code. We wish to give them a warning with 0.19, not break their code immediately. 🙂

jakobod:
Should be resolved now. I found some other comments as well.

Neverlord:
@jakobod some minor notes: you forgot to deprecate `to_integer` and we use lowercase in deprecation warings: ``[[deprecated("Use ...")]]`` -> ``[[deprecated("use ...")]]``. I've fixed those during merge, just for future reference. 🙂

PR #1293, opened 2021-09-09
Replace C++ enum code generator with CMake script

Neverlord:
We currently have auto-generated files in the repository. This is of course of big no-no, but we ended up with this setup since the files are generated from a C++ tool and the setup crashes and burns when doing cross-compilation.

By using a CMake script to do the code generation instead, we can safely generate the files during the build phase and drop the auto-generated code from the code base once again.


CMakeLists.txt
```diff
@@ -204,51 +203,24 @@ if(NOT TARGET uninstall)
...
-else()
-  function(caf_add_enum_consistency_check hpp_file cpp_file)
-    # nop
-  endfunction()
-endif()
-
 # -- utility functions ---------------------------------------------------------
 
+# adds a consistency check that verifies that `cpp_file` is still valid by
+# re-generating the file and comparing it to the existing file
```
josephnoir, line 209: This is wrong, I think. There is not consistency check. This simply generates the files, right?

cmake/caf-generate-enum-strings.cmake
```diff
@@ -0,0 +1,142 @@
...
+      scan_mode()
+    else()
+      # reached the end
+      if(NOT enum_name)
+        message(FATAL_ERROR "No enum found in input file")
+      endif()
+      if(NOT enum_values)
+        message(FATAL_ERROR "No enum values found for ${enum_name}")
+      endif()
+      write_enum_file()
```
josephnoir, line 126: (Why is this not called `generate_mode`? Or why do the `_mode` functions not have similarly speaking names, e.g., `seek_enum` and `scan_cases`.)

Neverlord: That's no longer a processing mode that processes lines. This step is writing the file and then returning from the function/script. But I did gave the two modes some better name.

josephnoir: :)

josephnoir (approved):

Issue #584, opened 2017-08-07
add delayed_request feature


skorokhod:
It's quite often use case when actor should repeat some requests to a server.  Zero timeout (or optional?) should give an ordinary request

like in this scenario:
```
void send_request_to_server(bool first_try, std::chrono::seconds retry_timeout)
{
    delayed_request(target, first_try ?  std::chrono::seconds(0) : retry_timeout, msg).then
    (
     [&](good_response& r)
     {
     }
    ,[&](eror& r)
     {
        send_request_to_server(false, retry_timeout);
     }
    );
}
```

josephnoir:
Looking at `delayed_send`, a `delayed_request` would accept an additional timeout after which the request is issued, i.e., a delay on the sender side in addition to the timeout after which the request fails.

So, using pseudocode, something like this:
```
void delayed_send(destination, delay, message);
void delayed_request(destination, delay, timeout, message);
```

Using an actor, your code would look something like the following, right?

```
self->send(self, try_atom::value);
self->become(
  [=](try_atom) {
    self->request(dest, failure_timeout, get_atom::value).then(
      [&](result&) {
        // handle result
      },
      [&](error& err) {
        self->delayed_send(self, retry_timeout, try_atom::value);
      }
    );
  }
);
```

But if you don't have an actor to send the delayed message to, you need a way to delay the request. At the same time you want to be able to handle the error before reissuing the request (otherwise you could choose your failure timeout accordingly)?

```
void send_request_to_server(bool first_try, seconds retry_timeout) {
  delayed_request(dest, first_try ? seconds(0) : retry_to, failure_to, msg).then(
    [&](result& r) {
      // handle result
    },
    [&](error& r) {
      // received after failure_to
      send_request_to_server(false, retry_timeout); // executed after retry_to
    }
  );
}
```

In this case, you would send the request immediately and then again in an interval of `retry_to + failure_to` until you receive a result.

Does this summarize the problem & proposal?

skorokhod:
Yes, you're right!

Neverlord:
I think rather than adding a narrow feature like `delayed_request`, we should rather implement a more generic `run_delayed` and `run_scheduled` function pair:

```c++
void send_request_to_server(std::chrono::seconds retry_timeout)
{
    request(target, caf::infinite, msg).then
    (
     [&](good_response& r)
     {
     }
    ,[&](eror& r)
     {
[...]

PR #1294, opened 2021-09-10
Implement run_delayed

Neverlord:
Relates #584.

This PR introduces two new abstraction: actions and disposables.

An action is just some function object plus an atomic flag for its state. The primary use case is tagging an action as disposed. Once an action became disposed it may no longer run. However, some actions can run multiple times. The use case here is some period task. Once executed, an action may get rescheduled to enable another run.

A disposable is a new handle type with one (relevant) member function: `dispose`. This allows users to cancel any operation they no longer care about.

The new actor clock implementation is entirely based on actions. The request and stream timeouts are now based on the new `run_delayed` infrastructure. Of course, there's also a `run_scheduled` function to allow actors to schedule something based on a total timeout instead of using a delay. Both functions return a `disposable` to allow users to drop a pending timeout. The new scaffold also enabled us to get rid of the weird has-timeout flag as well as the `timeout_msg` message type.

Once this has been merged, we can add additional layers on top of the new infrastructure such as `run_periodically` and `run_(scheduled|delayed)_weak`.


libcaf_core/test/actor_clock.cpp
```diff
@@ -21,188 +21,89 @@ using namespace std::chrono_literals;
...
-  return x.timeout_id == y.value;
-}
-
 } // namespace
 
 CAF_TEST_FIXTURE_SCOPE(timer_tests, fixture)
 
-CAF_TEST(single_receive_timeout) {
+CAF_TEST(run_delayed without dispose) {
   // Have AUT call t.set_receive_timeout().
```
josephnoir, line 69:
```suggestion
  // Have AUT call self->run_delayed().
```

libcaf_core/test/actor_clock.cpp
```diff
@@ -21,188 +21,89 @@ using namespace std::chrono_literals;
...
   self->send(aut, ok_atom_v);
   expect((ok_atom), from(self).to(aut).with(_));
-  CAF_CHECK_EQUAL(t.schedule().size(), 1u);
-  CAF_CHECK_EQUAL(t.actor_lookup().size(), 1u);
+  CAF_CHECK_EQUAL(t.schedule.size(), 1u);
   // Advance time to send timeout message.
```
josephnoir, line 73:
```suggestion
  // Advance time to trigger timeout.
```

libcaf_core/test/actor_clock.cpp
```diff
@@ -21,188 +21,89 @@ using namespace std::chrono_literals;
...
   t.advance_time(10s);
-  CAF_CHECK_EQUAL(t.schedule().size(), 0u);
-  CAF_CHECK_EQUAL(t.actor_lookup().size(), 0u);
+  CAF_CHECK_EQUAL(t.schedule.size(), 0u);
   // Have AUT receive the timeout.
```
josephnoir, line 76:
```suggestion
  // Have AUT receive the action.
```

libcaf_core/test/actor_clock.cpp
```diff
@@ -21,188 +21,89 @@ using namespace std::chrono_literals;
...
-  expect((timeout_msg), from(aut).to(aut).with(tid{42}));
+  expect((action), to(aut));
+  CAF_CHECK(state().run_delayed_called);
 }
 
-CAF_TEST(override_receive_timeout) {
+CAF_TEST(run_delayed with dispose before expire) {
   // Have AUT call t.set_receive_timeout().
```
josephnoir, line 82:
```suggestion
  // Have AUT call self->run_delayed().
```

libcaf_core/test/actor_clock.cpp
```diff
@@ -21,188 +21,89 @@ using namespace std::chrono_literals;
...
+  CAF_CHECK_EQUAL(t.schedule.size(), 0u);
   // Have AUT receive the timeout.
-  expect((timeout_msg), from(aut).to(aut).with(tid{43}));
+  disallow((action), to(aut));
+  CAF_CHECK(!state().run_delayed_called);
 }
 
-CAF_TEST(mixed_receive_and_multi_timeouts) {
+CAF_TEST(run_delayed with dispose after expire) {
   // Have AUT call t.set_receive_timeout().
```
josephnoir, line 96:
```suggestion
  // Have AUT call self->run_delayed().
```

libcaf_core/test/detached_actors.cpp
```diff
@@ -9,76 +9,109 @@
...
+//     return {
+//       [=](ok_atom) {
+//         self->delayed_send(self, 1ns, ok_atom_v);
+//       },
+//     };
+//   };
+//   auto a = sys.spawn<detached>(f);
+//   auto g = detail::make_scope_guard(
+//     [&] { self->send_exit(a, exit_reason::user_shutdown); });
+// }
```
josephnoir, line 117: Hmmm?

libcaf_core/src/detail/test_actor_clock.cpp
```diff
@@ -46,10 +50,34 @@ size_t test_actor_clock::trigger_timeouts() {
...
+    if (period.count() > 0) {
+      auto next = t + period;
+      while (next <= current_time)
+        next += period;
+      schedule.emplace(next, schedule_entry{std::move(f), period});
+    }
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
Neverlord: An action cannot *fail* in the sense that it'd be an error. It may get *disposed* asynchronously, in which case running it becomes a no-op. Whether the action actually did something or not is irrelevant to the clock here.

libcaf_core/src/detail/thread_safe_actor_clock.cpp
```diff
@@ -5,147 +5,69 @@
...
 }
 
-void thread_safe_actor_clock::set_request_timeout(time_point t,
-                                                  abstract_actor* self,
-                                                  message_id id) {
-  push(new request_timeout(t, self->ctrl(), id));
+disposable
+thread_safe_actor_clock::schedule_periodically(time_point first_run, action f,
+                                               duration_type period) {
+  queue_.emplace_back(new schedule_entry{first_run, f, period});
```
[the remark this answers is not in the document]
Neverlord: Yep, good catch.

libcaf_core/src/scheduled_actor.cpp
```diff
@@ -666,6 +633,13 @@ scheduled_actor::categorize(mailbox_element& x) {
     call_handler(down_handler_, this, dm);
     return message_category::internal;
   }
+  if (auto view = make_typed_message_view<action>(content)) {
+    auto ptr = get<0>(view).ptr();
+    CAF_ASSERT(ptr != nullptr);
+    CAF_LOG_DEBUG("run action");
+    ptr->run();
+    return message_category::internal;
+  }
```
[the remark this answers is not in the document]
Neverlord: Yep, you can send an actor any action and it'll run it. Otherwise you can't schedule arbitrary work on an actor, so this is absolutely a feature we need in CAF. Based on this mechanism, I've already implemented the new flow abstraction that leverages actions for back-pressure signaling. Action is an unsafe type, so the this is meant for internal use only.

libcaf_core/test/detached_actors.cpp
Neverlord: 🤷🏻‍♂️

Probably some leftover from previous experiments. Deleted.

libcaf_core/caf/actor_clock.hpp
```diff
@@ -26,53 +26,93 @@ class CAF_CORE_EXPORT actor_clock {
...
+  /// Schedules an action for periodic execution by an actor.
+  /// @param worker The actor that should run the action.
+  /// @param policy The strategy for dealing with a stalling worker.
+  /// @param f The action to schedule.
+  /// @param first_run The local time at which the action gets enqueued to the
+  ///                  mailbox of the worker for the first time.
+  /// @param period The time to wait between two messages to the actor.
+  disposable schedule_periodically(time_point first_run, action f,
+                                   weak_actor_ptr worker, duration_type period,
+                                   stall_policy policy);
```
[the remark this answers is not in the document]
Neverlord: Replaced worker with target, hope that makes it clearer.

libcaf_core/src/scheduled_actor.cpp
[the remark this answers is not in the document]
josephnoir: OK. It's the non-internal use that I worry about.

josephnoir (approved):

Issue #1304, opened 2021-10-15
Immediately after underlying networking changes (VPN switching ON/OFF), remote_actor<>() returns successfully, but the remote actor doesn't reply any request. 


xiewei20082008:
I found this issue in a production environment where frequent VPN switchings are involved.

The actor framework has a table to record the mapping relationship between actor nodes and socket handles. The root cause is that now a handshake package is dropped as long as it is from a known node in the mapping table.

The assumption, all handshake packages from a known node are redundant, might be wrong if the handshake is for a new TCP connection from a previous known node.

Neverlord:
After stepping through your patch again (was thinking about just adding the `close` statements mentioned in the PR) and reading the ticket again: I think I may have cheered too early. The patch now always switches to a new socket, completely bypassing the previous code. CAF picks a single connection and then sticks to it, rejecting any additional connection to the same node. CAF only uses a single connection and sticks to that one to avoid message reordering and other oddities.

In your case, it seems like one side (A) did not recognize that a connection failed while the other side (B) already tries reconnecting. I think the right approach would be to have A detect the broken connection first. If that takes too long (default TCP timeouts can take minutes) than there's a heartbeat parameter to help speed that up. And B then would need some sort of retry loop for the connection. With the proposed PR, I think there's actually a high chance that A doesn't ever realize that it did loose a connection, which also may lead to lost messages.

PR #1305, opened 2021-10-15
Only drop handshake if it comes from the same socket, otherwise updat…

xiewei20082008:
…e it.

https://github.com/actor-framework/actor-framework/issues/1304

Signed-off-by: Wei Xie <xiewei.fire@gmail.com>


Neverlord:
Thank you for tracking this down and providing a patch! 🙏

The diff looks good to me at a first glance. I'll review this in more detail as well as test (and probably integrate) it in the next couple days.

Neverlord:
As outlined in #1304: cannot merge as-is. On closer inspection, this messes up some assumptions CAF makes on ordering and (lack of) message loss between two peered nodes.

PR #1308, opened 2021-11-09
Add JSON reader example

Neverlord:
Just a small example on how to use the new JSON reader to read custom types from file.

### DIFF
--- a/examples/config/read-json.cpp
+++ b/examples/config/read-json.cpp
@@ -1,0 +1,87 @@
+// Illustrates how to read custom data types from JSON files.
+
+#include "caf/actor_system.hpp"
+#include "caf/actor_system_config.hpp"
+#include "caf/caf_main.hpp"
+#include "caf/json_reader.hpp"
+#include "caf/type_id.hpp"
+
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <string_view>
+
+constexpr std::string_view example_input = R"([
+  {
+    "id": 1,
+    "name": "John Doe"
+  },
+  {
+    "id": 2,
+    "name": "Jane Doe",
+    "email": "jane@doe.com"
+  }
+])";
+
+struct user {
+  uint32_t id;
+  std::string name;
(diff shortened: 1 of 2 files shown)


examples/config/read-json.cpp
```diff
@@ -0,0 +1,87 @@
...
+constexpr std::string_view example_input = R"([
+  {
+    "id": 1,
+    "name": "John Doe"
+  },
+  {
+    "id": 2,
+    "name": "Jane Doe",
+    "email": "jane@doe.com"
+  }
```
josephnoir, line 23: I think JSON has a `null` value as well. Does that work equivalently to omitting the key-value pair?

examples/config/read-json.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+CAF_END_TYPE_ID_BLOCK(example_app)
+
+int caf_main(caf::actor_system& sys) {
+  // Get file path from config (positional argument).
+  auto& cfg = sys.config();
+  if (cfg.remainder.size() != 1) {
+    std::cerr << "*** expected one positional argument: path to a JSON file\n";
+    return EXIT_FAILURE;
+  }
```
josephnoir, line 53: Maybe we should add support for that to our argument parser at some time. 🤔

josephnoir (approved):
No critique, just comments. Feel free to merge! (Not merging because I'm not sure if you can then still answer the comments.)

examples/config/read-json.cpp
Neverlord: It should. Let me try and I'll file a new PR with a fix in case it doesn't. :)

examples/config/read-json.cpp
Neverlord: You mean positional arguments? Yeah, would be a good addition especially since the positional argument won't show up in the helptext at the moment.

josephnoir: Exactly.

PR #1317, opened 2022-01-17
Add NetBSD support

niclasr:
Add support for NetBSD.
In order to get uuid from root disk
gpt can be used but using it to get
this require root privileges so use
random uuid for now.
NetBSD has cxxabi.h with __cxa_demangle.
Add process metrics for NetBSD.

### DIFF
--- a/libcaf_core/src/telemetry/importer/process.cpp
+++ b/libcaf_core/src/telemetry/importer/process.cpp
@@ -10,7 +10,7 @@
 
 // -- detect supported platforms -----------------------------------------------
 
-#if defined(CAF_MACOS) || defined(CAF_LINUX)
+#if defined(CAF_MACOS) || defined(CAF_LINUX) || defined(CAF_NETBSD)
 #  define CAF_HAS_PROCESS_METRICS
 #endif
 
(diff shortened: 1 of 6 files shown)


libcaf_core/src/telemetry/importer/process.cpp
```diff
@@ -212,6 +212,51 @@ sys_stats read_sys_stats() {
...
+
+#include <sys/sysctl.h>
+#include <unistd.h>
+
+namespace {
+
+sys_stats read_sys_stats() {
+  sys_stats result{0, 0, 0};
+  int mib[6];
+  struct kinfo_proc2 kip2;
```
Neverlord, line 225: @niclasr is this API only available on NetBSD or is this available on all/more BSD variants (FreeBSD in particular)?

niclasr: There are similar apis on Free and OpenBSD that use sysctl. they do differ in the mibs for requesting data and use kinfo_proc instead of kinfo_proc2. OpenBSD has the some variable names inside the structure(p_vm_rsssize and r_rtime_* ) that are the same as NetBSD . FreeBSD uses ki_* names in its structure (see sys/user.h) I could add such code as a pull request if you want that. When speaking of this I think it would be good to change CAF_BSD to CAF_FREEBSD and CAF_OPENBSD since all the BSD:s do differ as you have already seen with SO_NOSIGPIPE and MSG_NOSIGNAL.

Neverlord: @niclasr Thanks a lot for the extra context! I like the idea of having different macros for the different BSD flavors, but I think having `CAF_BSD` as a catch-all is also helpful in places. I've pulled this into `pr/1317` and tweaked it slightly. Could you take a look and see if I didn't break something? 🙂

niclasr: I have just looked at it and here are my comments:
CAF_NET_BSD can be removed for CAF_POSIX since it is set when CAF_BSD is set.
 libcaf_core/src/detail/set_thread_name.cpp is going to break during compilation since NetBSD does not have pthread_np.h but it is included when CAF_BSD is defined. Don't know how you want that solved, __has_include perhaps?
libcaf_core/telemetry/import/process.cpp breaks on "auto kip2 = kinfo_proc2" {} is need after it to work.
Also global_ticks_per_second and global_page_size is not defined guess it is from above in the linux section.
I must admit that I do not understad what the global* vars are for.
There is also a narrowing warning from the use of kip2_size in the initialization of mib. It can be removed by using sizeof(kip2) in the mib instead. attached is a git diff of the differences that made it compile and run cleanly for me.
[afnb-diff.txt](https://github.com/actor-framework/actor-framework/files/7909152/afnb-diff.txt)

Neverlord: I think we can avoid pulling in the entire Linux block in `process.cpp`. Next try: https://github.com/actor-framework/actor-framework/commit/66a76bdabc14c81cbac5efc0476df693fd309ef2.

niclasr: I have now tested 66a76bd and it builds and works as expected.

Neverlord: Merged. Thank you for contributing! 🙏

PR #1328, opened 2022-04-22
Remove overlooked use deprecated APIs

Neverlord:
Trivial fix, just filing for CI.


Neverlord:
Also found some leftover uses of `reacts_to` and `optional`.

PR #1332, opened 2022-04-27
Fix bookkeeping of the field name

Neverlord:
The `field_` of `json_reader` keeps track of the nested field name in the JSON object. This PR fixes a bug where the `field_.pop_back();` had no prior `field_.push_back(...)` in some cases.

### DIFF
--- a/libcaf_core/src/json_reader.cpp
+++ b/libcaf_core/src/json_reader.cpp
@@ -166,6 +166,7 @@ void json_reader::revert() {
     err_.reset();
     st_->clear();
     st_->emplace_back(root_);
+    field_.clear();
   }
 }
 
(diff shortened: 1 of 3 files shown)


libcaf_core/src/json_reader.cpp
```diff
@@ -297,13 +299,16 @@ bool json_reader::begin_field(std::string_view name,
...
+    // Always push since we'll do a .pop_back() in end_field().
+    field_.push_back(name);
     return false;
   }
 }
 
 bool json_reader::begin_field(std::string_view name, bool& is_present,
                               span<const type_id_t> types, size_t& index) {
   SCOPE(position::object);
+  field_.push_back(name);
```
josephnoir, line 311: In this function, `push_back(name)` can happen twice. Here and in line 320. Is that correct?

Neverlord: Oops. Fixed.

Neverlord:
@josephnoir anything left to do here?

josephnoir:
Don't think so.

PR #1334, opened 2022-04-29
Implement new promises and futures

Neverlord:
This is an "MVP" for promises and futures in CAF. They will also receive integration into the new flows (and some mentioning in the manual) at a later point. I want to have the minimal version available now, because I have already use cases in caf-net that are currently "blocked" by not having this.

### DIFF
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -144,13 +144,12 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
   }
 
   void run() override {
-    // Note: we do *not* set the state to disposed after running the function
-    // object. This allows the action to re-register itself when needed, e.g.,
-    // to implement time-based loops.
     if (state_.load() == action::state::scheduled) {
       f_();
       if constexpr (IsSingleShot)
         state_ = action::state::disposed;
+      // else: allow the action to re-register itself when needed by *not*
+      //       setting the state to disposed, e.g., to implement time loops.
     }
   }
 
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -121,7 +121,7 @@ private:
 } // namespace caf
 namespace caf::detail {
 
-template <class F>
+template <class F, bool IsSingleShot>
 struct default_action_impl : detail::atomic_ref_counted, action::impl {
   std::atomic<action::state> state_;
   F f_;
@@ -149,6 +149,8 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
     // to implement time-based loops.
     if (state_.load() == action::state::scheduled) {
       f_();
+      if constexpr (IsSingleShot)
+        state_ = action::state::disposed;
     }
   }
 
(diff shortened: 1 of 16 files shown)


libcaf_core/caf/async/promise.hpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+private:
+  using cell_type = detail::async_cell<T>;
+  using cell_ptr = std::shared_ptr<cell_type>;
+
+  explicit promise(cell_type cell) : cell_(std::move(cell)) {
+    CAF_ASSERT(cell_ != nullptr);
+    cell_->promises.fetch_add(1, std::memory_order_relaxed);
+  }
```
josephnoir, line 88: Where is this used, can a cell have multiple promises?

josephnoir (commented):
Looks good. Question: Is there a way to access `self*` in the callbacks? Since there is no other documentation, the example would currently be the go to place to check how to use this. Might be good to either have an example, pick it up in the docs, or add it to the test (if it works).

libcaf_core/caf/async/promise.hpp
```diff
@@ -0,0 +1,109 @@
...
+    return valid();
+  }
+
+  bool operator!() const noexcept {
+    return !valid();
+  }
+
+  /// @pre `valid()`
+  void set_value(T value) {
+    if (cell_) {
```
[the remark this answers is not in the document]
Neverlord: `valid()` is the precondition a user would have to check before calling the function, which essentially *is* this null-check.

Neverlord:
> Is there a way to access self* in the callbacks? 

Which callbacks? `on_success` and `on_error`? No, but from the context you bind the future from, you'll usually have access to that and can just pass a lambda that captures whatever you need. Pretty much like the unit tests do it.

Documentation: yes, sure thing. The entire async (and flow) API needs documentation once the API has settled down and we're pushing for the next release. :)

libcaf_core/caf/async/promise.hpp
Neverlord: Uh... good catch! That's a bug. I'll fix it.

libcaf_core/caf/action.hpp
```diff
@@ -149,6 +149,8 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
     // to implement time-based loops.
```
josephnoir, line 149: That comment is not accurate anymore. There is not a path that changes the state to disposed.

libcaf_core/caf/async/future.hpp
```diff
@@ -0,0 +1,138 @@
...
+          g(make_error(sec::broken_promise, "future found an invalid value"));
+          break;
+        case 1:
+          f(static_cast<const T&>(std::get<T>(cp->value)));
+          break;
+        case 2:
+          g(static_cast<const error&>(std::get<error>(cp->value)));
+      }
+    };
+    auto cb_action = make_single_shot_action(std::move(cb));
```
josephnoir, line 50: The result of this change is that even if multiple promises of the same future try to set a value, only (the most recent?) one is propagated?

josephnoir, line 50: The result of this change is that even if multiple promises of the same future try to set a value, only (the most recent?) one is propagated?

PR #1337, opened 2022-05-03
Fix actor_widget exit

alatdneg:
When exiting it'll try and send msgs to monitoring actors, to do this it needs a valid execution_unit.

&dummy is not valid, so we supply one.

### DIFF
--- a/libcaf_core/caf/mixin/actor_widget.hpp
+++ b/libcaf_core/caf/mixin/actor_widget.hpp
@@ -36,7 +36,7 @@ public:
 
   ~actor_widget() {
     if (companion_)
-      self()->cleanup(error{}, &dummy_);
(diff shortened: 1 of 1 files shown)


Neverlord:
After reviewing the file, I believe this patch is incomplete.

The `dummy_` member is still used elsewhere. I think we should either:
1. Properly initialize the `dummy_` by calling [`dummy_.system_ptr(&system)`](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/scoped_execution_unit.hpp#L25) in the actor widget's `init` function, or
2. Drop `dummy_` completely.

I lean towards 1. since it's less impact (one line fix), but I'd be fine with either one.

Neverlord:
Yep. 👍

PR #1355, opened 2022-09-27
Make mutable lambda functions can be accepted by future then

lgpseu:


Neverlord:
Welcome to CAF and thank you for contributing! 👋🙏

PR #1359, opened 2022-10-10
Fix datagram_servant_passivated_msg documentation

brunorijsman:
The documentation incorrectly refers to datagram_servant_closed_msg instead of datagram_servant_passivated_msg.


Neverlord:
@brunorijsman welcome to CAF and thank you for the fix! Sorry for the delay, I must've missed the notification on this. 🙂

PR #1368, opened 2023-01-05
Provide `*_weak` variants of `scheduled_actor::run_{delayed,scheduled}`

### LINKED ISSUE #1222: Support for sending messages that do not block actor destruction.
It is often desirable to send periodic self-check messages to an actor so that it can perform an action like releasing unused resources. Currently sending a delayed message appears to block the actor's destruction until the messages has been delivered and dispatched. One potential semi-solution to this is to use a manager actor which monitors client actors and sends pings to them. Yet this adds unnecessary complexity to the program and potentially degrades performance. Also, as the manager actors needs to ping itself, this only moves the problem another place.

So, as discussed on gitter, this issue proposes to add a method to send a "dealyed_weak_message" to the actor without incrementing its reference count.

dominiklohmann:
This adds variants to the very useful `run_delayed` and `run_scheduled` member functions of `scheduled_actor` that do not keep the actor alive until the action is triggered.

I chose the `_weak` suffix rather than a `weak_` prefix so they show up in auto-completion alongside their strong variants, which makes it easier for users to discover these functions and to think about which one they need to call.

As of opening the PR the new functions still lack test coverage, as I wasn't quite sure how to test them. Ideas are welcome, and the PR is configured to allow pushes from maintainers.

Fixes #1222 by allowing `self->run_delayed_weak(delay, [self] { self->send(...) }` as a weak version of `self->delayed_send(self, delay, ...)`.

### DIFF
--- a/libcaf_core/caf/scheduled_actor.hpp
+++ b/libcaf_core/caf/scheduled_actor.hpp
@@ -647,6 +647,32 @@ public:
     return run_scheduled(time_point_cast<duration_t>(when), make_action(what));
   }
 
+  /// Runs `what` asynchronously at some point after `when`, iff this actor is
+  /// alive.
+  /// @param when The local time until the actor waits before invoking the
+  ///             action. Due to scheduling delays, there will always be some
+  ///             additional wait time. Passing the current time or a past time
+  ///             immediately schedules the action for execution.
+  /// @param what The action to invoke after waiting on the timeout.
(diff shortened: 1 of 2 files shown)


Neverlord:
Sorry for the long delay here. Our CI infrastructure still hasn't recovered, so we've switched to Cirrus for now. Code looks good, just needs some tests. I've pulled this into an integration branch and add some coverage tests next week.

PR #1370, opened 2023-01-23
address unsafe sprintf usage

vpax:
This tweak is just to get rid of compile-time warnings when building Zeek under MacOS Monterey / modern clang.w


Neverlord:
Thanks for the fix!

Our CI is currently down as a result of a cyber attack against HAW Hamburg (@josephnoir is working on getting our infrastructure back up again), so it may still take a couple days before we get to merge it. It's a small change and I don't expect any issues, but I'd still like to have CI coverage on it. 🙂

PR #1378, opened 2023-02-18
Better integrate typed actors into the new flow API

### LINKED ISSUE #1365: Flow API does not integrate properly into the typed API
At least these things are currently lacking:

- Response handles lack an `as_single` and `as_observable` overload when working with typed actors
- Typed actor views have no `make_observable` member function

If we start digging, we may find more.

### LINKED ISSUE #1377: Actors terminate despite having open streams
Actors that subscribe to a stream (e.g., via `self->observe_as`) but don't set a behavior immediately terminate instead of waiting for the resulting flow to complete.

Neverlord:
Closes #1365 and closes #1377. The latter was actually found by the new unit tests, so it made sense to fix the issue alongside the original ticket.

With this taken care of, 0.19 is almost in reach. The biggest task that remains is bringing the documentation up to date. 🙂

### DIFF
--- a/libcaf_core/test/typed_actor_view.cpp
+++ b/libcaf_core/test/typed_actor_view.cpp
@@ -141,7 +141,7 @@ SCENARIO("typed actors may use the flow API") {
         CHECK(!*err);
       }
     }
-    WHEN("the actor creates a typed stream via to_stream") {
+    WHEN("the actor creates a typed stream via to_typed_stream") {
       THEN("other actors may observe the values") {
         auto res = std::make_shared<int>(0);
         auto err = std::make_shared<error>();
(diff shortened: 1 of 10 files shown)


libcaf_core/test/typed_actor_view.cpp
```diff
@@ -0,0 +1,183 @@
...
+                       .take(10)
+                       .compose(self->to_typed_stream("foo", 10ms, 10));
+          self->spawn(typed_stream_observer, str, res, err);
+        });
+        run();
+        CHECK_EQ(*res, 55);
+        CHECK(!*err);
+      }
+    }
+    WHEN("the actor creates a typed stream via to_stream") {
```
josephnoir, line 144:
```suggestion
    WHEN("the actor creates a typed stream via to_typed_stream") {
```

libcaf_core/test/typed_actor_view.cpp
```diff
@@ -0,0 +1,183 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#define CAF_SUITE typed_actor_view
```
josephnoir, line 5: Is there anything else to test for the `typed_actor_view` aside from stream / flow-releated functionality?

PR #1379, opened 2023-02-24
Provide better diagnostic for missing meta objects

Neverlord:
Let's be nicer to CAF users. 🙂

It's easy to forget initializing the meta objects table. When forgetting it, CAF just crashes in some internal function that the user never heard of. We know what's wrong, so not telling the user is just bad manners. Especially towards first-time CAF users that probably think CAF is just broken.

So, to improve usability of CAF, we now print a neat error message that gives users some guidance as to what went wrong.

Before: segfault. 🤨

After: 💡
```
[FATAL] critical error (/Users/neverlord/workspace/project/caf/libcaf_core/src/detail/meta_object.cpp:68): found no meta object for type ID 200!
        This usually means that run-time type initialization is missing.
        With CAF_MAIN, make sure to pass all custom type ID blocks.
        With a custom main, call (before any other CAF function):
        - caf::core::init_global_meta_objects()
        - <module>::init_global_meta_objects() for all loaded modules
        - caf::init_global_meta_objects<T>() for all custom ID blocks
```

We only have a few places where we can deal with non-existing type IDs. Mostly while deserializing some data into C++ objects, where we have to expect that the data might be corrupted. For these use cases, we have a new function now that may still return `nullptr`.

### DIFF
--- a/libcaf_core/src/detail/meta_object.cpp
+++ b/libcaf_core/src/detail/meta_object.cpp
@@ -51,11 +51,11 @@ span<const meta_object> global_meta_objects() {
   return {meta_objects, meta_objects_size};
 }
 
-const meta_object* global_meta_object(type_id_t id) {
+const meta_object& global_meta_object(type_id_t id) {
   if (id < meta_objects_size) {
     auto& meta = meta_objects[id];
     if (!meta.type_name.empty())
-      return &meta;
+      return meta;
   }
   CAF_CRITICAL_FMT(
     "found no meta object for type ID %d!\n"
@@ -66,7 +66,6 @@ const meta_object* global_meta_object(type_id_t id) {
     "        - <module>::init_global_meta_objects() for all loaded modules\n"
     "        - caf::init_global_meta_objects<T>() for all custom ID blocks",
     static_cast<int>(id));
-  return nullptr;
 }
 
 const meta_object* global_meta_object_or_null(type_id_t id) {
--- a/libcaf_core/src/detail/meta_object.cpp
+++ b/libcaf_core/src/detail/meta_object.cpp
@@ -54,7 +54,26 @@ span<const meta_object> global_meta_objects() {
 const meta_object* global_meta_object(type_id_t id) {
   if (id < meta_objects_size) {
     auto& meta = meta_objects[id];
-    return !meta.type_name.empty() ? &meta : nullptr;
+    if (!meta.type_name.empty())
+      return &meta;
+  }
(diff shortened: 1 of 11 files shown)


alatdneg:
Nice, that confused me no end when my serialise test code bombed out.

libcaf_core/src/detail/meta_object.cpp
```diff
@@ -54,7 +54,26 @@ span<const meta_object> global_meta_objects() {
...
+  CAF_CRITICAL_FMT(
+    "found no meta object for type ID %d!\n"
+    "        This usually means that run-time type initialization is missing.\n"
+    "        With CAF_MAIN, make sure to pass all custom type ID blocks.\n"
+    "        With a custom main, call (before any other CAF function):\n"
+    "        - caf::core::init_global_meta_objects()\n"
+    "        - <module>::init_global_meta_objects() for all loaded modules\n"
+    "        - caf::init_global_meta_objects<T>() for all custom ID blocks",
+    static_cast<int>(id));
+  return nullptr;
```
josephnoir, line 69: Is this unreachable code? Which to me would mean a compiler warning and it should not be here.

libcaf_core/src/type_id_list.cpp
```diff
@@ -13,8 +13,8 @@ namespace caf {
 size_t type_id_list::data_size() const noexcept {
   auto result = size_t{0};
   for (auto type : *this) {
-    auto meta_obj = detail::global_meta_object(type);
-    result += meta_obj->padded_size;
+    auto meta = detail::global_meta_object(type);
+    result += meta->padded_size;
```
josephnoir, line 17: Why? In the `message.cpp` diff it is still called `meta_obj` and in the `type_id.cpp` diff the result is called `ptr`.

libcaf_core/src/detail/meta_object.cpp
dominiklohmann: The function could also return a reference if it's marked `[[noreturn]]` given that `CAF_CRITICAL` always ends up calling `std::abort()`.

Neverlord: ACK to both.

libcaf_core/src/type_id_list.cpp
Neverlord: Should be consistently `meta` now wherever we access a meta object.

josephnoir (approved):

PR #1381, opened 2023-02-24
Limit how many actions actors may run in one go

### LINKED ISSUE #1364: Actors reading from SPSC buffers may end up in a loop
Actors that read from an SPSC buffer with a non-stop writer at the other end can end up in a loop where each pull action immediately causes another pull action.

Here's a potential patch:

```diff
diff --git a/libcaf_core/caf/scheduled_actor.hpp b/libcaf_core/caf/scheduled_actor.hpp
index dcea886fd..7a735a938 100644
--- a/libcaf_core/caf/scheduled_actor.hpp
+++ b/libcaf_core/caf/scheduled_actor.hpp
@@ -807,6 +807,10 @@ private:
   /// message.
   std::vector<action> actions_;
 
+  /// Flag that tells delay() to push actions to the mailbox if we are already
+  /// in run_actions.
+  bool running_actions_ = false;
+
   /// Stores resources that block the actor from terminating.
   std::vector<disposable> watched_disposables_;
 };
diff --git a/libcaf_core/src/scheduled_actor.cpp b/libcaf_core/src/scheduled_actor.cpp
index 6bc9ecd64..96b5576df 100644
--- a/libcaf_core/src/scheduled_actor.cpp
+++ b/libcaf_core/src/scheduled_actor.cpp
@@ -606,7 +606,12 @@ void scheduled_actor::schedule(action what) {
 }
 
 void scheduled_actor::delay(action what) {
-  actions_.emplace_back(std::move(what));
+  // If we are already in run_actions, we force the action through the mailbox
+  // in order to break hot loops that would otherwise starve any other activity.
+  if (!running_actions_)
+    actions_.emplace_back(std::move(what));
+  else
+    schedule(std::move(what));
 }
 
 disposable scheduled_actor::delay_until(steady_time_point abs_time,
@@ -1242,6 +1247,8 @@ void scheduled_actor::watch(disposable obj) {
 }
 
 void scheduled_actor::run_actions() {
+  running_actions_ = true;
+  auto guard = detail::make_scope_guard([this] { running_actions_ = false; });
   if (!actions_.empty()) {
     // Note: can't use iterators here since actions may add to the vector.
     for (auto index = size_t{0}; index < actions_.size(); ++index) {
```

Neverlord:
I ended up with a slightly different patch then the first one I've tried in #1364. Instead of using a flag, I'm using a counter now to allow some actions to go through immediately but then "interrupt" flows if they have cause a chain of >= 10 consecutive actions to trigger. The threshold basically balances between fairness and efficiency. 10 should be low enough to be on the "fair scheduling" side where actors can't hog the CPU for themselves for too long while not being too aggressive.

I think hard-coding this parameter should be fine. Maybe we can run some benchmarking at some point to see whether it sits in the right spot at 10. But that's a story for another day. 🙂

Close #1364.

### DIFF
--- a/libcaf_core/caf/defaults.hpp
+++ b/libcaf_core/caf/defaults.hpp
@@ -32,6 +32,11 @@ constexpr parameter<T> make_parameter(std::string_view name, T fallback) {
   return {name, fallback};
 }
 
+/// Configures how many actions scheduled_actor::delay may add to the internal
+/// queue for scheduled_actor::run_actions before being forced to push them to
+/// the mailbox instead.
+constexpr auto max_inline_actions_per_run = size_t{10};
+
 } // namespace caf::defaults
 
 namespace caf::defaults::stream {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   operators that use the `ucast` operator internally.
 - The `mcast` and `ucast` operators now stop calling `on_next` immediately when
   disposed.
+- Actors reading from external sources such as SPSC buffers via a local flow
+  could end up in a long-running read loop. To avoid potentially starving other
+  actors or activities, scheduled actors now limit the amount of actions that
+  may run in one iteration (#1364).
 
 ## [0.19.0-rc.1] - 2022-10-31
 
--- a/libcaf_core/caf/scheduled_actor.hpp
+++ b/libcaf_core/caf/scheduled_actor.hpp
@@ -829,6 +829,11 @@ private:
   /// message.
   std::vector<action> actions_;
 
+  /// Counter for scheduled_actor::delay to make sure
+  /// scheduled_actor::run_actions does not end up in a busy loop that might
+  /// starve other activities.
+  size_t delayed_actions_this_run_ = 0;
+
   /// Stores ongoing activities such as flows that block the actor from
   /// terminating.
   std::vector<disposable> watched_disposables_;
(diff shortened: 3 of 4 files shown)


libcaf_core/caf/defaults.hpp
```diff
@@ -32,6 +32,11 @@ constexpr parameter<T> make_parameter(std::string_view name, T fallback) {
   return {name, fallback};
 }
 
+/// Configures how many actions scheduled_actor::delay may add to the internal
+/// queue for scheduled_actor::run_actions before being forced to push them to
+/// the mailbox instead.
```
josephnoir, line 37: I'm assuming that "to the mailbox" means that other messages will be processed before the delayed actions. The doc of `void delay(action what)` In `coordinator.hpp` says:

> Delays execution of an action until all pending actions were executed.

Does that still fit? Are other tasks in the mailbox included in  "all pending actions"?

josephnoir (commented):
I don't think I completely understand how everything here is connected. During `run_actions()` new actions are added to `actions_` and that is why it is called `delayed_actions_during_this_run_`?

libcaf_core/caf/defaults.hpp
Neverlord: > Does that still fit?

Yep. Everything that's in `actions_` gets processed before taking the next message from the mailbox. So by pushing into the mailbox, we automatically handle those actions after whatever is in `actions_` at the moment.

Neverlord:
> During run_actions() new actions are added to actions_ and that is why it is called delayed_actions_during_this_run_?

Exactly. 🙂

PR #1382, opened 2023-02-24
Fix bug on SPSC buffer handshake

Neverlord:
While adding more coverage tests, I came across some bugs in the SPSC buffer. If the buffer has already been closed, consumers would end up waiting on the producer to appear forever. Same with a producer that appears after a consumer has dropped it's side already.

### DIFF
--- a/libcaf_core/caf/async/spsc_buffer.hpp
+++ b/libcaf_core/caf/async/spsc_buffer.hpp
@@ -155,7 +155,7 @@ public:
     CAF_ASSERT(consumer != nullptr);
     lock_type guard{mtx_};
     if (consumer_)
-      CAF_RAISE_ERROR("SPSC buffer already has a consumer");
+      CAF_RAISE_ERROR(std::logic_error, "SPSC buffer already has a consumer");
     consumer_ = std::move(consumer);
     if (producer_)
       ready();
@@ -167,10 +167,8 @@ public:
   void set_producer(producer_ptr producer) {
     CAF_ASSERT(producer != nullptr);
     lock_type guard{mtx_};
-    if (producer_) {
-      producer->on_consumer_cancel();
-      return;
-    }
+    if (producer_)
+      CAF_RAISE_ERROR(std::logic_error, "SPSC buffer already has a producer");
     producer_ = std::move(producer);
     if (consumer_)
       ready();
@@ -349,13 +347,12 @@ struct resource_ctrl : ref_counted {
   }
 
   buffer_ptr try_open() {
+    auto res = buffer_ptr{};
     std::unique_lock guard{mtx};
     if (buf) {
-      auto res = buffer_ptr{};
       res.swap(buf);
-      return res;
     }
-    return nullptr;
+    return res;
   }
 
   mutable std::mutex mtx;
@@ -428,6 +425,16 @@ public:
     return ctrl_ == nullptr;
   }
 
+  friend bool operator==(const consumer_resource& lhs,
+                         const consumer_resource& rhs) {
+    return lhs.ctrl_ == rhs.ctrl_;
+  }
+
+  friend bool operator!=(const consumer_resource& lhs,
+                         const consumer_resource& rhs) {
+    return lhs.ctrl_ != rhs.ctrl_;
+  }
+
 private:
   intrusive_ptr<resource_ctrl<T, false>> ctrl_;
 };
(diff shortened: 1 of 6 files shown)


libcaf_core/test/async/spsc_buffer.cpp
```diff
@@ -180,4 +214,322 @@ SCENARIO("SPSC buffers moves data between actors") {
...
+        auto outputs = std::vector<int>{};
+        {
+          auto [rd, wr] = async::make_spsc_buffer_resource<int>(6, 2);
+          sys.spawn([wr{wr}](actor_t* src) {
+            src
+              ->make_observable() //
+              .iota(1)
+              .subscribe(wr);
+          });
+          // Only difference to before: have the actor adding an observer that
```
josephnoir, line 344:
```suggestion
          // Only difference to before: have the actor add an observer that
```

libcaf_core/test/async/spsc_buffer.cpp
```diff
@@ -180,4 +214,322 @@ SCENARIO("SPSC buffers moves data between actors") {
...
+        CHECK(!rd2);
+        CHECK(wr);
+        CHECK_NE(wr.try_open(), nullptr);
+        CHECK(!wr);
+        CHECK_EQ(wr.try_open(), nullptr);
+        auto wr2 = async::producer_resource<int>{};
+        wr2 = wr;
+        CHECK(!wr2);
+        wr2 = std::move(wr);
+        CHECK(!wr2);
```
josephnoir, line 395: This not only calls `try_open` twice but checks the copy and move assignment as well?

libcaf_core/test/async/spsc_buffer.cpp
```diff
@@ -180,4 +214,322 @@ SCENARIO("SPSC buffers moves data between actors") {
...
+      }
+    }
+  }
+}
+
+SCENARIO("try_open on producer resources succeeds only once") {
+  GIVEN("a producer resource") {
+    WHEN("opening it twice") {
+      THEN("the second try_open fails") {
```
josephnoir, line 404: This happens implicitly? Maybe the text could more closely reflect what is happening here, e.g., "When creating two producers the second one receives a down message" ... or something more accurate.

libcaf_core/test/async/spsc_buffer.cpp
```diff
@@ -180,4 +214,322 @@ SCENARIO("SPSC buffers moves data between actors") {
...
+        CHECK(self->mailbox().empty());
+      }
+    }
+  }
+}
+
+SCENARIO("try_open on consumer resources succeeds only once") {
+  GIVEN("a consumer resource") {
+    WHEN("opening it twice") {
+      THEN("the second try_open fails") {
```
josephnoir, line 434: Same.

libcaf_core/caf/async/spsc_buffer.hpp
```diff
@@ -164,24 +157,24 @@ class spsc_buffer : public ref_counted {
...
 
   /// Producer callback for the initial handshake between producer and consumer.
   void set_producer(producer_ptr producer) {
     CAF_ASSERT(producer != nullptr);
     lock_type guard{mtx_};
-    if (producer_)
-      CAF_RAISE_ERROR("SPSC buffer already has a producer");
+    if (producer_) {
+      producer->on_consumer_cancel();
+      return;
```
josephnoir, line 172: Why does this propagate the error while calling `set_consumer` twice raises and error?

Neverlord: Yeah, good question. I went a bit back-and-forth on that one. We simply have no way to "gracefully" abort a consumer. But having the producer behave differently is also odd. So for consistency: ack, reverting to raise an error in both cases. Also throwing a `std::logic_error` here instead of `std::runtime_error`.

libcaf_core/caf/async/spsc_buffer.hpp
```diff
@@ -150,7 +142,8 @@ class spsc_buffer : public ref_counted {
   /// Closes the buffer by request of the consumer.
```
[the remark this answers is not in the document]
Neverlord: There is no abort flag. :)

A buffer can be closed or not. Optionally, the producer may set an error before closing. This is the difference between `close` and `abort`. That's why `close` calls `abort` with a default-constructor error (i.e., "no error").

A consumer cancels its subscription if it's no longer interested in the data, at which point the producer should dispose the buffer.

I'm not a big fan of these flags myself. We might revisit that design choice with https://github.com/actor-framework/actor-framework/issues/1385 at some later point.

PR #1401, opened 2023-04-17
Add `[[nodiscard]]` to `expected`

jachris:
Discarding a `caf::expected` is almost always an error. I would find it very useful to have this attribute on the declaration of the class itself. This small PR does exactly that. Feel free to close this if the absence of this attribute is intentional.

### DIFF
--- a/libcaf_core/caf/fwd.hpp
+++ b/libcaf_core/caf/fwd.hpp
@@ -27,7 +27,7 @@ template <class> class behavior_type_of;
 template <class> class callback;
 template <class> class cow_vector;
 template <class> class dictionary;
-template <class> class expected;
+template <class> class [[nodiscard]] expected;
 template <class> class intrusive_cow_ptr;
 template <class> class intrusive_ptr;
 template <class> class param;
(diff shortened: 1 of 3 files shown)


Neverlord:
Hi, and welcome to CAF! 👋

Funny, we had a discussion around `nodiscard` recently in chat with @dominiklohmann, who proposed that same patch. I actually think adding the attribute is a good idea.

However... the current `caf::expected` class is ultimately just a placeholder until we can define it as `expected = std::expected<T, error>` with C++23. We've even deprecated anything that goes beyond the STL interface with 0.19 to make that eventual transition smooth.

And the C++23 `expected` isn't marked as `nodiscard` (see https://en.cppreference.com/w/cpp/utility/expected and https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0323r12.html).

If we were to add `nodiscard` to the CAF implementation now and people start to rely on it, we'd eventually "silently remove" the attribute again by switching to the standard type. I think it's better to stick to whatever the standard is doing, even if we disagree on some of the choices. I personally think they should've kept the original name `map` over `transform`, for example. So... I'm sorry, but I gotta reject the PR.

dominiklohmann:
This made me think of other ways, and I actually found a solution that's easy to implement without modifying CAF code: https://github.com/tenzir/vast/pull/3086

PR #1403, opened 2023-04-23
Fix flow setup for WebSocket servers

### LINKED ISSUE #1402: Echo WebSocket server example broken
As reported in chat, `examples/web_socket/echo` is broken in 0.19. Connection aborts right after clients connect. With this patch:

```
diff --git a/examples/web_socket/echo.cpp b/examples/web_socket/echo.cpp
index 86163f5e0..b1b995b94 100644
--- a/examples/web_socket/echo.cpp
+++ b/examples/web_socket/echo.cpp
@@ -93,17 +93,25 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
                 // ... that simply pushes data back to the sender.
                 auto [pull, push] = ev.data();
                 pull.observe_on(self)
-                  .do_on_next([](const ws::frame& x) {
-                    if (x.is_binary()) {
-                      std::cout
-                        << "*** received a binary WebSocket frame of size "
-                        << x.size() << '\n';
-                    } else {
-                      std::cout
-                        << "*** received a text WebSocket frame of size "
-                        << x.size() << '\n';
-                    }
+                  .do_on_error([](const caf::error& what) { //
+                    std::cout << "*** connection closed: " << to_string(what)
+                              << "\n";
                   })
+                  .do_on_complete([] { //
+                    std::cout << "*** connection closed\n";
+                  })
+                  .do_on_next(
+                    [](const ws::frame& x) {
+                      if (x.is_binary()) {
+                        std::cout
+                          << "*** received a binary WebSocket frame of size "
+                          << x.size() << '\n';
+                      } else {
+                        std::cout
+                          << "*** received a text WebSocket frame of size "
+                          << x.size() << '\n';
+                      }
+                    })
                   .subscribe(push);
               });
           });
```

I get: `*** connection closed: caf::sec::disposed("destroyed producer_resource without opening it")`.

Neverlord:
No idea when/how this bug got in before releasing 0.19... 😕

Closes #1402.

### DIFF
--- a/robot/web_socket/echo.robot
+++ b/robot/web_socket/echo.robot
@@ -1,0 +1,61 @@
+*** Settings ***
+Documentation     A test suite for examples/web_socket/echo-server
+Library           Collections
+Library           Process
+Library           RequestsLibrary
+Library           WebSocketClient
+
+Suite Setup       Start Servers
+Suite Teardown    Stop Servers
+
+*** Variables ***
+${HTTP_URL}       http://localhost:55501
+${HTTPS_URL}      https://localhost:55502
+${WS_URL}         ws://localhost:55501
+${WSS_URL}        wss://localhost:55502
+${FRAME_COUNT}    ${10}
+${BINARY_PATH}    /path/to/the/server
+${SSL_PATH}       /path/to/the/pem/files
+
+*** Test Cases ***
+Test WebSocket Server
+    [Tags]    WebSocket
+    ${fd}=                  WebSocketClient.Connect  ${WS_URL}
+    WebSocketClient.Send    ${fd}    Hello
+    ${result}=              WebSocketClient.Recv    ${fd}
+    WebSocketClient.Close   ${fd}
+    Should Be Equal         ${result}    Hello
+
+Test WebSocket Over SSL Server
+    [Tags]    WebSocket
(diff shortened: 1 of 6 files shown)


robot/web_socket/echo.robot
```diff
@@ -0,0 +1,61 @@
...
+    WebSocketClient.Send    ${fd}    Hello
+    ${result}=              WebSocketClient.Recv    ${fd}
+    WebSocketClient.Close   ${fd}
+    Should Be Equal         ${result}    Hello
+
+*** Keywords ***
+Start Servers
+    ${res1}=    Start Process    ${BINARY_PATH}  -p  55501
+    Set Suite Variable    ${ws_server_process}    ${res1}
+    ${res2}=    Start Process    ${BINARY_PATH}  -p  55502  -k  ${SSL_PATH}/key.pem  -c  ${SSL_PATH}/cert.pem
```
josephnoir, line 42: There probably isn't a way to avoid a static port?

Neverlord: I don't know how. I don't think there's a Robot feature that gives you a currently unused port. But 555xx is in the ephemeral port range (https://en.wikipedia.org/wiki/Ephemeral_port) of IANA/Linux/BSD, so there really shouldn't be any conflicts when running the tests. Especially on CI.

PR #1405, opened 2023-05-08
Change installation location of caf-tools to ${CMAKE_INSTALL_BINDIR}

topazus:


Neverlord:
@topazus is this to make packaging easier?

topazus:
Yes. If someone makes the actor-framework package and enables building the utility tools, it would be good to make the `-tools` subpackage to store the tools without moving their location, which the binary files are normally in `/usr/bin`

PR #1411, opened 2023-05-15
Remove hard-coded default cipher lists for OpenSSL

### LINKED ISSUE #1410: crypto-policy-non-compliance-openssl /usr/lib64/libcaf_openssl.so.0.19.1 SSL_CTX_set_cipher_list, packaging for Fedora
I tried to make RPM package of the library on Fedora, rpmlint tool reported built RPM package with `actor-framework.x86_64: W: crypto-policy-non-compliance-openssl /usr/lib64/libcaf_openssl.so.0.19.1 SSL_CTX_set_cipher_list` which needs to be properly fixed.

ref: https://download.copr.fedorainfracloud.org/results/@fedora-review/fedora-review-2196274-actor-framework/fedora-rawhide-x86_64/05905027-actor-framework/fedora-review/review.txt
https://bugzilla.redhat.com/show_bug.cgi?id=2196274#c5
https://docs.fedoraproject.org/en-US/packaging-guidelines/CryptoPolicies/#_cc_applications

Neverlord:
@topazus with this change, CAF should be in line with https://docs.fedoraproject.org/en-US/packaging-guidelines/CryptoPolicies/#_cc_applications: don't call `SSL_CTX_set_cipher_list` unless the user explicitly asks for it.

Closes #1410.

### DIFF
--- a/libcaf_openssl/src/openssl/session.cpp
+++ b/libcaf_openssl/src/openssl/session.cpp
@@ -195,9 +195,9 @@ SSL_CTX* session::create_ssl_context() {
 #endif
   if (!ctx)
     CAF_RAISE_ERROR("cannot create OpenSSL context");
+  auto& cfg = sys_.config();
   if (sys_.openssl_manager().authentication_enabled()) {
     // Require valid certificates on both sides.
-    auto& cfg = sys_.config();
     if (!cfg.openssl_certificate.empty()
         && SSL_CTX_use_certificate_chain_file(ctx,
                                               cfg.openssl_certificate.c_str())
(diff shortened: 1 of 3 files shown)


libcaf_openssl/src/openssl/session.cpp
```diff
@@ -243,13 +241,12 @@ SSL_CTX* session::create_ssl_context() {
     SSL_CTX_set1_groups_list(ctx, "P-384");
 #  endif /* OPENSSL_VERSION_NUMBER < 0x10101000L */
 #endif
-#ifdef CAF_SSL_HAS_SECURITY_LEVEL
```
josephnoir, line 246: I think this is not used anywhere now and could be removed:
```
% rg CAF_SSL_HAS_SECURITY_LEVEL
libcaf_openssl/caf/openssl/session.hpp
21:#  define CAF_SSL_HAS_SECURITY_LEVEL
```

josephnoir (changes requested):

libcaf_openssl/src/openssl/session.cpp
Neverlord: Good catch! Removed. 👍

josephnoir (approved):

PR #1412, opened 2023-05-16
Fix initialization on Windows and enable Robot tests on Windows CI

### LINKED ISSUE #1409: Example "chat-server" have a 10093 runtime exception on windows
I build the projects on Win11 by native complier, and run example chat-server, I got a runtime exception, calling socket() at make_stream_socket_pair(), get a 10093([WSANOTINITIALISED]) error.

Please tell me where and when call WSAStartup(), thanks.

Neverlord:
Closes #1409.

@hoskylucky sorry for the inconvenience. This will be included 0.19.2, but feel free to try the branch (or `master` after the merge). 🙂

### DIFF
--- a/.ci/windows/Dockerfile
+++ b/.ci/windows/Dockerfile
@@ -13,8 +13,15 @@ RUN Set-ExecutionPolicy Bypass -Scope Process -Force; \
 RUN choco install -y --no-progress visualstudio2022buildtools
 RUN choco install -y --no-progress visualstudio2022-workload-vctools
 
-# Install Git and OpenSSL.
+# Install and OpenSSL and Python.
 RUN choco install -y --no-progress msysgit openssl
+RUN choco install -y --no-progress python
+
+# Install Robot Framework and dependencies.
+RUN python.exe -m pip install \
+      robotframework \
+      robotframework-requests \
+      robotframework-websocketclient
 
 # Add git to the global PATH
 SHELL [ "cmd", "/c" ]
--- a/.ci/windows/Dockerfile
+++ b/.ci/windows/Dockerfile
@@ -13,9 +13,8 @@ RUN Set-ExecutionPolicy Bypass -Scope Process -Force; \
 RUN choco install -y --no-progress visualstudio2022buildtools
 RUN choco install -y --no-progress visualstudio2022-workload-vctools
 
-# Install and OpenSSL and Python.
-RUN choco install -y --no-progress msysgit openssl
-RUN choco install -y --no-progress python
+# Install Git, OpenSSL and Python.
+RUN choco install -y --no-progress msysgit openssl python
 
 # Install Robot Framework and dependencies.
 RUN python.exe -m pip install \
--- a/examples/http/time-server.cpp
+++ b/examples/http/time-server.cpp
@@ -77,7 +77,8 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
   }
   // Note: the actor system will only wait for actors on default. Since we don't
   // start actors, we need to block on something else.
-  std::cout << "Server is up and running. Press <enter> to shut down.\n";
+  std::cout << "Server is up and running. Press <enter> to shut down."
+            << std::endl;
   getchar();
   std::cout << "Terminating.\n";
   return EXIT_SUCCESS;
--- a/libcaf_core/caf/actor_system.hpp
+++ b/libcaf_core/caf/actor_system.hpp
@@ -109,6 +109,39 @@ public:
   actor_system(const actor_system&) = delete;
   actor_system& operator=(const actor_system&) = delete;
 
+  /// Calls a cleanup function in its destructor for cleaning up global state.
+  class [[nodiscard]] global_state_guard {
+  public:
+    using void_fun_t = void (*)();
+
+    explicit global_state_guard(void_fun_t f) : fun_(f) {
+      // nop
+    }
+
+    global_state_guard(global_state_guard&& other) noexcept : fun_(other.fun_) {
+      other.fun_ = nullptr;
+    }
+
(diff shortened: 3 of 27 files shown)


.ci/windows/Dockerfile
```diff
@@ -13,8 +13,15 @@ RUN Set-ExecutionPolicy Bypass -Scope Process -Force; \
 RUN choco install -y --no-progress visualstudio2022buildtools
 RUN choco install -y --no-progress visualstudio2022-workload-vctools
 
-# Install Git and OpenSSL.
+# Install and OpenSSL and Python.
```
josephnoir, line 16: Isn't `msysgit` still git; or why did you remove the git comment?

examples/http/time-server.cpp
```diff
@@ -77,7 +77,8 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
   }
   // Note: the actor system will only wait for actors on default. Since we don't
   // start actors, we need to block on something else.
-  std::cout << "Server is up and running. Press <enter> to shut down.\n";
+  std::cout << "Server is up and running. Press <enter> to shut down."
+            << std::endl;
```
josephnoir, line 81: Just curious: Why?

libcaf_core/caf/actor_system.hpp
```diff
@@ -109,6 +109,39 @@ class CAF_CORE_EXPORT actor_system {
...
+    }
+
+    global_state_guard(global_state_guard&& other) noexcept : fun_(other.fun_) {
+      other.fun_ = nullptr;
+    }
+
+    global_state_guard& operator=(global_state_guard&& other) noexcept {
+      std::swap(fun_, other.fun_);
+      return *this;
+    }
```
josephnoir, line 128: I'm wondering if this could create a scenario where the move constructor will not call whatever was in `fun_` before, but the move assignment will. One swaps while the other one leaves `other.fun_` with a `nullptr`, thus discarding whatever `fun_` was previously?

libcaf_net/src/net/middleman.cpp
```diff
@@ -57,6 +59,15 @@ void middleman::init_global_meta_objects() {
   // nop
 }
 
+actor_system::global_state_guard middleman::init_host_system() {
+  this_host::startup();
+  ssl::startup();
+  return actor_system::global_state_guard{+[] {
```
josephnoir, line 65: What does the `+` do?

libcaf_core/caf/exec_main.hpp
```diff
@@ -105,8 +105,28 @@ int exec_main(F fun, int argc, char** argv) {
...
+}
+
+template <class... Module, class T, class... Ts>
+auto do_init_host_system(type_list<Module...>, type_list<T, Ts...>) {
+  if constexpr (detail::has_init_host_system_v<T>) {
+    return do_init_host_system(type_list<Module..., T>{}, type_list<Ts...>{});
+  } else {
+    return do_init_host_system(type_list<Module...>{}, type_list<Ts...>{});
+  }
+}
```
josephnoir, line 122: This dispatches the `init_host_system()` call to either `Module` or `T`?

josephnoir (changes requested):

examples/http/time-server.cpp
Neverlord: Because `stdout` isn't line-buffered on Windows. This had the effect that the examples ran but just didn't print anything when I tried them on Windows. Here's what past me had to say about the change in [a27358d](https://github.com/actor-framework/actor-framework/pull/1412/commits/a27358d36eaf2d329503e96122e25c8b0358e211):

> On Windows, STDOUT is "fully buffered" by default. This means that
messages don't appear after a newline on screen, like they do on UNIX
systems that use line buffering by default. Using `std::endl` makes sure
that written messages actually appear on screen for Windows users as
they should..

libcaf_net/src/net/middleman.cpp
Neverlord: Turn the lambda into a function pointer. https://stackoverflow.com/questions/18889028/a-positive-lambda-what-sorcery-is-this

libcaf_core/caf/exec_main.hpp
Neverlord: This entire thing is basically filtering the `type_list<...>`. Each `T` with `has_init_host_system_v<T> == true` ends up in the `Module...` list. Once the entire type list is consumed, the remainder is just `type_list<>` and the first overload kicks in to call `init_host_system()` on all remaining types. Or it just returns `std::make_tuple()` if nothing matched.

.ci/windows/Dockerfile
Neverlord: No idea what went wrong here. Fixed. :)

libcaf_core/caf/actor_system.hpp
Neverlord: The move constructor just transfers ownership. In the case of the move assignment operator, the operation should generally be equivalent to destroying `this` and then "re-create" it with the content of `other`.

So, when move-assigning something to a non-null guard, both cleanup functions should eventually be called. The two common ways to do that are by either just swapping with `other` or move `other` into a temporary first and then swapping. The latter makes sure that we release whatever resource is managed before leaving the operator. Since this guard is managing some global state, my thought was that it doesn't really matter when the calls to the cleanup functions happen as long as they do get called. So calling `swap` here seemed like the most efficient way to go about things.

libcaf_net/src/net/middleman.cpp
josephnoir: I looked on [cppreference](https://en.cppreference.com/w/cpp/language/lambda) and couldn't find it. Good to know!

josephnoir (approved):

PR #1416, opened 2023-05-27
Implement better utilities for timestamp parsing

Neverlord:
This adds time zone support when writing/reading ISO 8601 strings. So far, CAF handled timestamps quite poorly. We got away with it because they weren't many people exchanging timestamps with CAF over string input/output. This of course changed recently with HTTP, WebSocket and JSON support. This PR adds a dedicated class `caf::chrono::datetime` to represent times with UTC offset with a new parser for reading the subset of ISO 8601 that people most commonly use (i.e., `YYYY-MM-DDTHH:MM:SS.XXXXXX` with `Z` or `[+-]MM:SS` suffix).


libcaf_core/src/chrono.cpp
```diff
@@ -0,0 +1,288 @@
...
+  CAF_ASSERT(val < 1000);
+  *pos++ = static_cast<char>((val / 100) + '0');
+  *pos++ = static_cast<char>(((val % 100) / 10) + '0');
+  *pos++ = static_cast<char>((val % 10) + '0');
+  return pos;
+}
+
+char* print_localtime_impl(char* buf, size_t buf_size, time_t ts,
+                           int ns) noexcept {
+  // Max size of a timestamp is 29 bytes: "2019-12-31T23:59:59.111222333+01:00".
```
josephnoir, line 88:
```suggestion
  // Max size of a timestamp is 35 bytes: "2019-12-31T23:59:59.111222333+01:00".
```

libcaf_core/src/chrono.cpp
```diff
@@ -0,0 +1,288 @@
...
+         && is_valid_day(day, month, year) // Valid days depend on month & year.
+         && in_range(hour, 0, 23)          // Hour: 00 to 23.
+         && in_range(minute, 0, 59)        // Minute: 00 to 59.
+         && in_range(second, 0, 59)        // Second: 00 to 59.
+         && in_range(nanosecond, 0, 999'999'999);
+}
+
+bool datetime::equals(const datetime& other) const noexcept {
+  // Two timestamps can be equal even if they have different UTC offsets. Hence,
+  // we need to normalize them before comparing.
```
josephnoir, line 197: Is this covered by a test case?

libcaf_core/caf/chrono.hpp
```diff
@@ -0,0 +1,159 @@
...
+  /// constructed object is invalid.
+  bool valid() const noexcept;
+
+  /// Returns whether this object is equal to `other`.
+  bool equals(const datetime& other) const noexcept;
+
+  /// Parses a date and time string in ISO 8601 format.
+  /// @param str The string to parse.
+  /// @returns The parsed date and time.
+  error parse(std::string_view str);
```
josephnoir, line 115: For the [IP address](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/ipv4_address.hpp#L131) and [UUID](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/uuid.hpp#L154) classes this is a free function, do you think it makes sense to have this be a member in this case?

libcaf_core/caf/chrono.hpp
```diff
@@ -0,0 +1,159 @@
...
+    auto converted = sc::system_clock::from_time_t(to_time_t());
+    auto result = sc::time_point_cast<Duration>(converted);
+    auto ns = sc::nanoseconds{nanosecond};
+    result += sc::duration_cast<Duration>(ns);
+    return sc::time_point_cast<Duration>(result);
+  }
+
+private:
+  /// Converts this object to a time_t value.
+  time_t to_time_t() const noexcept;
```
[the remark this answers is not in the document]
Neverlord: This is just an implementation detail for `to_local_time`. I wouldn't want to commit to this API. Users that *really* want to have that conversion can just call `to_local_time` and convert that to `time_t` (basically chopping off sub-second precision).

libcaf_core/caf/chrono.hpp
```diff
@@ -0,0 +1,159 @@
...
+  template <class Duration = std::chrono::system_clock::duration>
+  auto to_local_time() const noexcept {
+    // TODO: consider using year_month_day once we have C++20.
+    namespace sc = std::chrono;
+    auto converted = sc::system_clock::from_time_t(to_time_t());
+    auto result = sc::time_point_cast<Duration>(converted);
+    auto ns = sc::nanoseconds{nanosecond};
+    result += sc::duration_cast<Duration>(ns);
+    return sc::time_point_cast<Duration>(result);
+  }
```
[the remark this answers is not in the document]
Neverlord: Negative years are not an error. They just mean the `datetime` is prior to the start of the epoch (1.1.1970). Calling this on an invalid object probably just gives you some "implementation-defined" garbage value, so it's not a good test. I've added a `@pre` entry to the docs.

libcaf_core/src/chrono.cpp
Neverlord: Good catch. I've added a new test case.

libcaf_core/caf/chrono.hpp
Neverlord: Uh, good thing that you caught that! 👍

Makes zero sense to have a member function here but a free function everywhere else. Fixed.

Neverlord:
The typo fixes here really weren't related to the changes. No idea why the spellchecker never complained about them before. Fixed them anyways as part of the PR to get CI passing.

josephnoir (approved):

Issue #1420, opened 2023-06-06
Update OS versions in the pipeline and bring back Jenkins


Neverlord:
- [x] Replace EOL Linux distributions with newer versions
- [x] Add FreeBSD 13 to the mix
- [x] Get back Jenkins and use it for the Linux workers

riemass:
Almalinux 8 still maintained reaching EOL 01 Mar 2029, but there is also Almalinux 9 reaching EOL 31 May 2032.
Also, CentOS 7 is still supported until 30 Jun 2024, with CentOS 9 being maintained and shifting to a roling release distribution. 

@Neverlord should I add Almalinux 9 and CentOS 9 Stream to the mix?

We can update:
- Fedora 35 -> 37 
- Fedora 36 -> 38

Since currently only 36 is being run by cirrus, I believe we should also only run 38 and delete 37/36.

Neverlord:
> @Neverlord should I add Almalinux 9 and CentOS 9 Stream to the mix?

Yes, but only after we switch the Linux builds back to our Jenkins. We have limited "credit" on Cirrus but not on our Jenkins. Let me resurrect our Jenkins setup this week. I'll make a PR to switch the Linux images over.

The plan is to have Windows, FreeBSD and macOS on Cirrus (because we don't want to maintain these ourselves) and all of the Linux versions over on Jenkins (they basically require no maintenance since we can launch them from Dockerfiles).

> We can update:
>
> - Fedora 35 -> 37
> - Fedora 36 -> 38

Yep, go ahead with those. 👍

You can also add FreeBSD 13 as a followup (plus macOS). Is FreeBSD 12 still supported? Otherwise it should get removed/replaced as well.

riemass:
FreeBSD is supported until end of this year (31.Dec.2023). We could have both until then.

Neverlord:
> @Neverlord should I add Almalinux 9 and CentOS 9 Stream to the mix?

@riemass on second thought, I think CentOS 9 Stream probably won't add much. It's a rolling release, isn't it? But I think Alpine would be a good idea since it's using a different C standard library. So I'd say AlmaLinux 9 + Alpine after #1431 lands and that should wrap up this ticket.

PR #1431, opened 2023-06-13
Bring back Jenkins for Linux builds

Neverlord:
Switch back to Jenkins for Linux builds:

- Add the `Jenkinsfile` back to the repo (and update it)
- Remove Linux builds from the Cirrus CI config
- Run Cirrus CI and GitHub actions only on PRs and commits to `master`

Relates #1420.

Jenkins also found build issues when building CAF with exceptions disabled. Fixed.


Jenkinsfile
```diff
@@ -0,0 +1,151 @@
...
+                'CAF_LOG_LEVEL:STRING=TRACE',
+                'CAF_ENABLE_ROBOT_TESTS:BOOL=ON',
+                'CAF_SANITIZERS:STRING=address',
+            ],
+            extraBuildEnv: [
+                'ASAN_OPTIONS=detect_leaks=1',
+            ],
+        ]],
+        // Extra debug build with static libs, UBSan and hardening flags.
+        ['fedora-38', [
```
riemass, line 92: Any way we could add some info on the build variant to the key found here. Like having the keys `fedora-38-static`, `fedora-38-noexcept` and `fedora-38-leak-check`. Here it is clear what is different from regular build, but when looking at Jenkins reports it would make more sense to see e.g. `fedora-38 passed` and `fedora-38-static` failed. 

If it means we have to keep multiple fedora-38 folders in the `.ci` directory then it's probably not worth it, but maybe something could be done in the `caf-continious-integration` library.

Neverlord: I like that idea. The string you put there is used to generate the path for finding the `Dockerfile`. However, we could implement a separator that allows adding extra information. For example: `fedora-38:asan`. The lib could then simply ignore the `:asan` part when looking for the Dockerfile. What do you think?

I'd like to do that as a followup, though, to keep this PR focused on the "bringing back" part.

riemass: The suffix separator would be easier to implement. Let's do that as some follow up then.

riemass (approved):

PR #1433, opened 2023-06-14
Tag debug builds on Jenkins with more context

### LINKED ISSUE #1432: Annotate non-release builds in Jenkins
Currently, all of our debug builds use the same Fedora image and thus all just show up as `fedora-38 debug`. However, these builds are actually using separate build flags. We have an ASAN build, a build with exceptions disabled, etc.

By having some way to add extra context to the build name, it would be easier to grasp which configuration exactly went wrong without looking at the build logs.

Context: https://github.com/actor-framework/actor-framework/pull/1431#discussion_r1228678831.

Neverlord:
Closes #1432.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ config = [
         'build',
         'tests',
     ],
-    // Default CMake flags by build type.
+    // Default CMake flags the builds.
     buildFlags: [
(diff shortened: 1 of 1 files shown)


riemass (approved):
Great, thanks :fire:

PR #1434, opened 2023-06-14
Remove windows from Cirrus. Add Github Action for Windows CI

riemass:
Draft, just to invoke the pipeline.

### DIFF
--- a/.ci/windows/build.cmd
+++ b/.ci/windows/build.cmd
@@ -1,7 +1,7 @@
 cmake.exe ^
   -S . ^
   -B build ^
-  -G "NMake Makefiles" ^
+  -G "Visual Studio 17 2022" ^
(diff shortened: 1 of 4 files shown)


riemass:
GH Actions don't give us number of cores as an environment variable, like cirrus does. There are workarounds to invoke a custom script to fetch the number. I've left that part out for now.

Neverlord (changes requested):
I would prefer keeping the `Visual Studio 17 2022` generator if at all possible. But it wouldn't be a showstopper.

riemass:
The tests failed because we can't locate robot framework. I'll fix that. Just wanted to point out that I'm glad we tested the check - it fails if the tests fail. (I actually wanted to trigger this manually to be sure)

Neverlord (approved):

PR #1442, opened 2023-06-16
Only exclude CAF options from --help output

Neverlord:
Tweak the behavior of `--help` and `--long-help`. I think the previous behavior to only print global options and options with a short name was quite unintuitive. It makes sense to not include the CAF options by default, since there are ~40 of them (depending on loaded modules) and most users would only care about the options that the application actually defines.


riemass (approved):
Definitely more intuitive this way.

PR #1445, opened 2023-06-17
Add missing compare overload

Neverlord:
Just a small addition for consistency. Also, I've reimplemented `int label::compare(const label& x)` so that all four `compare` overloads (the two in `label` and the two in `label_view`) now do the exact same thing.

### DIFF
--- a/libcaf_core/caf/telemetry/label.hpp
+++ b/libcaf_core/caf/telemetry/label.hpp
@@ -53,9 +53,15 @@ public:
 
   // -- comparison -------------------------------------------------------------
 
-  int compare(const label_view& x) const noexcept;
+  template <class T1, class T2>
+  static int compare(const T1& lhs, const T2& rhs) noexcept {
+    auto cmp1 = lhs.name().compare(rhs.name());
+    return cmp1 != 0 ? cmp1 : lhs.value().compare(rhs.value());
+  }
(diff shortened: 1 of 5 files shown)


libcaf_core/caf/telemetry/label.hpp
```diff
@@ -53,7 +53,15 @@ class CAF_CORE_EXPORT label : detail::comparable<label> {
 
   // -- comparison -------------------------------------------------------------
 
-  int compare(const label& x) const noexcept;
+  template <class T1, class T2>
+  static int compare(const T1& lhs, const T2& rhs) noexcept {
+    auto cmp1 = lhs.name().compare(rhs.name());
+    return cmp1 != 0 ? cmp1 : lhs.value().compare(rhs.value());
+  }
```
riemass, line 60: Could we move this implementation as a free function to the cpp file? The intent being to minimize overhead when including the hpp.

Neverlord: Unfortunately not, we need it from `label.cpp` and from `label_view.cpp`.

riemass: Too bad, still nice refactoring.

riemass (approved):

Issue #1440, opened 2023-06-15
Support config via environment variable


Neverlord:
Our current model for configurations is:

CLI Arguments > Config File > Hard-coded Defaults

However, another common way to configure applications is via environment variables. They usually sit between the config file and the CLI arguments:

CLI Arguments > Environment Variables > Config File > Hard-coded Defaults

Because they sit between CLI and the contents of the config file, user cannot easily implement this on their end.

dominiklohmann:
We've implemented something similar in a CAF-based application, but it took a lot of effort to get the precedence as you described.

If you're doing this, what are your plans for translating environment variables to options? Is it an automatic mapping (e.g., `CAF_FOO_BAR__BAZ` mapping to `caf.foo-bar.baz`), or is the mapping explicitly specified by the user when adding an option?

Neverlord:
> If you're doing this, what are your plans for translating environment variables to options? Is it an automatic mapping (e.g., `CAF_FOO_BAR__BAZ` mapping to `caf.foo-bar.baz`), or is the mapping explicitly specified by the user when adding an option?

Good question. Probably an auto-generated default with an option to override would make sense. Haven't thought about it too deeply yet. This came up in a different context recently and I've just wanted to add the ticket to not forget about it.

PR #1451, opened 2023-06-19
Enable --foo bar syntax

### LINKED ISSUE #1429: Allow `--foo bar` syntax in the CLI parser
Currently, the CAF parser accepts short options as `-f bar` or `-fbar` and long options as `--foo=bar`. Many parsers also allow `--foo bar`.

riemass:
Closes #1429

### DIFF
--- a/libcaf_core/caf/config_option.hpp
+++ b/libcaf_core/caf/config_option.hpp
@@ -105,8 +105,8 @@ private:
 };
 
 /// Finds `config_option` string with a matching long name in the argument list
-/// [`first`, `last`). Returns an `iterator` to the matching location and a 
-/// `string_view` of the option value if the entry is found, or a `iterator` 
+/// [`first`, `last`). Returns an `iterator` to the matching location and a
+/// `string_view` of the option value if the entry is found, or a `iterator`
 /// to `last` with an empty `string_view` otherwise.
 std::pair<std::vector<std::string>::const_iterator, std::string_view>
 find_by_long_name(const config_option& x,
--- a/libcaf_core/caf/config_option.hpp
+++ b/libcaf_core/caf/config_option.hpp
@@ -104,11 +104,10 @@ private:
   mutable void* value_;
 };
 
-/// Finds `config_option` string with a matching long name in the string
-/// argument list [`first`, `last`). Returns an `argument_iterator` to the
-/// matching location and a `string_view` of the option value if the entry is
-/// found, or a `arugment_iterator` to `last` with an empty `string_view`
-/// otherwise.
+/// Finds `config_option` string with a matching long name in the argument list
+/// [`first`, `last`). Returns an `iterator` to the matching location and a 
+/// `string_view` of the option value if the entry is found, or a `iterator` 
+/// to `last` with an empty `string_view` otherwise.
 std::pair<std::vector<std::string>::const_iterator, std::string_view>
 find_by_long_name(const config_option& x,
                   std::vector<std::string>::const_iterator first,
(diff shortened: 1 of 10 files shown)


libcaf_core/caf/config_option.hpp
```diff
@@ -104,37 +104,13 @@ class CAF_CORE_EXPORT config_option {
...
-  return {first, std::string_view{}};
-}
+/// Finds `config_option` string with a matching long name in the argument list
+/// [`first`, `last`). Returns an `iterator` to the matching location and a
+/// `string_view` of the option value if the entry is found, or a `iterator`
+/// to `last` with an empty `string_view` otherwise.
+std::pair<std::vector<std::string>::const_iterator, std::string_view>
+find_by_long_name(const config_option& x,
+                  std::vector<std::string>::const_iterator first,
+                  std::vector<std::string>::const_iterator last);
```
Neverlord, line 114: Good call turning this into a non-template function. It would also need `CAF_CORE_EXPORT`, but I would actually go one step further and lift this into a member function. Something like this:

```cpp
  // ... inside config_option ...

  /// An iterator over CLI arguments.
  using argument_iterator = std::vector<std::string>::const_iterator;

  /// Stores the result of a find operation. The option sets `begin == end`
  /// if the operation could not find a match.
  struct find_result {
    /// The begin of the matched range.
    argument_iterator begin;
  
    /// The end of the matched range.
    argument_iterator end;

    /// The value for the config option.
    std::string_view value;
  };

  /// Tries to find this option by its long name in `[first, last)`.
  find_result find_by_long_name(argument_iterator first,
                                argument_iterator last) const noexcept;
```

This also addresses an issue I have in the current caller code. Currently, `actor_system_config` goes looking for '=' and then guesses that it has to erase `{i, i + 2}` from the vector. By returning the matched range instead, we can do:

```cpp
auto [first, last, path] = find_by_long_name(*ptr, args.begin(), args.end());
if (first == last) {
  // ... not found ...
} else {
  auto path_str = std::string{path};
  args.erase(first, last);
}
```

libcaf_core/test/actor_system_config.cpp
```diff
@@ -150,6 +149,21 @@ CAF_TEST(file input overrides defaults but CLI args always win) {
...
+  CHECK(cfg.remainder.empty());
+  CHECK_EQ(get_or(cfg, "config-file", ""), "test-me");
+
+  cfg.clear();
+  err = cfg.parse(string_list{"--config-file", "test-me"});
+  CHECK_EQ(err, caf::sec::cannot_open_file);
+  CHECK(cfg.remainder.empty());
+  CHECK_EQ(get_or(cfg, "config-file", ""), "test-me");
+}
+
```
Neverlord, line 166: Unit tests shouldn't do file I/O. Did you consider adding this as a test to `robot/config/dump-config.robot`? The program should print something like `error while parsing CLI and file options: caf::sec::cannot_open_file("test-me")`, so we could just check the output contians `cannot_open_file`.

Neverlord (changes requested):

libcaf_core/caf/config_option.hpp
riemass: I had some similar ideas, returning a std::pair<iterator, iterator>. What I'd like is refactoring this use-case completely away and use only the newer config_option_set::parse function, but that would require a lot more effort and less gain. Maybe that could be done if we ever get to #1440.
It's probably for the best if I go with the proposed changes.

libcaf_core/test/actor_system_config.cpp
riemass: I've done robot. I'll remove this.

Issue #1394, opened 2023-03-29
WebSocket conformance testing


Neverlord:
By running the [Autobahn testsuite](https://github.com/crossbario/autobahn-testsuite) on our WebSocket implementation, we can probably identify some edge cases where our implementation doesn't comply to the RFC.

PR #1454, opened 2023-06-22
Fix autobahn No. 1.1.6 and 1.2.6

riemass:
Relates #1394 

The upper bound on medium data caused our tests to fail (it got encoded as large data). Autobahn also tests for values 0 and 125 so I've added these as well.

### DIFF
--- a/libcaf_net/test/detail/rfc6455.cpp
+++ b/libcaf_net/test/detail/rfc6455.cpp
@@ -60,6 +60,40 @@ CAF_TEST(masking) {
   CAF_CHECK_EQUAL(masked_data, data);
 }
 
+CAF_TEST(no mask key and no data) {
+  std::vector<uint8_t> data{};
+  byte_buffer out;
+  impl::assemble_frame(impl::binary_frame, 0, as_bytes(make_span(data)), out);
+  CAF_CHECK_EQUAL(out, bytes({
+                         0x82, // FIN + binary frame opcode
+                         0x00, // data size = 0
+                       }));
+  impl::header hdr;
+  CAF_CHECK_EQUAL(impl::decode_header(out, hdr), 2);
+  CAF_CHECK_EQUAL(hdr.fin, true);
+  CAF_CHECK_EQUAL(hdr.mask_key, 0u);
+  CAF_CHECK_EQUAL(hdr.opcode, impl::binary_frame);
+  CAF_CHECK_EQUAL(hdr.payload_len, data.size());
+}
+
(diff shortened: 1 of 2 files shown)


libcaf_net/test/detail/rfc6455.cpp
```diff
@@ -60,6 +60,40 @@ CAF_TEST(masking) {
   CAF_CHECK_EQUAL(masked_data, data);
 }
 
+CAF_TEST(no mask key and no data) {
+  std::vector<uint8_t> data{};
```
Neverlord, line 64: Style: calling the initializer is unnecessary. Looking at the file, I think it's just copy/paste, still wanting to point it out. In general:

```suggestion
  std::vector<uint8_t> data;
```

or

```suggestion
  auto data = std::vector<uint8_t>{};
```

libcaf_net/test/detail/rfc6455.cpp
```diff
@@ -96,6 +130,44 @@ CAF_TEST(valid mask key plus small data) {
...
+  CAF_CHECK_EQUAL(take(out, 10), bytes({
+                                   0x82, // FIN + binary frame opcode
+                                   0xFD, // MASKED + data size = 125
+                                   0xDE, 0xAD, 0xC0, 0xDE, // mask key,
+                                   0xFF, 0xFF, 0xFF, 0xFF, // masked data
+                                 }));
+  impl::header hdr;
+  CAF_CHECK_EQUAL(impl::decode_header(out, hdr), 6);
+  CAF_CHECK_EQUAL(hdr.fin, true);
+  CAF_CHECK_EQUAL(hdr.mask_key, 0xDEADC0DEu);
```
Neverlord, line 166:
```suggestion
  CAF_CHECK_EQUAL(hdr.mask_key, 0xDEADC0DE);
```

Integers with hexcode notation are always unsigned:

```
    using type_t = decltype(0xDEADC0DE);
    static_assert(std::is_same_v<type_t, unsigned>);
```

Neverlord (approved):
Just two nitpicks for future reference. Still approving, because there are multiple occurrences of these in the file prior to your additions. We should hunt these down eventually, but that's a different refactoring task.

Neverlord:
That was bad timing. You've pushed shortly after I've merged the branch and now GitHub is confused. I tried to fix it by force-pushing to the merged commit but GitHub still shows the wrong status. Please don't push to approved PRs. 🙂

This is not just "closed" but actually "merged".

PR #1458, opened 2023-06-26
Fix autobahn No. 2.5

riemass:
Ping (and control frames in general) can't cant have a payload larger than 125 bytes. Also, shutdown exit code in this case should be 1002 (protocol error).

### DIFF
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -57,6 +57,9 @@ struct CAF_NET_EXPORT rfc6455 {
                              uint8_t flags = fin_flag);
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
+
+  static bool is_control_frame(uint8_t opcode);
+
 };
 
 } // namespace caf::detail
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -58,8 +58,9 @@ struct CAF_NET_EXPORT rfc6455 {
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
 
-  static bool is_control_frame(uint8_t opcode);
-
+  static constexpr bool is_control_frame(uint8_t opcode) noexcept {
+    return opcode != text_frame && opcode != binary_frame;
+  }
 };
 
 } // namespace caf::detail
--- a/libcaf_net/src/detail/rfc6455.cpp
+++ b/libcaf_net/src/detail/rfc6455.cpp
@@ -84,6 +84,10 @@ ptrdiff_t rfc6455::decode_header(const_byte_span data, header& hdr) {
   bool masked = (byte2 & 0x80) != 0;
   auto len_field = byte2 & 0x7F;
   size_t header_length;
+  // control frames can only have payload up to 125 bytes
+  if (rfc6455::is_control_frame(hdr.opcode) && len_field > 125) {
+    return -1;
+  }
   if (len_field < 126) {
     header_length = 2 + (masked ? 4 : 0);
     hdr.payload_len = len_field;
@@ -133,6 +137,17 @@ ptrdiff_t rfc6455::decode_header(const_byte_span data, header& hdr) {
     default:
       return -1;
   }
+}
+
+bool rfc6455::is_control_frame(uint8_t opcode) {
+  switch (opcode) {
+    case ping:
+    case pong:
+    case connection_close:
+    case continuation_frame:
+      return true;
+  }
+  return false;
 }
 
 } // namespace caf::detail
--- a/libcaf_net/src/detail/rfc6455.cpp
+++ b/libcaf_net/src/detail/rfc6455.cpp
@@ -85,9 +85,8 @@ ptrdiff_t rfc6455::decode_header(const_byte_span data, header& hdr) {
   auto len_field = byte2 & 0x7F;
   size_t header_length;
   // control frames can only have payload up to 125 bytes
-  if (rfc6455::is_control_frame(hdr.opcode) && len_field > 125) {
+  if (rfc6455::is_control_frame(hdr.opcode) && len_field > 125)
     return -1;
-  }
   if (len_field < 126) {
     header_length = 2 + (masked ? 4 : 0);
     hdr.payload_len = len_field;
@@ -137,17 +136,6 @@ ptrdiff_t rfc6455::decode_header(const_byte_span data, header& hdr) {
     default:
       return -1;
   }
-}
-
-bool rfc6455::is_control_frame(uint8_t opcode) {
-  switch (opcode) {
-    case ping:
-    case pong:
-    case connection_close:
-    case continuation_frame:
-      return true;
-  }
-  return false;
 }
 
 } // namespace caf::detail
--- a/libcaf_net/src/net/web_socket/lower_layer.cpp
+++ b/libcaf_net/src/net/web_socket/lower_layer.cpp
@@ -6,6 +6,7 @@
 
 #include "caf/error.hpp"
 #include "caf/net/web_socket/status.hpp"
+#include "caf/sec.hpp"
 
 namespace caf::net::web_socket {
 
@@ -19,7 +20,11 @@ void lower_layer::shutdown() {
 }
 
 void lower_layer::shutdown(const error& reason) {
-  shutdown(status::unexpected_condition, to_string(reason));
+  if (reason.code() == static_cast<uint8_t>(sec::protocol_error)) {
+    shutdown(status::protocol_error, to_string(reason));
+  } else {
+    shutdown(status::unexpected_condition, to_string(reason));
+  }
 }
 
 } // namespace caf::net::web_socket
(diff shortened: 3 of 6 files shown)


libcaf_net/src/detail/rfc6455.cpp
```diff
@@ -84,6 +84,10 @@ ptrdiff_t rfc6455::decode_header(const_byte_span data, header& hdr) {
   bool masked = (byte2 & 0x80) != 0;
   auto len_field = byte2 & 0x7F;
   size_t header_length;
+  // control frames can only have payload up to 125 bytes
+  if (rfc6455::is_control_frame(hdr.opcode) && len_field > 125) {
+    return -1;
+  }
```
riemass, line 90: @Neverlord  This logic could also go to framing::consume? Only thing I think it's important is to detect it early and stop reading the payload.

libcaf_net/src/net/web_socket/lower_layer.cpp
```diff
@@ -19,7 +20,11 @@ void lower_layer::shutdown() {
 }
 
 void lower_layer::shutdown(const error& reason) {
-  shutdown(status::unexpected_condition, to_string(reason));
+  if (reason.code() == static_cast<uint8_t>(sec::protocol_error)) {
```
riemass, line 23: There are probably more edge cases I plan to add here on the go.

libcaf_net/test/net-test.cpp
```diff
@@ -82,7 +82,6 @@ ptrdiff_t mock_stream_transport::handle_input() {
     if (consumed < 0) {
       // Negative values indicate that the application encountered an
       // unrecoverable error.
-      up->abort(make_error(caf::sec::runtime_error, "consumed < 0"));
```
riemass, line 85: I wanted to check weather removing this is OK, since the framing layer will call abort on it's own. This just makes calling it twice with different error codes, so testing is harder.

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include "caf/net/web_socket/framing.hpp"
+
+#include "net-test.hpp"
+
+using namespace caf;
+using namespace std::literals;
+
+namespace {
+
+class app_t : public net::web_socket::upper_layer {
```
riemass, line 16: This is almost the same app_t as in client test. Only difference is in the abort call detection. I could elevate this to net-test? It seems different from the app_t in server tests, but I would really like to see only one app layer fake. Do you think it would be possible?

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+    uut = uut_layer.get();
+    transport = mock_stream_transport::make(std::move(uut_layer));
+  }
+
+  static const_byte_span make_test_data(uint64_t requested_size) {
+    static std::vector<std::byte> bytes{150, std::byte{0xFF}};
+    if (requested_size > bytes.size())
+      bytes.resize(requested_size, std::byte{0xFF});
+    return const_byte_span{bytes.data(), requested_size};
+  }
```
riemass, line 81: I don't know if static test data is acceptable, maybe I could elevate this into net-test and use it in rfc6455 tests? Or just to implement it in that file as well?

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+} // namespace
+
+BEGIN_FIXTURE_SCOPE(fixture)
+
+SCENARIO("the client sends a ping and receives a pong response") {
+  GIVEN("a valid WebSocket connection") {
+    std::vector<std::byte> ping_frame;
+    std::vector<std::byte> pong_frame;
+    CHECK_EQ(transport->start(nullptr), error{});
+    transport->configure_read(caf::net::receive_policy::up_to(2048u));
```
riemass, line 93: These two lines repeat in every test. Could I put them in the fixture ctor?

libcaf_net/src/detail/rfc6455.cpp
Neverlord: Intuitively it makes sense to have this here to have it centrally. But no preference either way.

libcaf_net/test/net/web_socket/framing.cpp
Neverlord: In general, a bit of duplication is OK in tests. However, if this exact app is used in multiple tests, feel free to move it to separate hpp/cpp files as a followup.

libcaf_net/test/net/web_socket/framing.cpp
Neverlord: If all of the tests will do exact this setup, sure. Makes the setup a bit less flexible, though.

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+TEST_CASE("calling shutdown with protocol_error sets status in close header") {
+  CHECK_EQ(transport->start(nullptr), error{});
+  transport->configure_read(caf::net::receive_policy::up_to(2048u));
+  uut->shutdown(make_error(sec::protocol_error));
+  detail::rfc6455::header hdr;
+  detail::rfc6455::decode_header(transport->output_buffer(), hdr);
+  CHECK_EQ(hdr.opcode, detail::rfc6455::connection_close);
+  CHECK(hdr.payload_len >= 2);
+  auto error_code = (static_cast<uint16_t>(transport->output_buffer()[2]) << 8)
+                    + static_cast<uint16_t>(transport->output_buffer()[3]);
```
Neverlord, line 147: 1) Since we only use this to compare values, I wouldn't bother with the uint16 type.
2) I think using `vector::at` here makes the code a bit easier to read. The bound checks are also nice to have in tests to get exceptions rather than ASAN traces.

```suggestion
  auto error_code = (transport->output_buffer().at(2) << 8)
                    + transport->output_buffer().at(3);
```

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+  CHECK_EQ(error_code,
+           static_cast<uint16_t>(net::web_socket::status::protocol_error));
```
Neverlord, line 149:
```suggestion
  CHECK_EQ(error_code,
           static_cast<int>(net::web_socket::status::protocol_error));
```

libcaf_net/test/net/web_socket/framing.cpp
```diff
@@ -0,0 +1,184 @@
...
+        detail::rfc6455::header hdr;
+        detail::rfc6455::decode_header(transport->output_buffer(), hdr);
+        MESSAGE("Buffer: " << transport->output_buffer());
+        CHECK_EQ(hdr.opcode, detail::rfc6455::connection_close);
+        CHECK(hdr.payload_len >= 2);
+        auto error_code
+          = (static_cast<uint16_t>(transport->output_buffer()[2]) << 8)
+            + static_cast<uint16_t>(transport->output_buffer()[3]);
+        CHECK_EQ(error_code, static_cast<uint16_t>(
+                               net::web_socket::status::protocol_error));
```
Neverlord, line 178: Same here: we don't need to take care of preserving uint16.

Neverlord (changes requested):
Mostly nitpicks, looks good otherwise.

libcaf_net/test/net/web_socket/framing.cpp
riemass: Right, we can use `int`. `static_cast` is still needed since we need to explicitly create `int` it from `std::byte`.

Neverlord: Ah, if this is a `std::byte`: the idiomatic conversion would be [`to_integer`](https://en.cppreference.com/w/cpp/types/byte).

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -57,6 +57,10 @@ struct CAF_NET_EXPORT rfc6455 {
                              uint8_t flags = fin_flag);
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
+
+  static bool is_control_frame(uint8_t opcode) {
```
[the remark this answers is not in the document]
riemass: I missed that.

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -57,6 +57,10 @@ struct CAF_NET_EXPORT rfc6455 {
                              uint8_t flags = fin_flag);
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
+
+  static constexpr bool is_control_frame(uint8_t opcode) {
```
Neverlord, line 61: The `noexcept` got lost from my last suggestion:

```suggestion
  static constexpr bool is_control_frame(uint8_t opcode) noexcept {
```

Neverlord (changes requested):

Neverlord (approved):

libcaf_net/caf/detail/rfc6455.hpp
riemass: Sorry about that. Should be fine now.

Issue #1452, opened 2023-06-19
Migrate from CAF_TEST to TEST_CASE


Neverlord:
Older tests still use the old `CAF_TEST` macro. We should refactor all of them to use `TEST_CASE` instead.

Neverlord:
We've decided to ditch the old unit test framework entirely, so this is obsolete now. 🙂

PR #1459, opened 2023-06-27
Refactor test cases and unify mock socket app

riemass:
Relates #1452. 
Refactor test cases with changes and suggestions from the last two PRs.

### DIFF
--- a/libcaf_net/test/net-test.cpp
+++ b/libcaf_net/test/net-test.cpp
@@ -120,16 +120,24 @@ ptrdiff_t mock_stream_transport::handle_input() {
 
 // -- mock_web_socket_app ------------------------------------------------------
 
-mock_web_socket_app::mock_web_socket_app(bool behave_as_server)
-  : behave_as_server(behave_as_server) {
+mock_web_socket_app::mock_web_socket_app(bool request_messages_on_start)
+  : request_messages_on_start(request_messages_on_start) {
   // nop
 }
 
 caf::error mock_web_socket_app::start(caf::net::web_socket::lower_layer* ll) {
   down = ll;
-  if (behave_as_server)
+  if (request_messages_on_start)
     down->request_messages();
   return caf::none;
+}
+
+void mock_web_socket_app::prepare_send() {
+  // nop
+}
+
+bool mock_web_socket_app::done_sending() {
+  return true;
 }
 
 caf::error
@@ -151,14 +159,8 @@ mock_web_socket_app::accept(const caf::net::http::request_header& hdr) {
 }
 
 void mock_web_socket_app::abort(const caf::error& reason) {
-  if (expect_abort_) {
-    has_aborted = true;
-    err = reason;
-    expect_abort_ = false;
-    down->shutdown(reason);
-  } else {
-    CAF_FAIL("app::abort called: " << reason);
-  }
+  abort_reason = reason;
+  down->shutdown(reason);
 }
 
 ptrdiff_t mock_web_socket_app::consume_text(std::string_view text) {
--- a/libcaf_net/test/net-test.cpp
+++ b/libcaf_net/test/net-test.cpp
@@ -118,6 +118,59 @@ ptrdiff_t mock_stream_transport::handle_input() {
   return result;
 }
 
+// -- mock_web_socket_app ------------------------------------------------------
+
+mock_web_socket_app::mock_web_socket_app(bool behave_as_server)
+  : behave_as_server(behave_as_server) {
+  // nop
+}
+
+caf::error mock_web_socket_app::start(caf::net::web_socket::lower_layer* ll) {
+  down = ll;
+  if (behave_as_server)
+    down->request_messages();
+  return caf::none;
+}
+
+caf::error
+mock_web_socket_app::accept(const caf::net::http::request_header& hdr) {
+  // Store the request information in cfg to evaluate them later.
+  auto& ws = cfg["web-socket"].as_dictionary();
+  put(ws, "method", to_rfc_string(hdr.method()));
+  put(ws, "path", std::string{hdr.path()});
+  put(ws, "query", hdr.query());
+  put(ws, "fragment", hdr.fragment());
+  put(ws, "http-version", hdr.version());
+  if (hdr.num_fields() > 0) {
+    auto& fields = ws["fields"].as_dictionary();
+    hdr.for_each_field([&fields](auto key, auto val) {
+      put(fields, std::string{key}, std::string{val});
+    });
+  }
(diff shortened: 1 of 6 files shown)


libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
...
+
+  std::string text_input;
+
+  caf::byte_buffer binary_input;
+
+  caf::net::web_socket::lower_layer* down = nullptr;
+
+  caf::settings cfg;
+
+  bool behave_as_server = false;
```
Neverlord, line 157: This looks a bit verbose to me. I'd prefer something a bit more concise:

```
  bool is_server = false;
```

However, this flag is confusing no matter the name. The class implements `web_socket::upper_layer::server`, so why would it be something else than a server?

Looking at the usage, I think this only determines whether the app should request items automatically when started. Right? In this case, it could be named something like:

```
  /// Configures whether the app calls `request_messages` on the lower layer
  /// automatically when started.
  bool request_messages_on_start = false;
```

libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
   caf::net::multiplexer* mpx_;
 };
 
+class mock_web_socket_app : public caf::net::web_socket::upper_layer::server {
+public:
+  // -- constructor ------------------------------------------------------------
+
+  mock_web_socket_app(bool behave_as_server);
```
Neverlord, line 113:
```suggestion
  explicit mock_web_socket_app(bool behave_as_server);
```

libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
...
+
+  void abort(const caf::error& reason) override;
+
+  ptrdiff_t consume_text(std::string_view text) override;
+
+  ptrdiff_t consume_binary(caf::byte_span bytes) override;
+
+  void expect_abort() {
+    expect_abort_ = true;
+  }
```
Neverlord, line 145: I would assume a getter if the function takes no arguments, not a setter:

```
  void expect_abort(bool value) {
    expect_abort_ = value;
  }
```

However, this is a public member variable. So it doesn't need getters or setters in the first place.

libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
...
+
+  caf::error start(caf::net::web_socket::lower_layer* ll) override;
+
+  // -- implementation ---------------------------------------------------------
+
+  caf::error accept(const caf::net::http::request_header& hdr) override;
+
+  void prepare_send() override {
+    // nop
+  }
```
Neverlord, line 131: Even if they are trivial, our convention is to put virtual functions always into the `.cpp` file (except for templates, of course).

libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
...
+
+  caf::settings cfg;
+
+  bool behave_as_server = false;
+
+  caf::error err;
+
+  bool has_aborted = false;
+
+  bool expect_abort_ = false;
```
Neverlord, line 163: Our convention is to use `_` suffix only for private member variables:

```suggestion
  bool expect_abort = false;
```

libcaf_net/test/net-test.hpp
```diff
@@ -105,6 +106,63 @@ class mock_stream_transport : public caf::net::octet_stream::lower_layer {
   caf::net::multiplexer* mpx_;
 };
 
+class mock_web_socket_app : public caf::net::web_socket::upper_layer::server {
```
Neverlord, line 109: Since this is a re-usable class now, a bit of docu would be nice. :)

```
/// Implements a trivial WebSocket application that stores the contents of all received text frames in ...
class mock_web_socket_app : public caf::net::web_socket::upper_layer::server {
```

libcaf_net/test/net-test.cpp
```diff
@@ -118,6 +118,59 @@ ptrdiff_t mock_stream_transport::handle_input() {
...
+}
+
+void mock_web_socket_app::abort(const caf::error& reason) {
+  if (expect_abort_) {
+    has_aborted = true;
+    err = reason;
+    expect_abort_ = false;
+    down->shutdown(reason);
+  } else {
+    CAF_FAIL("app::abort called: " << reason);
```
Neverlord, line 160: "Hiding" a `CAF_FAIL` in a `.cpp` file outside of the test suite itself feels a bit off. I think dropping the `expect_abort_` logic here and just checking `has_aborted` in the tests will make them much easier to maintain.

libcaf_net/test/net/web_socket/server.cpp
```diff
@@ -20,67 +20,10 @@ namespace {
...
-    binary_input.insert(binary_input.end(), bytes.begin(), bytes.end());
-    return static_cast<ptrdiff_t>(bytes.size());
-  }
-};
-
 struct fixture {
   fixture() {
     using namespace caf::net;
-    auto app_ptr = app_t::make();
+    auto app_ptr = mock_web_socket_app::make(/*behave_as_server=*/true);
```
Neverlord, line 26: Please avoid C-style comments.

```
    auto app_ptr = mock_web_socket_app::make(true);
```

If you are looking to make the code more readable, why not move away from passing a boolean flag and use a tag type instead?

```
    // in net-test.hpp
    struct request_messages_on_start_t {};
    constexpr auto request_messages_on_start = request_messages_on_start_{};

    // here
    auto app_ptr = mock_web_socket_app::make(request_messages_on_start);
```

And have to `make` overloads (one taking no argument and one taking the tag type).

Neverlord (changes requested):

libcaf_net/test/net-test.cpp
riemass: I left this as a safeguard since we might forget to expect that the connection did not abort, but I'll add it to the rest of the tests.

libcaf_net/test/net-test.hpp
riemass: Aside from the `request_messages`, one more virtual member function is implemented in the server interface, and when using it as a client, it will simply never be called. I'll rename the `bool`, should be clear enough.

Neverlord (approved):

PR #1460, opened 2023-06-27
Add utility for UTF-8 validation

Neverlord:
@riemass I've looked a bit around to see how hard verifying UTF-8 input. Wasn't that hard. 🙂

Hope that gets you started on the `autobahn` tests that require UTF-8 validation.


riemass (approved):

riemass:
Thanks a lot (even though you handled the fun part). The provided functions should be enough, since we only need to see if the payload is valid utf8.

PR #1463, opened 2023-06-29
Add GitHub action for code coverage via codecov

Neverlord:


Neverlord:
Overall coverage sits at 66% at the moment. The `caf-core` is at ~72% coverage (still not great), but first we should prioritize bringing up coverage in `caf-net`.

riemass (approved):
LGTM

PR #1469, opened 2023-07-04
Add a new abstraction: publishers

Neverlord:
Currently, it's very hard to connect multiple asynchronous subscribers to an `observable`. Users basically have to create an SPSC buffer for each subscriber and then set up everything manually. The new publisher abstraction removes the need for all that manual setup. A `publisher` is basically the concurrency-only version of a `stream`.

Since we use the term "publisher" now to mean something asynchronous, I've also renamed the `item_publisher` to `multicaster` to avoid confusion. I think the new name also better reflects the purpose of the class.


riemass (approved):
LGTM

PR #1475, opened 2023-07-09
Add format API scaffolding

Neverlord:
This is prep work to enable upcoming additions.

The new `detail::format` and `detail::format_to` functions enable format-based outputs in the new unit testing framework and gives us the option to re-write the logging API without relying on the `std::ostream` interfaces. These functions provide a thin wrapper around `std::format`. In case the standard version is not available, we fall back to a minimal implementation that delegates formatting of floats and integers to `snprintf`.


libcaf_core/caf/detail/format.hpp
```diff
@@ -0,0 +1,121 @@
...
+    return format_arg{arg};
+  } else if constexpr (std::is_same_v<T, std::string>) {
+    return format_arg{std::string_view{arg}};
+  } else if constexpr (std::is_same_v<T, std::string_view>) {
+    return format_arg{arg};
+  } else {
+    static_assert(std::is_pointer_v<T>, "unsupported argument type");
+    return format_arg{static_cast<const void*>(arg)};
+  }
+}
```
riemass, line 77: we could merge the same branches with `is_one_of`? Something in the sense of:
```
if constexpr (caf::detail::is_one_of_v<T, bool, char, const char*, std::string_view>) {
return format_arg{arg};
} // ...
```

riemass (approved):
Out of all the `codecov` reports, I would point out that we didn't test formatting `u64_t` types and never triggered `after_precision_index` (neither state or the member function).

Aside from that, great PR. One more thing to get rid of when we hit `c++20` :smile: I still approve, so you can estimate how serious this is.

CMakeLists.txt
```diff
@@ -158,6 +159,26 @@ if(NOT CMAKE_CROSSCOMPILING)
...
+    if("${snippet_output}" STREQUAL "ok")
+      message(STATUS "Enable std::format support")
+      set(CAF_USE_STD_FORMAT ON CACHE BOOL "Enable std::format support" FORCE)
+    else()
+      message(STATUS "Disable std::format support: NOT available")
+    endif()
+  endif()
+else()
+  set(CAF_USE_STD_FORMAT OFF CACHE BOOL "Enable std::format support" FORCE)
+endif()
```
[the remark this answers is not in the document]
Neverlord: Uh, good catch. This makes absolutely no sense, must be some refactoring artifact.

libcaf_core/caf/detail/format.hpp
Neverlord: Yeah, good idea!

riemass (approved):

PR #1477, opened 2023-07-10
Validate UTF-8 on WebSocket text payloads

riemass:
Relates #1394.

### DIFF
--- a/libcaf_core/caf/sec.hpp
+++ b/libcaf_core/caf/sec.hpp
@@ -105,8 +105,8 @@ enum class sec : uint8_t {
   incompatible_versions,
   /// Connection refused because of incompatible application IDs.
   incompatible_application_ids,
-  /// The middleman received a malformed BASP message from another node.
-  malformed_basp_message,
+  /// Received a malformed message from another node.
+  malformed_message,
   /// The middleman closed a connection because it failed to serialize or
   /// deserialize a payload.
   serializing_basp_payload_failed = 50,
--- a/libcaf_core/src/detail/rfc3629.cpp
+++ b/libcaf_core/src/detail/rfc3629.cpp
@@ -32,27 +32,6 @@ constexpr std::byte head(std::byte value) noexcept {
   }
 }
 
-// Takes the last N bits of `value`.
-template <size_t N>
-constexpr std::byte tail(std::byte value) noexcept {
-  if constexpr (N == 1) {
-    return value & 0b0000'0001_b;
-  } else if constexpr (N == 2) {
-    return value & 0b0000'0011_b;
-  } else if constexpr (N == 3) {
-    return value & 0b0000'0111_b;
-  } else if constexpr (N == 4) {
-    return value & 0b0000'1111_b;
-  } else if constexpr (N == 5) {
-    return value & 0b0001'1111_b;
-  } else if constexpr (N == 6) {
-    return value & 0b0011'1111_b;
-  } else {
-    static_assert(N == 7);
-    return value & 0b0111'1111_b;
-  }
-}
-
 // Checks whether `value` is an UTF-8 continuation byte.
 constexpr bool is_continuation_byte(std::byte value) noexcept {
   return head<2>(value) == 0b1000'0000_b;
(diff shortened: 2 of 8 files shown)


libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -42,9 +63,6 @@ constexpr bool is_continuation_byte(std::byte value) noexcept {
 bool validate_rfc3629(const std::byte* first, const std::byte* last) {
   while (first != last) {
     auto x = *first++;
-    // Null byte (terminator) is not allowed.
-    if (x == 0_b)
-      return false;
```
Neverlord, line 47: Why did you drop this?

libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -72,18 +90,25 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
         return false;
       continue;
     }
-    // 1111'0bxx: 4-byte sequence.
+    // 1111'0xxx: 4-byte sequence.
     if (head<5>(x) == 0b1111'0000_b) {
+      uint64_t code_point = std::to_integer<uint64_t>(tail<3>(x)) << 18;
```
Neverlord, line 95: I don't think we need to compute this here. We won't read this value until line 101. Why not move it below and merge with the first `|=`?

libcaf_net/src/net/web_socket/framing.cpp
```diff
@@ -201,6 +202,10 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
     case detail::rfc6455::text_frame: {
       std::string_view text{reinterpret_cast<const char*>(payload.data()),
                             payload.size()};
+      if (!detail::rfc3629::valid(text)) {
+        abort_and_shutdown(sec::runtime_error, "invalid UTF-8 sequence");
```
Neverlord, line 206:
```suggestion
        abort_and_shutdown(sec::runtime_error, "invalid UTF-8 sequence");
```

I think `runtime_error` is too generic. Maybe `invalid_encoding`? We could add this as a new enumerator to `sec`. I think that would be still general enough to be useful in other contests.

libcaf_core/src/detail/rfc3629.cpp
riemass: `0` is the first UTF code point, and is valid, at least according to Wikipedia and autobahn. 
There is an autobahn test that sends a payload of length 1 containing `0x00` so it made sense to remove it.

libcaf_net/src/net/web_socket/framing.cpp
riemass: I didn't dare add a new `sec`, but yeah, I agree that this would be best. There is `malformed_basp_message`, maybe generalize this to `malformed_message` and use it everywhere, protocol agnostic?

Neverlord: I like that idea. 👍

Go ahead and rename `malformed_basp_message` to `malformed_message`. It's definitely been overspecific. Also rename the enumerator in `io::basp::connection_state` to keep it consistent.

libcaf_core/src/detail/rfc3629.cpp
Neverlord: Well then. 🤷🏻‍♂️

libcaf_core/caf/sec.hpp
```diff
@@ -106,7 +106,7 @@ enum class sec : uint8_t {
   /// Connection refused because of incompatible application IDs.
   incompatible_application_ids,
   /// The middleman received a malformed BASP message from another node.
```
Neverlord, line 108:
```suggestion
  /// Received a malformed message from another node.
```

libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -72,13 +69,19 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
         return false;
       continue;
     }
-    // 1111'0bxx: 4-byte sequence.
+    // 1111'0xxx: 4-byte sequence.
     if (head<5>(x) == 0b1111'0000_b) {
+      // Out of valid UTF range
```
Neverlord, line 74: I had to read this comment a few times, at least to me it reads a bit ambiguous. Maybe this?

```suggestion
      // Check if code point is in the valid UTF range.
```

libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -72,13 +69,19 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
...
+      if (x > std::byte{0xf4})
+        return false;
       if (first == last || !is_continuation_byte(*first))
         return false;
       // No non-shortest form.
       if (x == 0b1111'0000_b && head<4>(*first) == 0b1000'0000_b)
         return false;
+      // Out of valid UTF range
```
Neverlord, line 82:
```suggestion
      // Check if code point is in the valid UTF range.
```

Neverlord (changes requested):

Neverlord:
@riemass the `codecov/patch` failing comes from the renaming. We'll ignore it for this PR, since it's really unrelated code parts that just happened to be touched for the re-purposing of the enumerator.

Neverlord (approved):

PR #1480, opened 2023-07-11
Integrate the new unit testing framework

Neverlord:
To avoid ABI symbol clashes, I've pushed the old unit testing classes into the namespace `caf::test::legacy` with an inline namespace to keep the API compatible.

All tests that use the old framework were moved to a "legacy" test binary. We'll probably keep this around for some time until we have fully shifted over. I picked one first test per module to migrate it to the new framework. Basically just picked what seemed trivial. The one exception is the OpenSSL module, because it has only two tests and neither one seemed easy to shift over. So I left that out for now.

Also added some missing exports.


libcaf_test/src/test/registry.cpp
```diff
@@ -22,8 +31,25 @@ registry::suites_map registry::suites() {
...
 
-std::unique_ptr<factory> registry::head_;
+ptrdiff_t registry::add(factory* new_factory) {
+  if (head_ == nullptr) {
+    head_ = new_factory;
+    tail_ = head_;
+  } else {
+    tail_->next_ = new_factory;
+    tail_ = new_factory;
+  }
```
riemass, line 41: Just a nitpick, but maybe you will like this simplification

```suggestion
  if (head_ == nullptr)
    head_ = new_factory;
  else
    tail_->next_ = new_factory;
  tail_ = new_factory;
```

libcaf_net/test/detail/rfc6455.cpp
```diff
@@ -18,193 +16,191 @@
...
+    result.emplace_back(static_cast<std::byte>(x));
+  return result;
+}
 
-BEGIN_FIXTURE_SCOPE(fixture)
+template <class T>
+auto take(const T& xs, size_t num_bytes) {
+  auto n = std::min(xs.size(), num_bytes);
+  return std::vector<typename T::value_type>{xs.begin(), xs.begin() + n};
+}
```
riemass, line 34: Should we add an anonymous namespace here?

Neverlord: No. 🙂

`SUITE("detail.rfc6455")` already makes sure that everything in the suite has internal linkage.

riemass (approved):

PR #1481, opened 2023-07-12
Implement fragmented UTF-8 validation

riemass:
Relates #1394 
Modified UTF-8 validation for soft and hard failure.

### DIFF
--- a/libcaf_core/caf/detail/rfc3629.hpp
+++ b/libcaf_core/caf/detail/rfc3629.hpp
@@ -22,6 +22,16 @@ public:
   static bool valid(std::string_view str) noexcept {
     return valid(as_bytes(make_span(str)));
   }
+
+  /// Checks whether `bytes` is a valid UTF-8 string.
+  static auto validate(const_byte_span bytes) noexcept
+    -> std::pair<size_t, bool>;
+
+  /// Checks whether `str` is a valid UTF-8 string.
+  static auto validate(std::string_view str) noexcept
+    -> std::pair<size_t, bool> {
+    return validate(as_bytes(make_span(str)));
+  }
 };
 
 } // namespace caf::detail
--- a/libcaf_core/caf/detail/rfc3629.hpp
+++ b/libcaf_core/caf/detail/rfc3629.hpp
@@ -24,12 +24,10 @@ public:
   }
 
   /// Checks whether `bytes` is a valid UTF-8 string.
-  static auto validate(const_byte_span bytes) noexcept
-    -> std::pair<size_t, bool>;
+  static std::pair<size_t, bool> validate(const_byte_span bytes) noexcept;
 
   /// Checks whether `str` is a valid UTF-8 string.
-  static auto validate(std::string_view str) noexcept
-    -> std::pair<size_t, bool> {
+  static std::pair<size_t, bool> validate(std::string_view str) noexcept {
     return validate(as_bytes(make_span(str)));
   }
 };
--- a/libcaf_core/src/detail/rfc3629.cpp
+++ b/libcaf_core/src/detail/rfc3629.cpp
@@ -39,8 +39,13 @@ constexpr bool is_continuation_byte(std::byte value) noexcept {
 
 // The following code is based on the algorithm described in
 // http://unicode.org/mail-arch/unicode-ml/y2003-m02/att-0467/01-The_Algorithm_to_Valide_an_UTF-8_String
-bool validate_rfc3629(const std::byte* first, const std::byte* last) {
+// Returns a pair consisting of the last iterator to where the range is valid,
+// and a boolean statin whether the validation failed because of incomplete
+// data, or other failuers like malformed byte encoding.
+std::pair<const std::byte*, bool> validate_rfc3629(const std::byte* first,
+                                                   const std::byte* last) {
   while (first != last) {
+    auto checkpoint = first;
     auto x = *first++;
     // First bit is zero: ASCII character.
     if (head<1>(x) == 0b0000'0000_b)
@@ -49,49 +54,61 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
     if (head<3>(x) == 0b1100'0000_b) {
       // No non-shortest form.
       if (head<7>(x) == 0b1100'0000_b)
-        return false;
-      if (first == last || !is_continuation_byte(*first++))
-        return false;
+        return std::pair(checkpoint, false);
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first++))
+        return std::pair(checkpoint, false);
       continue;
     }
     // 1110'xxxx: 3-byte sequence.
     if (head<4>(x) == 0b1110'0000_b) {
-      if (first == last || !is_continuation_byte(*first))
-        return false;
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first))
+        return std::pair(checkpoint, false);
       // No non-shortest form.
       if (x == 0b1110'0000_b && head<3>(*first) == 0b1000'0000_b)
-        return false;
+        return std::pair(checkpoint, false);
       // No surrogate characters.
       if (x == 0b1110'1101_b && head<3>(*first) == 0b1010'0000_b)
-        return false;
+        return std::pair(checkpoint, false);
       ++first;
-      if (first == last || !is_continuation_byte(*first++))
-        return false;
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first++))
+        return std::pair(checkpoint, false);
       continue;
     }
     // 1111'0xxx: 4-byte sequence.
     if (head<5>(x) == 0b1111'0000_b) {
       // Check if code point is in the valid UTF range.
       if (x > std::byte{0xf4})
-        return false;
-      if (first == last || !is_continuation_byte(*first))
-        return false;
+        return std::pair(checkpoint, false);
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first))
+        return std::pair(checkpoint, false);
       // No non-shortest form.
       if (x == 0b1111'0000_b && head<4>(*first) == 0b1000'0000_b)
-        return false;
+        return std::pair(checkpoint, false);
       // Check if code point is in the valid UTF range.
       if (x == std::byte{0xf4} && *first >= std::byte{0x90})
-        return false;
+        return std::pair(checkpoint, false);
       ++first;
-      if (first == last || !is_continuation_byte(*first++))
-        return false;
-      if (first == last || !is_continuation_byte(*first++))
-        return false;
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first++))
+        return std::pair(checkpoint, false);
+      if (first == last)
+        return std::pair(checkpoint, true);
+      if (!is_continuation_byte(*first++))
+        return std::pair(checkpoint, false);
       continue;
     }
-    return false;
+    return std::pair(checkpoint, false);
   }
-  return true;
+  return std::pair(last, false);
 }
 
 } // namespace
@@ -99,7 +116,13 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
 namespace caf::detail {
 
 bool rfc3629::valid(const_byte_span bytes) noexcept {
-  return validate_rfc3629(bytes.begin(), bytes.end());
+  return validate_rfc3629(bytes.begin(), bytes.end()).first == bytes.end();
+}
+
+auto rfc3629::validate(const_byte_span bytes) noexcept
+  -> std::pair<size_t, bool> {
+  auto [last, incomplete] = validate_rfc3629(bytes.begin(), bytes.end());
+  return std::pair(last - bytes.begin(), incomplete);
 }
 
 } // namespace caf::detail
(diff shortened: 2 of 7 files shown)


libcaf_core/caf/detail/rfc3629.hpp
```diff
@@ -22,6 +22,16 @@ class CAF_CORE_EXPORT rfc3629 {
   static bool valid(std::string_view str) noexcept {
     return valid(as_bytes(make_span(str)));
   }
+
+  /// Checks whether `bytes` is a valid UTF-8 string.
+  static auto validate(const_byte_span bytes) noexcept
+    -> std::pair<size_t, bool>;
```
Neverlord, line 28: I don't think we use trailing return type anywhere else in the code base. So for consistency:

```suggestion
  static std::pair<size_t, bool> validate(const_byte_span bytes) noexcept;
```

Same applies to the other functions.

libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -49,57 +54,75 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
     if (head<3>(x) == 0b1100'0000_b) {
       // No non-shortest form.
       if (head<7>(x) == 0b1100'0000_b)
-        return false;
-      if (first == last || !is_continuation_byte(*first++))
-        return false;
+        return std::pair(checkpoint, false);
```
Neverlord, line 57: 1) We prefer `{}` over `()` for initialization: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax.
2) We don't need to repeat the type name of the return type here:

```suggestion
        return {checkpoint, false};
```

Just tagging it here, but it applies to multiple sites.

Neverlord (changes requested):

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -162,3 +163,47 @@ TEST_CASE("invalid UTF-8 input") {
   CHECK(!valid_utf8(invalid_four_byte_9));
   CHECK(!valid_utf8(invalid_four_byte_10));
 }
+
+TEST_CASE("invalid UTF-8 input missing continuation bytes") {
+  CHECK_EQ(rfc3629::validate(invalid_two_byte_1), std::pair(0ul, true));
```
Neverlord, line 168: After looking at the error on MSVC (just in case you didn't think of this already): It might be a good idea to factor out the type.

```c++
using res_t = std::pair<size_t, bool>;
```

```suggestion
  CHECK_EQ(rfc3629::validate(invalid_two_byte_1), res_t(0, true));
```

This also helps making the code a bit cleaner since we can drop the `ul` suffixes. 🙂

riemass: Thanks a lot. I'll also shift to the new test framework to get rid of CHECK macros.

libcaf_core/src/detail/rfc3629.cpp
```diff
@@ -49,57 +54,74 @@ bool validate_rfc3629(const std::byte* first, const std::byte* last) {
...
 namespace caf::detail {
 
 bool rfc3629::valid(const_byte_span bytes) noexcept {
-  return validate_rfc3629(bytes.begin(), bytes.end());
+  return validate_rfc3629(bytes.begin(), bytes.end()).first == bytes.end();
+}
+
+std::pair<size_t, bool> rfc3629::validate(const_byte_span bytes) noexcept {
+  auto [last, incomplete] = validate_rfc3629(bytes.begin(), bytes.end());
+  return {last - bytes.begin(), incomplete};
```
Neverlord, line 124: This silently converts a singed to an unsigned type; should be explicit:

```suggestion
  return {static_cast<size_t>(last - bytes.begin()), incomplete};
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -2,15 +2,16 @@
...
-
 #include "caf/detail/rfc3629.hpp"
 
-#include "core-test.hpp"
+#include "caf/test/test.hpp"
 
 using namespace caf;
+using detail::rfc3629;
+
+using res_t = std::pair<size_t, bool>;
```
Neverlord, line 12: Now that we're on the new testing framework, what do you think of structuring this a bit more? Something like this:

```
TEST("rfc3629::valid checks whether an input is valid UTF-8") {
  SECTION("ASCII input") {
    ...
  }
  ...
}

TEST("rfc3629::validate stops at the first invalid byte") {
  SECTION("....") {
    ...
  }
}

TEST("rfc3629::validate stops at incomplete UTF-8 code points") {
  ...
}
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,87 @@ constexpr std::string_view ascii_2 = R"__(
...
+  check(!valid_utf8(invalid_four_byte_6));
+  check(!valid_utf8(invalid_four_byte_7));
+  check(!valid_utf8(invalid_four_byte_8));
+  check(!valid_utf8(invalid_four_byte_9));
+  check(!valid_utf8(invalid_four_byte_10));
+}
+
+TEST("invalid UTF-8 input missing continuation bytes") {
+  check_eq(rfc3629::validate(invalid_two_byte_1), res_t(0, true));
+  check_eq(rfc3629::validate(invalid_three_byte_1), res_t{0ul, true});
```
Neverlord, line 167: `res_t` has the type information, so we no longer need the `ul` suffies.

```suggestion
  check_eq(rfc3629::validate(invalid_three_byte_1), res_t{0, true});
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,87 @@ constexpr std::string_view ascii_2 = R"__(
...
+  check_eq(rfc3629::validate(invalid_four_byte_8), res_t{0ul, false});
+  check_eq(rfc3629::validate(invalid_four_byte_9), res_t{0ul, false});
+  check_eq(rfc3629::validate(invalid_four_byte_10), res_t{0ul, false});
+}
 
-TEST_CASE("ASCII input") {
-  CHECK(valid_utf8(ascii_1));
-  CHECK(valid_utf8(ascii_2));
+TEST("invalid UTF-8 input fails on the invalid byte") {
+  const_byte_span input{invalid_four_byte_9};
```
Neverlord, line 194: Coding convention: prefer `auto`:

```suggestion
  auto input = make_span(invalid_four_byte_9);
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,87 @@ constexpr std::string_view ascii_2 = R"__(
...
+  check_eq(rfc3629::validate(input.subspan(0, 2)), res_t{0ul, false});
+  input = invalid_four_byte_10;
```
Neverlord, line 196: I'd rather avoid re-assigning variables. It makes the code harder to comprehend / to understand control flow. In this, I think we don't even need this variable in the first place:

```
check_eq(rfc3629::validate(make_span(invalid_four_byte_10, 1)), res_t{0ul, false});
```

libcaf_net/caf/net/web_socket/framing.hpp
```diff
@@ -172,6 +174,9 @@ class CAF_NET_EXPORT framing : public octet_stream::upper_layer,
   /// Assembles fragmented payloads.
   binary_buffer payload_buf_;
 
+  /// Validated UTF-8 range for fragmented text payloads
+  size_t validation_point = 0ul;
```
Neverlord, line 178: I'm not quite sure whether I understand what this doing from the comment and declaration. Is this just the offset where we've stopped last time? In that case, naming it offset might be clearer:

```suggestion
  /// Stores where to resume the UTF-8 input validation.
  size_t validation_offset_ = 0;
```

libcaf_net/caf/net/web_socket/framing.hpp
```diff
@@ -152,10 +152,12 @@ class CAF_NET_EXPORT framing : public octet_stream::upper_layer,
     shutdown(err);
   }
 
+  bool validate_fragmented_utf8() noexcept;
```
Neverlord, line 155: From the function name, I would expect this to take a "fragmented UTF-8" as input and then validate it. But this function does bookkeeping, so maybe we find a name that makes this more clear? Also should have some documentation on it, since it's doing non-obvious side effects.

```suggestion
  /// Checks whether the current input is valid UTF-8. Stores the last position
  /// while scanning in order to avoid validating the same bytes again.
  bool payload_valid() noexcept;
```

Neverlord (changes requested):

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,107 @@ constexpr std::string_view ascii_2 = R"__(
...
-  CHECK(valid_utf8(valid_three_byte_1));
-  CHECK(valid_utf8(valid_three_byte_2));
-  CHECK(valid_utf8(valid_four_byte_1));
-  CHECK(valid_utf8(valid_four_byte_2));
+TEST("rfc3629::validate returns the end index if the range is valid") {
+  SECTION("valid ASCII input") {
+    check_eq(rfc3629::validate(ascii_1), res_t{ascii_1.size(), false});
+    check_eq(rfc3629::validate(ascii_2), res_t{ascii_2.size(), false});
+  }
+
```
Neverlord, line 172:
```suggestion
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,107 @@ constexpr std::string_view ascii_2 = R"__(
...
+TEST("rfc3629::validate stops at the first invalid byte") {
+  SECTION("UTF-8 input missing continuation bytes") {
+    check_eq(rfc3629::validate(invalid_two_byte_1), res_t{0, true});
+    check_eq(rfc3629::validate(invalid_three_byte_1), res_t{0, true});
+    check_eq(rfc3629::validate(invalid_three_byte_2), res_t{0, true});
+    check_eq(rfc3629::validate(invalid_four_byte_1), res_t{0, true});
+    check_eq(rfc3629::validate(invalid_four_byte_2), res_t{0, true});
+    check_eq(rfc3629::validate(invalid_four_byte_3), res_t{0, true});
+  }
+
```
Neverlord, line 192:
```suggestion
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,107 @@ constexpr std::string_view ascii_2 = R"__(
...
+    check_eq(rfc3629::validate(invalid_three_byte_8), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_4), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_5), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_6), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_7), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_8), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_9), res_t{0, false});
+    check_eq(rfc3629::validate(invalid_four_byte_10), res_t{0, false});
+  }
+
```
Neverlord, line 211:
```suggestion
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,107 @@ constexpr std::string_view ascii_2 = R"__(
...
+    byte_buffer data;
+    data.insert(data.end(), begin(valid_four_byte_1), end(valid_four_byte_1));
+    data.insert(data.end(), begin(valid_four_byte_2), end(valid_four_byte_2));
+    data.insert(data.end(), begin(valid_two_byte_1), end(valid_two_byte_1));
+    data.insert(data.end(), begin(invalid_four_byte_4),
+                end(invalid_four_byte_4));
+    check_eq(rfc3629::validate(data), res_t{10, false});
+  }
+}
 }
```
Neverlord, line 229:
```suggestion
}

} // SUITE("detail.rfc3629")
```

libcaf_core/test/detail/rfc3629.cpp
```diff
@@ -122,43 +123,107 @@ constexpr std::string_view ascii_2 = R"__(
...
-  CHECK(valid_utf8(valid_four_byte_1));
-  CHECK(valid_utf8(valid_four_byte_2));
+TEST("rfc3629::validate returns the end index if the range is valid") {
+  SECTION("valid ASCII input") {
+    check_eq(rfc3629::validate(ascii_1), res_t{ascii_1.size(), false});
+    check_eq(rfc3629::validate(ascii_2), res_t{ascii_2.size(), false});
+  }
+
+  SECTION("valid UTF-8 input") {
+    check_eq(rfc3629::validate(valid_two_byte_1), res_t{2, false});
```
Neverlord, line 174: I'd like to avoid hard-coding/guessing the sizes here. :)

I think we could use either `std::size()` here or change the types of these things to `std::array`.

Neverlord (changes requested):

libcaf_core/test/detail/rfc3629.cpp
riemass: I taught the `two` in `valid_two_byte_1` was enough to indicate the size, but sure, I can use size. The downside is it might break the line in two.

Neverlord: Yeah, you're right. I think I didn't read the variable name. Keep it that way. 👍

Neverlord (approved):

PR #1482, opened 2023-07-12
Use the Canonical Project Structure for caf_test

Neverlord:
For the Canonical Project Structure, see https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html.

With the new structure, we drop the `src` sub-directory and place the source files next to the header files. Further, we place the unit test next to the source file with `.test.cpp` suffix.

Instead of assembling all unit tests into a single large binary and then have `ctest` pick suites individually via CLI argument, we simply create one binary per unit test now. This makes the entire CMake setup much easier.

This also frees up the `test` folder (that we will probably rename to `tests`) for module tests.

@riemass as discussed earlier today. 🙂

I think this structure makes things a lot easier going forward. And the self-contained binaries per suite makes it much easier going forward to writes tests that use custom types in CAF messages. Currently, only because some tests require a type ID block, we have to put stuff for unit tests into headers and have the `main` pick up the type ID block. This setup eliminates all of that headache. This also allows us to get rid of the `SUITE` block now, that basically just contained the path to the file.

While we are building more binary targets now, I think the build times could actually improve overall because the individual binaries are very small and we no longer have to link a binary with dozens of test suites (and thus very many symbols).

I did the refactoring for the `caf_test` module because it is the smallest to see how it would look like. I think it's a change for the better. If we agree that this is the way to go, we can migrate the other modules one by one.


riemass (approved):
Looks good to me. I looked at the final project and I think it should be easier for file navigation, fuzzy finders can search for `*.test.cpp` and file tree plugins will display the three files together, with the `.test.cpp` last. Looking forward to some more subfolder heavy modules, but it should be OK. 

I'm all in favor of one binary per test. We will see if there are any drawbacks in the future. I've seen big link jobs that can take over 8GB of RAM. Not that we have them in CAF, but I also think this will speed up things. Also, one can execute only the binary we're currently working on without suite filters.

PR #1488, opened 2023-07-13
Drop unnecessary arguments in octet_stream::policy

Neverlord:
The interface of `octet_stream::policy` still asks for the socket - but the policy should know best which socket it manages. 🙂


riemass (approved):
LGTM

Issue #1479, opened 2023-07-11
Implement missing features in the new test framework


Neverlord:
To actually replace the old framework and to allow us to to port the existing tests to the new framework, we need to implement these missing features:

- [x] `-s <regex>` for selecting suites (high priority)
- [x] `-t <regex>` for selecting tests
- [x] `-n` for disabling colored output
- [x] `-r <seconds>` for aborting the test after reaching the time limit
- [x] missing `check` functions such as `check_le`
- [x] a `fail` function to unconditionally stop the test manually with a message (same as a failed requirement check)
- [x] missing `require` functions (all of them)
- [x] printing on various severities: currently there's `info`, but that should probably be renamed to `print_info` and then complemented by `print_debug` etc.

We will also have to implement new fixtures, but this set of functionality should at least allow us to start porting tests that don't require much setup to the new framework to get some experience with it.

PR #1490, opened 2023-07-13
Added filtering for suites based on regex

shariarriday:
Relates #1479.

Implemented filtering based on std::regex. But, it does not support std::string_view properly. Each call to regex_search requires a call to std::string constructor

### DIFF
--- a/libcaf_test/caf/test/runner.cpp
+++ b/libcaf_test/caf/test/runner.cpp
@@ -3,6 +3,7 @@
 // https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
 
 #include "caf/test/runner.hpp"
+
 #include "caf/config_option.hpp"
 #include "caf/config_option_adder.hpp"
 #include "caf/config_option_set.hpp"
@@ -68,13 +69,19 @@ int runner::run(int argc, char** argv) {
   }
   default_reporter->start();
   auto enabled = [](const std::regex& selected,
-                    const std::string_view& search_string) {
+                    std::string_view search_string) {
     return std::regex_search(search_string.begin(), search_string.end(),
                              selected);
   };
-  auto suite_regex = parse_regex_opt(get_or(cfg_, "suites", ".*"));
+  auto suite_regex = runner::to_regex(get_or(cfg_, "suites", ".*"));
+  if (!suite_regex) {
+    using detail::format_to;
+    auto err = std::ostream_iterator<char>{std::cerr};
+    format_to(err, "regex could not be parsed\n");
+    return EXIT_FAILURE;
+  }
   for (auto& [suite_name, suite] : suites_) {
-    if (suite_regex.has_value() && !enabled(suite_regex.value(), suite_name))
+    if (!enabled(*suite_regex, suite_name))
       continue;
     default_reporter->begin_suite(suite_name);
     for (auto [test_name, factory_instance] : suite) {
@@ -166,16 +173,15 @@ runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
   return {true, false};
 }
 
-std::optional<std::regex>
-runner::parse_regex_opt(const std::string& regex_string) {
+std::optional<std::regex> runner::to_regex(std::string_view regex_string) {
 #ifdef CAF_ENABLE_EXCEPTIONS
   try {
-    return std::regex(regex_string);
+    return std::regex{regex_string.begin(), regex_string.end()};
   } catch (...) {
     return std::nullopt;
   }
 #else
-  return std::regex(regex_string);
+  return std::regex{regex_string.begin(), regex_string.end()};
 #endif
 }
 
--- a/libcaf_test/caf/test/runner.cpp
+++ b/libcaf_test/caf/test/runner.cpp
@@ -3,12 +3,10 @@
 // https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
 
 #include "caf/test/runner.hpp"
-
 #include "caf/config_option.hpp"
 #include "caf/config_option_adder.hpp"
 #include "caf/config_option_set.hpp"
 #include "caf/detail/log_level.hpp"
-#include "caf/settings.hpp"
 #include "caf/test/context.hpp"
 #include "caf/test/nesting_error.hpp"
 #include "caf/test/reporter.hpp"
@@ -16,6 +14,7 @@
 
 #include <iostream>
 #include <optional>
+#include <regex>
 #include <string>
 
 namespace caf::test {
@@ -68,7 +67,17 @@ int runner::run(int argc, char** argv) {
     return EXIT_SUCCESS;
   }
   default_reporter->start();
+  auto enabled = [](const std::regex& selected,
+                    const std::string& search_string) {
+    return std::regex_search(search_string, selected);
+  };
+  std::regex suite_regex;
+  if (auto suite_regex_string = get_as<std::string>(cfg_, "suites")) {
+    suite_regex = std::regex(*suite_regex_string);
+  }
   for (auto& [suite_name, suite] : suites_) {
+    if (!enabled(suite_regex, {suite_name.data(), suite_name.size()}))
+      continue;
     default_reporter->begin_suite(suite_name);
     for (auto [test_name, factory_instance] : suite) {
       auto state = std::make_shared<context>();
@@ -109,27 +118,26 @@ int runner::run(int argc, char** argv) {
 
 runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
   using detail::format_to;
-  caf::settings cfg;
   std::vector<std::string> args_cpy{argv + 1, argv + argc};
   auto options = make_option_set();
-  auto res = options.parse(cfg, args_cpy);
+  auto res = options.parse(cfg_, args_cpy);
   auto err = std::ostream_iterator<char>{std::cerr};
   if (res.first != caf::pec::success) {
     format_to(err, "error while parsing argument '{}': {}\n\n{}\n", *res.second,
               to_string(res.first), options.help_text());
     return {false, true};
   }
-  if (get_or(cfg, "help", false)) {
+  if (get_or(cfg_, "help", false)) {
     format_to(err, "{}\n", options.help_text());
     return {true, true};
   }
-  if (get_or(cfg, "available-suites", false)) {
+  if (get_or(cfg_, "available-suites", false)) {
     format_to(err, "available suites:\n");
     for (auto& [suite_name, suite] : suites_)
       format_to(err, "- {}\n", suite_name);
     return {true, true};
   }
-  if (auto suite_name = get_as<std::string>(cfg, "available-tests")) {
+  if (auto suite_name = get_as<std::string>(cfg_, "available-tests")) {
     auto i = suites_.find(*suite_name);
     if (i == suites_.end()) {
       format_to(err, "no such suite: {}\n", *suite_name);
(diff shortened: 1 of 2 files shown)


libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +68,17 @@ int runner::run(int argc, char** argv) {
     return EXIT_SUCCESS;
   }
   default_reporter->start();
+  auto enabled = [](const std::regex& selected,
+                    const std::string& search_string) {
+    return std::regex_search(search_string, selected);
+  };
```
Neverlord, line 73: While `regex` doesn't support `string_view` directly, you can still use string views and pass in an iterator pair: https://en.cppreference.com/w/cpp/regex/regex_search.

```suggestion
  auto enabled = [](const std::regex& selected,
                    std::string_view search_string) {
    return std::regex_search(search_string.begin(), search_string.end(), selected);
  };
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +67,17 @@ int runner::run(int argc, char** argv) {
...
+  std::regex suite_regex;
+  if (auto suite_regex_string = get_as<std::string>(cfg_, "suites")) {
+    suite_regex = std::regex(*suite_regex_string);
+  }
```
Neverlord, line 77: This can be shortened:

```suggestion
  auto suite_regex = std::regex{get_or(cfg_, "suites", ".*")};
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -140,7 +148,11 @@ runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
       format_to(err, "- {}\n", test_name);
     return {true, true};
   }
-  if (auto verbosity = get_as<std::string>(cfg, "verbosity")) {
+  if (auto suite_regex_string = get_as<std::string>(cfg_, "suites")) {
+    format_to(err, "suite regex: {}\n", *suite_regex_string);
+    return {true, true};
+  }
```
Neverlord, line 154: I'm not sure if I understand what you are trying to do here. Isn't this failing the run if "suites" is defined? We also parse this option later again.

Neverlord (changes requested):

libcaf_test/caf/test/runner.cpp
shariarriday: I was planning to add some type of verification of the regex string as regex simply crashes the application if something weird is passed to the application. Something simple like "*" throws exception currently. Let me know your thoughts on this.

Neverlord: Good point. I think we can do the verification in a lambda / member function like `optional<regex> parse_regex_opt("suites")` that then internally catches the exception (or calls `abort` when exceptions are off). My thinking is that we will eventually have two regex parameters (when adding `-t`) and doing the verification here would mean repeating the code.

libcaf_test/caf/test/runner.cpp
```diff
@@ -3,19 +3,18 @@
 // https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
 
 #include "caf/test/runner.hpp"
```
Neverlord, line 5: This is the main header. So that's one of the few places where we keep the empty line. 🙂

```suggestion
#include "caf/test/runner.hpp"

```

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +67,15 @@ int runner::run(int argc, char** argv) {
     return EXIT_SUCCESS;
   }
   default_reporter->start();
+  auto enabled = [](const std::regex& selected,
+                    const std::string_view& search_string) {
```
Neverlord, line 71: It's idiomatic to pass `string_view` as value, since it's just a pointer and a size:

```suggestion
                    std::string_view search_string) {
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +67,15 @@ int runner::run(int argc, char** argv) {
...
+    return std::regex_search(search_string.begin(), search_string.end(),
+                             selected);
+  };
+  auto suite_regex = parse_regex_opt(get_or(cfg_, "suites", ".*"));
```
Neverlord, line 75: With this implementation, we silently drop errors if users pass in invalid options. We should stop the application here instead and only carry on if the regex is valid:

```suggestion
  auto suite_regex = parse_regex_opt(get_or(cfg_, "suites", ".*"));
  if (!suite_regex) {
    // print error and return
  }
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +67,15 @@ int runner::run(int argc, char** argv) {
...
   for (auto& [suite_name, suite] : suites_) {
+    if (suite_regex.has_value() && !enabled(suite_regex.value(), suite_name))
+      continue;
```
Neverlord, line 78: That's a bit on the verbose side. We usually just use the automatic conversion to `bool` and `operator*`. As stated above, the function should return on an invalid configuration. So here, we know that the regex is valid:

```suggestion
    if (!enabled(*suite_regex, suite_name))
      continue;
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -160,4 +166,17 @@ runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
   return {true, false};
 }
 
+std::optional<std::regex>
+runner::parse_regex_opt(const std::string& regex_string) {
```
Neverlord, line 170: Please use string view here.

libcaf_test/caf/test/runner.cpp
```diff
@@ -160,4 +166,17 @@ runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
...
+#ifdef CAF_ENABLE_EXCEPTIONS
+  try {
+    return std::regex(regex_string);
```
Neverlord, line 173: Prefer initialization with `{}`: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-list.

libcaf_test/caf/test/runner.cpp
```diff
@@ -160,4 +166,17 @@ runner::parse_cli_result runner::parse_cli(int argc, char** argv) {
...
+  } catch (...) {
+    return std::nullopt;
+  }
+#else
+  return std::regex(regex_string);
```
Neverlord, line 178: Same here: `{}`.

libcaf_test/caf/test/runner.hpp
```diff
@@ -26,8 +30,10 @@ class CAF_TEST_EXPORT runner {
 
 private:
   parse_cli_result parse_cli(int argc, char** argv);
+  std::optional<std::regex> parse_regex_opt(const std::string& regex_string);
```
Neverlord, line 33: This isn't accessing any state, so it can be `static`. With this name, I would expect that the argument is actually the name of the option, as was my suggestion. Passing in the option value is also ok, but then this should have a different name, e.g., `to_regex`.

Neverlord (changes requested):

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,7 +68,21 @@ int runner::run(int argc, char** argv) {
...
+  auto enabled = [](const std::regex& selected,
+                    std::string_view search_string) {
+    return std::regex_search(search_string.begin(), search_string.end(),
+                             selected);
+  };
+  auto suite_regex = runner::to_regex(get_or(cfg_, "suites", ".*"));
+  if (!suite_regex) {
+    using detail::format_to;
+    auto err = std::ostream_iterator<char>{std::cerr};
+    format_to(err, "regex could not be parsed\n");
```
Neverlord, line 80: As a user, I wouldn't know what this should tell me. The error should give the user context so that he can actually fix what's wrong, but this description omits the critical information which option this error belongs to. Is there a specific reason why you deviated from the wording we use in [`parse_cli`](https://github.com/actor-framework/actor-framework/blob/e4866c8638575bd912612acaecb3c6e71579dee8/libcaf_test/caf/test/runner.cpp#L131)?

Furthermore, this is a very generic error. By doing the error reporting here, we lost the context we had in `to_regex`. Specifically, the `what()` string from [`regex_error`](https://en.cppreference.com/w/cpp/regex/regex_error). Pushing the error reporting into the function would allow us to do better error reporting. Of course then the function needs the option name. Also, we'll add another option with a regex input (`-t`) that will need to do the same error reporting again.

libcaf_test/caf/test/runner.hpp
```diff
@@ -5,8 +5,12 @@
 #pragma once
 
 #include "caf/detail/test_export.hpp"
+#include "caf/settings.hpp"
 #include "caf/test/registry.hpp"
 
+#include <optional>
+#include <regex>
```
Neverlord, line 12: I didn't spot it last time. Given that `to_regex` is private static and since we never call it from this header, the member function could as well be implemented in `runner.cpp` as a free function in the anonymous namespace. That would avoid pulling in `<regex>` here for better header hygiene.

```suggestion
```

Neverlord (changes requested):

libcaf_test/caf/test/runner.cpp
shariarriday: Are you referring to this format for reporting errors 
`error while parsing argument '{}': {}\n\n{}\n` ?

libcaf_test/caf/test/runner.cpp
```diff
@@ -53,6 +53,24 @@ std::optional<unsigned> parse_log_level(std::string_view x) {
   return {};
 }
 
+std::optional<std::regex> to_regex(std::string_view regex_string) {
+#ifdef CAF_ENABLE_EXCEPTIONS
+  try {
+    return std::regex{regex_string.begin(), regex_string.end()};
+  } catch (std::regex_error& err) {
```
Neverlord, line 60: Minor nitpick, this can be const and we can generalize it to any standard exception:

```suggestion
  } catch (const std::exception& err) {
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -53,6 +53,24 @@ std::optional<unsigned> parse_log_level(std::string_view x) {
   return {};
 }
 
+std::optional<std::regex> to_regex(std::string_view regex_string) {
+#ifdef CAF_ENABLE_EXCEPTIONS
+  try {
```
Neverlord, line 58: To avoid redundant code in the catch blocks below:

```suggestion
#ifdef CAF_ENABLE_EXCEPTIONS
  using detail::format_to;
  auto err_out = [] {
    return std::ostream_iterator<char>{std::cerr};
  };
  try {
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -53,6 +53,24 @@ std::optional<unsigned> parse_log_level(std::string_view x) {
...
+    return std::regex{regex_string.begin(), regex_string.end()};
+  } catch (std::regex_error& err) {
+    using detail::format_to;
+    auto errStream = std::ostream_iterator<char>{std::cerr};
+    format_to(errStream, "error while parsing argument '{}': {}\n",
```
Neverlord, line 63: Aside from using the lambda here, please note that `errStream` is not according to our coding conventions (please use `snake_case`).

```suggestion
    format_to(err_out(), "error while parsing argument '{}': {}\n",
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -53,6 +53,24 @@ std::optional<unsigned> parse_log_level(std::string_view x) {
...
+              regex_string, err.what());
+    return std::nullopt;
+  } catch (...) {
+    return std::nullopt;
```
Neverlord, line 67: Unlikely, but we should print at least *something* here:

```suggestion
  } catch (...) {
    format_to(err_out(), "error while parsing argument '{}': unknown error\n",
              regex_string);
    return std::nullopt;
```

Neverlord (changes requested):
We're getting there. 🙂

In the "catch-all" case, we currently print nothing. While this is very unlikely to trigger, we should still cover our bases here and print at least *something* that tells the user why the program stopped.

PR #1493, opened 2023-07-14
Add option to disable colored output

shariarriday:

-n flag can be used to disabled colored output

### DIFF
--- a/libcaf_test/caf/test/reporter.cpp
+++ b/libcaf_test/caf/test/reporter.cpp
@@ -180,8 +180,8 @@ public:
   }
 
   auto colored() {
-    auto state = plain_output_ ? colorizing_iterator::off
-                               : colorizing_iterator::normal;
+    auto state = no_colors_ ? colorizing_iterator::off
+                            : colorizing_iterator::normal;
 
     return colorizing_iterator{state, &std::cout};
   }
@@ -393,8 +393,8 @@ public:
     level_ = level;
   }
 
-  void disable_colors(bool plain_output) override {
-    plain_output_ = plain_output;
+  void no_colors(bool new_value) override {
+    no_colors_ = new_value;
   }
 
   stats test_stats() override {
@@ -448,7 +448,7 @@ private:
   unsigned level_ = CAF_LOG_LEVEL_INFO;
 
   /// Configures the color output of the reporter.
-  bool plain_output_ = false;
+  bool no_colors_ = false;
 
   /// Stores the names of failed test suites.
   std::vector<std::string_view> failed_suites_;
--- a/libcaf_test/caf/test/reporter.cpp
+++ b/libcaf_test/caf/test/reporter.cpp
@@ -54,7 +54,6 @@ struct colorizing_iterator {
   };
   mode_t mode = normal;
   std::ostream* out;
-  mode_t start = mode;
   void put(char c) {
     switch (mode) {
       case normal:
@@ -102,10 +101,7 @@ struct colorizing_iterator {
         if (c == ')') {
           out->flush();
           *out << term::reset;
-          if (start == off)
-            mode = off;
-          else
-            mode = normal;
+          mode = normal;
           break;
         }
         out->put(c);
@@ -117,40 +113,24 @@ struct colorizing_iterator {
           out->put(c);
         break;
       case off_read_color:
+        out->flush();
         switch (c) {
-          case 'R':
-            mode = off_escape;
-            break;
-          case 'G':
-            mode = off_escape;
-            break;
-          case 'B':
-            mode = off_escape;
-            break;
-          case 'Y':
-            mode = off_escape;
-            break;
-          case 'M':
-            mode = off_escape;
-            break;
-          case 'C':
-            mode = off_escape;
-            break;
           case '0':
             mode = verbatim;
             return;
           default:
-            CAF_RAISE_ERROR("invalid color code");
-            break;
+            mode = off_escape;
+            return;
         }
         break;
       case off_escape:
         if (c != '(')
           CAF_RAISE_ERROR("expected ( after color code");
-        mode = color;
+        mode = off_color;
         break;
       case off_color:
         if (c == ')') {
+          out->flush();
           mode = off;
           break;
         }
@@ -200,10 +180,10 @@ public:
   }
 
   auto colored() {
-    if (plain_output_)
-      return colorizing_iterator{colorizing_iterator::off, &std::cout};
-    else
-      return colorizing_iterator{colorizing_iterator::normal, &std::cout};
+    auto state = plain_output_ ? colorizing_iterator::off
+                               : colorizing_iterator::normal;
+
+    return colorizing_iterator{state, &std::cout};
   }
 
   void stop() override {
@@ -413,7 +393,7 @@ public:
     level_ = level;
   }
 
-  void disable_colored(bool plain_output) override {
+  void disable_colors(bool plain_output) override {
     plain_output_ = plain_output;
   }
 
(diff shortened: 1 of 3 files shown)


libcaf_test/caf/test/reporter.cpp
```diff
@@ -53,6 +53,7 @@ struct colorizing_iterator {
   };
   mode_t mode = normal;
   std::ostream* out;
+  mode_t start = mode;
```
Neverlord, line 56: We don't need the extra state. See below.

```suggestion
```

libcaf_test/caf/test/reporter.cpp
```diff
@@ -100,7 +101,10 @@ struct colorizing_iterator {
         if (c == ')') {
           out->flush();
           *out << term::reset;
-          mode = normal;
+          if (start == off)
+            mode = off;
+          else
+            mode = normal;
```
Neverlord, line 107: I believe you have tried to fix a case where the uncolored mode jumps back to colored here. But the issue is actually in `case off_escape`. It should not go to state `color` but to `off_color` instead.

Basically, the `off_` states mirror the regular states and should only transition to other `off_` states (or to `verbatim`).

```suggestion
          mode = normal;
```

libcaf_test/caf/test/reporter.hpp
```diff
@@ -76,6 +76,9 @@ class CAF_TEST_EXPORT reporter {
 
   virtual void verbosity(unsigned level) = 0;
 
+  /// Disables the colored output mode for the reporter
+  virtual void disable_colored(bool plain_output) = 0;
```
Neverlord, line 80: "colored" is an adjective (or past tense). It can't stand on its own like this. I'd suggest `disable_colors` or `disable_colored_output` instead.

When naming the function in this active form, I find it illogical that it takes a parameter at all. So either stick to the `disable_*` command verb and take no argument or rename it to something like `colored_output(bool value)`.

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,6 +68,13 @@ int runner::run(int argc, char** argv) {
     return EXIT_SUCCESS;
   }
   default_reporter->start();
+  default_reporter->disable_colored(get_or(cfg_, "no-colors", false));
```
Neverlord, line 71: Reading this line here, I can see why you chose to give the function an argument. But then I'd go with something like this:

```c++
  default_reporter->colored_output(!get_or(cfg_, "no-colors", false));
```

If you prefer the `disable` style, I think this would make more sense:

```c++
  if (get_or(cfg_, "no-colors", false)))
    default_reporter->disable_colors();
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -68,6 +68,13 @@ int runner::run(int argc, char** argv) {
...
+  auto enabled = [](const std::regex& selected,
+                    const std::string_view& search_string) {
+    return std::regex_search(search_string.begin(), search_string.end(),
+                             selected);
+  };
+  auto suite_regex = parse_regex_opt(get_or(cfg_, "suites", ".*"));
```
Neverlord, line 77: This seems unrelated. Probably slipped in from the other PR?

libcaf_test/caf/test/reporter.cpp
```diff
@@ -112,7 +116,32 @@ struct colorizing_iterator {
...
+            break;
+          case 'C':
+            mode = off_escape;
+            break;
+          case '0':
+            mode = verbatim;
+            return;
+          default:
+            CAF_RAISE_ERROR("invalid color code");
+            break;
```
Neverlord, line 143: All cases do the same thing. So either fall through, or just ignore everything but `0`:

```
          default:
            mode = off_escape;
            break;
          case '0':
            mode = verbatim;
```

libcaf_test/caf/test/reporter.cpp
```diff
@@ -170,7 +199,10 @@ class default_reporter : public reporter {
   }
 
   auto colored() {
-    return colorizing_iterator{colorizing_iterator::normal, &std::cout};
+    if (plain_output_)
+      return colorizing_iterator{colorizing_iterator::off, &std::cout};
+    else
+      return colorizing_iterator{colorizing_iterator::normal, &std::cout};
```
Neverlord, line 205: The if-else really only translates the boolean to the state flag, so putting this upfront makes the control flow a bit more clear IMHO:

```suggestion
    auto state = plain_output_ ? colorizing_iterator::off : colorizing_iterator::normal;
    return colorizing_iterator{state, &std::cout};
```

Neverlord (changes requested):

libcaf_test/caf/test/reporter.cpp
```diff
@@ -380,6 +392,10 @@ class default_reporter : public reporter {
     level_ = level;
   }
 
+  void disable_colors(bool plain_output) override {
+    plain_output_ = plain_output;
```
Neverlord, line 396: This is a setter, so the associated member should have the same name with a `_` suffix according to our [style guide](https://github.com/actor-framework/actor-framework/blob/master/CONTRIBUTING.md#naming):

```c++
  void disable_colors(bool new_value) override {
    disable_colors_ = new_value;
  }
```

or:

```c++
  void plain_output(bool new_value) override {
    plain_output_ = new_value;
  }
```

Given that this gets filled from a config option named `no-colors`, the most logical choice would actually probably be:

```c++
  void no_colors(bool new_value) override {
    no_colors_ = new_value;
  }
```

libcaf_test/caf/test/reporter.cpp
```diff
@@ -112,15 +112,24 @@ struct colorizing_iterator {
           out->put(c);
         break;
       case off_read_color:
-        mode = off_escape;
+        out->flush();
```
Neverlord, line 115: Why did you add the `flush` here?

Neverlord (changes requested):

libcaf_test/caf/test/reporter.cpp
shariarriday: You have mentioned that off_* will be similar to it's colored states. We are using flush to print everything in read_color state. I think that this behavior needs to be the same in off_read_color state in order to maintain it's parallel to read_color state.

libcaf_test/caf/test/reporter.cpp
```diff
@@ -170,7 +179,10 @@ class default_reporter : public reporter {
   }
 
   auto colored() {
-    return colorizing_iterator{colorizing_iterator::normal, &std::cout};
+    auto state = no_colors_ ? colorizing_iterator::off
+                            : colorizing_iterator::normal;
+
+    return colorizing_iterator{state, &std::cout};
```
Neverlord, line 185:
```suggestion
    return colorizing_iterator{state, &std::cout};
```

libcaf_test/caf/test/reporter.hpp
```diff
@@ -76,6 +76,9 @@ class CAF_TEST_EXPORT reporter {
 
   virtual void verbosity(unsigned level) = 0;
 
+  /// Disables the colored output mode for the reporter
```
Neverlord, line 79:
```suggestion
  /// Sets whether the reporter disables colored output even when writing to a TTY.
```

libcaf_test/caf/test/reporter.cpp
```diff
@@ -430,6 +446,9 @@ class default_reporter : public reporter {
   /// Configures the verbosity of the reporter.
   unsigned level_ = CAF_LOG_LEVEL_INFO;
 
+  /// Configures the color output of the reporter.
```
Neverlord, line 449:
```suggestion
  /// Configures whether we render text without colors.
```

Neverlord (changes requested):
Minor nitpicks, otherwise LGTM now. 👍

PR #1496, opened 2023-07-18
Ignore *.test.cpp files in codecov

riemass:


riemass:
I checked the test report and the `*.test.cpp` files are not present, just to be sure.

Neverlord (approved):
👍

PR #1497, opened 2023-07-18
Stop handling continuation frames as control frames

riemass:
Relates #1394 

This caused a bug when a continuation frame had more than 126 bytes of data. Fixes the tests that transfer 4MB payloads.

### DIFF
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -59,8 +59,7 @@ struct CAF_NET_EXPORT rfc6455 {
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
 
   static constexpr bool is_control_frame(uint8_t opcode) noexcept {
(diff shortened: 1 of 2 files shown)


libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -59,7 +59,8 @@ struct CAF_NET_EXPORT rfc6455 {
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
 
   static constexpr bool is_control_frame(uint8_t opcode) noexcept {
-    return opcode != text_frame && opcode != binary_frame;
+    return opcode != text_frame && opcode != binary_frame
+           && opcode != continuation_frame;
```
Neverlord, line 63: At this point, wouldn't it make more sense to implement it this way?

```suggestion
    return opcode > binary_frame;
```

Neverlord (changes requested):

Neverlord (approved):

PR #1498, opened 2023-07-18
Decode and validate WebSocket closing payload

riemass:
Relates #1394 
With this and #1497 we have the majority of tests green and a small number of yellow tests, that are non-strict due to the nature of distributed systems.

### DIFF
--- a/libcaf_net/caf/net/web_socket/framing.cpp
+++ b/libcaf_net/caf/net/web_socket/framing.cpp
@@ -257,8 +257,7 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
     case detail::rfc6455::connection_close: {
       auto result = decode_closing_payload(payload);
       if (!result) {
-        abort(result.error());
-        shutdown(result.error());
+        abort_and_shutdown(result.error());
         return -1;
       }
       abort_and_shutdown(sec::connection_closed);
--- a/libcaf_net/caf/net/web_socket/framing.cpp
+++ b/libcaf_net/caf/net/web_socket/framing.cpp
@@ -12,45 +12,41 @@ namespace caf::net::web_socket {
 
 // -- static utility functions -------------------------------------------------
 
-expected<std::pair<uint16_t, const_byte_span>>
-framing::decode_closing_payload(const_byte_span payload) {
-  if (payload.size() == 0)
-    return std::make_pair(0, payload);
-  if (payload.size() < 2)
+error framing::validate_closing_payload(const_byte_span payload) {
+  if (payload.empty())
+    return error{};
+  if (payload.size() == 1)
     return make_error(caf::sec::protocol_error,
-                      "no status flag in closing payload");
-  auto status = (std::to_integer<uint_fast16_t>(payload[0]) << 8)
+                      "closing status code must have two bytes");
+  auto status = (std::to_integer<uint16_t>(payload[0]) << 8)
                 + std::to_integer<uint16_t>(payload[1]);
   if (!detail::rfc3629::valid(payload.subspan(2)))
     return make_error(sec::protocol_error,
-                      "malformed text message in closing payload");
-  if (status < 1000 || status >= 5000)
-    return make_error(sec::protocol_error,
-                      "invalid status code in closing payload");
+                      "malformed UTF-8 text message in closing payload");
   // statuses between 3000 and 4999 are allowed and application specific
   if (status >= 3000 && status < 5000)
-    return std::make_pair(status, payload.subspan(2));
-  // statuses between 1000 and 2999 are protocol defined
-  web_socket::status status_code;
-  auto success = from_integer(status, status_code);
-  if (!success)
-    return make_error(sec::protocol_error,
-                      "invalid status code in closing payload");
-  switch (status_code) {
-    case status::normal_close:
-    case status::going_away:
-    case status::protocol_error:
-    case status::invalid_data:
-    case status::inconsistent_data:
-    case status::policy_violation:
-    case status::message_too_big:
-    case status::missing_extensions:
-    case status::unexpected_condition:
-      return std::make_pair(status, payload.subspan(2));
-    default:
-      return make_error(sec::protocol_error,
-                        "invalid status code in closing payload");
+    return error{};
+  // statuses between 1000 and 2999 need to be protocol defined, and status
+  // codes lower then 1000 and greater or equal then 5000 are invalid.
+  auto status_code = web_socket::status{0};
+  if (from_integer(status, status_code)) {
+    switch (status_code) {
+      case status::normal_close:
+      case status::going_away:
+      case status::protocol_error:
+      case status::invalid_data:
+      case status::inconsistent_data:
+      case status::policy_violation:
+      case status::message_too_big:
+      case status::missing_extensions:
+      case status::unexpected_condition:
+        return error{};
+      default:
+        break;
+    }
   }
+  return make_error(sec::protocol_error,
+                    "invalid status code in closing payload");
 }
 
 // -- octet_stream::upper_layer implementation ---------------------------------
@@ -254,15 +250,13 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
                           size_t frame_size) {
   // opcodes are checked for validity when decoding the header
   switch (opcode) {
-    case detail::rfc6455::connection_close: {
-      auto result = decode_closing_payload(payload);
-      if (!result) {
-        abort_and_shutdown(result.error());
+    case detail::rfc6455::connection_close:
+      if (auto error = validate_closing_payload(payload); error) {
+        abort_and_shutdown(error);
         return -1;
       }
       abort_and_shutdown(sec::connection_closed);
       return -1;
-    }
     case detail::rfc6455::text_frame: {
       std::string_view text{reinterpret_cast<const char*>(payload.data()),
                             payload.size()};
--- a/libcaf_net/caf/net/web_socket/framing.cpp
+++ b/libcaf_net/caf/net/web_socket/framing.cpp
@@ -10,6 +10,49 @@
 
 namespace caf::net::web_socket {
 
+// -- static utility functions -------------------------------------------------
+
+expected<std::pair<uint16_t, const_byte_span>>
+framing::decode_closing_payload(const_byte_span payload) {
+  if (payload.size() == 0)
+    return std::make_pair(0, payload);
+  if (payload.size() < 2)
+    return make_error(caf::sec::protocol_error,
+                      "no status flag in closing payload");
+  auto status = (std::to_integer<uint_fast16_t>(payload[0]) << 8)
+                + std::to_integer<uint16_t>(payload[1]);
+  if (!detail::rfc3629::valid(payload.subspan(2)))
+    return make_error(sec::protocol_error,
+                      "malformed text message in closing payload");
+  if (status < 1000 || status >= 5000)
+    return make_error(sec::protocol_error,
+                      "invalid status code in closing payload");
+  // statuses between 3000 and 4999 are allowed and application specific
+  if (status >= 3000 && status < 5000)
+    return std::make_pair(status, payload.subspan(2));
+  // statuses between 1000 and 2999 are protocol defined
+  web_socket::status status_code;
+  auto success = from_integer(status, status_code);
+  if (!success)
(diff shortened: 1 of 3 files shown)


libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
+    case status::inconsistent_data:
+    case status::policy_violation:
+    case status::message_too_big:
+    case status::missing_extensions:
+    case status::unexpected_condition:
+      return std::make_pair(status, payload.subspan(2));
+    default:
+      return make_error(sec::protocol_error,
+                        "invalid status code in closing payload");
+  }
```
Neverlord, line 53: When looking at https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.2, don't we actually have to accept everything in range 1000 to 4999? That would make the code just look like this:

```suggestion
  if (status < 1000 || status >= 5000)
    return make_error(sec::protocol_error,
                      "invalid status code in closing payload");
  return std::make_pair(status, payload.subspan(2));
```

Or am I missing something here?

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -211,9 +254,16 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
                           size_t frame_size) {
   // opcodes are checked for validity when decoding the header
   switch (opcode) {
-    case detail::rfc6455::connection_close:
+    case detail::rfc6455::connection_close: {
+      auto result = decode_closing_payload(payload);
+      if (!result) {
+        abort(result.error());
+        shutdown(result.error());
```
Neverlord, line 261: Why isn't this calling `abort_and_shutdown`?

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Status codes in the range 1000-2999 are reserved for definition by the WebSocket protocol. At [section 7.4.1](https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.1) we see that only a limited set are defined at this moment and we should reject the other values. Even from the defined status code values, 1004 is undefined, and 1005 and 1006 must not be set as close control by definition, so we need to rule these out.

libcaf_net/caf/net/web_socket/framing.cpp
riemass: My mistake.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
 namespace caf::net::web_socket {
 
+// -- static utility functions -------------------------------------------------
+
+expected<std::pair<uint16_t, const_byte_span>>
+framing::decode_closing_payload(const_byte_span payload) {
+  if (payload.size() == 0)
+    return std::make_pair(0, payload);
+  if (payload.size() < 2)
+    return make_error(caf::sec::protocol_error,
```
Neverlord, line 20: This is currently not covered by the tests. Mind adding an additional test that sends a single byte to cover this branch as well?

Neverlord (changes requested):

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Added.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -211,9 +254,15 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
                           size_t frame_size) {
   // opcodes are checked for validity when decoding the header
   switch (opcode) {
-    case detail::rfc6455::connection_close:
+    case detail::rfc6455::connection_close: {
+      auto result = decode_closing_payload(payload);
+      if (!result) {
+        abort_and_shutdown(result.error());
+        return -1;
+      }
```
Neverlord, line 262: We don't need the result outside of the `if` statement, so the scope for this variable can be narrowed further:

```suggestion
      if (auto result = decode_closing_payload(payload); !result) {
        abort_and_shutdown(result.error());
        return -1;
      }
```

Also avoids the extra `{}` for the `case` block.

libcaf_net/caf/net/web_socket/framing.hpp
```diff
@@ -43,6 +43,11 @@ class CAF_NET_EXPORT framing : public octet_stream::upper_layer,
   /// Stored as currently active opcode to mean "no opcode received yet".
   static constexpr size_t nil_code = 0xFF;
 
+  // -- static utility functions -----------------------------------------------
+
+  static expected<std::pair<uint16_t, const_byte_span>>
+  decode_closing_payload(const_byte_span payload);
```
Neverlord, line 49: I couldn't find a use for the actual result of this function outside of tests. It seems like we only really care about error vs. no error, right? In this case, we might consider just returning an `error` here to check for default-constructed (i.e., no-error) vs actual error:

```
  /// Checks whether ...
  /// @returns A default-constructed `error` if the payload is valid ...
  static error validate_closing_payload(const_byte_span payload);
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
+    return make_error(caf::sec::protocol_error,
+                      "no status flag in closing payload");
+  auto status = (std::to_integer<uint_fast16_t>(payload[0]) << 8)
+                + std::to_integer<uint16_t>(payload[1]);
+  if (!detail::rfc3629::valid(payload.subspan(2)))
+    return make_error(sec::protocol_error,
+                      "malformed text message in closing payload");
+  if (status < 1000 || status >= 5000)
+    return make_error(sec::protocol_error,
+                      "invalid status code in closing payload");
```
Neverlord, line 29: I believe this is redundant. The `if`-block below pass for anything in range [3000, 5000). Then, we call `from_integer`, which converts known values to an `enum` value. This will always fail for any `x < 1000` and any `x > 5000` by definition.

```suggestion
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
+  // statuses between 3000 and 4999 are allowed and application specific
+  if (status >= 3000 && status < 5000)
+    return std::make_pair(status, payload.subspan(2));
+  // statuses between 1000 and 2999 are protocol defined
+  web_socket::status status_code;
+  auto success = from_integer(status, status_code);
+  if (!success)
```
Neverlord, line 36: I can't find a reason to bind this result to a named variable, since it's only checked once:

```suggestion
  if (!from_integer(status, status_code))
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
+    return make_error(sec::protocol_error,
+                      "malformed text message in closing payload");
+  if (status < 1000 || status >= 5000)
+    return make_error(sec::protocol_error,
+                      "invalid status code in closing payload");
+  // statuses between 3000 and 4999 are allowed and application specific
+  if (status >= 3000 && status < 5000)
+    return std::make_pair(status, payload.subspan(2));
+  // statuses between 1000 and 2999 are protocol defined
+  web_socket::status status_code;
```
Neverlord, line 34: Nitpick, but I think we should stick to AAA (almost always auto):

```suggestion
  auto status_code = web_socket::status{0};
```

0 is an invalid value, but at least it's a defined value. 🙂

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,49 @@
...
+// -- static utility functions -------------------------------------------------
+
+expected<std::pair<uint16_t, const_byte_span>>
+framing::decode_closing_payload(const_byte_span payload) {
+  if (payload.size() == 0)
+    return std::make_pair(0, payload);
+  if (payload.size() < 2)
+    return make_error(caf::sec::protocol_error,
+                      "no status flag in closing payload");
+  auto status = (std::to_integer<uint_fast16_t>(payload[0]) << 8)
```
Neverlord, line 22: Auto-completion mistake? It's inconsistent with the line immediately below and also doesn't make intuitive sense to me.

```suggestion
  auto status = (std::to_integer<uint16_t>(payload[0]) << 8)
```

Neverlord (changes requested):

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Mistake, you're probably right about the autocomplete.

libcaf_net/caf/net/web_socket/framing.hpp
riemass: Currently there is no need to report the status code to the user, so we can do it this way for clarity.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,45 @@
 
 namespace caf::net::web_socket {
 
+// -- static utility functions -------------------------------------------------
+
+error framing::validate_closing_payload(const_byte_span payload) {
+  if (payload.empty())
+    return error{};
```
Neverlord, line 17:
```suggestion
    return {};
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,45 @@
...
+    return make_error(caf::sec::protocol_error,
+                      "closing status code must have two bytes");
+  auto status = (std::to_integer<uint16_t>(payload[0]) << 8)
+                + std::to_integer<uint16_t>(payload[1]);
+  if (!detail::rfc3629::valid(payload.subspan(2)))
+    return make_error(sec::protocol_error,
+                      "malformed UTF-8 text message in closing payload");
+  // statuses between 3000 and 4999 are allowed and application specific
+  if (status >= 3000 && status < 5000)
+    return error{};
```
Neverlord, line 28:
```suggestion
    return {};
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,45 @@
...
+      case status::normal_close:
+      case status::going_away:
+      case status::protocol_error:
+      case status::invalid_data:
+      case status::inconsistent_data:
+      case status::policy_violation:
+      case status::message_too_big:
+      case status::missing_extensions:
+      case status::unexpected_condition:
+        return error{};
```
Neverlord, line 43:
```suggestion
        return {};
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -10,6 +10,45 @@
 
 namespace caf::net::web_socket {
 
+// -- static utility functions -------------------------------------------------
+
+error framing::validate_closing_payload(const_byte_span payload) {
+  if (payload.empty())
+    return error{};
+  if (payload.size() == 1)
```
Neverlord, line 18: The status below says "must have two bytes", so shouldn't this read the following?

```suggestion
  if (payload.size() != 2)
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -212,6 +251,10 @@ ptrdiff_t framing::handle(uint8_t opcode, byte_span payload,
   // opcodes are checked for validity when decoding the header
   switch (opcode) {
     case detail::rfc6455::connection_close:
+      if (auto error = validate_closing_payload(payload); error) {
+        abort_and_shutdown(error);
```
Neverlord, line 255: We generally name this `err`, it's also good to avoid `error` because it introduces ambiguity with `caf::error`.

```suggestion
      if (auto err = validate_closing_payload(payload); err) {
        abort_and_shutdown(err);
```

Neverlord (changes requested):
Just some nitpicks, looks good otherwise now. 👍

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Sadly, no. Empty payload is allowed, meaning no status code at all, 2 bytes means only the status code is present, and everything after the second byte is the closing message.
I'll change the error message though.

PR #1499, opened 2023-07-18
Remove ripemd160

### LINKED ISSUE #1425: Get rid of ripemd 160
We only have a single use case for this in `hashed_node_id::local`: https://github.com/actor-framework/actor-framework/blob/0.19.1/libcaf_core/src/node_id.cpp#L120. The whole thing with using system-information and hashing stuff is obsolete anyways and we had to add some random bytes to make it work in containerized environments. The entire function can just generate random bytes instead.

riemass:
Closes #1425
Remove ripemd and system info, use random generated bytes instead.


libcaf_core/caf/node_id.cpp
```diff
@@ -95,29 +95,18 @@ bool hashed_node_id::can_parse(std::string_view str) noexcept {
 
 node_id hashed_node_id::local(const actor_system_config&) {
```
riemass, line 96: Should I remove the `actor_system_config` parameter?

libcaf_core/caf/node_id.cpp
```diff
@@ -95,29 +95,18 @@ bool hashed_node_id::can_parse(std::string_view str) noexcept {
...
-  auto seeded_hd_serial_and_mac_addr = join(macs, "") + detail::get_root_uuid();
-  // By adding 8 random ASCII characters, we make sure to assign a new (random)
-  // ID to a node every time we start it. Otherwise, we might run into issues
-  // where a restarted node produces identical actor IDs than the node it
-  // replaces. Especially when running nodes in a container, because then the
-  // process ID is most likely the same.
+  // We add an global incrementing counter to make sure two actor systems in the
+  // same process won't have the same node ID - even if the user manipulates the
+  // system to always produce the same seed for its randomness.
+  auto sys_seed = static_cast<uint8_t>(system_id.fetch_add(1));
```
riemass, line 101: What is bogging me is that every time we call the `local` function we get a new value. So is this part still needed?

Neverlord: Hm, good question. The chance that `random_device` returns the same value is probably very small.

riemass: There's the possibility from the comment -  the user manipulating the system to always produce the same seed. Then again, if I'm the user that manipulates the system to produce the same seed, I probably want the same random numbers generated.

PR #1501, opened 2023-07-18
Update contributions guide

Neverlord:
I've updated the contributions guide and removed everything that we already enforce with `clang-format` (commit 1). I've also tweaked the `.clang-format` config to sort includes according to our style guide and let it run on the code base (commit 2). Finally, I had to add a few missing includes that became apparent after the automatic changes from `clang-format` (commit 3).

@riemass: unfortunately, `clang-format` doesn't recognize the main header when using the `#include <...>` syntax. So I've opted to leave it at `#include "..."` for CAF headers.


CONTRIBUTING.md
```diff
@@ -57,28 +44,57 @@ Commit Message Style
...
+  correctly. We also measure code coverage and require that new code is covered
+  by tests.
+
+- We require at least one code review approval before merging a pull request. We
+  take the review phase seriously and have a high bar for code quality. Multiple
+  rounds of review are common.
+
+- In case there are multiple rounds of review, please do not squash the commits
+  until everything is resolved. This makes it easier for reviewers to see what
+  has changed since the last review.
```
riemass, line 64: Guess I was the inspiration for this :D

riemass (approved):
Good update. Should we add a statement about the local headers and relative paths, something in the line of:
- Include directives using the `#include "caf/foo.hpp"` syntax should start from the 'caf` directory and must not use relative file paths.

Just to give you a heads up, I removed a few include directives in #1499 and it might cause a merge conflict.

CONTRIBUTING.md
Neverlord: No worries, happened before you as well. ;)

Neverlord:
> Good update. Should we add a statement about the local headers and relative paths, something in the line of:
> 
> * Include directives using the `#include "caf/foo.hpp"` syntax should start from the 'caf` directory and must not use relative file paths.

While it's not explicitly spelled out like this, I think this paragraph still covers this convention:

> - Use `#include "caf/foo/bar.hpp"` for CAF headers and `#include <foo/bar.hpp>`
  for third-party headers.

I rather err on the side of fewer rules. But we can still make it more explicit later if we see a need.

> Just to give you a heads up, I removed a few include directives in #1499 and it might cause a merge conflict.

Took care of it.

PR #1502, opened 2023-07-19
Add watchdog and option to abort application when test exceeds maximum runtime

shariarriday:

The test application aborts when any test exceeds the maximum runtime limit. The limit is set with -m option

### DIFF
--- a/libcaf_test/caf/test/runner.cpp
+++ b/libcaf_test/caf/test/runner.cpp
@@ -29,13 +29,9 @@ namespace {
 
 class watchdog {
 public:
-  static void start(int secs);
-  static void stop();
-
-private:
   watchdog(int secs) {
     using detail::format_to;
-    thread_ = std::thread{[=] {
+    thread_ = secs <= 0 ? std::thread() : std::thread{[=] {
       caf::detail::set_thread_name("test.watchdog");
       auto tp = std::chrono::high_resolution_clock::now()
                 + std::chrono::seconds(secs);
@@ -58,27 +54,15 @@ private:
       canceled_ = true;
       cv_.notify_all();
     }
-    thread_.join();
+    if (thread_.joinable())
+      thread_.join();
   }
 
-  volatile bool canceled_ = false;
+  bool canceled_ = false;
   std::mutex mtx_;
   std::condition_variable cv_;
   std::thread thread_;
 };
-
-namespace {
-watchdog* s_watchdog;
-}
-
-void watchdog::start(int secs) {
-  if (secs > 0)
-    s_watchdog = new watchdog(secs);
-}
-
-void watchdog::stop() {
-  delete s_watchdog;
-}
 
 config_option_set make_option_set() {
   config_option_set result;
@@ -158,6 +142,7 @@ int runner::run(int argc, char** argv) {
                              selected);
   };
   const auto max_runtime = get_or(cfg_, "max-runtime", 0);
+  const auto watchdog_ptr = std::make_unique<watchdog>(max_runtime);
   for (auto& [suite_name, suite] : suites_) {
     if (!enabled(*suite_regex, suite_name))
       continue;
@@ -165,7 +150,6 @@ int runner::run(int argc, char** argv) {
     for (auto [test_name, factory_instance] : suite) {
       if (!enabled(*test_regex, test_name))
         continue;
-      watchdog::start(max_runtime);
       auto state = std::make_shared<context>();
 #ifdef CAF_ENABLE_EXCEPTIONS
       try {
@@ -195,7 +179,6 @@ int runner::run(int argc, char** argv) {
         state->clear_stacks();
       } while (state->can_run());
 #endif
-      watchdog::stop();
     }
     default_reporter->end_suite(suite_name);
   }
--- a/libcaf_test/caf/test/runner.cpp
+++ b/libcaf_test/caf/test/runner.cpp
@@ -29,9 +29,25 @@ namespace {
 
 class watchdog {
 public:
-  watchdog(int secs) {
+  explicit watchdog(int secs) {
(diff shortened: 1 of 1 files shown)


libcaf_test/caf/test/runner.cpp
```diff
@@ -100,13 +157,15 @@ int runner::run(int argc, char** argv) {
...
   };
+  const auto max_runtime = get_or(cfg_, "max-runtime", 0);
   for (auto& [suite_name, suite] : suites_) {
     if (!enabled(*suite_regex, suite_name))
       continue;
     default_reporter->begin_suite(suite_name);
     for (auto [test_name, factory_instance] : suite) {
       if (!enabled(*test_regex, test_name))
         continue;
+      watchdog::start(max_runtime);
```
Neverlord, line 168: I think there's a conceptual disconnect. Probably the explanation for why you have implemented the watchdog in this way. The watchdog should terminate the program if the whole run reaches the timeout. Your train of thought seemed to be that this is per test. 

Since the watchdog is actually about the program runtime, the initialization should move to *before* the `for` loop and then just be an RAII type.

Also, your implementation is buggy. This will allocate one watchdog for each test that we start, leaking all but the last one.

Neverlord (changes requested):

libcaf_test/caf/test/runner.cpp
shariarriday: Yes, I was planning to add timer for each test as the task mentioned "aborting test after reaching the time limit". But, as I am calling the watchdog::stop() after completing each test, it should delete existing watchdog without leaking any memory. But, I am moving this implementation to a RAII based approach and running the watchdog once for the whole program.

libcaf_test/caf/test/runner.cpp
```diff
@@ -13,23 +13,64 @@
...
 
 namespace caf::test {
 
 namespace {
 
+class watchdog {
+public:
+  watchdog(int secs) {
+    using detail::format_to;
+    thread_ = secs <= 0 ? std::thread() : std::thread{[=] {
```
Neverlord, line 34: At least to me, the ternary operator makes the control code somewhat hard to follow. There's also a lot going on in the constructor that's not actually related to constructing the object. I think writing it like this makes the intent much more clear:

```suggestion
    if (secs > 0)
      thread_ = std::thread{[this, secs] { run(secs); });
```

And the current body of the lambda should go to a private `run` member function.

libcaf_test/caf/test/runner.cpp
```diff
@@ -13,23 +13,64 @@
...
+  ~watchdog() {
+    {
+      std::lock_guard<std::mutex> guard{mtx_};
+      canceled_ = true;
+      cv_.notify_all();
+    }
+    if (thread_.joinable())
+      thread_.join();
+  }
+
```
Neverlord, line 60: I believe this is missing a `private` here.

```suggestion

private:
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -13,23 +13,64 @@
...
+};
+
 config_option_set make_option_set() {
   config_option_set result;
   config_option_adder{result, "global"}
     .add<bool>("available-suites,a", "print all available suites")
     .add<bool>("help,h?", "print this help text")
     .add<bool>("no-colors,n", "disable coloring (ignored on Windows)")
-    .add<size_t>("max-runtime,r", "set a maximum runtime in seconds")
+    .add<size_t>("max-runtime,m", "set a maximum runtime in seconds")
```
Neverlord, line 73: I agree that `m` is the more logical choice. But please leave it at `r` for backwards compatibility with the old framework. We will have them both side-by-side for a while, so we should be careful about breaking the CLI.

libcaf_test/caf/test/runner.cpp
```diff
@@ -100,6 +141,8 @@ int runner::run(int argc, char** argv) {
     return std::regex_search(search_string.begin(), search_string.end(),
                              selected);
   };
+  const auto max_runtime = get_or(cfg_, "max-runtime", 0);
+  const auto watchdog_ptr = std::make_unique<watchdog>(max_runtime);
```
Neverlord, line 145:
```suggestion
  watchdog runtime_guard{get_or(cfg_, "max-runtime", 0)};
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -13,23 +13,64 @@
...
 #include <string>
+#include <thread>
 
 namespace caf::test {
 
 namespace {
 
+class watchdog {
+public:
+  watchdog(int secs) {
```
Neverlord, line 32:
```suggestion
  explicit watchdog(int secs) {
```

Neverlord (changes requested):

libcaf_test/caf/test/runner.cpp
Neverlord: Ah, I know why you've changed it. I had `-m` in the ticket. Sorry about that, my mistake (I've updated the description in the ticket).

libcaf_test/caf/test/runner.cpp
```diff
@@ -13,23 +13,64 @@
...
+    }};
+  }
+  ~watchdog() {
+    {
+      std::lock_guard<std::mutex> guard{mtx_};
+      canceled_ = true;
+      cv_.notify_all();
+    }
+    if (thread_.joinable())
+      thread_.join();
```
Neverlord, line 58: If no thread is running, we don't need to do any locking at all:

```suggestion
    if (!thread_.joinable())
      return;
    {
      std::lock_guard<std::mutex> guard{mtx_};
      canceled_ = true;
      cv_.notify_all();
    }
    thread_.join();
```

PR #1503, opened 2023-07-20
Add check for all binary predicates

shariarriday:

Add check_ne, check_le, check_ge, check_lt and check_gt functions

### DIFF
--- a/libcaf_test/caf/test/runnable.hpp
+++ b/libcaf_test/caf/test/runnable.hpp
@@ -58,7 +58,7 @@ public:
   bool check_eq(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs == rhs) {
       reporter::instance->pass(location);
       return true;
@@ -73,7 +73,7 @@ public:
   bool check_ne(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs != rhs) {
       reporter::instance->pass(location);
       return true;
@@ -88,7 +88,7 @@ public:
   bool check_lt(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs < rhs) {
       reporter::instance->pass(location);
       return true;
@@ -103,7 +103,7 @@ public:
   bool check_le(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs <= rhs) {
       reporter::instance->pass(location);
       return true;
@@ -118,7 +118,7 @@ public:
   bool check_gt(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs > rhs) {
       reporter::instance->pass(location);
       return true;
@@ -133,7 +133,7 @@ public:
   bool check_ge(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_save_comparison(lhs, rhs);
+    assert_save_comparison<T0, T1>();
     if (lhs >= rhs) {
       reporter::instance->pass(location);
       return true;
@@ -188,7 +188,7 @@ public:
   }
 
   template <class T0, class T1>
-  static void assert_save_comparison(T0&&, T1&&) {
+  static void assert_save_comparison() {
     if constexpr (std::is_integral_v<T0> && std::is_integral_v<T1>) {
       static_assert(std::is_signed_v<T0> == std::is_signed_v<T1>,
                     "comparing signed and unsigned integers is unsafe");
--- a/libcaf_test/caf/test/runnable.hpp
+++ b/libcaf_test/caf/test/runnable.hpp
@@ -189,7 +189,7 @@ public:
 
   template <class T0, class T1>
   static void assert_save_comparison(T0&&, T1&&) {
-    if (std::is_integral_v<T0> && std::is_integral_v<T1>) {
+    if constexpr (std::is_integral_v<T0> && std::is_integral_v<T1>) {
       static_assert(std::is_signed_v<T0> == std::is_signed_v<T1>,
                     "comparing signed and unsigned integers is unsafe");
     }
--- a/libcaf_test/caf/test/runnable.hpp
+++ b/libcaf_test/caf/test/runnable.hpp
@@ -19,18 +19,6 @@
 
 namespace caf::test {
 
-namespace {
-
-template <class T0, class T1>
-inline void assert_integral_comparison(const T0& lhs, const T1& rhs) {
-  if (std::is_integral_v<T0> && std::is_integral_v<T1>) {
-    static_assert(std::is_signed_v<T0> == std::is_signed_v<T1>,
-                  "comparing signed and unsigned integers is unsafe");
-  }
-}
-
-}
-
 /// A runnable definition of a test case or scenario.
 class CAF_TEST_EXPORT runnable {
 public:
@@ -70,7 +58,7 @@ public:
   bool check_eq(const T0& lhs, const T1& rhs,
                 const detail::source_location& location
                 = detail::source_location::current()) {
-    assert_integral_comparison(lhs, rhs);
+    assert_save_comparison(lhs, rhs);
     if (lhs == rhs) {
       reporter::instance->pass(location);
       return true;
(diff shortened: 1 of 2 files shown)


Neverlord:
This is adding the new check functions, but not using them anywhere. 🙂

A good place to test them is `libcaf_test/caf/test/test.test.cpp`. Doesn't have to be something fancy, just refactoring the `tests can contain checks` test would suffice (e.g. drop the for loop and use each check once).

libcaf_test/caf/test/test.test.cpp
```diff
@@ -7,13 +7,33 @@
 
 using caf::test::block_type;
 
-TEST("tests can contain checks") {
+TEST("tests can contain different type of checks") {
```
Neverlord, line 10:
```suggestion
TEST("tests can contain different types of checks") {
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -7,13 +7,33 @@
...
-    check_eq(i, i);
+  /// tests all of the possible scenario for check_ne
+  check_ne(0u, 0u);
+  check_ne(0u, 1u);
   auto stats = rep->test_stats();
-  check_eq(stats.passed, 3u);
-  check_eq(stats.failed, 0u);
+  /// tests all of the possible scenario for check_eq
+  check_eq(stats.passed, 1u);
+  check_eq(stats.passed, 0u);
```
Neverlord, line 18: I find these tests very confusing. Why are we checking `stats.passed` here? The `check_ne` checks above make sense and use obvious values. Also, this entire suite isn't really testing anything. At the end of it all, this test overrides the test stats without ever checking that the things that are supposed to fail actually fail.

Here's an idea:

```c++
  // Either as a lambda here or as a member function in `runnable`.
  auto should_fail = [this](auto expr, const detail::source_location& location
    auto before = rep->test_stats();
    expr();
    auto after = rep->test_stats();
    check_eq(before.passed, after.passed, location);
    if (check_eq(before.failed + 1, after.failed, location)) {
      rep->test_stats({before.passed + 1, before.failed});
  };
```

This gives us a clean way to single-out tests that we expect to fail and to selectively "fix" the `failed` count. What do you think? With this utility, we can write tests like this:

```suggestion
  /// Test `check_eq`.
  check_eq(1, 1);
  should_fail([this] { check_eq(1, 2); });
  should_fail([this] { check_eq(2, 1); });
```

libcaf_test/caf/test/runnable.hpp
```diff
@@ -70,6 +70,91 @@ class CAF_TEST_EXPORT runnable {
...
 
+  /// Checks whether `lhs` and `rhs` are unequal.
+  template <class T0, class T1>
+  void check_ne(const T0& lhs, const T1& rhs,
+                const detail::source_location& location
+                = detail::source_location::current()) {
+    if (std::is_integral_v<T0> && std::is_integral_v<T1>) {
+      static_assert(std::is_signed_v<T0> == std::is_signed_v<T1>,
+                    "comparing signed and unsigned integers is unsafe");
+    }
```
Neverlord, line 81: This is repeated in pretty much all of the checks. I think this should be factored out into a (private) function.

Neverlord (changes requested):

libcaf_test/caf/test/test.test.cpp
shariarriday: Yes, that makes sense as we are not confirming if the things that should fail, actually fails. But, my current understanding is that we are using these tests to verify the binary predicates themselves and I thought that adding a new class that returns a value will not be a good idea. Do you recommend adding a new class `test.cpp` which will be tested using various predicates in this file?

libcaf_test/caf/test/runnable.hpp
```diff
@@ -19,6 +19,18 @@
 
 namespace caf::test {
 
+namespace {
+
+template <class T0, class T1>
+inline void assert_integral_comparison(const T0& lhs, const T1& rhs) {
```
Neverlord, line 25: Opening up the anonymous namespace in a header is kinda weird. Please move this to `runnable` as a (static) private member function and maybe rename to something a bit more "high level" like `assert_save_comparison`? Also, this doesn't need actual arguments since it's just operating on the types.

libcaf_test/caf/test/runnable.hpp
```diff
@@ -358,6 +187,18 @@ class CAF_TEST_EXPORT runnable {
   }
 #endif
 
+  template <class T>
+  void should_fail(T expr, const caf::detail::source_location& location
+                           = caf::detail::source_location::current()) {
```
Neverlord, line 192:
```suggestion
  /// Runs `expr()` and assumes that it contains a single check that will fail.
  template <class Expr>
  void should_fail(Expr&& expr, const caf::detail::source_location& location
                           = caf::detail::source_location::current()) {
```

libcaf_test/caf/test/runnable.hpp
```diff
@@ -358,6 +187,18 @@ class CAF_TEST_EXPORT runnable {
...
+    auto* rep = caf::test::reporter::instance;
```
Neverlord, line 193:
```suggestion
    auto* rep = reporter::instance;
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -7,33 +7,37 @@
...
-  check_le(stats.passed, 0u);
-  check_le(stats.passed, 1u);
-  check_le(stats.passed, 2u);
-  /// tests all of the possible scenario for check_lt
-  check_lt(stats.passed, 0u);
-  check_lt(stats.passed, 1u);
-  check_lt(stats.passed, 2u);
-  info("reset error count to not fail the test");
-  caf::test::reporter::instance->test_stats({16, 0});
+  SECTION("checks functionality of check_ne") {
```
Neverlord, line 13: Tests and sections should have descriptive titles, not a description of what the section does (applies to all section titles). For example:

```suggestion
  SECTION("check_ne checks for inequality") {
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -7,33 +7,37 @@
...
+    check_ne(0u, 1u);
+    should_fail([this]() { check_ne(0u, 0u); });
```
Neverlord, line 15: Did you simply keep the unsigned types because they've been there before? I think just using integers on both sides is cleaner and adds less noise:

```suggestion
    should_fail([this]() { check_ne(0, 0); });
```

Neverlord (changes requested):
Just some nitpicks this time. 👍

libcaf_test/caf/test/runnable.hpp
```diff
@@ -199,6 +187,14 @@ class CAF_TEST_EXPORT runnable {
       rep->test_stats({before.passed + 1, before.failed});
   }
 
+  template <class T0, class T1>
+  static void assert_save_comparison(T0&&, T1&&) {
```
Neverlord, line 191: As I said before, this function does not need actual function arguments:

```suggestion
  static void assert_save_comparison() {
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -10,33 +10,33 @@ using caf::test::block_type;
...
+  SECTION("check_le checks that lhs is lesser than or equals to rhs") {
+    check_le(0, 0);
+    check_le(1, 2);
+    should_fail([this]() { check_le(2, 1); });
   }
-  SECTION("checks functionality of check_lt") {
-    check_lt(1u, 2u);
-    should_fail([this]() { check_lt(1u, 1u); });
-    should_fail([this]() { check_lt(2u, 1u); });
+  SECTION("check_lt checks that lhs is lesser than rhs") {
```
Neverlord, line 36:
```suggestion
  SECTION("check_lt checks that lhs is less than rhs") {
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -10,33 +10,33 @@ using caf::test::block_type;
...
+  SECTION("check_gt checks that lhs is greater than rhs") {
+    check_gt(2, 1);
+    should_fail([this]() { check_gt(0, 0); });
+    should_fail([this]() { check_gt(1, 2); });
   }
-  SECTION("checks functionality of check_le") {
-    check_le(0u, 0u);
-    check_le(1u, 2u);
-    should_fail([this]() { check_le(2u, 1u); });
+  SECTION("check_le checks that lhs is lesser than or equals to rhs") {
```
Neverlord, line 31:
```suggestion
  SECTION("check_le checks that lhs is less than or equal to rhs") {
```

Neverlord (changes requested):

libcaf_test/caf/test/runnable.hpp
shariarriday: We can not use this approach as the compiler will not be able to deduce the types in this case. This would work if the whole class was templated.

Neverlord: Of course the deduction won't work if there aren't arguments. You need to call it differently: `assert_save_comparison<T0, T1>()`.

shariarriday: Sorry, I did not understand that you meant to use the classes as template argument directly. I have updated the code accoordingly.

libcaf_test/caf/test/runnable.hpp
```diff
@@ -103,6 +175,26 @@ class CAF_TEST_EXPORT runnable {
...
+    auto before = rep->test_stats();
+    expr();
+    auto after = rep->test_stats();
+    check_eq(before.passed, after.passed, location);
+    if (check_eq(before.failed + 1, after.failed, location))
+      rep->test_stats({before.passed + 1, before.failed});
+  }
+
+  template <class T0, class T1>
+  static void assert_save_comparison() {
```
Neverlord, line 191: It seems like you've missed this part from my earlier comment:

> I think this should be factored out into a (**private**) function.

This is a pure implementation detail and should not be part of the public API.

Neverlord (changes requested):

libcaf_test/caf/test/runnable.hpp
shariarriday: Yes, sorry I seem to have missed it. I have updated the PR accordingly.

Issue #1389, opened 2023-03-12
More Flow Operators


Neverlord:
CAF 0.19 introduced ReactiveX-style flow processing with a base set of the most useful operators. There are more operators defined by ReactiveX and eventually we should implement all of them:

## Creating Observables

- [ ] Start: https://reactivex.io/documentation/operators/start.html

## Transforming Observables

- [ ] GroupBy: https://reactivex.io/documentation/operators/groupby.html
- [ ] Scan: https://reactivex.io/documentation/operators/scan.html
- [ ] Window: https://reactivex.io/documentation/operators/window.html

## Filtering Observables

- [ ] TakeLast: https://reactivex.io/documentation/operators/takelast.html
- [ ] SkipLast: https://reactivex.io/documentation/operators/skiplast.html
- [ ] Sample: https://reactivex.io/documentation/operators/sample.html
- [ ] Last: https://reactivex.io/documentation/operators/last.html
- [ ] IgnoreElements: https://reactivex.io/documentation/operators/ignoreelements.html
- [ ] First: https://reactivex.io/documentation/operators/first.html
- [ ] ElementAt: https://reactivex.io/documentation/operators/elementat.html
- [ ] Debounce: https://reactivex.io/documentation/operators/debounce.html

## Combining Observables

- [ ] And/Then/When: https://reactivex.io/documentation/operators/and-then-when.html
- [ ] CombineLatest: https://reactivex.io/documentation/operators/combinelatest.html
- [ ] Join: https://reactivex.io/documentation/operators/join.html
- [ ] StartWith: https://reactivex.io/documentation/operators/startwith.html
- [ ] Switch: https://reactivex.io/documentation/operators/switch.html

## Error Handling Operators

- [ ] Catch: https://reactivex.io/documentation/operators/catch.html
- [ ] Retry: https://reactivex.io/documentation/operators/retry.html

Neverlord:
@shariarriday these should be straightforward to implement as a `step` (like `take`):

- First (actually, this is just `take(1)`)
- TakeLast
- Last
- IgnoreElements
- ElementAt
- SkipLast

Neverlord:
The `catch` operator is actually more like a family of operators. For example, [RxJava](https://github.com/ReactiveX/RxJava/wiki/Error-Handling-Operators) splits this into `onErrorComplete`, `onErrorResumeNext`, `onErrorReturn`, `onErrorReturnItem` and `onExceptionResumeNext`.

[...]

PR #1504, opened 2023-07-22
Add take last flow for observables

shariarriday:
https://github.com/actor-framework/actor-framework/issues/1389

Take last has been implemented. The current implementation saves all the data in vector and returns them on_complete. This can be improved upon.


libcaf_core/caf/flow/step/all.hpp
```diff
@@ -9,4 +9,5 @@
 #include "caf/flow/step/reduce.hpp"
 #include "caf/flow/step/skip.hpp"
 #include "caf/flow/step/take.hpp"
-#include "caf/flow/step/take_while.hpp"
+#include "caf/flow/step/take_last.hpp"
+#include "caf/flow/step/take_while.hpp"
```
Neverlord, line 13: Please fix your editor. Proper text files must have a newline at the end.

libcaf_core/caf/flow/observable.hpp
```diff
@@ -223,6 +223,11 @@ class observable_def {
     return add_step(step::take<output_type>{n});
   }
 
+  /// @copydoc observable::take_last
+  auto take_last(size_t n) && {
+    return add_step(step::take_last<output_type>{n});
+  }
```
Neverlord, line 229: There is no `take_last` in `observable`. You didn't add it there.

libcaf_core/caf/flow/step/take_last.hpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
+    auto take_last_begin = elements_.begin();
+    if (elements_.size() > last_elements_count_)
+      take_last_begin += (elements_.size() - last_elements_count_);
+    std::for_each(take_last_begin, elements_.end(),
+                  [&](auto item) { next.on_next(item, steps...); });
+    next.on_complete(steps...);
```
Neverlord, line 42: 1) This logic is unnecessary, since the buffer shouldn't hold more elements than requested.
2) `next.on_next` may return `false`, in which case this operator needs to stop calling anything on `next`.

libcaf_core/caf/flow/step/take_last.hpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  take_last(take_last&&) = default;
+  take_last(const take_last&) = default;
+  take_last& operator=(take_last&&) = default;
+  take_last& operator=(const take_last&) = default;
+
+  template <class Next, class... Steps>
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
+    elements_.emplace_back(std::move(item));
```
Neverlord, line 31: This implementation has multiple issues. If you only take `m` elements, this still requires to store all `n` elements. This is a total waste of both memory and CPU cycles (due to unnecessary copying operations for frequent re-allocations).

This operator should not buffer more than `m` elements at any point in time. It also shouldn't need to allocate memory more than once, because we know the maximum space requirement at construction time and can use a simple ring buffer.

libcaf_core/test/flow/mixed.cpp
```diff
@@ -69,6 +69,31 @@ SCENARIO("sum up all the multiples of 3 or 5 below 1000") {
   }
 }
 
+SCENARIO("take the last 5 digits in a series of 100 digits") {
```
Neverlord, line 72: Please move these tests to a new `caf/flow/observable.test.cpp` that is using the new testing framework. The tests are ok, but they operate on the blueprints. There should be at least one additional test that calls `take_last` on an actual `observable`.

Neverlord (changes requested):

libcaf_core/caf/flow/step/take_last.hpp
shariarriday: I had planned this implementation but I came across this doc in ReactiveX (http://introtorx.com/Content/v1.0.10621.0/05_Filtering.html#SkipLastTakeLast) and decided that we need to cache all the values. For ring buffer I was planning to use boost circular buffer https://www.boost.org/doc/libs/1_81_0/doc/html/circular_buffer.html. But, as we currently don't use anything boost the closest thing would be std::deque (https://en.cppreference.com/w/cpp/container/deque) which still has some performance issues. Can you tell me your thoughts on this?

libcaf_core/caf/flow/step/take_last.hpp
shariarriday: In the scenario that next.on_next returns false but the buffer is not empty should we call next.on_error ?

libcaf_core/caf/flow/step/take_last.hpp
Neverlord: Pulling in a new dependency for this would be overkill. All we need here is an index for the current writing position, a bit of module math and a flag that tells us whether we've filled the cache at least once (to know whether we `< n` elements).

libcaf_core/caf/flow/step/take_last.hpp
Neverlord: No. If `on_next` returns `false`, we may not call anything on `next` again. It's done processing the flow.

libcaf_core/caf/flow/step/take_last.hpp
shariarriday: In that case, I think that it is better to create a class for ring buffer data structure instead of just adding the code directly into the observable class in case we need it for other classes in future. What are your thoughts on this?

Neverlord: Yeah, good point. We already have a ring buffer class, but that is for synchronized producers/consumers. I've filed #1508 to give that old class a better name. When that's merged, you can go ahead an implement a new class `detail::ring_buffer` for this use case. But keep the interface simple. Probably all we need is `push_back`, `size`, `front` and `pop_front`.

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "caf/config.hpp"
+
+#include <vector>
+
+namespace caf::detail {
+
+// A simple ring buffer implementation
+template <class T>
+class ring_buffer {
```
Neverlord, line 15: A few observations on this container:

- Your implementation will crash when constructing a ring-buffer of size `0` and calling `push_back` on it. None of your tests covers this case.
- Having a `vector` member adds unnecessary overhead. A `std::unique_ptr<T[]>` plus a `size_` member are sufficient for our needs since we don't need re-allocations and capacity management.
- I found the member names a bit confusing. The `curr_back_` member is the writing position, so it's actually the past-the-end position, not the position of the `back` element. I would find `write_pos_` much more intuitive.
- The `curr_front_` member is redundant and I would argue dangerous since it can run out of sync. We can compute this index from the writing position at any time: `auto front_index = (write_pos_ + max_size_ - size_) % max_size_`.
- This implementation monotonically increments the write position, why did you implement it this way instead of `write_pos_ = (write_pos_ + 1) % max_size_`?

I think this container is asking for no more than three `size_t` members (`size_`, `max_size_` and `write_pos_`) plus a `data_` pointer. Your current implementation uses five `size_t` members plus a pointer (`std::vector` has capacity, size and a data pointer).

Neverlord (changes requested):

libcaf_core/caf/detail/ring_buffer.hpp
shariarriday: I think that using monotonically increasing write position may give us information related to how much data is written into the buffer if we need to access it. We can remove this easily and the front index can be calculated easily in that scenario.

shariarriday: Also, I think there is another issue with the current implementation. There are no checks before pop_front()/front() to handle empty buffers. I think that this should throw an error as we should not return any value from buffer when it is empty.

Neverlord: Calling `front` on an empty container is undefined behavior. No need for checks here (the standard library doesn't do checks here either).

Handling a zero-sized buffer gracefully does make sense to stay consistent with the rest of the language (`auto ptr = new int[0]` is valid C++). I think a zero-sized buffer should simply stay `empty`, no matter how often you call `push_back` on it.

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,76 @@
...
+      buf_[i] = other.buf_[i];
+  }
+
+  T& front() {
+    return buf_[(write_pos_ + max_size_ - size_ + 1) % max_size_];
+  }
+
+  void pop_front() {
+    if (empty())
+      return;
```
Neverlord, line 40:
```suggestion
    CAF_ASSERT(!empty());
```

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,115 @@
...
+  check_eq(buf.front(), 2);
+  buf.pop_front();
+  check_eq(buf.full(), false);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 3);
+  buf.pop_front();
+  check_eq(buf.empty(), true);
+}
+
+TEST("circular buffer functions properly") {
```
Neverlord, line 60: Can you please give names to tests that gives a better idea on what is actually being tested? Without a definition for "properly", this test name basically says nothing.

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,115 @@
...
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 4);
+  buf.push_back(9);
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 5);
+}
+
+TEST("empty buffer is handled correctly") {
```
Neverlord, line 93: What is "correctly"? As a reader, I don't know what this should mean or what this test is about. From looking at the test, I think something like "calling push_back on a buffer with size 0 does nothing"?

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include "caf/test/caf_test_main.hpp"
+#include "caf/test/test.hpp"
+
+using namespace caf;
+
+using detail::ring_buffer;
+
+TEST("push_back adds element") {
+  ring_buffer<int> buf(3);
```
Neverlord, line 15: Please take this to heart, I've mentioned it before: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax

```suggestion
  ring_buffer<int> buf{3};
```

libcaf_core/caf/flow/step/take_last.hpp
```diff
@@ -0,0 +1,58 @@
...
+template <class T>
+class take_last {
+public:
+  using input_type = T;
+
+  using output_type = T;
+
+  explicit take_last(size_t num)
+    : last_elements_count_(num),
+      elements_(caf::detail::ring_buffer<output_type>(num)) {
```
Neverlord, line 23: Why are you creating a temporary to initialize this member?

```suggestion
      elements_(num) {
```

libcaf_core/caf/flow/step/all.hpp
Neverlord: This still isn't fixed as far as I can tell.

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+  template <class... Ts>
+  std::vector<int> ls(Ts... xs) {
+    return std::vector<int>{xs...};
+  }
+};
+
+WITH_FIXTURE(fixture) {
+
+TEST("take the last 5 digits in a series of 100 digits") {
```
Neverlord, line 30: We aren't processing strings here. Did you mean "numbers" instead of "digits"?

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,113 @@
...
+  SECTION("blueprint") {
+    auto result = std::vector<std::vector<int>>{};
+    ctx->make_observable().range(1, 100).take_last(5).to_vector().for_each(
```
Neverlord, line 33: We push to a vector anyways, so we don't need to use `to_vector`.

```suggestion
    ctx->make_observable().range(1, 100).take_last(5).for_each(
```

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,113 @@
...
+    check_eq(result.size(), 1u);
+    check_eq(result.at(0), ls(96, 97, 98, 99, 100));
+  }
+  SECTION("observable") {
+    auto result = std::vector<std::vector<int>>{};
+    ctx->make_observable()
+      .range(1, 100)
+      .as_observable()
+      .take_last(5)
+      .to_vector()
```
Neverlord, line 47:
```suggestion
```

Neverlord (changes requested):

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,113 @@
...
+TEST("take the last 5 digits in a series of 100 digits") {
+  SECTION("blueprint") {
+    auto result = std::vector<std::vector<int>>{};
+    ctx->make_observable().range(1, 100).take_last(5).to_vector().for_each(
+      [&](const caf::cow_vector<int>& result_cow) {
+        result.emplace_back(result_cow.std());
+      });
+    ctx->run();
+    check_eq(result.size(), 1u);
+    check_eq(result.at(0), ls(96, 97, 98, 99, 100));
```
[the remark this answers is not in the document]
shariarriday: We are checking result which is declared inside the scope. If we move it outside the scope the variable is not accessible and we can not run the tests properly. From my understanding you meant to do something like the following which will not compile due to out of scope "result"

```
TEST("take the last 5 numbers in a series of 3 numbers") {
  SECTION("blueprint") {
    auto result = std::vector<std::vector<int>>{};
    ctx->make_observable().range(1, 2).take_last(5).for_each(
      [&](const caf::cow_vector<int>& result_cow) {
        result.emplace_back(result_cow.std());
      });
  }
  SECTION("observable") {
    auto result = std::vector<std::vector<int>>{};
    ctx->make_observable().range(1, 2).as_observable().take_last(5).for_each(
      [&](const caf::cow_vector<int>& result_cow) {
        result.emplace_back(result_cow.std());
      });
  }
  ctx->run();
  check_eq(result.size(), 1u);
  check_eq(result.at(0), ls(1, 2));
}
```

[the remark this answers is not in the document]
Neverlord: Did you try to move the `result` variable just above the sections?

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <memory>
+
+namespace caf::detail {
+
+// A simple ring buffer implementation
+template <class T>
+class ring_buffer {
+public:
+  explicit ring_buffer(size_t max_size)
+    : max_size_(max_size), size_(0), write_pos_(0) {
```
shariarriday, line 18: In visual studio I observed that write_pos_ and size_ were initialized with garbage value which introduces undefined behavior, I chose to keep it in this way for that reason.

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,85 @@
...
+  ring_buffer& operator=(ring_buffer&& other) noexcept = default;
+
+  explicit ring_buffer(const ring_buffer& other) {
+    max_size_ = other.max_size_;
+    write_pos_ = other.write_pos_;
+    size_ = other.size_;
+    buf_ = std::make_unique<T[]>(max_size_);
+    if (max_size_ > 0)
+      std::copy(other.buf_.get(), other.buf_.get() + other.max_size_,
+                buf_.get());
```
Neverlord, line 34:
```suggestion
    if (max_size_ > 0) {
      buf_ = std::make_unique<T[]>(max_size_);
      std::copy(other.buf_.get(), other.buf_.get() + other.max_size_,
                buf_.get());
```

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  ring_buffer& operator=(const ring_buffer& other) {
+    ring_buffer copy_ring_buffer{other.max_size_};
+    if (max_size_ > 0)
+      std::copy(other.buf_.get(), other.buf_.get() + other.max_size_,
+                copy_ring_buffer.buf_);
```
Neverlord, line 41: This copy-assignment operator is broken in multiple ways. Which very clearly tells me that there is no test for it.

What are you trying to do here? For the copy-and-swap idiom, this misses the `swap` part. Also, the copy constructor for the copy already takes care of copying the content, what are you calling `std::copy` for? Last but not least, this is UB since you don't return from the function (`return *this`?).

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,85 @@
...
+  bool empty() const noexcept {
+    return size_ == 0;
+  }
+
+  size_t size() const noexcept {
+    return size_;
+  }
+
+private:
+  /// write index for the buffer, new elements will be added in this index
```
Neverlord, line 75: As I mentioned before, please use full sentences. Properly capitalized and with a `.` at the end. Leave a whitespace between the declarations.

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  check_eq(buf.front(), 2);
+  buf.pop_front();
+  check_eq(buf.full(), false);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 3);
+  buf.pop_front();
+  check_eq(buf.empty(), true);
+}
+
+TEST("circular buffer overwrites oldest element after it is full") {
```
Neverlord, line 60:
```suggestion
TEST("pushing to a full buffer overwrites the oldest element") {
```

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 4);
+  buf.push_back(9);
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 5);
+}
+
+TEST("empty ring buffer ignores call to push_back") {
```
Neverlord, line 118: If empty ring buffer ignored `push_back`, we had a problem. Did you mean buffers with capacity 0?

```suggestion
TEST("push_back does nothing for ring buffer with a capacity of 0") {
```

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  info("empty buffer is initialized");
+  check_eq(buf.size(), 0u);
+  for (int i = 1; i <= 3; ++i) {
+    buf.push_back(i);
+  }
+  info("buffer size after adding some elements");
+  check_eq(buf.size(), 0u);
+}
+
+TEST("buffer size is calculated correctly") {
```
Neverlord, line 129: There is no context here what "correctly" means.

```suggestion
TEST("size() returns the number of elements in a buffer") {
```

libcaf_core/caf/flow/step/take_last.hpp
```diff
@@ -0,0 +1,55 @@
...
+    next.on_complete(steps...);
+  }
+
+  template <class Next, class... Steps>
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    next.on_error(what, steps...);
+  }
+
+private:
+  caf::detail::ring_buffer<output_type> elements_;
```
Neverlord, line 52: You are in a sub-namespace of `caf`, no need to spell it out again:

```suggestion
  detail::ring_buffer<output_type> elements_;
```

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+    ctx->make_observable().range(1, 3).as_observable().take_last(5).for_each(
+      [&](const int& result_observable) {
+        result.emplace_back(result_observable);
+      });
+  }
+  ctx->run();
+  check_eq(result.size(), 3u);
+  check_eq(result, ls(1, 2, 3));
+}
+}
```
Neverlord, line 86:
```suggestion
}

} // WITH_FIXTURE(fixture)
```

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,85 @@
...
+#pragma once
+
+#include "caf/config.hpp"
+
+#include <algorithm>
+#include <memory>
+
+namespace caf::detail {
+
+// A simple ring buffer implementation
```
Neverlord, line 14:
```suggestion
/// A simple ring buffer implementation.
```

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 3);
+  buf.push_back(6);
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 4);
+}
+
+TEST("pop_front removes element") {
```
Neverlord, line 38:
```suggestion
TEST("pop_front removes the oldest element") {
```

libcaf_core/caf/detail/ring_buffer.test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 4);
+  buf.push_back(9);
+  check_eq(buf.full(), true);
+  check_eq(buf.empty(), false);
+  check_eq(buf.front(), 5);
+}
+
+TEST("new elements overwrite old elements that are popped") {
```
Neverlord, line 85: This title doesn't make sense to me. If we pop an element it's no longer in the buffer. How can it override something? Did you mean to say something like "pop_front removes the oldest element from the buffer"? Looking at the implementation, I think this is what it's actually testing for.

libcaf_core/caf/flow/step/take_last.test.cpp
```diff
@@ -0,0 +1,88 @@
...
+using namespace caf;
+
+using caf::flow::make_observer;
+
+struct fixture {
+  flow::scoped_coordinator_ptr ctx = flow::make_scoped_coordinator();
+
+  template <class... Ts>
+  std::vector<int> ls(Ts... xs) {
+    return std::vector<int>{xs...};
```
Neverlord, line 24: Can be shortened, never touches `this`:

```suggestion
  static auto ls(Ts... xs) {
    return std::vector<int>{xs...};
```

Neverlord (changes requested):

Neverlord (changes requested):
In case you didn't see: there is no coverage for theses cases:

- Calling `on_error` on the new `take_last` step.
- Having a `next` that only consumes a substet of values.

Translated to "high-level" terms, I think there should be at least two more tests. Here are some ideas:

- `TEST("the take_last operator forwards errors")`
- `TEST("calling take(3) on take_last(5) ignores the last two items")`

I hope you get the idea. 🙂

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+  explicit ring_buffer(size_t max_size) : max_size_(max_size) {
+    if (max_size > 0)
+      buf_ = std::make_unique<T[]>(max_size);
+  }
+
+  ring_buffer(ring_buffer&& other) noexcept = default;
+
+  ring_buffer& operator=(ring_buffer&& other) noexcept = default;
+
+  explicit ring_buffer(const ring_buffer& other) {
```
Neverlord, line 27: Copy constructors are never explicit:

```suggestion
  ring_buffer(const ring_buffer& other) {
```

libcaf_core/caf/detail/ring_buffer.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+  size_t size() const noexcept {
+    return size_;
+  }
+
+  friend void swap(ring_buffer& first, ring_buffer& second) noexcept {
+    std::swap(first.buf_, second.buf_);
+    std::swap(first.size_, second.size_);
+    std::swap(first.max_size_, second.max_size_);
+    std::swap(first.write_pos_, second.write_pos_);
```
Neverlord, line 78: This code is not DRY:

```suggestion
    using std::swap;
    swap(first.buf_, second.buf_);
    swap(first.size_, second.size_);
    swap(first.max_size_, second.max_size_);
    swap(first.write_pos_, second.write_pos_);
```

libcaf_core/caf/flow/step/take_last.hpp
```diff
@@ -0,0 +1,55 @@
...
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
+    elements_.push_back(item);
+    return true;
+  }
+
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
+    while (!elements_.empty()) {
+      if (!next.on_next(elements_.front(), steps...))
+        break;
```
Neverlord, line 40: This is a bug: we must not call *anything* on `next` if it returns `false`:

```suggestion
        return;
```

Neverlord (changes requested):
Three more comments/fixes. Looks good otherwise now.

PR #1506, opened 2023-07-24
Send error if a request fails to deserialize

### LINKED ISSUE #1428: Failing to deserialize a request should trigger an error at the sender
Currently, when a type sent as part of a request to a remote actor fails to deserialize, the sender of the request receives no error. In fact, if the timeout for the request is set to infinite, the sender may never receive a reply at all.

To reproduce the issue, create a type like this that can be serialized but not deserialized:

```c++
struct foo {
  template <class Inspector>
  friend auto inspect(Inspector&, foo&) -> bool {
    return not Inspector::is_loading;
  }
};
```

Then use it in a request to an actor in another actor system than the sending actor (so the message must go through serialization and deserialization):

```c++
self->request(remote_dest, infinite, foo{}).then(
  []() {
    // This (expectedly) never gets called…
  },
  [](const error& err) {
    // …and (unexpectedly) neither does this.
  });
```

We're still using CAF 0.18.x, so please ignore if the issue is fixed for CAF 0.19.x.

If need be, I can also provide a unit test for this, and would appreciate some pointers to writing a test that involves two actor systems, whether that be in one process or two.

riemass:
Closes #1428.

### DIFF
--- a/libcaf_io/caf/io/basp/remote_message_handler.hpp
+++ b/libcaf_io/caf/io/basp/remote_message_handler.hpp
@@ -104,8 +104,7 @@ public:
     if (!source.apply(msg)) {
       CAF_LOG_ERROR("failed to read message content:" << source.get_error());
       src->enqueue(nullptr, mid.response_id(),
-                   make_message(make_error(sec::malformed_message)),
-                   nullptr);
+                   make_message(make_error(sec::malformed_message)), nullptr);
       return;
     }
     telemetry::timer::observe(mm_metrics.deserialization_time, t0);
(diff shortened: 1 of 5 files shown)


libcaf_io/caf/io/basp/remote_message_handler.hpp
```diff
@@ -103,6 +103,8 @@ class remote_message_handler {
     auto t0 = telemetry::timer::clock_type::now();
     if (!source.apply(msg)) {
       CAF_LOG_ERROR("failed to read message content:" << source.get_error());
+      detail::sync_request_bouncer srb{sec::malformed_message};
+      srb(src, mid);
```
Neverlord, line 107: This would currently only send this error message when using requests. I think failing to deserialize a message is severe enough that we should always send the error.

```suggestion
      src->enqueue(nullptr, mid.response_id(),
                   make_message(make_error(sec::malformed_message)),
                   nullptr);
```

Neverlord (changes requested):

Neverlord (approved):

PR #1507, opened 2023-07-25
Fix fallback when inspecting optional types

### LINKED ISSUE #1427: Optional field causes crash in caf::optional_inspector_access
[Stacktrace](https://github.com/actor-framework/actor-framework/files/11733367/stacktrace.txt) from the actual crash we saw (in this case ASAN caught a stack overflow and the stack trace repeats itself endlessly, but it will also just segfault without sanitizers), but a minimal example that should cause the same behavior is this:

```
struct Example {
    std::optional<uint64_t> field;
}

template<typename Inspector>
bool inspect(Inspector& ins, Example& msg) {
    return ins.object(msg).fields(ins.field("field", msg.field).fallback(std::nullopt));
}
```

riemass:
Closes #1427 

This one was sneaky, we had a recursive call between `impl::save_field` and `detail::save_field` when using optional-like types (`unique_ptr`, `intrusive_ptr`, etc).

### DIFF
--- a/libcaf_core/caf/inspector_access.hpp
+++ b/libcaf_core/caf/inspector_access.hpp
@@ -396,7 +396,10 @@ struct optional_inspector_access {
   template <class Inspector, class IsPresent, class Get>
   static bool save_field(Inspector& f, std::string_view field_name,
                          IsPresent& is_present, Get& get) {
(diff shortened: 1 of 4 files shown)


Neverlord (approved):
Thanks for tracking this down. 👍

PR #1511, opened 2023-07-29
Make actor mailboxes configurable

### LINKED ISSUE #1473: Make actor mailboxes configurable
Having configurable mailboxes would allow us to change the mailbox of actors without changing the implementation. We have two immediate use cases for this:

- Actor groups (see #1453).
- Deterministic testing. Currently, the setup for doing deterministic testing is quite convoluted and tricky. If we could tell all actors to put their messages to a single, global queue, things would become much easier.

Neverlord:
- Add a new interface for actor mailboxes.
- Add a new interface for mailbox factories.
- Add a new parameter to the `actor_config` for setting a custom mailbox factory.

Closes #1473.

@riemass technically, this is breaking "public" API by removing a whole bunch of classes from `caf::intrusive`. However, I *highly* doubt that any user depends on classes in this namespace directly. So I took some liberties in streamlining everything under `caf::intrusive` and to get rid of a whole lot of cruft. This turned out to be a bit of a larger change set, but I tried to break up the overall work into smaller pieces of work and have one commit per chunk. So maybe reviewing the commit diffs one by one makes it easier to review.


libcaf_core/caf/intrusive/linked_list.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+  /// Removes the first element from the queue and returns it.
+  unique_pointer pop_front() {
+    unique_pointer result;
+    if (!empty()) {
+      auto ptr = promote(head_.next);
+      head_.next = ptr->next;
+      if (--size_ == 0) {
+        CAF_ASSERT(head_.next == &(tail_));
+        tail_.next = &(head_);
```
riemass, line 116: Not sure about the parentheses?
```suggestion
        CAF_ASSERT(head_.next == &tail_);
        tail_.next = &head_;
```

riemass (changes requested):
Great work. Now that I've went through the implementation, it seems a lot cleaner, not only because of the interface, but the code needed some segregation of concerns. 
Here are some points, as good as I could review them. 
Also, are there some pros/cons on why you axed the DRR stuff?

Neverlord:
> Also, are there some pros/cons on why you axed the DRR stuff?

To give a bit of context: the old design had messages and stream data flow through the same mailbox and the actor would then use DRR (deficit-round-robin) to make sure it somewhat evenly multiplexes the various inputs. It sounds like in theory, but it actually overcomplicated things because the mailbox did too many things. After switching to the new flow API, we simply no longer have any need for the DRR stuff and can use a much simpler implementation.

libcaf_core/caf/intrusive/linked_list.hpp
```diff
@@ -0,0 +1,285 @@
...
+    size_ = 0;
+  }
+
+  // -- member variables ------------------------------------------------------
+
+  /// Dummy before-the-first-element node.
+  node_type head_;
+
+  /// Dummy past-the-last-element node. The `tail_->next` pointer is pointing to
+  /// the last element in the queue or to `head_` if the queue is empty.
```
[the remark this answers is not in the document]
Neverlord: ACK on the `queue => list` rename, but the the dummy really is the past-the-end node, i.e., `end()` returns a pointer to `tail_`.

libcaf_core/caf/intrusive/linked_list.hpp
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  // -- member variables ------------------------------------------------------
+
+  /// Dummy before-the-first-element node.
+  node_type head_;
+
+  /// Dummy past-the-last-element node. The `tail_->next` pointer is pointing to
+  /// the last element in the list or to `head_` if the list is empty.
+  node_type tail_;
```
riemass, line 277: Could we add these changes, and then skip calling `init` in the default and rvalue ctors?

```suggestion
  node_type head_ = &tail_;

  /// Dummy past-the-last-element node. The `tail_->next` pointer is pointing to
  /// the last element in the list or to `head_` if the list is empty.
  node_type tail_ = &head_;
```

riemass (approved):
:1st_place_medal:

riemass (approved):
Great find in the last minute.

PR #1514, opened 2023-08-05
Move legacy tests to proper subfolder

riemass:
This PR basically moves the content of `test` to `tests/legacy` for each caf module. 
The first commit is just the `mv` commands. The second commit integrates these changes into CMake and should be the focus in the review.


CMakeLists.txt
```diff
@@ -124,6 +124,17 @@ if(CAF_ENABLE_TESTING)
       endif()
     endforeach()
   endfunction()
+  function(caf_add_legacy_test_suites target)
```
Neverlord, line 127: You can actually remove `caf_add_test_suites` now. We won't use it anymore. Same for the `TEST_SOURCES` and `TEST_SUITES` markers.

Neverlord (changes requested):

Neverlord (approved):

PR #1515, opened 2023-08-05
Turn warnings into errors in CI

### LINKED ISSUE #1436: Build with -Werror in the pipeline
To get the code quality up and make the most out of our various compilers, we should enable `-Werror` on CI. Probably enabling this flag build-by-build is the best strategy to avoid instant overload. Eventually, we might also want to tighten up the warning settings as well.

riemass:
Closes #1436 

Covers Windows, Linux and FreeBSD.


libcaf_net/caf/net/http/arg_parser.hpp
```diff
@@ -29,7 +29,7 @@ struct builtin_arg_parser {
     if (auto err = detail::parse(str, tmp); !err)
       return tmp;
     else
-      return {};
+      return std::nullopt;
```
Neverlord, line 32: I'd say `return {};` is the preferred way, because it doesn't rely on an implicit conversion. What's the warning here?

libcaf_net/caf/net/http/request_header.hpp
```diff
@@ -121,7 +121,7 @@ class CAF_NET_EXPORT request_header {
       if (auto res = caf::get_as<T>(val))
         return std::move(*res);
     }
-    return {};
+    return std::nullopt;
```
Neverlord, line 124: Same question.

libcaf_test/caf/test/test.test.cpp
```diff
@@ -9,7 +9,7 @@ using caf::test::block_type;
 
 TEST("tests can contain different types of checks") {
   auto* rep = caf::test::reporter::instance;
-  auto stats = rep->test_stats();
+  [[maybe_unused]] auto stats = rep->test_stats();
```
Neverlord, line 12: If it's unused, we can just remove it.

```suggestion
```

Neverlord (changes requested):

libcaf_net/caf/net/http/arg_parser.hpp
riemass: Here is the exact warning. Same for all `return {}` locations. 
```
warning: '<anonymous>' may be used uninitialized in this function [-Wmaybe-uninitialized]
     return {};
```
I could replace it with return `std::optional{};`

Neverlord (approved):

PR #1517, opened 2023-08-06
Add example for caf::after

### LINKED ISSUE #1395: Add example for after(...)
This came up in chat recently: we currently have no example that shows how to use `after(...)`.

riemass:
Closes #1395.

### DIFF
--- a/examples/message_passing/after.cpp
+++ b/examples/message_passing/after.cpp
@@ -1,4 +1,4 @@
-// This example shows how to use caf::after
+// This example shows how to use caf::after.
 
 #include "caf/after.hpp"
 
@@ -16,7 +16,7 @@ using std::endl;
 
 using namespace caf;
 
-// Sends a random number of printable characters to buddy and exits
+// Sends a random number of printable characters to `sink` and then quits.
 void generator(event_based_actor* self, actor sink) {
   std::random_device rd;
   std::minstd_rand gen{rd()};
--- a/examples/message_passing/after.cpp
+++ b/examples/message_passing/after.cpp
@@ -11,7 +11,7 @@ using std::endl;
 
 using namespace caf;
 
-// Sends random caracters to buddy, and then waits for a letter back
+// Sends random characters to buddy, and then waits for a letter back
 behavior noisy_buddy(event_based_actor* self, actor buddy) {
   using namespace std::chrono_literals;
   std::random_device rd;
--- a/examples/message_passing/after.cpp
+++ b/examples/message_passing/after.cpp
@@ -1,0 +1,59 @@
+// This example shows how to use caf::after
+
+#include "caf/all.hpp"
+
+#include <algorithm>
+#include <chrono>
+#include <iostream>
+#include <random>
+using std::cout;
+using std::endl;
+
+using namespace caf;
+
+// Sends random caracters to buddy, and then waits for a letter back
+behavior noisy_buddy(event_based_actor* self, actor buddy) {
+  using namespace std::chrono_literals;
+  std::random_device rd;
+  std::minstd_rand gen{rd()};
+  const auto count = std::uniform_int_distribution<>(20, 100)(gen);
+  std::uniform_int_distribution<> dis(33, 126);
+  for (auto i = 0; i < count; i++) {
+    self->send(buddy, static_cast<char>(dis(gen)));
+  }
+  return {[self](std::string letter) {
+    cout << "Received a letter:" << endl << letter << endl;
+    self->quit();
+  }};
+}
+
+struct aggregator_state {
+  std::string letter;
+  caf::actor dest;
+};
+
+// Aggregates incoming characters and stores the sender, replies with the
+// reversed string when inactive for 100ms
+behavior aggregator(stateful_actor<aggregator_state>* self) {
+  using namespace std::chrono_literals;
+  return {
+    [=](char c) mutable {
+      self->state.dest = caf::actor_cast<caf::actor>(self->current_sender());
+      self->state.letter.push_back(c);
+    },
+    // trigger if we dont receive a message for 100ms
+    caf::after(100ms) >>
+      [=] {
+        std::reverse(self->state.letter.begin(), self->state.letter.end());
+        self->send(self->state.dest, self->state.letter);
+        cout << "bye" << endl;
+        self->quit();
+      }};
+}
+
+void caf_main(actor_system& system) {
+  auto agg = system.spawn(aggregator);
+  system.spawn(noisy_buddy, agg);
+}
+
+CAF_MAIN()
--- a/examples/message_passing/after.cpp
+++ b/examples/message_passing/after.cpp
@@ -2,58 +2,56 @@
 
 #include "caf/all.hpp"
 
-#include <algorithm>
 #include <chrono>
 #include <iostream>
 #include <random>
+
 using std::cout;
 using std::endl;
 
 using namespace caf;
 
-// Sends random characters to buddy, and then waits for a letter back
-behavior noisy_buddy(event_based_actor* self, actor buddy) {
-  using namespace std::chrono_literals;
+// Sends a random number of printable characters to buddy and exits
+void generator(event_based_actor* self, actor buddy) {
   std::random_device rd;
   std::minstd_rand gen{rd()};
-  const auto count = std::uniform_int_distribution<>(20, 100)(gen);
-  std::uniform_int_distribution<> dis(33, 126);
+  const auto count = std::uniform_int_distribution<>{20, 100}(gen);
+  std::uniform_int_distribution<> dis{33, 126};
   for (auto i = 0; i < count; i++) {
     self->send(buddy, static_cast<char>(dis(gen)));
   }
-  return {[self](std::string letter) {
-    cout << "Received a letter:" << endl << letter << endl;
-    self->quit();
-  }};
 }
 
-struct aggregator_state {
-  std::string letter;
-  caf::actor dest;
-};
-
-// Aggregates incoming characters and stores the sender, replies with the
-// reversed string when inactive for 100ms
-behavior aggregator(stateful_actor<aggregator_state>* self) {
+// Collects the incoming characters until either the awaited_size of characters
+// is received, or no new characters arrive for 100ms
+behavior collector(stateful_actor<std::string>* self, size_t awaited_size) {
   using namespace std::chrono_literals;
+  self->state.reserve(awaited_size);
   return {
-    [=](char c) mutable {
-      self->state.dest = caf::actor_cast<caf::actor>(self->current_sender());
-      self->state.letter.push_back(c);
+    [=](char c) {
+      self->state.push_back(c);
+      if (self->state.size() == awaited_size) {
+        cout << "Received message length: " << self->state.size() << endl
+             << "Message content: " << self->state << endl;
+        self->quit();
+      }
     },
     // trigger if we dont receive a message for 100ms
     caf::after(100ms) >>
-      [=] {
-        std::reverse(self->state.letter.begin(), self->state.letter.end());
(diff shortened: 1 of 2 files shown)


examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+using std::endl;
+
+using namespace caf;
+
+// Sends random characters to buddy, and then waits for a letter back
+behavior noisy_buddy(event_based_actor* self, actor buddy) {
+  using namespace std::chrono_literals;
+  std::random_device rd;
+  std::minstd_rand gen{rd()};
+  const auto count = std::uniform_int_distribution<>(20, 100)(gen);
```
Neverlord, line 19: Please stick to uniform initialization syntax.

```suggestion
  const auto count = std::uniform_int_distribution<>{20, 100}(gen);
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+  std::uniform_int_distribution<> dis(33, 126);
```
Neverlord, line 20:
```suggestion
  std::uniform_int_distribution<> dis{33, 126};
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+  for (auto i = 0; i < count; i++) {
+    self->send(buddy, static_cast<char>(dis(gen)));
+  }
+  return {[self](std::string letter) {
+    cout << "Received a letter:" << endl << letter << endl;
+    self->quit();
+  }};
```
Neverlord, line 27: We always put a `,` after message handlers when initializing behaviors. Primarily because it'll make `clang-format` do a better job.

```suggestion
  },};
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+  std::string letter;
+  caf::actor dest;
+};
+
+// Aggregates incoming characters and stores the sender, replies with the
+// reversed string when inactive for 100ms
+behavior aggregator(stateful_actor<aggregator_state>* self) {
+  using namespace std::chrono_literals;
+  return {
+    [=](char c) mutable {
```
Neverlord, line 40:
```suggestion
    [self](char c) {
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+      self->state.dest = caf::actor_cast<caf::actor>(self->current_sender());
+      self->state.letter.push_back(c);
+    },
+    // trigger if we dont receive a message for 100ms
+    caf::after(100ms) >>
+      [=] {
```
Neverlord, line 46:
```suggestion
      [self] {
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+        std::reverse(self->state.letter.begin(), self->state.letter.end());
+        self->send(self->state.dest, self->state.letter);
+        cout << "bye" << endl;
+        self->quit();
+      }};
```
Neverlord, line 51:
```suggestion
      },};
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+  using namespace std::chrono_literals;
+  return {
+    [=](char c) mutable {
+      self->state.dest = caf::actor_cast<caf::actor>(self->current_sender());
+      self->state.letter.push_back(c);
+    },
+    // trigger if we dont receive a message for 100ms
+    caf::after(100ms) >>
+      [=] {
+        std::reverse(self->state.letter.begin(), self->state.letter.end());
```
Neverlord, line 47: Why reverse the output?

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+behavior noisy_buddy(event_based_actor* self, actor buddy) {
+  using namespace std::chrono_literals;
+  std::random_device rd;
+  std::minstd_rand gen{rd()};
+  const auto count = std::uniform_int_distribution<>(20, 100)(gen);
+  std::uniform_int_distribution<> dis(33, 126);
+  for (auto i = 0; i < count; i++) {
+    self->send(buddy, static_cast<char>(dis(gen)));
+  }
+  return {[self](std::string letter) {
```
Neverlord, line 24: This will make unnecessary copies of the string.

```suggestion
  return {[self](const std::string& letter) {
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,59 @@
...
+  caf::actor dest;
+};
+
+// Aggregates incoming characters and stores the sender, replies with the
+// reversed string when inactive for 100ms
+behavior aggregator(stateful_actor<aggregator_state>* self) {
+  using namespace std::chrono_literals;
+  return {
+    [=](char c) mutable {
+      self->state.dest = caf::actor_cast<caf::actor>(self->current_sender());
```
Neverlord, line 41: This comes with many asterisks. For an example that introduces `after`, I think staying away from advanced tools like `actor_cast` would be a better choice.

In general, I think the setup with `aggregator` sending something back doesn't really add much and obfuscates what we actually want to show.

What would you think of streamlining this as follows:
- Rename aggregator to collector and simply let it print after:
  - The buffer reached some maximum (say, 40 characters).
  - A timeout occurs (and the buffer is not empty).
- Rename noisy_buddy to generator and have it simply terminate after it's done emitting messages (by returning `void`).

I think that would make the control flow and purpose of each piece much more clear. Also would allow spawning multiple generators and give users an easy path forward to extend/play with this example.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+        cout << "Received message length: " << self->state.size() << endl
+             << "Message content: " << self->state << endl;
+        self->quit();
+      }
+    },
+    // trigger if we dont receive a message for 100ms
+    caf::after(100ms) >>
+      [self]() {
+        cout << "Timeout reached!" << endl;
+        if (!self->state.empty()) {
```
Neverlord, line 43: I think this should block should also call `self->state.clear()`.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+// Collects the incoming characters until either the awaited_size of characters
+// is received, or no new characters arrive for 100ms
+behavior collector(stateful_actor<std::string>* self, size_t awaited_size) {
+  using namespace std::chrono_literals;
+  self->state.reserve(awaited_size);
+  return {
+    [=](char c) {
```
Neverlord, line 31: We don't do this consistently yet, but there are reasons to avoid `[=]`:
- It's greedy.
- C++20 deprecated the implicit capture of `this` via `[=]`. So there's already one case where an explicit capture is required by the language. Just listing all is more consistent.

```suggestion
    [self](char c) {
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <iostream>
+#include <random>
+
+using std::cout;
+using std::endl;
+
+using namespace caf;
+
+// Sends a random number of printable characters to buddy and exits
+void generator(event_based_actor* self, actor buddy) {
```
Neverlord, line 15: A `generator` having a buddy gives little insight into what this relation actually is. Did you consider naming it something like "sink" to make this relation obvious on first read?

```
// Sends a random number of printable characters to `sink` and exits.
void generator(event_based_actor* self, actor sink) {
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
+// This example shows how to use caf::after
+
+#include "caf/all.hpp"
```
Neverlord, line 3: The `all.hpp` header was added a long time ago when CAF was *a lot* smaller than it is today. I would no longer consider it best practice to use it. I already thought of deprecating it, but all of this becomes a non-issue if we finally do the switch to C++ modules.

Long story short: please include `caf_main`, `event_based_actor`, `actor_system`, etc.

examples/message_passing/after.cpp
riemass: I've modified the example, since it was more of a rewrite I'll resolve the previous conversations, but I did integrate the feedback where applicable.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+  const auto count = std::uniform_int_distribution<>{20, 100}(gen);
+  std::uniform_int_distribution<> dis{33, 126};
+  for (auto i = 0; i < count; i++) {
+    self->send(buddy, static_cast<char>(dis(gen)));
+  }
+}
+
+// Collects the incoming characters until either the awaited_size of characters
+// is received, or no new characters arrive for 100ms
+behavior collector(stateful_actor<std::string>* self, size_t awaited_size) {
```
Neverlord, line 27: Not a fan of this `awaited_size`. I think it gets into the way of toying with this example by adding more generators for example.

My suggestion to print after a certain threshold was more about making sure that the buffer isn't growing indefinitely. So it would be more of a "flush threshold", but for an example I think we don't need this as a spawn parameter.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+  using namespace std::chrono_literals;
+  self->state.reserve(awaited_size);
+  return {
+    [=](char c) {
+      self->state.push_back(c);
+      if (self->state.size() == awaited_size) {
+        cout << "Received message length: " << self->state.size() << endl
+             << "Message content: " << self->state << endl;
+        self->quit();
```
Neverlord, line 36: I wouldn't terminate here.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+      }
+    },
+    // trigger if we dont receive a message for 100ms
```
Neverlord, line 39: That's just the definition of `after`.

```suggestion
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,57 @@
...
+    caf::after(100ms) >>
```
Neverlord, line 40: Since this kills the actor now, I would bump the timeout to something like 500ms.

```suggestion
    after(500ms) >>
```

Neverlord (changes requested):

examples/message_passing/after.cpp
riemass: Makes more sense.

examples/message_passing/after.cpp
riemass: Fair enough.

examples/message_passing/after.cpp
riemass: I needed `self` and the `awaited_size`, which means everything. Since I've decided to drop the latter self sounds great. Also didn't know about deprecating the capture of `this`.

examples/message_passing/after.cpp
riemass: Since the next step is quit, the clear would be redundant. If I would remove the quit then the example wouldn't terminate but call the timeouts indefinitely.

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <iostream>
+#include <random>
+#include <string>
+
+using std::cout;
+using std::endl;
+
+using namespace caf;
+
+// Sends a random number of printable characters to buddy and exits
```
Neverlord, line 19:
```suggestion
// Sends a random number of printable characters to `sink` and then quits.
```

examples/message_passing/after.cpp
```diff
@@ -0,0 +1,61 @@
+// This example shows how to use caf::after
```
Neverlord, line 1:
```suggestion
// This example shows how to use caf::after.
```

Neverlord (changes requested):

PR #1518, opened 2023-08-06
Implement new fixture for deterministic testing

Neverlord:
@riemass with the new mailbox interface, we can now implement a new testing fixture for deterministic testing. The new setup uses a single, global queue where all actors will push their messages to. This makes it super easy to re-implement `expect` and `allow` without macros. I have also implemented a new `prepone` function that allows us to process messages out-of-order: something we couldn't do before. 🙂

I've also found a bug in `intrusive_ptr` (hence the `bugfix` label). If you do `lhs == rhs` with one side being an `intrusive_ptr` and the other side being a raw pointer, CAF tried to construct an `intrusive_ptr` for the right-hand side. I happened to have two types where that failed, so I got a compiler error. Technically unrelated to the rest of the PR, but I needed that fix in here to get it to compile.

@shariarriday I've needed `fail` for this, so I've added it. However, it currently throws `std::logic_error`. If you get to it, please update the implementation to use `requirement_failed`, as discussed. I've also improved `should_fail` and added a new `should_fail_with_exception` (since I needed it).

### DIFF
--- a/libcaf_test/caf/test/fixture/deterministic.test.cpp
+++ b/libcaf_test/caf/test/fixture/deterministic.test.cpp
@@ -23,6 +23,15 @@ struct my_int {
   int value;
 };
 
+template <class Inspector>
+auto inspect(Inspector& f, my_int& x) {
+  return f.object(x).fields(f.field("value", x.value));
+}
+
+bool operator==(my_int lhs, my_int rhs) noexcept {
+  return lhs.value == rhs.value;
+}
+
 bool operator==(my_int lhs, int rhs) noexcept {
   return lhs.value == rhs;
 }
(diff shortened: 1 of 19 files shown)


libcaf_test/caf/test/runnable.hpp
```diff
@@ -173,20 +215,52 @@ class CAF_TEST_EXPORT runnable {
...
+      } catch (...) {
+        caught = true;
+      }
+    } else {
+      try {
+        expr();
+      } catch (const Exception&) {
+        caught = true;
+      } catch (...) {
+        // TODO: print error message
```
riemass, line 241: Can you resolve this TODO?

Neverlord: No, the function doesn't exist yet. Needs the printing functions from https://github.com/actor-framework/actor-framework/issues/1479.

libcaf_test/caf/test/fixture/deterministic.test.cpp
```diff
@@ -0,0 +1,346 @@
...
+    check_eq(*count, 6);
+    should_fail_with_exception([this, worker] { //
+      expect<int32_t>().with(3).to(worker);
+    });
+  }
+  SECTION("expect() matches the types of the next message") {
+    anon_send(worker, "4");
+    should_fail_with_exception([this, worker] { //
+      expect<std::string>().to(worker);
+    });
```
riemass, line 106: Just to make sure I got this: `should_fail_with_exception` is for the purpose of testing the testing framework. In regular unit tests it should not be use and expect behaves like a `require` - it either succeeds or aborts the test.

riemass (approved):
Had a few small questions just to get things straight. Kudos for the new way of deterministic testing and the message expectations. 

Might be a code smell but first time I went through the PR, `expect`, `allow`, `with`, `evaluator` sounded too generic and nothing beside code comments said that they provided message matching and conditional dispatch. I wanted to suggest something like `expect_message<int32_t>().with_matching(3)`, but then I got used to the syntax and I actually started preferring the shorter test expressions. You might give it another judgement. 

Also, I find the test actor clock interesting. Might give it another test case showcasing stuff like `delayed_send`, or how to manually trigger/avoid a `caf::after` timeout. If the behavior I'm suggesting is indeed possible, we might move `test_actor_clock` from `caf_core` to `caf_test`.

There is also the old `test_coordinator`. I remember you stating to move it to the test module as well, but it might get the axe soon enough.

libcaf_test/caf/test/fixture/deterministic.test.cpp
```diff
@@ -0,0 +1,346 @@
...
+    check_eq(mail_count(worker), 2u);
+    check_eq(*count, 1);
+    expect<int32_t>().to(worker);
+    check_eq(mail_count(worker), 1u);
+    check_eq(*count, 3);
+    expect<int32_t>().to(worker);
+    check_eq(mail_count(worker), 0u);
+    check_eq(*count, 6);
+    should_fail_with_exception([this, worker] { //
+      expect<int32_t>().with(3).to(worker);
```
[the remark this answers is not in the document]
Neverlord: I read this as: "there is no doxgen comment on `with`". I'll add one. :)

Neverlord:
> Kudos for the new way of deterministic testing and the message expectations.

I should've been a bit more clear. This is a re-implementation of what we already have in the legacy tests via `dsl.hpp`. The syntax slightly changed since we no longer rely on macros:

```cpp
// legacy API
expect((int, int), from(foo).to(bar));
// new API
expect<int, int>().from(foo).to(bar);
```

> I actually started preferring the shorter test expressions

I'm all for shorter names. 🙂

This is an opt-in feature. You need to explicitly pull in the fixture. So I think shorter names are better. The context from the setup should be enough to know that `expect` is about messages.

> Also, I find the actor clock interesting. You might give it another test case showcasing stuff like delayed_send or how to manually trigger/avoid a caf::after timeout. If the behavior I'm suggesting is indeed possible, we might move test_actor_clock from libcaf_core to libcaf_test.

Absolutely possible. There are a couple functions for that in the legacy API that we should bring over. Since this is already a large PR, I'd prefer to do that as a followup, though. Full ACK on the `test_actor_clock`. If we remove the test scheduler from the core, we should also get rid of the test actor clock and move it to the test module.

PR #1519, opened 2023-08-07
Add ignore elements flow operator

shariarriday:

Implements ignore elements.

### DIFF
--- a/libcaf_core/caf/flow/step/ignore_elements.hpp
+++ b/libcaf_core/caf/flow/step/ignore_elements.hpp
@@ -1,0 +1,46 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/detail/ring_buffer.hpp"
+#include "caf/fwd.hpp"
+
+#include <cstddef>
+
+namespace caf::flow::step {
+
+template <class T>
+class ignore_elements {
+public:
+  using input_type = T;
+
+  using output_type = T;
+
+  explicit ignore_elements() {
+    // nop
+  }
+
+  ignore_elements(ignore_elements&&) = default;
+  ignore_elements(const ignore_elements&) = default;
+  ignore_elements& operator=(ignore_elements&&) = default;
+  ignore_elements& operator=(const ignore_elements&) = default;
+
+  template <class Next, class... Steps>
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
+    return true;
+  }
+
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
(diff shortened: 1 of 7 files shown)


libcaf_core/caf/flow/step/ignore_elements.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  explicit ignore_elements() {
+    // nop
+  }
+
+  ignore_elements(ignore_elements&&) = default;
+  ignore_elements(const ignore_elements&) = default;
+  ignore_elements& operator=(ignore_elements&&) = default;
+  ignore_elements& operator=(const ignore_elements&) = default;
+
```
Neverlord, line 29: This has no state: rule of 0.

```suggestion
```

libcaf_core/caf/flow/step/ignore_elements.hpp
```diff
@@ -0,0 +1,46 @@
...
+  template <class Next, class... Steps>
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
```
Neverlord, line 31: Don't you get warnings for unused parameters on MSVC here? If not, we need to figure out the appropriate flags.

```suggestion
  bool on_next(const input_type&, Next&, Steps&...) {
```

libcaf_core/caf/flow/step/ignore_elements.test.cpp
```diff
@@ -0,0 +1,102 @@
...
+      .ignore_elements()
+      .take(5)
+      .for_each([&](const int& result_observable) {
+        result.emplace_back(result_observable);
+      });
+  }
+  ctx->run();
+  check_eq(result.size(), 0u);
+}
+
```
Neverlord, line 51: I think you made these analogous to the `take_last` tests. Good idea in general, but for this operator, there is no edge case to check here. Hence, this is redundant and adds no extra coverage. I'd say we can just drop it.

```suggestion
```

libcaf_core/caf/flow/step/ignore_elements.hpp
```diff
@@ -0,0 +1,46 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/detail/ring_buffer.hpp"
+#include "caf/fwd.hpp"
+
+#include <cstddef>
```
Neverlord, line 10: Header hygiene: we shouldn't include unnecessary headers.

```suggestion
#include "caf/fwd.hpp"
```

Neverlord (changes requested):

libcaf_core/caf/flow/step/ignore_elements.hpp
shariarriday: I don't see any warning for unused parameters. I will look into appropriate flags for MSVC for this.

PR #1525, opened 2023-08-11
Fail early on invalid UTF-8 WebSocket text frames

riemass:
Relates #1394 
With this the last two tests pass (they where passing non strictly before).

Take into account that with this we get multiple calls from the transport layer to the web_socket layer, even when we can know that the whole frame didn't arrive. This means that for a text message, on every call we:
- decode the header, check for malformed header 
- unmask the data, make the data owned (we need to either own the data or to mask it back up), 
- combine it with existing unvalidated bytes in case of an continuation frame (in which case we do need to own our data), 
- validate the utf-8.
Not my proudest peace of work. The mentioned overhead could be avoided at the cost of introducing additional state and complicating to the framing layer.

@Neverlord It's your call. I might even suggest we leave the 2 tests as non-strictly passing and fail on the end of the frame. I can't think of any security risk if we wait a little longer to fail. However I do want to have a big green dot saying fully autobahn compliant :smile:

### DIFF
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -34,7 +34,7 @@ struct CAF_NET_EXPORT rfc6455 {
     uint32_t mask_key = 0;
     uint64_t payload_len = 0;
 
-    // utility funcitons
+    // utility functions.
     explicit operator bool() const noexcept {
       return opcode != opcode_type::nil_code;
     }
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -42,9 +42,9 @@ struct CAF_NET_EXPORT rfc6455 {
 
   // -- utility functions ------------------------------------------------------
 
-  static void mask_data(uint32_t key, span<char> data, size_t skip = 0);
+  static void mask_data(uint32_t key, span<char> data, size_t offset = 0);
 
-  static void mask_data(uint32_t key, byte_span data, size_t skip = 0);
+  static void mask_data(uint32_t key, byte_span data, size_t offset = 0);
 
   static void assemble_frame(uint32_t mask_key, span<const char> data,
                              byte_buffer& out);
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -17,26 +17,34 @@ namespace caf::detail {
 struct CAF_NET_EXPORT rfc6455 {
   // -- member types -----------------------------------------------------------
 
+  enum class opcode_type : uint8_t {
+    continuation_frame = 0x00,
+    text_frame = 0x01,
+    binary_frame = 0x02,
+    connection_close = 0x08,
+    ping = 0x09,
+    pong = 0x0A,
+    /// Invalid opcode to mean "no opcode received yet".
+    nil_code = 0xFF,
+  };
+
   struct header {
-    bool fin;
-    uint8_t opcode;
-    uint32_t mask_key;
-    uint64_t payload_len;
+    bool fin = false;
+    opcode_type opcode = opcode_type::nil_code;
+    uint32_t mask_key = 0;
+    uint64_t payload_len = 0;
+
+    // utility funcitons
+    explicit operator bool() const noexcept {
+      return opcode != opcode_type::nil_code;
+    }
+
+    bool operator!() const noexcept {
+      return !static_cast<bool>(*this);
+    }
   };
 
   // -- constants --------------------------------------------------------------
-
-  static constexpr uint8_t continuation_frame = 0x00;
-
-  static constexpr uint8_t text_frame = 0x01;
-
-  static constexpr uint8_t binary_frame = 0x02;
-
-  static constexpr uint8_t connection_close = 0x08;
-
-  static constexpr uint8_t ping = 0x09;
-
-  static constexpr uint8_t pong = 0x0A;
 
   static constexpr uint8_t fin_flag = 0x80;
 
@@ -52,14 +60,14 @@ struct CAF_NET_EXPORT rfc6455 {
   static void assemble_frame(uint32_t mask_key, const_byte_span data,
                              byte_buffer& out);
 
-  static void assemble_frame(uint8_t opcode, uint32_t mask_key,
+  static void assemble_frame(opcode_type opcode, uint32_t mask_key,
                              const_byte_span data, byte_buffer& out,
                              uint8_t flags = fin_flag);
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
 
-  static constexpr bool is_control_frame(uint8_t opcode) noexcept {
-    return opcode > binary_frame;
+  static constexpr bool is_control_frame(opcode_type opcode) noexcept {
+    return opcode > opcode_type::binary_frame;
   }
 };
 
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -17,7 +17,7 @@ namespace caf::detail {
 struct CAF_NET_EXPORT rfc6455 {
   // -- member types -----------------------------------------------------------
 
-  enum class opcode_type : uint8_t {
+  enum opcode_type : uint8_t {
     continuation_frame = 0x00,
     text_frame = 0x01,
     binary_frame = 0x02,
@@ -25,17 +25,17 @@ struct CAF_NET_EXPORT rfc6455 {
     ping_frame = 0x09,
     pong_frame = 0x0A,
     /// Invalid opcode to mean "no opcode received yet".
-    nil_code = 0xFF,
+    invalid_frame = 0xFF,
   };
 
   struct header {
     bool fin = false;
-    opcode_type opcode = opcode_type::nil_code;
+    uint8_t opcode = invalid_frame;
     uint32_t mask_key = 0;
     uint64_t payload_len = 0;
 
     constexpr bool valid() const noexcept {
-      return opcode != opcode_type::nil_code;
+      return opcode != invalid_frame;
     }
   };
 
@@ -55,14 +55,14 @@ struct CAF_NET_EXPORT rfc6455 {
   static void assemble_frame(uint32_t mask_key, const_byte_span data,
                              byte_buffer& out);
 
-  static void assemble_frame(opcode_type opcode, uint32_t mask_key,
+  static void assemble_frame(uint8_t opcode, uint32_t mask_key,
                              const_byte_span data, byte_buffer& out,
                              uint8_t flags = fin_flag);
 
   static ptrdiff_t decode_header(const_byte_span data, header& hdr);
 
-  static constexpr bool is_control_frame(opcode_type opcode) noexcept {
-    return opcode > opcode_type::binary_frame;
+  static constexpr bool is_control_frame(uint8_t opcode) noexcept {
+    return opcode > binary_frame;
   }
 };
 
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -42,9 +42,9 @@ struct CAF_NET_EXPORT rfc6455 {
 
   // -- utility functions ------------------------------------------------------
 
-  static void mask_data(uint32_t key, span<char> data);
+  static void mask_data(uint32_t key, span<char> data, size_t skip = 0);
 
-  static void mask_data(uint32_t key, byte_span data);
+  static void mask_data(uint32_t key, byte_span data, size_t skip = 0);
 
   static void assemble_frame(uint32_t mask_key, span<const char> data,
                              byte_buffer& out);
--- a/libcaf_net/caf/detail/rfc6455.hpp
+++ b/libcaf_net/caf/detail/rfc6455.hpp
@@ -17,17 +17,6 @@ namespace caf::detail {
 struct CAF_NET_EXPORT rfc6455 {
   // -- member types -----------------------------------------------------------
 
-  enum opcode_type : uint8_t {
-    continuation_frame = 0x00,
-    text_frame = 0x01,
-    binary_frame = 0x02,
-    connection_close_frame = 0x08,
-    ping_frame = 0x09,
-    pong_frame = 0x0A,
-    /// Invalid opcode to mean "no opcode received yet".
-    invalid_frame = 0xFF,
-  };
-
   struct header {
     bool fin = false;
     uint8_t opcode = invalid_frame;
@@ -40,6 +29,21 @@ struct CAF_NET_EXPORT rfc6455 {
   };
 
   // -- constants --------------------------------------------------------------
+
+  static constexpr uint8_t continuation_frame = 0x00;
+
+  static constexpr uint8_t text_frame = 0x01;
+
+  static constexpr uint8_t binary_frame = 0x02;
+
+  static constexpr uint8_t connection_close_frame = 0x08;
+
+  static constexpr uint8_t ping_frame = 0x09;
+
+  static constexpr uint8_t pong_frame = 0x0A;
+
+  /// Invalid opcode to mean "no opcode received yet".
+  static constexpr uint8_t invalid_frame = 0xFF;
 
   static constexpr uint8_t fin_flag = 0x80;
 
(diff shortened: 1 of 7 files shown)


libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -11,6 +11,23 @@
...
+/// while scanning in order to avoid validating the same bytes again.
+bool payload_valid(const_byte_span payload, size_t& offset) noexcept {
+  // validate from the index where we left off last time
+  auto [index, incomplete] = detail::rfc3629::validate(payload.subspan(offset));
+  offset += index;
+  if (offset == payload.size())
+    return true;
+  // incomplete will be true if the last code point is missing continuation
+  // bytes but might be valid
+  return incomplete;
```
Neverlord, line 26:
```suggestion
  // Incomplete will be true if the last code point is missing continuation
  // bytes but might be valid.
  return offset == payload.size() || incomplete;
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -11,6 +11,23 @@
 
 namespace caf::net::web_socket {
 
+namespace {
+
+/// Checks whether the current input is valid UTF-8. Stores the last position
+/// while scanning in order to avoid validating the same bytes again.
+bool payload_valid(const_byte_span payload, size_t& offset) noexcept {
+  // validate from the index where we left off last time
```
Neverlord, line 19:
```suggestion
  // Continue from the index where we left off last time.
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,7 +80,7 @@ void framing::abort(const error& reason) {
   up_->abort(reason);
 }
 
-ptrdiff_t framing::consume(byte_span buffer, byte_span) {
+ptrdiff_t framing::consume(byte_span buffer, byte_span delta) {
   // Make sure we're overriding any 'exactly' setting.
   down_->configure_read(receive_policy::up_to(2048));
   // Parse header.
```
Neverlord, line 86: We've switched the model from "get all bytes at once" to "incrementally parse". However, we still call `decode_header` unconditionally each time. And this function isn't trivial, so it'll add quite a bit of overhead if the client sends small chunks. We can tackle this as a followup, but I think we should address it.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -95,26 +112,58 @@ ptrdiff_t framing::consume(byte_span buffer, byte_span) {
...
     }
   }
-  // Wait for more data if necessary.
+  // Calculate at what point of the received buffer the delta payload begins.
+  auto offset = static_cast<ptrdiff_t>((buffer.size() - delta.size()))
+                - hdr_bytes;
+  // Offset < zero  - the delta buffer contains header bytes.
+  // Delta is empty - we didn't process the whole input last time we got called.
+  if (delta.empty() || offset < 0)
+    offset = 0;
```
Neverlord, line 129: `delta` should never be empty. This would mean that the transport is calling you twice with the same data. Is this check only due to #1537?

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -95,26 +112,58 @@ ptrdiff_t framing::consume(byte_span buffer, byte_span) {
...
+  // We read frame_size at most. This can leave unprocessed input.
+  auto payload = buffer.subspan(hdr_bytes, std::min(buffer.size() - hdr_bytes,
+                                                    hdr.payload_len));
```
Neverlord, line 132: The logic here is quite complex. Calculating the offset, treating special cases and so on. I think this really comes down to leaving the header in the buffer in the first place. So maybe it's actually best to address this right away after all.

So here are some thought that come to mind how we could simplify this logic.

If we keep the parsed header as state in `framing`, we would only need to remember whether we are currently parsing a header or a payload. If we parse a header, we just wait until the transport has handed us enough bytes. Then we return `hdr_bytes`. This means the transport will drop the header bytes for us and we no longer need to "work around it". The next time the transport calls `consume`, we know that we are parsing a payload and can simply call `mask_data` on the delta.

This would also allow us to split the consume function into `consume_header` and `consume_payload` and then `consume` would only dispatch based on the current state.

Does that make sense?

Neverlord (changes requested):

libcaf_net/caf/net/web_socket/framing.cpp
riemass: #1537 is one cause of this, as you stated, but the other valid scenario is if multiple frames arrive at once from the transport layer, and we process one frame per call. We have a unit test for this [here](https://github.com/actor-framework/actor-framework/blob/5933a5f7612ab93d45ede32f91c2e1a1f76f360c/libcaf_net/tests/legacy/net/web_socket/framing.cpp#L234)

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Will do. Discussion in the below comment.

libcaf_net/caf/net/web_socket/framing.cpp
riemass: This is fixed now.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+    payload_buf_.insert(payload_buf_.end(), payload.begin(), payload.end());
+  }
+  auto ret = static_cast<ptrdiff_t>(hdr_.payload_len);
+  hdr_.opcode = detail::rfc6455::opcode_type::nil_code;
+  return ret;
+}
+
+ptrdiff_t framing::consume(byte_span buffer, byte_span delta) {
+  // Make sure we're overriding any 'exactly' setting.
+  down_->configure_read(receive_policy::up_to(2048));
```
Neverlord, line 217: Why is this still unconditionally here?

I would expect something like this:
- call `down_->configure_read(receive_policy::up_to(2048));` in `init`
- call `down_->configure_read(receive_policy::up_to(payload_size));` after reading the header in `consume_header`
- call `up_to(2048)` again in `consume_payload` when preparing for the next header

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+}
+
+ptrdiff_t framing::consume_payload(byte_span buffer, byte_span delta) {
+  // Calculate at what point of the received buffer the delta payload begins.
+  auto offset = static_cast<ptrdiff_t>(buffer.size() - delta.size());
+  // We read frame_size at most. This can leave unprocessed input.
+  auto payload = buffer.first(std::min(buffer.size(), hdr_.payload_len));
+  // Unmask the arrived data.
+  if (hdr_.mask_key != 0)
+    detail::rfc6455::mask_data(hdr_.mask_key, payload, offset);
```
Neverlord, line 146: Why can we receive more data then the payload len here? I would expect to see simply `detail::rfc6455::mask_data(hdr_.mask_key, buffer, offset);` and have `consume_header` prepare the transport with `up_to(hdr.payload_len)`. Is there a bug preventing this?

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+      payload_buf_.insert(payload_buf_.end(), payload.begin() + offset,
+                          payload.end());
+      if (!payload_valid(payload_buf_, validation_offset_)) {
+        abort_and_shutdown(sec::malformed_message, "invalid UTF-8 sequence");
+        return -1;
+      }
+    }
+    // Wait for more data if necessary.
+    if (buffer.size() < hdr_.payload_len) {
+      down_->configure_read(receive_policy::up_to(hdr_.payload_len));
```
Neverlord, line 166: Again, I would expect that the `consume_header` function does this and we simply return `0` here until all data arrived.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+      return 0;
+    }
+  } else {
+    // Wait for more data if necessary.
+    if (buffer.size() < hdr_.payload_len) {
+      down_->configure_read(receive_policy::exactly(hdr_.payload_len));
```
Neverlord, line 172: Once again: should be done by `consume_header`.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
     CAF_LOG_DEBUG("expected a continuation frame");
     abort_and_shutdown(sec::protocol_error, "expected a continuation frame");
     return -1;
   }
-  payload_buf_.insert(payload_buf_.end(), payload.begin(), payload.end());
-  if (opcode_ == detail::rfc6455::text_frame && !payload_valid()) {
-    abort_and_shutdown(sec::malformed_message, "invalid UTF-8 sequence");
-    return -1;
+  if (opcode_ != detail::rfc6455::opcode_type::text_frame) {
+    payload_buf_.insert(payload_buf_.end(), payload.begin(), payload.end());
```
Neverlord, line 209: I'd prefer not touching `payload_buf_` at all here to have a clear separation of concerns.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+  }
+  auto ret = static_cast<ptrdiff_t>(hdr_.payload_len);
+  hdr_.opcode = detail::rfc6455::opcode_type::nil_code;
+  return ret;
+}
+
+ptrdiff_t framing::consume(byte_span buffer, byte_span delta) {
+  // Make sure we're overriding any 'exactly' setting.
+  down_->configure_read(receive_policy::up_to(2048));
+  if (!hdr_) {
```
Neverlord, line 218: As per previous suggestion:

```suggestion
  if (!hdr_.valid()) {
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+    auto hdr_bytes = consume_header(buffer, delta);
+    if (hdr_bytes <= 0) {
+      hdr_.opcode = detail::rfc6455::opcode_type::nil_code;
+      return hdr_bytes;
+    }
```
Neverlord, line 223: If `hdr_bytes` is `0`, then `hdr_.opcode` must be unchanged, right? If `hdr_bytes` is `-1`, then we abort anyways. So why setting `hdr_.opcode` here?

```suggestion
    if (hdr_bytes <= 0) {
      hdr_.opcode = detail::rfc6455::opcode_type::nil_code;
      return hdr_bytes;
    }
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+    if (hdr_.payload_len == 0
+        && consume_payload(buffer.first(0), delta.first(0)) < 0)
```
Neverlord, line 225: Can't we just call `handle` here instead?

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -63,119 +78,167 @@ void framing::abort(const error& reason) {
...
+      return -1;
+    return hdr_bytes;
+  } else {
+    return consume_payload(buffer, delta);
   }
```
Neverlord, line 230:
```suggestion
  }
  return consume_payload(buffer, delta);
```

Neverlord (changes requested):

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -17,49 +17,57 @@ namespace caf::detail {
...
 
+  enum class opcode_type : uint8_t {
+    continuation_frame = 0x00,
+    text_frame = 0x01,
+    binary_frame = 0x02,
+    connection_close = 0x08,
+    ping = 0x09,
+    pong = 0x0A,
+    /// Invalid opcode to mean "no opcode received yet".
+    nil_code = 0xFF,
```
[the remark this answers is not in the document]
riemass: The 4-bit range doesn't include an invalid opcode and one of the further revisions of the standard might define the `0x0F` value to have another meaning. That's way I prefer to go out of the 4-bit boundary for this use case.

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -17,49 +17,57 @@ namespace caf::detail {
 struct CAF_NET_EXPORT rfc6455 {
   // -- member types -----------------------------------------------------------
 
+  enum class opcode_type : uint8_t {
+    continuation_frame = 0x00,
+    text_frame = 0x01,
+    binary_frame = 0x02,
+    connection_close = 0x08,
+    ping = 0x09,
+    pong = 0x0A,
```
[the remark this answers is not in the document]
riemass: Good call.

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -17,49 +17,57 @@ namespace caf::detail {
...
+    uint64_t payload_len = 0;
+
+    // utility functions.
+    explicit operator bool() const noexcept {
+      return opcode != opcode_type::nil_code;
+    }
+
+    bool operator!() const noexcept {
+      return !static_cast<bool>(*this);
+    }
```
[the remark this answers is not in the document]
riemass: The first one felt more pragmatic, but yeah this one is more readable and I'll go with that.

libcaf_net/caf/net/web_socket/framing.cpp
riemass: Basically at the start of processing, and at every new call to consume, we set the receive policy to `ut_to(2048);`. But I agree, I can work around this now and no need for `consume_payload` to be concerned about this.

libcaf_net/caf/net/web_socket/framing.cpp
riemass: :+1:

libcaf_net/caf/net/web_socket/framing.cpp
riemass: I'm on it. Full ack.

libcaf_net/caf/net/web_socket/framing.cpp
riemass: :+1:

libcaf_net/caf/net/web_socket/framing.cpp
riemass: In `decode_header` we can decode and store the `opcode` field, and still return 0 if some other bytes are missing. I could modify the `decode_header` function to be atomic, and only change `hdr_` if it succeeded.

libcaf_net/caf/detail/rfc6455.hpp
```diff
@@ -17,49 +17,57 @@ namespace caf::detail {
...
+
   struct header {
-    bool fin;
-    uint8_t opcode;
-    uint32_t mask_key;
-    uint64_t payload_len;
+    bool fin = false;
+    opcode_type opcode = opcode_type::nil_code;
+    uint32_t mask_key = 0;
+    uint64_t payload_len = 0;
```
riemass, line 35: @Neverlord basically I only need the `opcode` initialized here. Does it make sense if I initialize all fields, or should I, for performance reasons, only init the `opcode`?

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -246,11 +159,141 @@ bool framing::end_text_message() {
...
+    payload_buf_.clear();
+    validation_offset_ = 0;
+    return result;
+  }
+  if (opcode_ == detail::rfc6455::invalid_frame)
+    opcode_ = hdr_.opcode;
+  // Clean up the state since we finished processing this frame
+  down_->configure_read(receive_policy::up_to(2048));
+  hdr_.opcode = detail::rfc6455::invalid_frame;
+  return static_cast<ptrdiff_t>(hdr_.payload_len);
```
Neverlord, line 289: This code is not DRY: the same lines are copied at the end of the function. Could be factored out into a lambda, e.g.:

```
auto receive_next_header = [this] {
  down_->configure_read(receive_policy::up_to(2048));
  hdr_.opcode = detail::rfc6455::invalid_frame;
  return static_cast<ptrdiff_t>(hdr_.payload_len);
};
```

And then these lines would just become `return receive_next_header();`.

Neverlord (changes requested):

Neverlord (approved):

PR #1528, opened 2023-08-12
Implement requirement_failed exception

shariarriday:

The fail function has been implemented in https://github.com/actor-framework/actor-framework/pull/1518 but the proper `requirement_failed` error was not raised. Proper error has been added here.

@Neverlord I think that the error can be added in a different file `libcaf_test/caf/test/errors.hpp`. I would like your opinion on this. Furthermore, I have not called fail() from the tests as I think the tests will be covered in the next PR when requirement_predicates are implemented.

### DIFF
--- a/libcaf_test/caf/test/requirement_error.cpp
+++ b/libcaf_test/caf/test/requirement_error.cpp
@@ -1,0 +1,31 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/test/requirement_error.hpp"
+
+#include "caf/detail/format.hpp"
+
+namespace caf::test {
+
+std::string requirement_error::message() const {
+  switch (code_) {
+    case code::failed:
+      return detail::format("requirement failed at {}:{}", loc_.file_name(),
+                            static_cast<int>(loc_.line()));
+  }
+}
+
+[[noreturn]] void
+requirement_error::raise_impl(requirement_error::code what,
+                              const detail::source_location& loc) {
+#ifdef CAF_ENABLE_EXCEPTIONS
+  throw requirement_error{what, loc};
+#else
+  auto msg = nesting_error{what, parent, child, loc}.message();
+  fprintf(stderr, "[FATAL] critical error: %s\n", msg.c_str());
+  abort();
+#endif
+}
+
+} // namespace caf::test
--- a/libcaf_test/caf/test/requirement_error.cpp
+++ b/libcaf_test/caf/test/requirement_error.cpp
@@ -10,7 +10,7 @@ namespace caf::test {
 
 std::string requirement_error::message() const {
   return detail::format("requirement failed at {}:{}", loc_.file_name(),
-                        static_cast<int>(loc_.line()));
+                        loc_.line());
 }
 
 [[noreturn]] void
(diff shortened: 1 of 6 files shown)


Neverlord (changes requested):
Please use `nesting_error` as a template and don't specialize `std::exception`. We don't want to use `CAF_RAISE_ERROR` but `requirement_failed::raise()`.

The exception only needs a source location for now. Also needs a new `catch` case in `runner.cpp` that only calls `default_reporter->end_test();`: we don't need to report this exception since we already print output to the user before throwing it.

libcaf_test/caf/test/requirement_error.hpp
```diff
@@ -0,0 +1,55 @@
...
+  [[noreturn]] static void raise_failed(const detail::source_location& loc
+                                        = detail::source_location::current()) {
+    raise_impl(code::failed, loc);
+  }
+
+private:
+  enum class code {
+    failed,
+  };
+
```
Neverlord, line 41: I don't think this has any practical purpose. Nor can I think of good use case for having multiple kinds of requirement failures.

```suggestion
```

libcaf_test/caf/test/requirement_error.hpp
```diff
@@ -0,0 +1,55 @@
...
+  constexpr requirement_error(code what,
+                              const detail::source_location& loc) noexcept
+    : code_(what), loc_(loc) {
+    // nop
+  }
```
Neverlord, line 46: As mentioned prior, I don't think `code` has any practical use:

```suggestion
  constexpr explicit requirement_error(const detail::source_location& loc) noexcept
    : loc_(loc) {
    // nop
  }
```

libcaf_test/caf/test/requirement_error.hpp
```diff
@@ -0,0 +1,55 @@
...
+  /// Returns a human-readable error message.
+  std::string message() const;
+
+  /// Returns the source location of the error.
+  constexpr const detail::source_location& location() const noexcept {
+    return loc_;
+  }
+
+  /// Throws a `requirement_error` to indicate that requirement check failed.
+  [[noreturn]] static void raise_failed(const detail::source_location& loc
```
Neverlord, line 32:
```suggestion
  [[noreturn]] static void raise(const detail::source_location& loc
```

Neverlord (changes requested):

libcaf_test/caf/test/requirement_error.hpp
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /// Throws a `requirement_error` to indicate that requirement check failed.
+  [[noreturn]] static void raise(const detail::source_location& loc
+                                 = detail::source_location::current()) {
+    raise_impl(loc);
+  }
+
+private:
+  constexpr requirement_error(const detail::source_location& loc) noexcept
```
Neverlord, line 38:
```suggestion
  constexpr explicit requirement_error(const detail::source_location& loc) noexcept
```

libcaf_test/caf/test/requirement_error.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "caf/test/requirement_error.hpp"
+
+#include "caf/detail/format.hpp"
+
+namespace caf::test {
+
+std::string requirement_error::message() const {
+  return detail::format("requirement failed at {}:{}", loc_.file_name(),
+                        static_cast<int>(loc_.line()));
```
Neverlord, line 13:
```suggestion
                        loc_.line());
```

Neverlord (changes requested):

Neverlord (approved):

PR #1530, opened 2023-08-15
Add require function for all predicates

shariarriday:

@Neverlord a new `info()` function was added to in `reporter.cpp` to show the information without failing the test, as `fail()` will fail the test anyways. An alternative will be a `fail()` function which can take appropriate parameters as input and pass them to the `reporter.fail()` function. Let me know your opinion on this.

### DIFF
--- a/libcaf_test/caf/test/test.test.cpp
+++ b/libcaf_test/caf/test/test.test.cpp
@@ -41,6 +41,8 @@ TEST("tests can contain different types of checks") {
   info("this test had {} checks", rep.test_stats().total());
 }
 
+#ifdef CAF_ENABLE_EXCEPTIONS
+
 TEST("tests fail when requirement errors occur") {
   auto& rep = caf::test::reporter::instance();
   SECTION("require_eq fails when lhs != rhs") {
@@ -81,6 +83,8 @@ TEST("tests fail when requirement errors occur") {
   }
   info("this test had {} checks", rep.test_stats().total());
 }
+
+#endif // CAF_ENABLE_EXCEPTIONS
 
 TEST("failed checks increment the failed counter") {
   check_eq(1, 2);
--- a/libcaf_test/caf/test/test.test.cpp
+++ b/libcaf_test/caf/test/test.test.cpp
@@ -4,6 +4,7 @@
 #include "caf/test/test.hpp"
 
 #include "caf/test/caf_test_main.hpp"
+#include "caf/test/requirement_error.hpp"
 
 using caf::test::block_type;
 
@@ -36,6 +37,47 @@ TEST("tests can contain different types of checks") {
     check_lt(1, 2);
     should_fail([this]() { check_lt(1, 1); });
     should_fail([this]() { check_lt(2, 1); });
+  }
+  info("this test had {} checks", rep.test_stats().total());
+}
+
+TEST("tests fail when requirement errors occur") {
+  auto& rep = caf::test::reporter::instance();
+  SECTION("require_eq fails when lhs != rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_eq(1, 2); });
+    require_eq(1, 1);
+  }
+  SECTION("require_ne fails when lhs == rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_ne(1, 1); });
+    require_ne(1, 2);
+  }
+  SECTION("require_le fails when lhs > rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_le(2, 1); });
+    require_le(1, 2);
+    require_le(2, 2);
+  }
+  SECTION("require_lt fails when lhs >= rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_lt(2, 2); });
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_lt(2, 1); });
+    require_lt(1, 2);
+  }
+  SECTION("require_ge fails when lhs < rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_ge(1, 2); });
+    require_ge(2, 1);
+    require_ge(2, 2);
+  }
+  SECTION("require_gt fails when lhs <= rhs") {
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_gt(1, 1); });
+    should_fail_with_exception<caf::test::requirement_error>(
+      [this]() { require_gt(1, 2); });
+    require_gt(2, 1);
   }
   info("this test had {} checks", rep.test_stats().total());
 }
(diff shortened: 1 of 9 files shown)


Neverlord:
I don't think I can follow your though process on this. By calling `info`, we will print an actual error output with severity `info`. From a user perspective, I don't think this makes sense.

I'm also not sure whether I can follow your reasoning for the implementation. Let's pick `require_eq` as an example. Here's what `check_eq` does:

```c++
  /// Checks whether `lhs` and `rhs` are equal.
  template <class T0, class T1>
  bool check_eq(const T0& lhs, const T1& rhs,
                const detail::source_location& location
    if (lhs == rhs) {
      reporter::instance().pass(location);
      return true;
    }
    reporter::instance().fail(binary_predicate::eq, stringify(lhs),
                              stringify(rhs), location);
    return false;
  }
```

Basically: check if the values are equal. If not, print an error. If we think of `require_eq`, the only delta to `check_eq` is that it'll throw an exception instead of returning `false`. The current patch adds this:

```c++
  /// Evaluates whether `lhs` and `rhs` are equal and fails otherwise.
  template <class T0, class T1>
  void require_eq(const T0& lhs, const T1& rhs,
                  const detail::source_location& location
    if (lhs == rhs) {
      reporter::instance().pass(location);
      return;
    }
    reporter::instance().info(binary_predicate::eq, stringify(lhs),
                              stringify(rhs), location);
    fail({"requirement failed", location});
  }
```

So it's basically a copy of `check_eq` except that we demote the error message that's most useful to `info` and then print a generic `requirement failed` text with `error` severity by calling `fail`.

Naïvely, I would have implemented `require_eq` like this:

```c++
  /// Evaluates whether `lhs` and `rhs` are equal and fails otherwise.
  template <class T0, class T1>
  void require_eq(const T0& lhs, const T1& rhs,
                  const detail::source_location& location
    if (!check_eq(lhs, rhs, location))
      requirement_error::raise(location);
  }
```

Am I missing something?

Neverlord:
> Because, If we use the following implementation the user is shown a silent error which just stops running tests without giving additional information. I tried to address this concern in the current implementation.

How do you mean silent? It's printing an error at error severity and then stops the test. Do you mean "silent" in the sense that there's no text saying that the test aborted due to a requirement failing?

> On this note, I think that we can add some verbose statement in the `fail` statement and update the current log structure as well which is planned in #1479 .

As it is, this new `info` adds a lot of redundancy that imposes a maintenance burden. Currently, `fail` has a very clear and specific use case, I'd like to keep it that way.

It seems like we agree that calling the `check_*` functions and then throwing gives us pretty much the behavior we want. If I read your intentions correctly, your main concern is that there's no text being printed to the user that explicitly states that we abort due to a requirement failure. I think that's a good point. However, I'd argue the proper place for adding this would be in `runner.cpp` where we catch the exception. Then we do this is one central place without adding redundancy to the code base.

Also, please rename `requirement_error` to `requirement_failed` while we're at it. IMHO, this reflects the purpose of this exception type better. Sorry for not catching this on the earlier PR. 🙂

shariarriday:
I think I understand your concerns and I will update the PR accordingly.

> However, I'd argue the proper place for adding this would be in runner.cpp where we catch the exception. Then we do this is one central place without adding redundancy to the code base.

Regarding this, I had initially planned to add a statement in the catch block. But, this approach will print the log even when the `fail()` function is called. I discarded as I thought that it is a problem, but do we need to consider it as an issue now?

Neverlord:
> I think I understand your concerns and I will update the PR accordingly.
> 
> > However, I'd argue the proper place for adding this would be in runner.cpp where we catch the exception. Then we do this is one central place without adding redundancy to the code base.
> 
> Regarding this, I had initially planned to add a statement in the catch block. But, this approach will print the log even when the `fail()` function is called. I discarded the approach previously as I thought that it is a problem, but do we need to consider it as an issue now?

You probably mean because `fail` with increment the `failed` counter (https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/reporter.cpp#L329)? That's a good point. Why not add the `print_error` function now that we meant to add anyways? Then we can simply use that to avoid touching the `failed` count. That's basically this function: https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/reporter.cpp#L376, but printing with error verbosity.

shariarriday:
> You probably mean because fail with increment the failed counter (https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/reporter.cpp#L329)? That's a good point. Why not add the print_error function now that we meant to add anyways? Then we can simply use that to avoid touching the failed count. That's basically this function: https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/reporter.cpp#L376, but printing with error verbosity.

Thanks, I think that this will be the best approach to resolve this issue. 

Let me just reiterate our final discussion to ensure that we are on the same page. The requirement checks will be implemented in the following way which will call the `fail()` functions.

```
/// Evaluates whether `lhs` and `rhs` are equal and fails otherwise.
  template <class T0, class T1>
  void require_eq(const T0& lhs, const T1& rhs,
                  const detail::source_location& location
    if (!check_eq(lhs, rhs, location))
      fail({"requirement check failed", location});
  }
```

The fail function will be updated to use print_error which will not increase the count of failed tests and print the log with correct verbosity.

`reporter::instance().print_error(msg, fwl.location);`

Neverlord:
That's not quite what I meant. Since `fail` touches the `failed` counter, I'd not touch it. I'd implement `require_eq` like this:

```
   template <class T0, class T1>
   void require_eq(const T0& lhs, const T1& rhs,
                   const detail::source_location& location
     if (!check_eq(lhs, rhs, location))
       requirement_failed::raise(location);
   }
```

Then, we print in `runner.cpp` after catching the exception, e.g.:

```
default_reporter->print_error(ex.message(), ex.location());
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -40,6 +41,51 @@ TEST("tests can contain different types of checks") {
   info("this test had {} checks", rep.test_stats().total());
 }
 
+#ifdef CAF_ENABLE_EXCEPTIONS
+
+TEST("tests fail when requirement errors occur") {
+  auto& rep = caf::test::reporter::instance();
```
Neverlord, line 47: We are using `caf::test::requirement_failed` quite a lot, adding a `using` here allows us to reduce the verbosity below.

```suggestion
TEST("tests fail when requirement errors occur") {
  using caf::test::requirement_failed;
  auto& rep = caf::test::reporter::instance();
```

libcaf_test/caf/test/test.test.cpp
```diff
@@ -40,6 +41,51 @@ TEST("tests can contain different types of checks") {
...
+  SECTION("require_eq fails when lhs != rhs") {
+    should_fail_with_exception<caf::test::requirement_failed>(
```
Neverlord, line 49:
```suggestion
    should_fail_with_exception<requirement_failed>(
```

Same for the others.

Neverlord (changes requested):
Just one nitpick to reduce a bit of redundancy. 👍

PR #1532, opened 2023-08-16
Re-implement inject in the deterministic fixture

Neverlord:
I've also ported the `action` unit test to the new framework to have a use test actually using the new `inspect`.

### DIFF
--- a/libcaf_test/caf/test/fixture/deterministic.hpp
+++ b/libcaf_test/caf/test/fixture/deterministic.hpp
@@ -318,6 +318,73 @@ public:
     message_predicate<Ts...> with_;
   };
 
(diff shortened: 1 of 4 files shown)


libcaf_test/caf/test/fixture/deterministic.hpp
```diff
@@ -379,6 +446,25 @@ class CAF_TEST_EXPORT deterministic {
     return evaluator<Ts...>{this, loc, evaluator_algorithm::allow};
   }
 
+  /// Helper class for `inject` that only provides `with`.
+  struct inject_helper {
+    deterministic*fix;
```
riemass: This should fix the `clang-format`
```suggestion
    deterministic* fix;
```

riemass (approved):

PR #1533, opened 2023-08-19
Add print for debug, warning and trace severities

shariarriday:

Complete issue by adding the remaining print functions for debug, warning and trace severities.

### DIFF
--- a/libcaf_test/caf/test/reporter.cpp
+++ b/libcaf_test/caf/test/reporter.cpp
@@ -373,8 +373,8 @@ public:
               ' ', indent_, location.file_name(), location.line(), msg);
   }
 
-  void info(std::string_view msg,
-            const detail::source_location& location) override {
+  void print_info(std::string_view msg,
+                  const detail::source_location& location) override {
     using detail::format_to;
     if (level_ < CAF_LOG_LEVEL_INFO)
       return;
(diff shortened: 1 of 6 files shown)


libcaf_test/caf/test/reporter.hpp
```diff
@@ -72,7 +72,19 @@ class CAF_TEST_EXPORT reporter {
...
+  virtual void
+  print_debug(std::string_view msg, const detail::source_location& location)
+    = 0;
+
+  virtual void
+  print_warning(std::string_view msg, const detail::source_location& location)
+    = 0;
+
+  virtual void
+  print_trace(std::string_view msg, const detail::source_location& location)
```
Neverlord, line 87:
```suggestion
```

Let's skip `trace` for now. Tracing should affect the indentation level, so we'll probably need to split it into begin/end functions and/or work with guards. We have no immediate use case yet, so let's kick this down the road.

Neverlord (changes requested):

libcaf_test/caf/test/runnable.hpp
```diff
@@ -62,9 +62,9 @@ class CAF_TEST_EXPORT runnable {
   void info(detail::format_string_with_location fwl, Ts&&... xs) {
```
Neverlord, line 62: The user-facing function should also be renamed:

```suggestion
  void print_info(detail::format_string_with_location fwl, Ts&&... xs) {
```

Plus adding `print_debug` and `print_warning`.

Neverlord (changes requested):

libcaf_test/caf/test/reporter.cpp
```diff
@@ -373,8 +373,8 @@ class default_reporter : public reporter {
               ' ', indent_, location.file_name(), location.line(), msg);
   }
 
-  void info(std::string_view msg,
-            const detail::source_location& location) override {
+  void print_info(std::string_view msg,
```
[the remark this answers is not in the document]
shariarriday: I had initially thought of making a single `print` function that can call `print_info`, `print_debug` or `print_error` based on the verbosity that is passed in the function. But, I think that this does not provide any benefit to the users of the library or in the code base as we still have to use the copy and pasted `print_*` libraries privately. 

We can probably keep a single `print` function that prints based on verbosity by dynamically configuring the print string in a single function. But, there is a small drawback to that approach, which is that it will be much harder to add different type of print functionalities (like trace) in the future. Let me know your preference regarding this.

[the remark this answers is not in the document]
Neverlord: Yeah, good point. Trace wouldn't play nice with a single `print` function since we need enter and leave events.

Then let's at least reduce the code duplication with a private print function, e.g.:

```c++
  void print_debug(std::string_view msg,
                   const detail::source_location& location) override {
    print_impl(CAF_LOG_LEVEL_DEBUG, 'C', "debug", msg, location);
  }
```

Whereas `B` is the color code (cyan in this case), and `debug` is the text for the message. I hope the rest is self-explanatory regarding how to implement `print_impl`.

libcaf_test/caf/test/reporter.cpp
```diff
@@ -375,54 +375,37 @@ class default_reporter : public reporter {
...
-    using detail::format_to;
-    if (level_ < CAF_LOG_LEVEL_INFO)
-      return;
-    set_live();
-    format_to(colored(),
-              "{0:{1}}$M(info):\n"
-              "{0:{1}}  loc: $C({2}):$Y({3})$0\n"
-              "{0:{1}}  msg: {4}\n",
-              ' ', indent_, location.file_name(), location.line(), msg);
+    print_impl(CAF_LOG_LEVEL_INFO, 'C', "info", msg, location);
```
Neverlord, line 378:
```suggestion
    print_impl(CAF_LOG_LEVEL_INFO, 'M', "info", msg, location);
```

libcaf_test/caf/test/reporter.hpp
```diff
@@ -114,6 +114,12 @@ class CAF_TEST_EXPORT reporter {
 
   /// Creates a default reporter that writes to the standard output.
   static std::unique_ptr<reporter> make_default();
+
+private:
+  virtual void print_impl(unsigned level, char color_code,
+                          std::string_view prefix, std::string_view msg,
+                          const detail::source_location& location)
+    = 0;
```
Neverlord, line 122: Why should the interface impose an implementation detail?

```suggestion
```

libcaf_test/caf/test/reporter.cpp
```diff
@@ -375,54 +375,37 @@ class default_reporter : public reporter {
...
-              ' ', indent_, location.file_name(), location.line(), msg);
+    print_impl(CAF_LOG_LEVEL_DEBUG, 'B', "debug", msg, location);
   }
 
   void print_warning(std::string_view msg,
                      const detail::source_location& location) override {
+    print_impl(CAF_LOG_LEVEL_WARNING, 'Y', "warning", msg, location);
+  }
+
+  void print_impl(unsigned level, char color_code, std::string_view prefix,
```
Neverlord, line 396: This shouldn't be public. Also, I find the `prefix` parameter confusing. It's not used as a prefix at all, I believe something like `level_name` would be more descriptive.

Neverlord (changes requested):

Neverlord (approved):

PR #1541, opened 2023-08-27
Add element_at observable step

shariarriday:

Adds element_at observable step.

### DIFF
--- a/libcaf_core/caf/flow/observable_decl.hpp
+++ b/libcaf_core/caf/flow/observable_decl.hpp
@@ -122,6 +122,9 @@ public:
   /// Returns a transformation that selects all but the first `n` items.
   transformation<step::skip<T>> skip(size_t n);
 
+  /// Returns a transformation that selects only the item with `n` index.
+  transformation<step::element_at<T>> element_at(size_t n);
+
   /// Returns a transformation that discards only the last `n` items.
   transformation<step::skip_last<T>> skip_last(size_t n);
 
--- a/libcaf_core/caf/flow/observable_decl.hpp
+++ b/libcaf_core/caf/flow/observable_decl.hpp
@@ -122,7 +122,7 @@ public:
   /// Returns a transformation that selects all but the first `n` items.
   transformation<step::skip<T>> skip(size_t n);
 
-  /// Returns a transformation that selects only the item with `n` index.
+  /// Returns a transformation that selects only the item at index `n`.
   transformation<step::element_at<T>> element_at(size_t n);
 
   /// Returns a transformation that discards only the last `n` items.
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/flow/step/element_at.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace caf::flow::step {
+
+template <class T>
+class element_at {
+public:
+  using input_type = T;
+
+  using output_type = T;
+
+  explicit element_at(size_t ind) : element_index_(ind) {
```
Neverlord, line 20:
```suggestion
  explicit element_at(size_t index) : element_index_(index) {
```

libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -122,6 +122,9 @@ class observable {
   /// Returns a transformation that selects all but the first `n` items.
   transformation<step::skip<T>> skip(size_t n);
 
+  /// Returns a transformation that selects only the item with `n` index.
```
Neverlord, line 125:
```suggestion
  /// Returns a transformation that selects only the item at index `n`.
```

libcaf_core/caf/flow/step/element_at.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  element_at(element_at&&) = default;
+  element_at(const element_at&) = default;
+  element_at& operator=(element_at&&) = default;
+  element_at& operator=(const element_at&) = default;
+
+  template <class Next, class... Steps>
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
+    if (element_index_ == current_index_) {
+      if (!next.on_next(item, steps...))
```
Neverlord, line 32: After finding the correct item, this operator should call `on_complete` and return `false`.

Neverlord (changes requested):

libcaf_core/caf/flow/step/element_at.hpp
```diff
@@ -0,0 +1,55 @@
...
+  element_at& operator=(const element_at&) = default;
+
+  template <class Next, class... Steps>
+  bool on_next(const input_type& item, Next& next, Steps&... steps) {
+    if (element_index_ == current_index_) {
+      if (next.on_next(item, steps...))
+        on_complete(next, steps...);
+      return false;
+    }
+    current_index_++;
```
Neverlord, line 36: As a general use: always use prefix increment *unless* you are actually using the result of the postfix increment. For integers and pointers, it shouldn't make a difference on modern compilers. However, it's a good habit to have regardless, because it does make a difference when dealing with custom iterator types.

```suggestion
    ++current_index_;
```

Neverlord (changes requested):

PR #1547, opened 2023-09-05
Add delegate tests with deterministic fixture

shariarriday:
relates https://github.com/actor-framework/actor-framework/issues/1051

Initial tests have been written that checks that delegates passes error and successful responses to the original sender.

### DIFF
--- a/libcaf_core/caf/delegate.test.cpp
+++ b/libcaf_core/caf/delegate.test.cpp
@@ -16,18 +16,20 @@ namespace fixture = caf::test::fixture;
 WITH_FIXTURE(fixture::deterministic) {
 
 #ifdef CAF_ENABLE_EXCEPTIONS
-TEST("delegates pass on the response properly") {
+TEST("delegates pass on the response properly to the sender") {
   using namespace caf;
-  auto worker = sys.spawn([] {
-    return behavior{[](int32_t value) { return value * 2; },
-                    [](std::string value) {
-                      return make_error(caf::sec::runtime_error,
-                                        "Invalid type");
-                    }};
+  auto count = std::make_shared<int32_t>(0);
+  auto worker = sys.spawn([count] {
+    return behavior{
+      [count](int32_t value) {
+        *count = *count + value;
+        return *count;
+      },
+    };
   });
   auto delegator = sys.spawn(
     [](event_based_actor* self, caf::actor worker) {
-      return caf::behavior{
+      return behavior{
         [self, worker](int32_t x) {
           auto promise = self->make_response_promise<int>();
           return promise.delegate(worker, x);
@@ -39,20 +41,65 @@ TEST("delegates pass on the response properly") {
       };
     },
     worker);
-  scoped_actor self{sys};
-  SECTION("successful requests are propagated to sender") {
-    self->send(delegator, 2);
-    check_eq(mail_count(delegator), 1u);
-    check(dispatch_messages());
-    self->receive([&](int32_t result) { check_eq(result, 4); });
-  }
-  SECTION("failed requests are propagated to sender") {
-    self->send(delegator, "2");
-    check_eq(mail_count(delegator), 1u);
-    check(dispatch_messages());
-    self->receive([&](caf::error& err) {
-      check_eq(err, make_error(sec::runtime_error, "Invalid type"));
+  SECTION("handled messages are received by client") {
+    auto client = sys.spawn([delegator](event_based_actor* self) {
+      self->send(delegator, 2);
+      self->send(delegator, 3);
+      return behavior{[](int32_t) {}};
     });
+    expect<int32_t>().with(2).from(client).to(delegator);
+    expect<int32_t>().with(2).from(client).to(worker);
+    expect<int32_t>().with(2).from(worker).to(client);
+    check_eq(*count, 2);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(5).from(worker).to(client);
+    check_eq(*count, 5);
+  }
+  SECTION("unhandled messages respond with error") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "2");
+      return behavior{[](int32_t) {}};
+    });
+    expect<std::string>().with("2").from(client).to(delegator);
+    expect<std::string>().with("2").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+  }
+  SECTION("consecutive messages are handled after unhandled message") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "2");
+      self->send(delegator, 3);
+      self->send(delegator, 4);
+      return behavior{[](int32_t) {}};
+    });
+    expect<std::string>().with("2").from(client).to(delegator);
+    expect<std::string>().with("2").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(3).from(worker).to(client);
+    check_eq(*count, 3);
+    expect<int32_t>().with(4).from(client).to(delegator);
+    expect<int32_t>().with(4).from(client).to(worker);
+    expect<int32_t>().with(4).from(worker).to(client);
+    check_eq(*count, 7);
+  }
+  SECTION("consecutive unhandlded messages respond with error") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "2");
+      self->send(delegator, "3");
+      self->send(delegator, "4");
+      return behavior{[](int32_t) {}};
+    });
+    expect<std::string>().with("2").from(client).to(delegator);
+    expect<std::string>().with("2").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+    expect<std::string>().with("3").from(client).to(delegator);
+    expect<std::string>().with("3").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+    expect<std::string>().with("4").from(client).to(delegator);
+    expect<std::string>().with("4").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
   }
 }
 #endif // CAF_ENABLE_EXCEPTIONS
--- a/libcaf_core/caf/delegate.test.cpp
+++ b/libcaf_core/caf/delegate.test.cpp
@@ -11,12 +11,9 @@
 #include "caf/sec.hpp"
 #include "caf/send.hpp"
 
-namespace fixture = caf::test::fixture;
+WITH_FIXTURE(caf::test::fixture::deterministic) {
 
-WITH_FIXTURE(fixture::deterministic) {
-
-#ifdef CAF_ENABLE_EXCEPTIONS
-TEST("delegates pass on the response properly to the sender") {
+TEST("delegation moves responsibility for a request to another actor") {
   using namespace caf;
   auto count = std::make_shared<int32_t>(0);
   auto worker = sys.spawn([count] {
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "caf/test/test.hpp"
+
+#include "caf/event_based_actor.hpp"
+#include "caf/scoped_actor.hpp"
+#include "caf/sec.hpp"
+#include "caf/send.hpp"
+
+namespace fixture = caf::test::fixture;
+
+WITH_FIXTURE(fixture::deterministic) {
```
Neverlord, line 16:
```suggestion
WITH_FIXTURE(caf::test::fixture) {
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#ifdef CAF_ENABLE_EXCEPTIONS
```
Neverlord, line 18: Why is this here?

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+TEST("delegates pass on the response properly to the sender") {
```
Neverlord, line 19: Please avoid things like "does correctly/properly". Without context, there is no specification for correctness or what "properly" would imply.

Here's a suggestion:

```suggestion
TEST("delegation moves responsibility for a request to another actor") {
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+  auto worker = sys.spawn([count] {
+    return behavior{
+      [count](int32_t value) {
+        *count = *count + value;
+        return *count;
+      },
+    };
+  });
+  auto delegator = sys.spawn(
+    [](event_based_actor* self, caf::actor worker) {
```
Neverlord, line 31: Why use a parameter if we can just capture the variable?

```suggestion
    [worker](event_based_actor* self) {
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+      return behavior{
+        [self, worker](int32_t x) {
+          auto promise = self->make_response_promise<int>();
+          return promise.delegate(worker, x);
```
Neverlord, line 35: This certainly works, but it's quite verbose. Why go through a promise?

```suggestion
          return self->delegate(worker, x);
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+        },
+        [self, worker](std::string x) {
+          auto promise = self->make_response_promise<error>();
+          return promise.delegate(worker, x);
```
Neverlord, line 39: Same.

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+        },
+      };
+    },
+    worker);
+  SECTION("handled messages are received by client") {
```
Neverlord, line 44:
```suggestion
  SECTION("the delegatee responds to the original sender") {
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    expect<int32_t>().with(2).from(worker).to(client);
+    check_eq(*count, 2);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(5).from(worker).to(client);
+    check_eq(*count, 5);
+  }
+  SECTION("unhandled messages respond with error") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "2");
```
Neverlord, line 61: I'd use a different string to make a bit more obvious that we are no longer dealing with integers.

```suggestion
      self->send(delegator, "foo");
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    expect<int32_t>().with(2).from(client).to(delegator);
+    expect<int32_t>().with(2).from(client).to(worker);
+    expect<int32_t>().with(2).from(worker).to(client);
+    check_eq(*count, 2);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(5).from(worker).to(client);
+    check_eq(*count, 5);
+  }
+  SECTION("unhandled messages respond with error") {
```
Neverlord, line 59:
```suggestion
  SECTION("errors at the delegatee are sent to the original sender") {
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+          return promise.delegate(worker, x);
+        },
+      };
+    },
+    worker);
+  SECTION("handled messages are received by client") {
+    auto client = sys.spawn([delegator](event_based_actor* self) {
+      self->send(delegator, 2);
+      self->send(delegator, 3);
+      return behavior{[](int32_t) {}};
```
Neverlord, line 48: We always put trailing commas for behaviors.

```suggestion
      return behavior{[](int32_t) {},};
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    check_eq(*count, 2);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(5).from(worker).to(client);
+    check_eq(*count, 5);
+  }
+  SECTION("unhandled messages respond with error") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "2");
+      return behavior{[](int32_t) {}};
```
Neverlord, line 62:
```suggestion
      return behavior{[](int32_t) {},};
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(3).from(worker).to(client);
+    check_eq(*count, 3);
+    expect<int32_t>().with(4).from(client).to(delegator);
+    expect<int32_t>().with(4).from(client).to(worker);
+    expect<int32_t>().with(4).from(worker).to(client);
+    check_eq(*count, 7);
+  }
```
Neverlord, line 86: This test makes no sense to me. The `client` does not override the error handler. Hence, it will terminate after receiving the unexpected message error. Consequently, Line 80 and line 84 can never pass, since `client` will close its mailbox when shutting down and the response from `worker` will be discarded.

In general, this test doesn't seem to cover anything that we haven't covered before. I'd just remove it unless there's an edge case that you have tried to test here and that I'm missing.

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,109 @@
...
+  SECTION("consecutive unhandlded messages respond with error") {
```
Neverlord, line 87: This test also uses tries to use an actor after it has been terminated. Same as above: doesn't seem to bring something new to the table.

Neverlord (changes requested):

libcaf_core/caf/delegate.test.cpp
shariarriday: In the issue description it was mentioned that 

> If actor B1 does not have a handler for this message and it falls into the deadletter (in akka terminology), the original sender actor in the chain on side A (A1) hangs and is no longer able to receive any messages.

I attempted to reproduce this as from my understanding the behavior for the actor would be to receive messages. But, if we are expecting that no further message will be received by the `client` , the test can only cover one more scenario that has not been tested and can be added, that is to check that the error handler for the delegate has been called. Shall I rewrite the tests to reproduce that scenario instead?

Neverlord: I think we have the scenario in the ticket covered with test in line 59, no? If that test is passing, I think the comment by tariksmajic is pointing to the right direction. In the setup described by the original poster, the actor might not actually be hanging but in fact shut down properly.

To double-check this, add another actor to the setup for line 59 that monitors the client. After the client receives the error, we want the actor (a good name might be `observer`) to receive the `down_msg`.

libcaf_core/caf/delegate.test.cpp
shariarriday: Did you mean `WITH_FIXTURE(caf::test::fixture::deterministic) {` ?

libcaf_core/caf/delegate.test.cpp
shariarriday: In the deterministic.test.cpp file all the tests using expect<> was written with exceptions enabled. We also need to add some `should_fail_with_exception` in the tests in future in my opinion. Shall I remove them and group the checks with `should_fail_with_exception` seperately?

Neverlord: We won't need `should_fail_with_exception` here. That was for testing the fixture itself. If we need something that checks that a message is *not* present, we'll add another predicate. The old framework has `disallow` for this, so we'll probably bring that back.

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,78 @@
...
+      };
+    });
+    auto observer_client = sys.spawn([&](event_based_actor* self) {
+      self->monitor(client);
+      self->set_down_handler([=](down_msg& dm) {
+        check_eq(dm.reason, sec::unexpected_message);
+        check_eq(dm.source, client);
+        check_eq(*count, 0);
+        self->quit();
+      });
```
Neverlord, line 64: I'd leave the test logic outside of the actor. Technically this works, but it spreads the checks out more than necessary. You can leave an empty down handler to suppress the printed warning.

```suggestion
      self->set_down_handler([](down_msg&) {});
```

Neverlord (changes requested):

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,77 @@
...
+  SECTION("the delegatee sends errors to the original sender") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
+      self->send(delegator, "foo");
+      return behavior{
+        [](int32_t) {},
+      };
+    });
+    auto observer = sys.spawn([client](event_based_actor* self) {
+      self->monitor(client);
+      self->set_down_handler([](down_msg&) {});
```
Neverlord, line 60:
```suggestion
      self->set_down_handler([](const down_msg&) {});
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,77 @@
...
+      return behavior{
+        [](int32_t) {},
+      };
+    });
+    expect<std::string>().with("foo").from(client).to(delegator);
+    expect<std::string>().with("foo").from(client).to(worker);
+    expect<error>().with(sec::unexpected_message).from(worker).to(client);
+    expect<down_msg>()
+      .with(down_msg{client->address(), sec::unexpected_message})
```
Neverlord, line 69: Always call `.address()` over `->address()`. The latter is private API (and will crash on a null handle).

```suggestion
      .with(down_msg{client.address(), sec::unexpected_message})
```

libcaf_core/caf/delegate.test.cpp
```diff
@@ -0,0 +1,77 @@
...
+    expect<int32_t>().with(2).from(client).to(worker);
+    expect<int32_t>().with(2).from(worker).to(client);
+    check_eq(*count, 2);
+    expect<int32_t>().with(3).from(client).to(delegator);
+    expect<int32_t>().with(3).from(client).to(worker);
+    expect<int32_t>().with(5).from(worker).to(client);
+    check_eq(*count, 5);
+  }
+  SECTION("the delegatee sends errors to the original sender") {
+    auto client = sys.spawn([count, delegator](event_based_actor* self) {
```
Neverlord, line 52:
```suggestion
    auto client = sys.spawn([delegator](event_based_actor* self) {
```

Neverlord (changes requested):

PR #1548, opened 2023-09-06
Fix function object lifetimes in actions

### LINKED ISSUE #698: Canceled timeouts hold onto actor references until they expire, thus blocking program shutdown
This is the first issue I'm filing with actor-framework. I'm new to most of its concepts and this might be an error on my end.

The following test program is supposed to chain two actors together. The message flow is supposed to be caf_main -> parent -> child -> parent handler -> caf_main handler. The timeout in the call from the parent to the child seems to hold up the main process from exiting. Setting it to `seconds(2)` makes the whole process take 2 seconds to exit. Setting it to `infinite` makes it exit immediately. Is this expected?

I convinced myself through a bunch of aout's that the control flow is as I expect it to be.

```
#include <chrono>
#include <string>
#include <iostream>
#include <unordered_map>

#include "caf/all.hpp"

using namespace caf;
using namespace std;
using namespace std::chrono;

using ping_atom = atom_constant<atom("ping")>;
using pong_atom = atom_constant<atom("pong")>;

behavior fwd_actor(event_based_actor* self, const actor* child) {
  return {
    [=](ping_atom) -> result<int> {
      auto rp = self->make_response_promise<int>();
      if (child != nullptr) {
        aout(self) << "ping parent -> child" << endl;
        self->request(*child, seconds(2), ping_atom::value).then([=](int i) mutable {
          aout(self) << "parent -> child done, result: " << i << endl;
          rp.deliver(i);
        }, [=](error& e){
          aout(self) << self->system().render(e) << std::endl;
        });
        aout(self) << "parent done" << endl;
      } else {
        aout(self) << "ping in the child" << endl;
        rp.deliver(2);
      }
    return rp;
    },
  };
}

void caf_main(actor_system& system) {
  scoped_actor self(system);
  auto child = system.spawn(fwd_actor, nullptr);
  auto parent = system.spawn(fwd_actor, &child);
  self->request(parent, seconds(4), ping_atom::value).receive(
    [&](int i) {
      aout(self) << "main call done, result: " << i << endl;
    },
    [&](error& e) {
      cout << self->system().render(e) << std::endl;
  });

  aout(self) << "Done" << endl;
}

CAF_MAIN();
```

Compile with `g++ -g --std=c++17 -l caf_core -o chain chain.cpp`.

```
time ./chain
ping parent -> child
parent done
ping in the child
parent -> child done, result: 2
main call done, result: 2
Done
./chain  0.01s user 0.03s system 2% cpu 2.034 total
```

riemass:
Closes #698.
When disposing actions, we would tag the state enum as disposed, but still keep the function object until all the intrusive pointers go out of scope. Since the function can have it's own state, like a strong actor pointer, we would block the system until the action would timeout, even thought it is disposed.

### DIFF
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -10,7 +10,6 @@
 #include "caf/detail/core_export.hpp"
 #include "caf/disposable.hpp"
 #include "caf/make_counted.hpp"
-#include "caf/raise_error.hpp"
 
 #include <atomic>
 #include <cstddef>
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -136,7 +136,7 @@ class default_action_impl : public detail::atomic_ref_counted,
                             public action::impl {
 public:
   default_action_impl(F fn)
-    : state_(action::state::scheduled), f_{std::move(fn)} {
+    : state_(action::state::scheduled), f_(std::move(fn)) {
     // nop
   }
 
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -199,7 +199,7 @@ public:
     if (!state_.compare_exchange_strong(expected, action::state::running))
       return;
     f_();
-    // Once run, we can stay in the running state or swith to deferred dispose.
+    // Once run, we can stay in the running state or switch to deferred dispose.
     expected = state_.load();
     if (expected == action::state::running
         && state_.compare_exchange_strong(expected, action::state::scheduled))
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -170,9 +170,6 @@ public:
     }
   }
 
-  void handle_deferred_dispose() {
-  }
-
   bool disposed() const noexcept override {
     auto current_state = state_.load();
     return current_state == action::state::disposed
@@ -199,16 +196,14 @@ public:
       return;
     f_();
     // Once run, we can stay in the running state or switch to deferred dispose.
-    expected = state_.load();
-    if (expected == action::state::running
-        && state_.compare_exchange_strong(expected, action::state::scheduled))
+    expected = action::state::running;
+    if (state_.compare_exchange_strong(expected, action::state::scheduled))
       return;
-    if (expected == action::state::deferred_dispose
-        && state_.compare_exchange_strong(expected, action::state::disposed)) {
-      f_.~F();
-      return;
-    }
-    CAF_RAISE_ERROR("Invalid state found after running an action.");
+    CAF_ASSERT(expected == action::state::deferred_dispose);
+    [[maybe_unused]] auto ok
+      = state_.compare_exchange_strong(expected, action::state::disposed);
+    CAF_ASSERT(ok);
+    f_.~F();
   }
 
   void run() override {
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -132,17 +132,22 @@ inline bool operator!=(const action& lhs, const action& rhs) noexcept {
 namespace caf::detail {
 
 template <class F, bool IsSingleShot>
-struct default_action_impl : detail::atomic_ref_counted, action::impl {
-  std::atomic<action::state> state_;
-  F f_;
-
+class default_action_impl : public detail::atomic_ref_counted,
+                            public action::impl {
+public:
   default_action_impl(F fn)
-    : state_(action::state::scheduled), f_(std::move(fn)) {
+    : state_(action::state::scheduled), f_{std::move(fn)} {
     // nop
   }
 
+  ~default_action_impl() {
+    if (state_.load() == action::state::scheduled)
+      f_.~F();
+  }
+
   void dispose() override {
-    state_ = action::state::disposed;
+    state_.store(action::state::disposed);
+    f_.~F();
   }
 
   bool disposed() const noexcept override {
@@ -156,8 +161,9 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
   void run() override {
     if (state_.load() == action::state::scheduled) {
       f_();
-      if constexpr (IsSingleShot)
-        state_ = action::state::disposed;
+      if constexpr (IsSingleShot) {
+        dispose();
+      }
       // else: allow the action to re-register itself when needed by *not*
       //       setting the state to disposed, e.g., to implement time loops.
     }
@@ -178,6 +184,12 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
   friend void intrusive_ptr_release(const default_action_impl* ptr) noexcept {
     ptr->deref();
   }
+
+private:
+  std::atomic<action::state> state_;
+  union {
+    F f_;
+  };
 };
 
 } // namespace caf::detail
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -146,8 +146,9 @@ public:
   }
 
   void dispose() override {
-    state_.store(action::state::disposed);
-    f_.~F();
+    auto expected = action::state::scheduled;
+    if (state_.compare_exchange_weak(expected, action::state::disposed))
+      f_.~F();
   }
 
   bool disposed() const noexcept override {
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -10,6 +10,7 @@
 #include "caf/detail/core_export.hpp"
 #include "caf/disposable.hpp"
 #include "caf/make_counted.hpp"
+#include "caf/raise_error.hpp"
 
 #include <atomic>
 #include <cstddef>
@@ -24,8 +25,10 @@ public:
 
   /// Describes the current state of an `action`.
   enum class state {
-    disposed,  /// The action may no longer run.
-    scheduled, /// The action is scheduled for execution.
+    scheduled,        /// The action is scheduled for execution.
+    running,          /// The action is currently running in another thread.
+    deferred_dispose, /// The action is currently running, and will be disposed.
+    disposed,         /// The action may no longer run.
   };
 
   /// Internal interface of `action`.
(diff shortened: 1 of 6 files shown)


riemass:
> Ok, I can kinda see how that could be related... But where's a test for it? 🙂

I've confirmed it with the code from the ticket. I've pushed a unit test. The last check, dropping the strong refs to 0, is what failed without the committed code.

libcaf_core/caf/action.hpp
```diff
@@ -132,17 +132,22 @@ inline bool operator!=(const action& lhs, const action& rhs) noexcept {
...
 
+  ~default_action_impl() {
+    if (state_.load() == action::state::scheduled)
+      f_.~F();
+  }
+
   void dispose() override {
-    state_ = action::state::disposed;
+    state_.store(action::state::disposed);
+    f_.~F();
```
Neverlord, line 150: Should try to set `state_` to `disposed` atomically and only call the destructor when transitioning from another state.

```
    state_.store(action::state::disposed);
    f_.~F();
```

libcaf_core/tests/legacy/mixin/requester.cpp
```diff
@@ -363,4 +363,23 @@ TEST_CASE("GH-1299 regression blocking") {
   }
 }
 
+TEST_CASE("GH-698 regression responding to a request drops the strong "
+          "reference count") {
+  auto weak_client_ref = [this] {
+    return actor_cast<weak_actor_ptr>(
+      sys.spawn([this](event_based_actor* self) {
+        self->request(adding_server, std::chrono::seconds{10}, 1, 2)
```
Neverlord, line 371:
```suggestion
        self->request(adding_server, 10s, 1, 2)
```

(plus `using namespace std::literals`)

libcaf_core/tests/legacy/mixin/requester.cpp
```diff
@@ -363,4 +363,23 @@ TEST_CASE("GH-1299 regression blocking") {
   }
 }
 
+TEST_CASE("GH-698 regression responding to a request drops the strong "
+          "reference count") {
```
Neverlord, line 367:
```suggestion
TEST_CASE("actors cancel request timeouts after receiving a response") {
```

libcaf_core/tests/legacy/mixin/requester.cpp
```diff
@@ -363,4 +363,23 @@ TEST_CASE("GH-1299 regression blocking") {
...
+  auto weak_client_ref = [this] {
+    return actor_cast<weak_actor_ptr>(
+      sys.spawn([this](event_based_actor* self) {
```
Neverlord, line 370: This is not what I would expect here. By only having a weak ref to this actor, we will terminate it as unreachable once it processes the response message.

The reported bug was about *not* terminating the actor. The fixed version must work with just:

```suggestion
  auto client = sys.spawn([this](event_based_actor* self) {
```

libcaf_core/tests/legacy/mixin/requester.cpp
```diff
@@ -363,4 +363,23 @@ TEST_CASE("GH-1299 regression blocking") {
...
+      }));
+  }();
+  CHECK_EQ(weak_client_ref->strong_refs, 1u);
+  run_once();
+  CHECK_EQ(weak_client_ref->strong_refs, 2u);
+  expect((int, int), from(weak_client_ref).to(adding_server).with(1, 2));
+  CHECK_EQ(weak_client_ref->strong_refs, 2u);
+  expect((int), from(adding_server).to(weak_client_ref));
+  CHECK_EQ(weak_client_ref->strong_refs, 0u);
+  CHECK_EQ(*result, result_type{3});
```
Neverlord, line 382: I don't get why we are tracking the reference count here if the issue was about timeouts not being cancelled. I would have expected something like this:

```
  run_once();
  expect((int, int), from(client).to(adding_server).with(1, 2));
  CHECK(sched.has_pending_timeout());
  expect((int), from(adding_server).to(client));
  CHECK(!sched.has_pending_timeout());
```

I would also expect a second test that covers the error case, i.e., if the client receives an error instead of an actual result.

Neverlord (changes requested):

libcaf_core/tests/legacy/mixin/requester.cpp
riemass: I guess I can keep the client in scope and expect the ref count to drop to 1.

libcaf_core/tests/legacy/mixin/requester.cpp
Neverlord: Ahh! Yeah, now it makes sense. Sorry for being a bit slow.

To make the test a bit more "digestible", I'd recommend explicitly triggering the `request` to start with a ref count of 1. Something like this:

```c++
  auto server = sys.spawn(...);
  auto client = actor_cast<strong_actor_ptr>(sys.spawn(...));
  run();
  CHECK_EQ(client->strong_refs, 1u);
  inject((actor), with(server).to(client));
  CHECK(sched.has_pending_timeout());
  expect((int, int), from(client).to(server));
  expect((int), from(server).to(client));
  CHECK(!sched.has_pending_timeout());
  // The scheduler may no longer hold a reference to the client.
  CHECK_EQ(client->strong_refs, 1u);
```

libcaf_core/tests/legacy/mixin/requester.cpp
Neverlord: All right, since it's a quite subtle issue, let's just name the test `TEST_CASE("GH-698 regression")`.

libcaf_core/caf/action.hpp
```diff
@@ -156,8 +162,9 @@ struct default_action_impl : detail::atomic_ref_counted, action::impl {
   void run() override {
     if (state_.load() == action::state::scheduled) {
       f_();
-      if constexpr (IsSingleShot)
-        state_ = action::state::disposed;
+      if constexpr (IsSingleShot) {
+        dispose();
+      }
```
Neverlord, line 167: This code can now run into UB, e.g. segfault or do other weird stuff.

- Thread 1 is calling run() and passes line 164
- Thread 2 is calling dispose() and runs the destructor of f_
- Thread 1 calls f_(): boom; accessing a destructed object

There's an easy fix for single-shot actions, but multi-shot actions are tricky to implement lock-free. The multi-shot type is used much more widely in the code base, so we need a good solution here.

Neverlord (changes requested):

libcaf_core/caf/action.hpp
```diff
@@ -132,34 +136,87 @@ inline bool operator!=(const action& lhs, const action& rhs) noexcept {
...
+    auto expected = action::state::scheduled;
+    if (!state_.compare_exchange_strong(expected, action::state::disposed))
+      return;
+    f_();
+    f_.~F();
+    return;
+  }
+
+  void run_multi_shot() {
+    // We can only run a scheduled action
```
Neverlord, line 197:
```suggestion
    // We can only run a scheduled action.
```

libcaf_core/caf/action.test.cpp
```diff
@@ -35,6 +38,25 @@ SCENARIO("actions wrap function calls") {
...
+        called = true;
+        uut_ptr->dispose();
+        check_eq(uut_ptr->ptr()->current_state(),
+                 action::state::deferred_dispose);
+      });
+      uut_ptr = &uut;
+      THEN("it transitions to deferred_dispose") {
+        check(uut.scheduled());
+        uut.run();
+        check(called);
```
Neverlord, line 54: Good idea writing the test like this to guarantee the action ending up the deferred dispose path. 👍

One nitpick: I'd keep the checks in the `THEN` block since we spell out the expected transition there but don't actually check it:

```suggestion
      });
      uut_ptr = &uut;
      THEN("it transitions to deferred_dispose") {
        check(uut.scheduled());
        uut.run();
        check(called);
        check_eq(uut_ptr->ptr()->current_state(),
                 action::state::deferred_dispose);
```

libcaf_core/caf/action.test.cpp
```diff
@@ -35,6 +38,25 @@ SCENARIO("actions wrap function calls") {
...
+    WHEN("disposing an running action") {
+      auto called = false;
+      action* uut_ptr = nullptr;
+      auto uut = make_action([this, &called, &uut_ptr] {
+        called = true;
+        uut_ptr->dispose();
+        check_eq(uut_ptr->ptr()->current_state(),
+                 action::state::deferred_dispose);
+      });
+      uut_ptr = &uut;
```
Neverlord, line 50: I suppose this whole `uut_ptr` setup is to have the action refer back to itself? Unless I miss something, I think it could still be a bit less verbose like this:

```suggestion
      action uut;
      uut = make_action([&called, &uut] {
        called = true;
        uut.dispose();
      });
```

libcaf_core/caf/action.test.cpp
```diff
@@ -55,6 +77,41 @@ SCENARIO("actions wrap function calls") {
...
+    WHEN("running the action once") {
+      auto called = false;
+      auto uut = make_single_shot_action([&called] { called = true; });
+      THEN("it transitions to disposed") {
+        check(uut.scheduled());
+        uut.run();
+        check(called);
+        check(uut.disposed());
+      }
+      AND_THEN("run no longer calls the lambda") {
```
Neverlord, line 90:
```suggestion
      AND_THEN("running the action again has no effect") {
```

libcaf_core/caf/action.hpp
```diff
@@ -132,34 +136,87 @@ inline bool operator!=(const action& lhs, const action& rhs) noexcept {
...
+    expected = state_.load();
+    if (expected == action::state::running
+        && state_.compare_exchange_strong(expected, action::state::scheduled))
+      return;
+    if (expected == action::state::deferred_dispose
+        && state_.compare_exchange_strong(expected, action::state::disposed)) {
+      f_.~F();
+      return;
+    }
+    CAF_RAISE_ERROR("Invalid state found after running an action.");
```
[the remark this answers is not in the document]
riemass: I see what you did there. I agree with the changes, but I think without the line:
`auto expected = state_.load();`
expected would be left as `action::state::scheduled`, since compare exchange does nothing to expected if it succeeds.

libcaf_core/caf/action.test.cpp
riemass: Good point. I also need the check_eq inside the lambda.

libcaf_core/caf/action.test.cpp
riemass: I can't really refactor it like this since after finishing `run` the action would go to the `disposed` state. I need a check that is triggered during the `run` and asks for `deferred_dispose`.

Neverlord: Yeah, good point. Didn't think of that. :)

Neverlord (approved):

Issue #588, opened 2017-08-12
Allow user-defined logger classes


Neverlord:
Original discussion see [stackoverflow](https://stackoverflow.com/questions/45589227/customizing-c-caf-framework).

rsmmr:
+1 

It'd  be great if an application could leverage CAF's logging capabilities for its own debug logging even if CAF isn't specifically compiled with debugging support .

Neverlord:
> It'd be great if an application could leverage CAF's logging capabilities for its own debug logging even if CAF isn't specifically compiled with debugging support.

FYI, this is now the case with [0.17](https://github.com/actor-framework/actor-framework/releases/tag/0.17.0). Users can't replace the logger as of yet, but they can use CAF's logging infrastructure regardless of CAF's build configuration now.

PR #1549, opened 2023-09-06
Refactor logger to be an interface

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/588

@Neverlord , I would like to get your feedback regarding the refactor. The code will not compile due to the tests in `libcaf_core\tests\legacy\logger.cpp`. The public interface has changed and some functions are not testable now. We could approach this in two ways, one would be to make the tested functions public again and then they will compile and the tests will pass too. Another would be to remove these tests as they are not relevant anymore.

### DIFF
--- a/libcaf_core/caf/logger.cpp
+++ b/libcaf_core/caf/logger.cpp
@@ -213,18 +213,19 @@ public:
     }
   };
 
+  /// Represents a single format string field.
+  struct field {
+    field_type kind;
+    std::string text;
+  };
+
+  /// Stores a parsed format string as list of fields.
+  using line_format = std::vector<field>;
+
   // -- constructors, destructors, and assignment operators --------------------
 
   default_logger(actor_system& sys) : t0_(make_timestamp()), system_(sys) {
     // nop
-  }
-
-  virtual ~default_logger() {
-    stop();
-    // tell system our dtor is done
-    std::unique_lock<std::mutex> guard{system_.logger_dtor_mtx()};
-    system_.logger_dtor_done(true);
-    system_.logger_dtor_cv().notify_one();
   }
 
   // -- logging ----------------------------------------------------------------
@@ -376,16 +377,9 @@ public:
     this->deref();
   }
 
-private:
-  // -- called by the actor_system when running with a test coordinator --------
-
-  void inline_output(bool value) noexcept {
-    cfg_.inline_output = value;
-  }
-
   // -- initialization ---------------------------------------------------------
 
-  void init(actor_system_config& cfg) {
+  void init(const actor_system_config& cfg) override {
     CAF_IGNORE_UNUSED(cfg);
     namespace lg = defaults::logger;
     using std::string;
@@ -399,6 +393,7 @@ private:
       if (auto lst = get_as<string_list>(cfg, key))
         var = std::move(*lst);
     };
+    cfg_.inline_output = get_or(cfg, "caf.scheduler.policy", "") == "testing";
     cfg_.file_verbosity = get_verbosity("caf.logger.file.verbosity");
     cfg_.console_verbosity = get_verbosity("caf.logger.console.verbosity");
     cfg_.verbosity = std::max(cfg_.file_verbosity, cfg_.console_verbosity);
@@ -544,7 +539,7 @@ private:
     }
   }
 
-  void start() {
+  void start() override {
     parent_thread_ = std::this_thread::get_id();
     if (verbosity() == CAF_LOG_LEVEL_QUIET)
       return;
@@ -596,7 +591,7 @@ private:
     }
   }
 
-  void stop() {
+  void stop() override {
     if (cfg_.inline_output) {
       log_last_line();
       return;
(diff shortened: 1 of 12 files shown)


libcaf_core/caf/logger.hpp
```diff
@@ -265,84 +202,19 @@ class CAF_CORE_EXPORT logger : public ref_counted {
...
-
-  // Filters events by component name for file output.
-  std::vector<std::string> file_filter_;
-
-  // Filters events by component name for console output.
-  std::vector<std::string> console_filter_;
+  virtual void start() = 0;
 
   // References the parent system.
   actor_system& system_;
```
Neverlord, line 217: There shouldn't be any state in the interface. Also shouldn't inherit from `ref_counted` for that reason.

For the ref counting, there should be `ref_logger` and `deref_logger` as pure virtual functions. You can use https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/flow/coordinated.hpp as an example.

libcaf_core/caf/logger.hpp
```diff
@@ -265,84 +202,19 @@ class CAF_CORE_EXPORT logger : public ref_counted {
...
-
-  void inline_output(bool value) noexcept {
-    cfg_.inline_output = value;
-  }
-
-  // -- initialization ---------------------------------------------------------
+  virtual void init(actor_system_config& cfg) = 0;
 
-  void init(actor_system_config& cfg);
+  virtual void inline_output(bool value) = 0;
```
Neverlord, line 212: Please drop. This whole inline output setup is only used by unit tests. We'll implement a separate logger for unit tests.

Neverlord (changes requested):

shariarriday:
@Neverlord I have removed the tests in `libcaf_core/tests/legacy/logger.cpp` as they are not appropriate for `default_logger` and new logger will be implemented for unit tests according to the previous discussions.

libcaf_core/caf/logger.cpp
```diff
@@ -40,6 +40,177 @@ thread_local actor_id current_actor_id;
 // Stores a pointer to the system-wide logger.
 thread_local intrusive_ptr<logger> current_logger_ptr;
 
+// Default logger implementation.
+class default_logger : public logger, detail::atomic_ref_counted {
```
Neverlord, line 44:
```suggestion
// Default logger implementation.
class default_logger : public logger, public detail::atomic_ref_counted {
```

Also, this class is in the anonymous namespace but the declaration is outside of the namespace. In any case, there is no need to implement the member functions out-of-line anymore. Please move the implementations to the class definition.

libcaf_core/caf/logger.cpp
```diff
@@ -248,26 +423,35 @@ logger* logger::current_logger() {
...
+
+default_logger::~default_logger() {
   stop();
   // tell system our dtor is done
-  std::unique_lock<std::mutex> guard{system_.logger_dtor_mtx_};
-  system_.logger_dtor_done_ = true;
-  system_.logger_dtor_cv_.notify_one();
+  std::unique_lock<std::mutex> guard{system_.logger_dtor_mtx()};
+  system_.logger_dtor_done(true);
+  system_.logger_dtor_cv().notify_one();
```
Neverlord, line 451: Just a note, this setup is too complicated (has nothing to do with your changes). When the refactoring itself is done, I'll think of a better way we can do this so that people don't need to know about internals like this when subtyping `logger`.

libcaf_core/caf/logger.cpp
```diff
@@ -248,26 +423,35 @@ logger* logger::current_logger() {
...
   if (level > cfg_.verbosity)
     return false;
   return std::none_of(global_filter_.begin(), global_filter_.end(),
                       [=](std::string_view name) { return name == cname; });
 }
 
-logger::logger(actor_system& sys) : system_(sys), t0_(make_timestamp()) {
+logger::logger() {
   // nop
 }
```
Neverlord, line 435: This does nothing now, please remove.

```suggestion
```

libcaf_core/caf/logger.hpp
```diff
@@ -209,53 +176,19 @@ class CAF_CORE_EXPORT logger : public ref_counted {
...
-
-  unsigned console_verbosity() const noexcept {
-    return cfg_.console_verbosity;
-  }
+  virtual bool accepts(unsigned level, std::string_view component_name) = 0;
 
   // -- static utility functions -----------------------------------------------
 
+  /// Creates a new logger instance.
+  static logger* make(actor_system& sys);
```
Neverlord, line 184: Returning a raw pointer that transfers ownership is a major red flag:

```suggestion
  /// Creates a new logger instance.
  static intrusive_ptr<caf::logger> make(actor_system& sys);
```

Neverlord (changes requested):

libcaf_core/caf/logger.cpp
shariarriday: I had initially chosen to move these functions inside the logger destructor. Maybe, we can call some overloaded functions which can call these methods instead of calling them in the destructor directly? It still does not fix the issue completely but it may be worth refactoring for now.

Neverlord (approved):

Issue #1474, opened 2023-07-07
Deprecate automatic stashing


Neverlord:
Currently, event-based actors automatically "stash" messages if users skip them in a default handler. Stashing basically means: a skipped message gets put into a stash until an incoming message matched the behavior. Then, all messages from the stash get put back into the mailbox and re-matched again for the next run.

There are some use cases where this work flow makes sense, but all actors have this stash, even if they never use it. It's also easy to shoot yourself in the foot with this implicit skipping, because the constant re-scanning can become expensive. Instead of doing this implicitly, we should instead:

- Provide an explicit `caf::stash` class to allow actors that actually need this feature to use it manually.
- Deprecate `skippable_result` as well as the default handler in general (`set_default_handler`) and remove with version 1.0. Event-based actors that receive unexpected inputs should terminate.
- If users want to deal with unexpected inputs in any other way (like stashing): we should bring back "catch-all" handlers for event-based actors. We don't need to have a special syntax for this, I think. A handler with signature `[](message msg) { }` (or any cv-qualified version thereof) could be treated as catch-all.

Maybe we can also use this opportunity to get rid of the other special handlers as well. Issue #1423 already suggests using an alternative approach for down messages. Basically just remove the old way instead of "overloading" it. We could get rid of the `exit` handler as well: just add a `[](const exit_msg&)` handler to your behavior if you want to "override" the default behavior. In my experience, actors that override the exit handler *very* rarely switch behaviors. The original idea of separating the exit handlers from the behavior was to allow actors to have "constant" exit handler with changing behaviors, but several years later, I think that's just a non-issue. By going down this route, we could get rid of all of these member variables:

```
default_handler default_handler_;
error_handler error_handler_;
down_handler down_handler_;
node_down_handler node_down_handler_;
exit_handler exit_handler_;
```

I think it's also an easier mental model to have (and teach). There's the `behavior` of an actor that has your message callbacks. If an actors fails to match a message, it panicks (i.e., terminates). Period. Currently, it's: "The `behavior` has your message callbacks, but then there are these 5 special cases where CAF treats messages differently."

Neverlord:
We have deprecated `set_default_handler`, which in turns means we have deprecated the use of `skip` here. The new way is using a `mail_cache`.

PR #1559, opened 2023-09-13
Allow passing catch-all handlers to behaviors

Neverlord:
The idea is to render the current "default handlers" on actors obsolete (eventually).

Relates #1474.


libcaf_core/caf/behavior.test.cpp
```diff
@@ -0,0 +1,155 @@
...
+    auto f = behavior{
+      [](message) { return 1; },
+      [](int) { return 2; },
+      [](int, int) { return 3; },
+      [](int, int, int) { return 4; },
+    };
+    check_eq(res_of(f, m1), 1);
+    check_eq(res_of(f, m2), 1);
+    check_eq(res_of(f, m3), 1);
+  }
```
riemass, line 111: If we've put the `message` handler last, the output would be 2, 3, 4? If my intuition is right, maybe cover that in a test.

riemass (approved):
Just a question:

libcaf_core/caf/behavior.test.cpp
Neverlord: That'd be the `the first matching handler wins` test... no?

PR #1560, opened 2023-09-14
Add case insensitive compare to request header

### LINKED ISSUE #1554: Add case-insensitive header field access in http request_header
According to RFC 9110 header field names are case-insensitive: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1-3

However the `request_header` class only allows case-sensitive access of header fields via the key in https://github.com/actor-framework/actor-framework/blob/master/libcaf_net/caf/net/http/request_header.hpp.

Currently the only way to search for case-insensitive keys is to either compare each key with `request_header::for_each_field()` (but then you can't return early when the field was found) or iterate manually over each field with `request_header::num_fields()` and `request_header::field_at()`.

riemass:
Closes #1554.
All field access is case insensitive now. I've also checked for const correctness.

### DIFF
--- a/libcaf_net/caf/net/http/request_header.hpp
+++ b/libcaf_net/caf/net/http/request_header.hpp
@@ -23,9 +23,6 @@ namespace caf::net::http {
 /// query, fragment, version, and fields.
 class CAF_NET_EXPORT request_header {
 public:
-  friend class http::server;
-  friend class web_socket::server;
-
   /// Default constructor.
   request_header() = default;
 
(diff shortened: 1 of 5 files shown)


libcaf_net/caf/net/http/request_header.test.cpp
```diff
@@ -0,0 +1,71 @@
...
+  }
+  SECTION("non existing field returns an empty view") {
+    check_eq(hdr.field("Foo"), "");
+  }
+  SECTION("field access by position") {
+    check_eq(hdr.field_at(0), std::pair("Host"sv, "localhost:8090"sv));
+    check_eq(hdr.field_at(1), std::pair("User-Agent"sv, "AwesomeLib/1.0"sv));
+    check_eq(hdr.field_at(2), std::pair("Accept-Encoding"sv, "gzip"sv));
+    check_eq(hdr.field_at(3), std::pair("Number"sv, "150"sv));
+    check_throws([&hdr]() { hdr.field_at(4); });
```
Neverlord, line 46:
```suggestion
#ifdef CAF_ENABLE_EXCEPTIONS
    check_throws([&hdr]() { hdr.field_at(4); });
#endif
```

libcaf_net/caf/net/http/request_header.test.cpp
```diff
@@ -0,0 +1,71 @@
...
+  SECTION("fields access is case insensitive") {
+    check_eq(hdr.field("HOST"), "localhost:8090");
+    check_eq(hdr.field("USER-agent"), "AwesomeLib/1.0");
+    check_eq(hdr.field("accept-ENCODING"), "gzip");
+  }
+  SECTION("non existing field returns an empty view") {
+    check_eq(hdr.field("Foo"), "");
+  }
+  SECTION("field access by position") {
+    check_eq(hdr.field_at(0), std::pair("Host"sv, "localhost:8090"sv));
```
Neverlord, line 42:
```suggestion
    check_eq(hdr.field_at(0), std::pair{"Host"sv, "localhost:8090"sv});
```

(same for the others)

libcaf_net/caf/net/http/request_header.hpp
```diff
@@ -151,6 +148,15 @@ class CAF_NET_EXPORT request_header {
   std::pair<status, std::string_view> parse(std::string_view raw);
 
 private:
+  // Finds a field by using case insensitive key comparison. Returns the
+  // iterator pointing to the found field, or to the end if no field is found.
+  std::vector<std::pair<std::string_view, std::string_view>>::const_iterator
+  find_by_key_icase(std::string_view key) const noexcept {
```
Neverlord, line 154:
```suggestion
  // An unsorted "map" type for storing key/value pairs.
  using fields_map = std::vector<std::pair<std::string_view, std::string_view>>;

  // Finds a field by using case insensitive key comparison. Returns the
  // iterator pointing to the found field, or to the end if no field is found.
  fields_map::const_iterator
  find_by_key_icase(std::string_view key) const noexcept {
```

(and for the `fields_` member)

Neverlord (changes requested):
Looks good, just a few nitpicks. 👍

libcaf_net/caf/net/http/request_header.hpp
```diff
@@ -164,7 +170,7 @@ class CAF_NET_EXPORT request_header {
   std::string_view version_;
 
   /// A shallow map for looking up individual header fields.
-  unordered_flat_map<std::string_view, std::string_view> fields_;
+  std::vector<std::pair<std::string_view, std::string_view>> fields_;
```
riemass, line 173:
```suggestion
  fields_map fields_;
```

Neverlord (approved):
Good job. 👍

Squashed and rebase only because of the conflict in the changelog (my bad).

Issue #1557, opened 2023-09-12
Coverage: message_builder has <40% coverage


Neverlord:
The class `message_builder` has a legacy test suite, but it's coverage is rather poor: https://app.codecov.io/gh/actor-framework/actor-framework/commit/1dd7a1f6e8bfa7a4586c965663b34f5b39733bdc/blob/libcaf_core/caf/message_builder.cpp.

- Port the existing test to the new framework (plus move the file)
- Get rid of the `STEP`-thingy, just print debug messages
- Add a new scenario for appending from an existing message

PR #1561, opened 2023-09-14
Add test for message_builder based on new framework

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/1557

The tests have been refactored a little bit to minimize duplication. The old tests have not been removed yet and kept as is.


libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "caf/message_builder.hpp"
+
+#include "caf/test/caf_test_main.hpp"
+#include "caf/test/test.hpp"
+
+using namespace caf;
+
+TEST("message_builder can build messages incrementally with "
+     "default-constructor") {
```
Neverlord, line 13:
```suggestion
TEST("message_builder can build messages incrementally") {
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+  message_builder builder;
+  check(builder.empty());
+  check(builder.to_message().empty());
+  check_eq(builder.size(), 0u);
+  SECTION("builder contains the values added by append") {
```
Neverlord, line 18:
```suggestion
  SECTION("append() adds values to a builder") {
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+      check_eq(msg.types(), (make_type_id_list<int32_t, int32_t, int32_t>()));
+      check_eq(to_string(msg.types()), "[int32_t, int32_t, int32_t]");
+      check_eq(to_string(msg), "message(1, 2, 3)");
+    }
+    SECTION("move_to_message produces the same message again") {
+      auto msg = builder.move_to_message();
+      check_eq(msg.types(), (make_type_id_list<int32_t, int32_t, int32_t>()));
+      check_eq(to_string(msg.types()), "[int32_t, int32_t, int32_t]");
+      check_eq(to_string(msg), "message(1, 2, 3)");
+    }
```
Neverlord, line 43: I wouldn't nest this (as well as the other sections below) into the section that tests `append`.

Furthermore, the current setup ties the test for the builder to:
- the output format of `to_string`, which may change in the future
- the unrelated class `const_typed_message_view` (further below)

To test the individual values, please stick to interfaces provided by `message_builder` and `message`. The `move_to_message` test also needs to test that the builder actually no longer holds copies of old values.

```suggestion
    SECTION("move_to_message moves all elements into a message") {
      builder.append("foo"s);
      builder.append(make_counted<counted_int>(42));
      auto msg = builder.move_to_message();
      check_eq(msg.types(), (make_type_id_list<std::string, counted_int>()));
      if (check_eq(to_string(msg.types()), "[std::string, counted_int]")) {
        check_eq(msg.get_as<std::string>(0), "foo");
        const auto& ci = msg.get_as<counted_int_ptr>(1);
        check_eq(ci->value(), 42);
        check_eq(ci->get_reference_count(), 1u);
      }
    }
    SECTION("to_message copies all elements into a message") {
      builder.append("foo"s);
      builder.append(make_counted<counted_int>(42));
      check_eq(builder.size(), 2u);
      auto msg = builder.to_message();
      check_eq(msg.types(), make_type_id_list<std::string, counted_int>());
      if (check_eq(to_string(msg.types()), "[std::string, counted_int]")) {
        check_eq(msg.get_as<std::string>(0), "foo");
        const auto& ci = msg.get_as<counted_int_ptr>(1);
        check_eq(ci->value(), 42);
        check_eq(ci->get_reference_count(), 2u);
      }
    }
```

Whereas `counted_int` is something along those lines:

```cpp
class counted_int : public ref_counted {
public:
  explicit counted_int(int initial_value) : value_(initial_value) {
    // nop
  }

  int value() const {
    return value_;
  }

  void value(int new_value) {
    value_ = new_value;
  }

  template <class Inspector>
  friend bool inspect(Inspector& f, counted_int& x) {
    return f.object(x).fields(f.field("value", x.value_));
  }

private:
  int value_;
};

using counted_int_ptr = intrusive_ptr<counted_int>;

template <class Inspector>
bool inspect(Inspector& f, counted_int_ptr& x) {
  if constexpr (Inspector::is_loading) {
    if (!x)
      x = make_counted<counted_int>(0);
    return inspect(f, *x);
  } else {
    if (!x) {
      auto tmp = counted_int{0};
      return inspect(f, tmp);
    }
    return inspect(f, *x);
  }
}
```

(make sure to not forget the type ID block)

Neverlord (changes requested):

libcaf_core/caf/message_builder.test.cpp
shariarriday: In the updated implementation we will still be relying on the output format of to_string(). Do you think that we should look into updating that too?

```
     if (check_eq(to_string(msg.types()), "[std::string, counted_int]")) {
         check_eq(msg.get_as<std::string>(0), "foo");
         const auto& ci = msg.get_as<counted_int_ptr>(1);
         check_eq(ci->value(), 42);
         check_eq(ci->get_reference_count(), 2u);
      }
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,163 @@
...
+  CAF_ADD_TYPE_ID(counted_int_block, (counted_int))
+  CAF_ADD_TYPE_ID(counted_int_block, (counted_int_ptr))
+
+CAF_END_TYPE_ID_BLOCK(counted_int_block)
+
+TEST("message_builder can build messages incrementally") {
+  message_builder builder;
+  check(builder.empty());
+  check(builder.to_message().empty());
+  check_eq(builder.size(), 0u);
```
Neverlord, line 67:
```suggestion
  SECTION("a default-constructed builder is empty") {
    check(builder.empty());
    check(builder.to_message().empty());
    check_eq(builder.size(), 0u);
  }
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,163 @@
...
+    auto msg = builder.to_message();
+    if (check_eq(msg.types(),
+                 make_type_id_list<std::string, counted_int_ptr>())) {
+      check_eq(msg.get_as<std::string>(0), "foo");
+      const auto& ci = msg.get_as<counted_int_ptr>(1);
+      check_eq(ci->value(), 42);
+      check_eq(ci->get_reference_count(), 2u);
+    }
+  }
+  SECTION("clear() removes all messages in the builder") {
```
Neverlord, line 141:
```suggestion
  SECTION("clear() removes all messages from the builder") {
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,163 @@
...
+    std::string quote = "He who laughs at himself "
+                        "never runs out of things to laugh at.";
+    builder.append(quote);
+    auto msg = builder.to_message();
+    check_eq(msg.types(), (make_type_id_list<std::string>()));
+    using view_t = const_typed_message_view<std::string>;
+    if (auto tup = view_t(msg); check(tup.operator bool())) {
+      check_eq(get<0>(tup), quote);
+    }
+  }
```
Neverlord, line 160: Doesn't test anything new since we use strings previously.

```suggestion
```

Neverlord (changes requested):
Just a few more nitpicks. 🙂

Neverlord:
All right, looking at the coverage report, these are currently missing:

- Constructing from an iterator pair (range)
- Using `append_from` and then calling `move_to_message`

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,191 @@
...
+    check(builder2.to_message().empty());
+    SECTION("builder appends the values when first < msg.size()") {
+      builder2.append_from(msg, 0, msg.size());
+      auto msg2 = builder2.to_message();
+      check_eq(msg2.types(), make_type_id_list<int32_t, int32_t, int32_t>());
+      check_eq(msg2.get_as<int>(0), 1);
+      check_eq(msg2.get_as<int>(1), 2);
+      check_eq(msg2.get_as<int>(2), 3);
+    }
+    SECTION("builder does not append the value when first >= msg.size()") {
```
Neverlord, line 106:
```suggestion
    SECTION("builder does nothing when first >= msg.size()") {
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,191 @@
...
+    builder.append(vec.begin(), vec.end());
+    check_eq(builder.size(), 3u);
+    builder.clear();
+    check(builder.empty());
+    check(builder.to_message().empty());
+    check_eq(builder.size(), 0u);
+  }
+}
+
+TEST("message_builder can build messages from iterator pair") {
```
Neverlord, line 179:
```suggestion
TEST("message_builder can build messages from iterator pairs") {
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,191 @@
...
+  std::vector<int32_t> xs{1, 2, 3};
+  message_builder builder(xs.begin(), xs.end());
+  check(!builder.empty());
+  check_eq(builder.size(), 3u);
+  auto msg = builder.to_message();
+  check_eq(msg.types(), make_type_id_list<int32_t, int32_t, int32_t>());
+  check_eq(msg.get_as<int>(0), 1);
+  check_eq(msg.get_as<int>(1), 2);
+  check_eq(msg.get_as<int>(2), 3);
```
Neverlord, line 188:
```suggestion
  if (check_eq(msg.types(), make_type_id_list<int32_t, int32_t, int32_t>())) {
    check_eq(msg.get_as<int32_t>(0), 1);
    check_eq(msg.get_as<int32_t>(1), 2);
    check_eq(msg.get_as<int32_t>(2), 3);
  }
```

libcaf_core/caf/message_builder.test.cpp
```diff
@@ -0,0 +1,191 @@
...
+    auto msg = builder.move_to_message();
+    if (check_eq(msg.types(),
+                 make_type_id_list<std::string, counted_int_ptr>())) {
+      check_eq(msg.get_as<std::string>(0), "foo");
+      const auto& ci = msg.get_as<counted_int_ptr>(1);
+      check_eq(ci->value(), 42);
+      check_eq(ci->get_reference_count(), 1u);
+    }
+  }
+  SECTION("move_to_message copies message appended with append_form") {
```
Neverlord, line 130: This test overall became a bit confusing by adding too many sections to it. At the very least, I'd like to see the `move_to_message` tests move out of this test. Since these tests are a bit more involved, I think they would benefit from structuring them as scenarios:

```
SCENARIO: move_to_message move-constructs message elements
  GIVEN: a builder object
    WHEN: appending primitive values
      THEN: move_to_message copies values
        - builder.append with integers
    WHEN: appending movable values
      THEN: move_to_message moves values into place
        - builder.append with counted_int_ptr
  GIVEN: a builder object and a message
    WHEN: appending from a the message with multiple references
      THEN: move_to_message copies the value from the message
        - msg = make_message(make_counted<counted_int> ...
        - use append_from and hold onto the reference in `msg`
    WHEN: appending from a message with no other references
      THEN: move_to_message moves the value from the message
        - same as above but use append_from(make_message(... to have no extra ref to the message
```

Should make it more obvious what's actually being tested here, don't you think?

Neverlord (changes requested):

libcaf_core/caf/message_builder.test.cpp
shariarriday: Yes, I have updated the restructured tests and added an additional test according to the given scenarios.

Neverlord (approved):

PR #1564, opened 2023-09-19
Verify type IDs when creating messages

### LINKED ISSUE #973: Qt widget mixin segfaults on unexpected messages.
Send it a unexpected message and it get's very unhappy, as soon as I add a placeholder to receive it it's happy again.

Is there special behaviour required here?

Neverlord:
Closes #973.

@riemass can you double-check that there aren't more places where CAF could crash from accessing the meta object table unchecked? We do verify all IDs in `load_data` (`message.cpp`), so this was the only place that looked like it needed the checked access. Also left some comments at the other places where we can safely skip the checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,9 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   positive value from `consume`.
 - When constructing a `behavior` or `message_handler`, callbacks that take a
   `message` argument are now treated as catch-all handlers.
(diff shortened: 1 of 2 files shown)


riemass (approved):

riemass:
I've looked the code up for access to the `global_meta_objects` span, there where a few more uses in `message_data`, `message` and one in `actor_config` but all seem either guarded or guaranteed to be valid at call site.

Issue #1435, opened 2023-06-14
Add clang-tidy to our CI pipeline


Neverlord:
Clang tidy has a lot of checks to choose from. As a start, we should probably use only the `bugprone` subset.

riemass:
This isn't in the bugprone subset, but I'll add https://clang.llvm.org/extra/clang-tidy/checks/google/explicit-constructor.html

PR #1565, opened 2023-09-20
Additional clang-tidy fixes

riemass:
Relates #1435.

After this merge, the only remaining bugprone check is `bugprone-exception-escape` which would take some more time to fix, as discussed before. After that we can enable the clang-tidy github action. 

@Neverlord as you might see one of the checks was related to same body of different branches in if/else conditions. I've fixed most of these by omitting the else if the previous branch returns. The unchecked optional access is mostly nolint since we designed our own expected type. 

Also the `use-after-move` gives some discussion. The standard says that moved from objects are valid, but left in an unspecified state and should have a write operation to make them usable again (write in terms of assigning a new value or calling `clear`). However in some of our implementations we leave the object in a valid state and use it afterwards without resetting the value at move site.

### DIFF
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -44,7 +44,8 @@ void ChatWidget::init(actor_system& system, const std::string& name,
       auto str = std::string_view{reinterpret_cast<const char*>(bytes.data()),
                                   bytes.size()};
       if (std::all_of(str.begin(), str.end(), ::isprint)) {
-        print(QString::fromUtf8(str.data(), str.size()));
+        print(
+          QString::fromUtf8(str.data(), static_cast<qsizetype>(str.size())));
       } else {
         QString msg = "<non-ascii-data of size ";
         msg += QString::number(bytes.size());
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -44,8 +44,9 @@ void ChatWidget::init(actor_system& system, const std::string& name,
       auto str = std::string_view{reinterpret_cast<const char*>(bytes.data()),
                                   bytes.size()};
       if (std::all_of(str.begin(), str.end(), ::isprint)) {
-        print(
-          QString::fromUtf8(str.data(), static_cast<qsizetype>(str.size())));
+        auto qstr = QString::fromUtf8(str.data(),
+                                      static_cast<qsizetype>(str.size()));
+        print(qstr);
       } else {
         QString msg = "<non-ascii-data of size ";
         msg += QString::number(bytes.size());
--- a/libcaf_core/caf/async/future.hpp
+++ b/libcaf_core/caf/async/future.hpp
@@ -62,7 +62,8 @@ public:
       }
     }
     if (fire_immediately)
-      event.first->schedule(std::move(event.second));
+      event.first->schedule(      // NOLINT(bugprone-use-after-move)
+        std::move(event.second)); // NOLINT(bugprone-use-after-move)
     auto res = std::move(cb_action).as_disposable();
     ctx_->watch(res);
     return res;
--- a/libcaf_core/caf/async/future.hpp
+++ b/libcaf_core/caf/async/future.hpp
@@ -62,8 +62,9 @@ public:
       }
     }
     if (fire_immediately)
-      event.first->schedule(      // NOLINT(bugprone-use-after-move)
-        std::move(event.second)); // NOLINT(bugprone-use-after-move)
+      // Note: clang-tidy gets confused by fire_immediately.
+      // NOLINTNEXTLINE(bugprone-use-after-move)
+      event.first->schedule(std::move(event.second));
     auto res = std::move(cb_action).as_disposable();
     ctx_->watch(res);
     return res;
--- a/libcaf_core/caf/async/promise.hpp
+++ b/libcaf_core/caf/async/promise.hpp
@@ -27,8 +27,8 @@ public:
   }
 
   promise& operator=(const promise& other) noexcept {
-    promise copy{other};
-    cell_.swap(copy.cell_);
+    if (this != &other)
+      return *this = promise{other};
     return *this;
   }
 
--- a/libcaf_core/caf/binary_deserializer.cpp
+++ b/libcaf_core/caf/binary_deserializer.cpp
@@ -116,7 +116,7 @@ bool binary_deserializer::begin_field(std::string_view,
                     "received type index out of bounds");
       return false;
     }
-    index = static_cast<unsigned char>(tmp);
+    index = static_cast<size_t>(tmp);
     return true;
   };
   if (types.size() < max_value<int8_t>) {
@@ -146,7 +146,7 @@ bool binary_deserializer::begin_field(std::string_view, bool& is_present,
       return false;
     }
     is_present = true;
-    index = static_cast<unsigned char>(tmp);
+    index = static_cast<size_t>(tmp);
     return true;
   };
   if (types.size() < max_value<int8_t>) {
--- a/libcaf_core/caf/binary_deserializer.cpp
+++ b/libcaf_core/caf/binary_deserializer.cpp
@@ -116,7 +116,7 @@ bool binary_deserializer::begin_field(std::string_view,
                     "received type index out of bounds");
       return false;
     }
-    index = static_cast<size_t>(tmp);
+    index = static_cast<unsigned char>(tmp);
     return true;
   };
   if (types.size() < max_value<int8_t>) {
@@ -146,7 +146,7 @@ bool binary_deserializer::begin_field(std::string_view, bool& is_present,
       return false;
     }
     is_present = true;
-    index = static_cast<size_t>(tmp);
+    index = static_cast<unsigned char>(tmp);
     return true;
   };
   if (types.size() < max_value<int8_t>) {
--- a/libcaf_core/caf/catch_all.hpp
+++ b/libcaf_core/caf/catch_all.hpp
@@ -22,12 +22,11 @@ struct catch_all {
 
   F handler;
 
-  catch_all(const catch_all& x) = default;
-
   catch_all(catch_all&& x) = default;
 
-  template <class T, class = std::enable_if_t<!std::is_same_v<T, catch_all>>>
-  catch_all(T&& x) : handler(std::forward<T>(x)) {
+  template <class T, class = std::enable_if_t<
+                       !std::is_same_v<std::decay_t<T>, catch_all>>>
+  explicit catch_all(T&& x) : handler(std::forward<T>(x)) {
     // nop
   }
 
--- a/libcaf_core/caf/catch_all.hpp
+++ b/libcaf_core/caf/catch_all.hpp
@@ -16,9 +16,9 @@ template <class F>
 struct catch_all {
   using fun_type = std::function<skippable_result(message&)>;
 
-  static_assert(
-    std::is_convertible_v<F, fun_type>,
-    "catch-all handler must have signature skippable_result (message&)");
+  static_assert(std::is_convertible_v<F, fun_type>,
+                "catch-all handler must have signature "
+                "skippable_result (message&)");
 
   F handler;
 
--- a/libcaf_core/caf/catch_all.hpp
+++ b/libcaf_core/caf/catch_all.hpp
@@ -16,17 +16,17 @@ template <class F>
 struct catch_all {
   using fun_type = std::function<skippable_result(message&)>;
 
-  static_assert(std::is_convertible_v<F, fun_type>,
-                "catch-all handler must have signature "
-                "skippable_result (message&)");
+  static_assert(
+    std::is_convertible_v<F, fun_type>,
+    "catch-all handler must have signature skippable_result (message&)");
 
   F handler;
 
-  catch_all(catch_all&& x) : handler(std::move(x.handler)) {
-    // nop
-  }
+  catch_all(const catch_all& x) = default;
 
-  template <class T>
+  catch_all(catch_all&& x) = default;
+
+  template <class T, class = std::enable_if_t<!std::is_same_v<T, catch_all>>>
   catch_all(T&& x) : handler(std::forward<T>(x)) {
     // nop
   }
--- a/libcaf_core/caf/chrono.cpp
+++ b/libcaf_core/caf/chrono.cpp
@@ -46,7 +46,7 @@ time_t tm_to_time_t(tm& time_buf) noexcept {
 
 namespace {
 
-int utc_offset(const tm& time_buf) noexcept {
+long int utc_offset(const tm& time_buf) noexcept {
   return time_buf.tm_gmtoff;
 }
 
--- a/libcaf_core/caf/chrono.cpp
+++ b/libcaf_core/caf/chrono.cpp
@@ -46,7 +46,7 @@ time_t tm_to_time_t(tm& time_buf) noexcept {
 
 namespace {
 
-long int utc_offset(const tm& time_buf) noexcept {
+auto utc_offset(const tm& time_buf) noexcept {
   return time_buf.tm_gmtoff;
 }
 
--- a/libcaf_core/caf/detail/sync_ring_buffer.test.cpp
+++ b/libcaf_core/caf/detail/sync_ring_buffer.test.cpp
@@ -29,7 +29,7 @@ std::vector<int> consumer(int_buffer& buf, size_t num) {
 }
 
 void producer(int_buffer& buf, int first, int last) {
-  for (auto i = first; i != last; ++i)
+  for (auto i = first; i != last; ++i) // NOLINT(bugprone-use-after-move)
     buf.push_back(std::move(i));
 }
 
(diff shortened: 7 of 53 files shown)


libcaf_core/caf/catch_all.hpp
```diff
@@ -16,17 +16,17 @@ template <class F>
...
+  static_assert(
+    std::is_convertible_v<F, fun_type>,
+    "catch-all handler must have signature skippable_result (message&)");
 
   F handler;
 
-  catch_all(catch_all&& x) : handler(std::move(x.handler)) {
-    // nop
-  }
+  catch_all(const catch_all& x) = default;
```
Neverlord, line 25: Why did you add a copy constructor? `F` might be move-only type.

libcaf_core/caf/catch_all.hpp
```diff
@@ -16,17 +16,17 @@ template <class F>
...
+
+  catch_all(catch_all&& x) = default;
 
-  template <class T>
+  template <class T, class = std::enable_if_t<!std::is_same_v<T, catch_all>>>
   catch_all(T&& x) : handler(std::forward<T>(x)) {
```
Neverlord, line 30: Shouldn't this be `decay_t<T>`? Also... can we still make this explicit without breaking something?

```suggestion
  template <class T, class = std::enable_if_t<!std::is_same_v<T, catch_all>>>
  catch_all(T&& x) : handler(std::forward<T>(x)) {
```

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
```diff
@@ -29,7 +29,7 @@ std::vector<int> consumer(int_buffer& buf, size_t num) {
 }
 
 void producer(int_buffer& buf, int first, int last) {
-  for (auto i = first; i != last; ++i)
+  for (auto i = first; i != last; ++i) // NOLINT(bugprone-use-after-move)
     buf.push_back(std::move(i));
```
Neverlord, line 33: These are integers, there's no point in moving them anyways.

```suggestion
  for (auto i = first; i != last; ++i)
    buf.push_back(i);
```

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
```diff
@@ -56,6 +56,7 @@ TEST("push_back adds one element to the ring buffer") {
   check(!buf.full());
   check_eq(buf.size(), 0u);
   print_debug("fill buffer");
+  // NOLINTNEXTLINE(bugprone-use-after-move)
   for (int i = 0; i < static_cast<int>(int_buffer_size - 1); ++i)
     buf.push_back(std::move(i));
```
Neverlord, line 61: Same here

```suggestion
  for (int i = 0; i < static_cast<int>(int_buffer_size - 1); ++i)
    buf.push_back(i);
```

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
```diff
@@ -75,7 +76,7 @@ TEST("get_all returns all elements from the ring buffer") {
     return vector_type(i, e);
   };
   print_debug("add five element");
-  for (int i = 0; i < 5; ++i)
+  for (int i = 0; i < 5; ++i) // NOLINT(bugprone-use-after-move)
     buf.push_back(std::move(i));
```
Neverlord, line 80:
```suggestion
  for (int i = 0; i < 5; ++i)
    buf.push_back(i);
```

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
```diff
@@ -88,7 +89,7 @@ TEST("get_all returns all elements from the ring buffer") {
   check_eq(buf.size(), 0u);
   print_debug("add 60 elements (wraps around)");
   vector_type expected;
-  for (int i = 0; i < 60; ++i) {
+  for (int i = 0; i < 60; ++i) { // NOLINT(bugprone-use-after-move)
```
Neverlord, line 92:
```suggestion
  for (int i = 0; i < 60; ++i) {
```

libcaf_core/caf/hash/fnv.hpp
```diff
@@ -144,7 +144,7 @@ class fnv : public save_inspector_base<fnv<T>> {
   /// Convenience function for computing an FNV1a hash value for given
   /// arguments in one shot.
   template <class... Ts>
-  static T compute(Ts&&... xs) noexcept {
+  static T compute(Ts&&... xs) {
```
Neverlord, line 147: What was clang-tidy complaining about? Can't we just make more things noexcept? I'd like to keep this `noexcept`, because there this actually cannot throw.

libcaf_core/caf/node_id.hpp
```diff
@@ -288,7 +288,7 @@ namespace std {
 
 template <>
 struct hash<caf::node_id> {
-  size_t operator()(const caf::node_id& x) const noexcept {
```
Neverlord, line 291: Also should stay noexcept.

libcaf_core/caf/string_algorithms.cpp
```diff
@@ -94,11 +94,12 @@ void replace_all(std::string& str, std::string_view what,
   };
   auto i = next(str.begin());
   while (i != str.end()) {
-    auto before = static_cast<size_t>(std::distance(str.begin(), i));
-    str.replace(i, i + what.size(), with.begin(), with.end());
+    auto before = std::distance(str.begin(), i);
+    str.replace(i, i + static_cast<ptrdiff_t>(what.size()), with.begin(),
```
Neverlord, line 98: We do this cast twice now, so I'd store the result of the cast in a variable.

libcaf_core/caf/unit.hpp
```diff
@@ -20,7 +20,7 @@ struct unit_t : detail::comparable<unit_t> {
 
   constexpr unit_t& operator=(const unit_t&) noexcept = default;
 
-  template <class T>
+  template <class T, class = std::enable_if_t<!std::is_same_v<T, unit_t>>>
```
Neverlord, line 23:
```suggestion
  template <class T, class = std::enable_if_t<!std::is_same_v<std::decay_t<T>, unit_t>>>
```

libcaf_core/caf/uri.hpp
```diff
@@ -292,7 +292,7 @@ namespace std {
 
 template <>
 struct hash<caf::uri> {
-  size_t operator()(const caf::uri& x) const noexcept {
```
Neverlord, line 295: Should stay noexcept.

libcaf_core/tests/legacy/detail/parser/read_number.cpp
```diff
@@ -56,7 +56,9 @@ struct range_consumer {
 
 struct res_t {
   std::variant<pec, double, int64_t> val;
-  template <class T>
+  res_t(const res_t&) = default;
+  res_t(res_t&&) = default;
+  template <class T, class E = std::enable_if_t<!std::is_same_v<T, res_t>>>
```
Neverlord, line 61:
```suggestion
  template <class T, class E = std::enable_if_t<!std::is_same_v<std::decay_t<T>, res_t>>>
```

Also, can we mark this `explicit` or would this break stuff?

libcaf_core/tests/legacy/detail/parser/read_number_or_timespan.cpp
```diff
@@ -29,8 +29,10 @@ struct number_or_timespan_parser_consumer {
 
 struct res_t {
   std::variant<pec, double, int64_t, timespan> val;
-  template <class T>
-  explicit res_t(T&& x) : val(std::forward<T>(x)) {
+  res_t(const res_t&) = default;
+  res_t(res_t&&) = default;
+  template <class T, class E = std::enable_if_t<!std::is_same_v<T, res_t>>>
+  res_t(T&& x) : val(std::forward<T>(x)) {
```
Neverlord, line 35:
```suggestion
  template <class T, class E = std::enable_if_t<!std::is_same_v<std::decay<T>, res_t>>>
  res_t(T&& x) : val(std::forward<T>(x)) {
```

`explicit`?

libcaf_net/caf/net/http/server_factory.hpp
```diff
@@ -219,7 +219,7 @@ class server_factory
     routes.push_back(make_route([producer](responder& res) {
       if (!producer->push(std::move(res).to_request())) {
         auto err = make_error(sec::runtime_error, "flow disconnected");
-        res.router()->shutdown(err);
+        res.router()->shutdown(err); // NOLINT(bugprone-use-after-move)
```
Neverlord, line 222: Can't we drop the move above? This actually looks suspicious to me.

Neverlord (changes requested):

libcaf_core/caf/async/future.hpp
```diff
@@ -62,7 +62,8 @@ class bound_future {
       }
     }
     if (fire_immediately)
-      event.first->schedule(std::move(event.second));
+      event.first->schedule(      // NOLINT(bugprone-use-after-move)
+        std::move(event.second)); // NOLINT(bugprone-use-after-move)
```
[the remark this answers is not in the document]
riemass: It's the conditional move you mentioned that triggers the warning. clang-tidy doesn't pick up the fact that, considering the `fire_immediately` boolean, this code will never be run if we moved before.

libcaf_core/caf/binary_deserializer.cpp
```diff
@@ -116,7 +116,7 @@ bool binary_deserializer::begin_field(std::string_view,
                     "received type index out of bounds");
       return false;
     }
-    index = static_cast<size_t>(tmp);
+    index = static_cast<unsigned char>(tmp);
```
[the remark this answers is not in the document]
riemass: Here is the line and the related doc:
```
warning: 'signed char' to 'size_t' (aka 'unsigned long') conversion; consider casting to 'unsigned char' first. [bugprone-signed-char-misuse]
```
Explanation found on [link](https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-signed-char-misuse.html).

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
riemass: That was my first thought as well but `sync_ring_buffer` only has a rvalue `push_back` so this wouldn't compile...

libcaf_core/caf/hash/fnv.hpp
riemass: Looking at the code of this function, `f.apply` is not `noexcept` so this is warning us about `bugprone-exception-escape`. This check probably generates the most warnings in the whole code base.

libcaf_core/caf/node_id.hpp
riemass: Since this is calling the above function, `fnv::compute` which can throw because of `f.apply`, we get a warning here about an exception being thrown.

libcaf_core/tests/legacy/detail/parser/read_number.cpp
riemass: This one I remember trying with explicit and it breaks stuff. Since it's only used in this local file I might go and name the constructors.

libcaf_net/caf/net/http/server_factory.hpp
riemass: We depend on the contract that "If the `producer::push` fails it won't break the rvalue argument". It looks like something hard to enforce, but I left the code as is.

libcaf_core/caf/detail/sync_ring_buffer.test.cpp
Neverlord: Then why not just give it a temporary when it asks for it via `buf.push_back(int{i});`?

libcaf_core/caf/binary_deserializer.cpp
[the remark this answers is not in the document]
Neverlord: That warning only makes sense for the `int8_t` case. In all other cases, this will truncate the index.

libcaf_core/caf/async/promise.hpp
```diff
@@ -27,8 +27,8 @@ class promise {
   }
 
   promise& operator=(const promise& other) noexcept {
-    promise copy{other};
-    cell_.swap(copy.cell_);
+    if (this != &other)
+      return *this = promise{other};
```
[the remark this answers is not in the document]
riemass: The default constructor for `promise` increments the number of promises by 1, and omitting it here breaks the unit tests, so we can't just do it like suggested. Either keep the copy-swap or add the missing line for `fetch_add`, and I like this approach more because of DRY.

libcaf_core/caf/catch_all.hpp
riemass: Fixed.

libcaf_core/caf/node_id.hpp
riemass: I'll return the noexcept and ignore clang-tidy, for the best.

Neverlord: `fnv` should be fully noexcept. `clang-tidy` shouldn't complain in the first place. :)

libcaf_core/caf/binary_deserializer.cpp
```diff
@@ -146,7 +146,7 @@ bool binary_deserializer::begin_field(std::string_view, bool& is_present,
       return false;
     }
     is_present = true;
-    index = static_cast<size_t>(tmp);
+    index = static_cast<unsigned char>(tmp);
```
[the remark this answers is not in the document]
riemass: Looking at the code below, this is probably a false positive.

libcaf_core/caf/hash/fnv.hpp
riemass: Returning noexcept.

libcaf_core/caf/string_algorithms.cpp
riemass: First static cast is `what.size()` and the second `with.size()` if that's what you meant?

libcaf_core/caf/chrono.cpp
```diff
@@ -46,7 +46,7 @@ time_t tm_to_time_t(tm& time_buf) noexcept {
 
 namespace {
 
-int utc_offset(const tm& time_buf) noexcept {
+long int utc_offset(const tm& time_buf) noexcept {
```
Neverlord, line 49:
```suggestion
auto utc_offset(const tm& time_buf) noexcept {
```

libcaf_core/caf/async/future.hpp
Neverlord, line 66:
```suggestion
      // Note: clang-tidy gets confused by fire_immediately.
      // NOLINTNEXTLINE(bugprone-use-after-move)
      event.first->schedule(std::move(event.second));
```

examples/qtsupport/chatwidget.cpp
```diff
@@ -44,7 +44,8 @@ void ChatWidget::init(actor_system& system, const std::string& name,
       auto str = std::string_view{reinterpret_cast<const char*>(bytes.data()),
                                   bytes.size()};
       if (std::all_of(str.begin(), str.end(), ::isprint)) {
-        print(QString::fromUtf8(str.data(), str.size()));
+        print(
+          QString::fromUtf8(str.data(), static_cast<qsizetype>(str.size())));
```
Neverlord, line 48:
```suggestion
        auto qstr = QString::fromUtf8(str.data(), static_cast<qsizetype>(str.size()));
        print(qstr);
```

libcaf_core/caf/catch_all.hpp
```diff
@@ -16,18 +16,17 @@ template <class F>
 struct catch_all {
   using fun_type = std::function<skippable_result(message&)>;
 
-  static_assert(std::is_convertible_v<F, fun_type>,
-                "catch-all handler must have signature "
-                "skippable_result (message&)");
+  static_assert(
+    std::is_convertible_v<F, fun_type>,
+    "catch-all handler must have signature skippable_result (message&)");
```
Neverlord, line 21: Why did you change the formatting?

libcaf_core/caf/inspector_access_type.hpp
```diff
@@ -56,7 +56,7 @@ constexpr auto inspect_access_type() {
   //        > inspector_access > inspect > trivial > defaults.
   if constexpr (is_allowed_unsafe_message_type_v<T>)
     return inspector_access_type::unsafe{};
-  else if constexpr (std::is_array<T>::value)
+  else if constexpr (std::is_array<T>::value || is_stl_tuple_type_v<T>)
```
Neverlord, line 59: This is a breaking change. Builtin-inspects must have precedence over the default implementation.

Neverlord (changes requested):
There's quite a bit going on here and changes like that cast and the re-order in the inspector meta programming code would actually result in bugs / change in behavior. Can we please break up this PR?

There's unused `this` pointers in lambdas, the noexcept stuff, the move stuff and some re-ordering of if statements. The latter requires a lot more care than the other things (the new casts also require a lot of care).

The `this` cleanup and the suppression of some of the `move` uses are trivial, so I'd like to get them out of the way. The `noexcept` is mostly straightforward but produces a large diff, so pulling this out would also make this much easier to review.

libcaf_core/caf/catch_all.hpp
riemass: Must've been a mistake and clang-tidy didn't return it the way it was before.

libcaf_core/caf/inspector_access_type.hpp
riemass: I've disabled the `bugprone-branch-clone` in the end.

riemass:
Pulled out everything into separate PRs, closing this.

PR #1568, opened 2023-09-22
Manually set system openssl in windows runner

riemass:
Test setting manually the OpenSSL path. The path has changed slightly since last time we used this method.


Neverlord (approved):
Thanks for the quick fix! 🙏

Issue #1444, opened 2023-06-17
Overload `dump_content` in the examples


Neverlord:
Our examples should promote best practices. One such best practice is to overload `dump_content()` when adding config options with default values. This will allow users to discover hard-coded default values by calling an application with `--dump-config`.

I did this for one example to show what I mean:

```diff
diff --git a/examples/flow/iota.cpp b/examples/flow/iota.cpp
index 6652f7809..f4982a236 100644
--- a/examples/flow/iota.cpp
+++ b/examples/flow/iota.cpp
@@ -9,18 +9,25 @@
 
 namespace {
 
+constexpr size_t default_num_values = 10;
+
 struct config : caf::actor_system_config {
   config() {
     opt_group{custom_options_, "global"} //
-      .add(n, "num-values,n", "number of values produced by the source");
+      .add<size_t>("num-values,n", "number of values produced by the source");
   }
 
-  size_t n = 10;
+  caf::settings dump_content() const override {
+    auto result = actor_system_config::dump_content();
+    caf::put_missing(result, "num-values", default_num_values);
+    return result;
+  }
 };
 
 // --(rst-main-begin)--
 void caf_main(caf::actor_system& sys, const config& cfg) {
-  sys.spawn([n = cfg.n](caf::event_based_actor* self) {
+  auto n = get_or(cfg, "num-values", default_num_values);
+  sys.spawn([n](caf::event_based_actor* self) {
     self
       // Get an observable factory.
       ->make_observable()
```

We need to go through all of the examples to make this consistent.

PR #1574, opened 2023-10-04
Overload dump_content for examples

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/1444

Updated examples with `dump_content` to ensure `--dump-config` prints all of the default config values.

### DIFF
--- a/examples/web_socket/stock-ticker.cpp
+++ b/examples/web_socket/stock-ticker.cpp
@@ -28,7 +28,7 @@ static constexpr uint16_t default_port = 8080;
 
 static constexpr size_t default_max_connections = 128;
 
-static constexpr caf::timespan default_interval = caf::timespan{1s};
+static constexpr caf::timespan default_interval = 1s;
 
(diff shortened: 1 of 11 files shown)


examples/web_socket/stock-ticker.cpp
```diff
@@ -28,6 +28,8 @@ static constexpr uint16_t default_port = 8080;
 
 static constexpr size_t default_max_connections = 128;
 
+static constexpr caf::timespan default_interval = caf::timespan{1s};
```
Neverlord, line 31: Shouldn't it work without the type repetition?

```suggestion
static constexpr caf::timespan default_interval = 1s;
```

Neverlord (changes requested):
Looking good so far, just one nitpick. 👍

examples/web_socket/stock-ticker.cpp
shariarriday: Thanks, this has been updated in the PR.

Issue #1573, opened 2023-10-03
Add File Data Sources for Flows


Neverlord:
With many building blocks for the flow API in place now, we also need a few data sources. The most obvious one is text files.

A simple pattern for this would be a detached actor reading from a file. The file content is then passed to a user-defined step function, character by character. API could look something like this:

```c++
// Read from a file, convert the content to all-uppercase.
auto src = caf::flow::from_text_file(sys, "/my/file/path").compose([](auto&& chars) {
  return std::move(chars).map([](char ch) { return static_cast<char>(toupper(ch)); });
});
// src is a caf::async::publisher<char>
```

Of course we should also add a couple utilities to do the usual stuff like collecting characters into a string and then emit line by line.

Simply reading a file line-by-line as flow should be as simple as this:

```c++
// Read from a file, convert the content to all-uppercase.
auto src = caf::flow::from_text_file(sys, "/my/file/path").lines();
// src is a caf::async::publisher<caf::cow_string>
```

Feedback most welcome, just thinking loudly. 🙂

Neverlord:
Reading text files is in, we can consider adding things like binary files plus writing to files.

Neverlord:
First set of sources is available. Sinks will be a different topic.

PR #1577, opened 2023-10-05
Make to_stream available on observable

Neverlord:
This makes the API more intuitive to use for users, but the main reason is actually #1573: I've played a bit with the design and the current iteration spins up an (inactive) actor, creates an `observable` on it and then hands that to the user for further setup. Since we don't want to hand the `self` pointer to users, the only option is calling `to_stream` directly on the `observable` in case the user wants to consume items from a stream (the alternative is producing a `publisher`).


riemass (approved):
LGTM. Only one nit about coverage below.

PR #1581, opened 2023-10-05
Fix clang-tidy implicit move constructor

riemass:
Clang-tidy complained that having a constructor accepting a forward reference may act as a move constructor. 
Pulled out of #1565.

### DIFF
--- a/libcaf_core/caf/others.hpp
+++ b/libcaf_core/caf/others.hpp
@@ -18,7 +18,7 @@ struct others_t {
 
   template <class F>
   catch_all<F> operator>>(F fun) const {
-    return {fun};
+    return catch_all<F>{fun};
   }
 };
 
--- a/libcaf_core/caf/others.hpp
+++ b/libcaf_core/caf/others.hpp
@@ -17,7 +17,7 @@ struct others_t {
   }
 
   template <class F>
-  catch_all<F> operator>>(F fun) const {
+  auto operator>>(F fun) const {
     return catch_all<F>{fun};
   }
 };
--- a/libcaf_core/tests/legacy/detail/parser/read_number.cpp
+++ b/libcaf_core/tests/legacy/detail/parser/read_number.cpp
@@ -56,10 +56,6 @@ struct range_consumer {
 
 struct res_t {
   std::variant<pec, double, int64_t> val;
-  template <class T>
-  explicit res_t(T x) noexcept : val(x) {
-    // nop
-  }
 };
 
 std::string to_string(const res_t& x) {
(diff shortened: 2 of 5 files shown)


libcaf_core/caf/others.hpp
```diff
@@ -18,7 +18,7 @@ struct others_t {
 
   template <class F>
   catch_all<F> operator>>(F fun) const {
```
Neverlord, line 20:
```suggestion
  auto operator>>(F fun) const {
```

libcaf_core/tests/legacy/detail/parser/read_number.cpp
```diff
@@ -56,8 +56,11 @@ struct range_consumer {
 
 struct res_t {
   std::variant<pec, double, int64_t> val;
-  template <class T>
-  res_t(T&& x) : val(std::forward<T>(x)) {
+  res_t(const res_t&) = default;
+  res_t(res_t&&) = default;
+  template <class T,
+            class E = std::enable_if_t<!std::is_same_v<std::decay_t<T>, res_t>>>
+  explicit res_t(T&& x) : val(std::forward<T>(x)) {
```
Neverlord, line 63: Actually, all of these are primitive types. Can't we do it like this?

```suggestion
  res_t(const res_t&) noexcept = default;
  
  template <class T,
            class E = std::enable_if_t<!std::is_same_v<T, res_t>>>
  explicit res_t(T x) noexcept : val(x) {
```

libcaf_core/tests/legacy/detail/parser/read_number_or_timespan.cpp
```diff
@@ -29,7 +29,10 @@ struct number_or_timespan_parser_consumer {
 
 struct res_t {
   std::variant<pec, double, int64_t, timespan> val;
-  template <class T>
+  res_t(const res_t&) = default;
+  res_t(res_t&&) = default;
+  template <class T,
+            class E = std::enable_if_t<!std::is_same_v<std::decay_t<T>, res_t>>>
```
Neverlord, line 35: Same as the other `res_t`: primitive types don't need all of this move/forward complexity.

Neverlord (changes requested):

Neverlord (changes requested):
Actually, the more I look at these `res_t` things... I think we shouldn't define constructors at all:

```c++
struct res_t {
  std::variant<...> val;
};
```

We can just use member-initialization syntax, e.g., `res_t{{42}}`.

Neverlord (approved):

PR #1586, opened 2023-10-06
Implement basic file I/O source interface

Neverlord:
Relates #1573.

### DIFF
--- a/libcaf_core/caf/async/file.hpp
+++ b/libcaf_core/caf/async/file.hpp
@@ -1,0 +1,170 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/actor_system.hpp"
+#include "caf/detail/type_traits.hpp"
+#include "caf/event_based_actor.hpp"
+#include "caf/flow/string.hpp"
+#include "caf/scheduled_actor/flow.hpp"
+
+#include <cstdio>
+
+namespace caf::detail {
+
+/// A generator that emits characters from a file.
+class file_reader {
+public:
+  using output_type = char;
+
+  explicit file_reader(std::string path) : path_(std::move(path)) {
+    // nop
+  }
+
+  file_reader(file_reader&& other) noexcept {
+    swap(other);
+  }
+
+  file_reader& operator=(file_reader&& other) noexcept {
+    if (this != &other)
+      swap(other);
+    return *this;
+  }
+
+  file_reader(const file_reader& other) : path_(other.path_) {
+    // Note: intentionally don't copy the file handle.
+  }
+
+  file_reader& operator=(const file_reader& other) {
+    path_ = other.path_;
+    if (file_ != nullptr) {
+      fclose(file_);
+      file_ = nullptr;
+    }
+    return *this;
+  }
+
+  ~file_reader() {
+    if (file_ != nullptr)
+      fclose(file_);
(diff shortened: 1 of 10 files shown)


libcaf_core/caf/async/fwd.hpp
```diff
@@ -39,4 +41,15 @@ class spsc_buffer;
...
+
+template <class T>
+struct is_publisher : std::false_type {};
+
+template <class T>
+struct is_publisher<publisher<T>> : std::true_type {};
+
+template <class T>
+inline constexpr bool is_publisher_v = is_publisher<T>::value;
+
```
riemass, line 54: Not fond of putting implementation in fwd headers, just for header hygiene. There's one more comment below.

libcaf_core/caf/async/file.hpp
```diff
@@ -0,0 +1,150 @@
...
+};
+
+/// Bundles factories for asynchronous sources and sinks that operate on files.
+class file {
+public:
+  explicit file(actor_system& sys) : sys_(&sys) {
+    // nop
+  }
+
+  auto read_chars(std::string path) {
```
riemass, line 126: Like discussed, we can move the `path` argument to the main constructor. Applies to `read_lines` as well.

libcaf_core/caf/async/file.hpp
```diff
@@ -0,0 +1,150 @@
...
+public:
+  source_runner(actor_system* sys, Factory gen)
+    : sys_(sys), gen_(std::move(gen)) {
+    // nop
+  }
+
+  template <class F>
+  auto run(F&& init) && {
+    auto [self, launch] = sys_->spawn_inactive<event_based_actor, detached>();
+    auto res = init(gen_(self));
```
riemass, line 101: Just a discussion question: would forwarding have any benefit here since you took `init` as an universal reference?  I sometimes see code doing `std::forward<F>(init)(...)` but I always take it with a grain of salt.

libcaf_core/caf/async/file.hpp
```diff
@@ -0,0 +1,150 @@
...
+      return std::move(src).to_publisher();
+    });
+  }
+
+private:
+  actor_system* sys_;
+  Factory gen_;
+};
+
+/// Bundles factories for asynchronous sources and sinks that operate on files.
```
riemass, line 119: Since this is the the entry point, I would suggest adding an extra note stating that this is a convenience API that wraps file IO around a blocking actor, meaning it spawns a new thread.

riemass (changes requested):
Neat feature. I can definitely see it's place in processing pipelines. Left a few comments.

libcaf_core/caf/async/file.hpp
Neverlord: Added.

libcaf_core/caf/async/file.hpp
Neverlord: It would make a difference if you would pass an `F` with multiple `operator()` overloads like this:

```c++
  auto operator()() const& { ... }
  auto operator()() && { ... }
```

Fixed.

libcaf_core/caf/async/fwd.hpp
```diff
@@ -6,6 +6,8 @@
 
 #include "caf/fwd.hpp"
 
+#include <type_traits>
+
```
riemass, line 10: We can move this to `detail/type_traits.hpp` as well.

riemass (changes requested):
Nagging about macro hygiene.

riemass (approved):
Great :+1:

PR #1588, opened 2023-10-06
Add on_error_return_item step along with tests

shariarriday:

Added a new step for observable on_error_return_item.

### DIFF
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -330,6 +330,11 @@ public:
     return add_step(step::on_error_complete<output_type>{});
   }
 
+  auto on_error_return_item(output_type&& f) {
+    return add_step(
+      step::on_error_return_item<output_type>{std::forward<output_type>(f)});
+  }
+
   /// Materializes the @ref observable.
   observable<output_type> as_observable() && {
     return materialize();
@@ -633,6 +638,12 @@ transformation<step::map<F>> observable<T>::map(F f) {
 template <class T>
 transformation<step::on_error_complete<T>> observable<T>::on_error_complete() {
   return transform(step::on_error_complete<T>{});
+}
+
+template <class T>
+transformation<step::on_error_return_item<T>>
+observable<T>::on_error_return_item(T&& t) {
+  return transform(step::on_error_return_item<T>{std::forward<T>(t)});
 }
 
 template <class T>
--- a/libcaf_core/caf/flow/observable.test.cpp
+++ b/libcaf_core/caf/flow/observable.test.cpp
@@ -249,6 +249,65 @@ TEST("on_error_complete() suppresses errors") {
   }
 }
 
+TEST("(on_error_return_item() returns item on error") {
+  SECTION("on_error_return_item() does nothing if no error occurs") {
+    SECTION("blueprint") {
+      check_eq(collect(range(1, 1).on_error_return_item(42)), vector{1});
+      check_eq(collect(range(1, 2).on_error_return_item(42)), vector{1, 2});
+      check_eq(collect(range(1, 3).on_error_return_item(42)), vector{1, 2, 3});
+      check_eq(collect(range(1, 0).on_error_return_item(42)), nil);
+    }
+    SECTION("observable") {
+      check_eq(collect(mat(range(1, 1)).on_error_return_item(42)), vector{1});
+      check_eq(collect(mat(range(1, 2)).on_error_return_item(42)),
+               vector{1, 2});
+      check_eq(collect(mat(range(1, 3)).on_error_return_item(42)),
+               vector{1, 2, 3});
+      check_eq(collect(mat(range(1, 0)).on_error_return_item(42)), nil);
+    }
+  }
+  SECTION("on_error_return_item() returns item when an error occurs") {
+    SECTION("blueprint") {
+      check_eq(collect(obs_error(sec::runtime_error)
+                         .concat(range(1, 1))
+                         .on_error_return_item(42)),
+               vector{42});
+      check_eq(collect(obs_error(sec::runtime_error)
+                         .concat(range(1, 2))
+                         .on_error_return_item(42)),
+               vector{42});
+      check_eq(collect(range(1, 2)
+                         .concat(obs_error(sec::runtime_error))
+                         .concat(range(1, 2))
+                         .on_error_return_item(42)),
+               vector{1, 2, 42});
+      check_eq(collect(range(1, 3)
+                         .concat(obs_error(sec::runtime_error))
+                         .concat(range(1, 3))
+                         .on_error_return_item(42)),
+               vector{1, 2, 3, 42});
+    }
+    SECTION("observable") {
+      check_eq(collect(mat(obs_error(sec::runtime_error).concat(range(1, 1)))
+                         .on_error_return_item(42)),
+               vector{42});
+      check_eq(collect(mat(obs_error(sec::runtime_error).concat(range(1, 2)))
+                         .on_error_return_item(42)),
+               vector{42});
+      check_eq(collect(mat(range(1, 2)
+                             .concat(obs_error(sec::runtime_error))
+                             .concat(range(1, 2)))
+                         .on_error_return_item(42)),
+               vector{1, 2, 42});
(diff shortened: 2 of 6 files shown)


libcaf_core/caf/flow/step/on_error_return_item.hpp
```diff
@@ -0,0 +1,47 @@
...
+  void on_complete(Next& next, Steps&... steps) {
+    next.on_complete(steps...);
+  }
+
+  template <class Next, class... Steps>
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    if (next.on_next(item_, steps...))
+      next.on_complete(steps...);
+    else
+      next.on_error(what, steps...);
```
Neverlord, line 40: If `next.on_next` returns false, we must not call anything on `next` again.

```suggestion
```

libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -249,6 +249,83 @@ TEST("on_error_complete() suppresses errors") {
...
+      check_eq(collect(mat(range(1, 2).concat(obs_error(sec::runtime_error)))
+                         .on_error_return_item(42)
+                         .take(2)),
+               make_error(sec::runtime_error));
+      check_eq(collect(mat(range(1, 3).concat(obs_error(sec::runtime_error)))
+                         .on_error_return_item(42)
+                         .take(3)),
+               make_error(sec::runtime_error));
+    }
+  }
```
Neverlord, line 326: The sequence must always behave the same, whether or not we cut it short. Merge the tests into the sections above (and they shouldn't result in an error).

libcaf_core/caf/flow/observable.hpp
```diff
@@ -330,6 +330,11 @@ class observable_def {
     return add_step(step::on_error_complete<output_type>{});
   }
 
+  auto on_error_return_item(output_type&& f) {
+    return add_step(
+      step::on_error_return_item<output_type>{std::forward<output_type>(f)});
+  }
+
```
Neverlord, line 337: Could it be that you mistook this for a universal ref? The `output_type` is a known type here, so this will always be an rvalue ref. Please just pass it by value and then move into place.

```suggestion
  auto on_error_return_item(output_type item) {
    return add_step(
      step::on_error_return_item<output_type>{std::move(item)});
  }

```

libcaf_core/caf/flow/observable.hpp
```diff
@@ -635,6 +640,12 @@ transformation<step::on_error_complete<T>> observable<T>::on_error_complete() {
   return transform(step::on_error_complete<T>{});
 }
 
+template <class T>
+transformation<step::on_error_return_item<T>>
+observable<T>::on_error_return_item(T&& t) {
+  return transform(step::on_error_return_item<T>{std::forward<T>(t)});
+}
```
Neverlord, line 647: Same here.

```suggestion
transformation<step::on_error_return_item<T>>
observable<T>::on_error_return_item(T item) {
  return transform(step::on_error_return_item<T>{std::move(item)});
}
```

(plus adjusting the definition)

libcaf_core/caf/flow/step/on_error_return_item.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace caf::flow::step {
+
+template <class T>
+class on_error_return_item {
+public:
+  using input_type = T;
+
+  using output_type = T;
+
+  explicit on_error_return_item(T&& item) : item_(std::forward<T>(item)) {
```
Neverlord, line 21: This is an rvalue ref, calling `forward` on it makes little sense. The common pattern is pass by value and then move into place.

```suggestion
  explicit on_error_return_item(T item) : item_(std::move(item)) {
```

libcaf_core/caf/flow/step/on_error_return_item.hpp
```diff
@@ -0,0 +1,47 @@
...
+    // nop
+  }
+
```
Neverlord, line 24: This class should re-enable the defaults for copy & move assign/construct.

Neverlord (changes requested):

libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -249,6 +249,83 @@ TEST("on_error_complete() suppresses errors") {
...
+               vector{1, 2});
+      check_eq(collect(mat(range(1, 3)).on_error_return_item(42)),
+               vector{1, 2, 3});
+      check_eq(collect(mat(range(1, 0)).on_error_return_item(42)), nil);
+    }
+  }
+  SECTION("on_error_return_item() returns item when an error occurs") {
+    SECTION("blueprint") {
+      check_eq(collect(range(1, 0)
+                         .concat(obs_error(sec::runtime_error))
```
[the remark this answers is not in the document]
shariarriday: I had to use the 0 size range to ensure that the value is returned properly. The `fail` operator does not return items and an empty vector is returned when `on_complete` is called. If this is not the expected behavior, I think that we need to update the `fail` operator accordingly. https://github.com/actor-framework/actor-framework/blob/93d522fc51110c77320c27f452ab7a3417a80aed/libcaf_core/caf/flow/op/fail.hpp#L30-L33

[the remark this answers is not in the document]
Neverlord: > I had to use the 0 size range to ensure that the value is returned properly.

Now I'm confused. What value? `range(1, 0).concat(obs_error())` should be identical to just `obs_error()`, no? `range(1, 0)` should call `on_complete()` immediately (since we asked it to produce 0 values), then, `concat(...)` will switch to the next observable, which calls `on_error()` immediately. End result should be an observable that simply calls `on_error()`... which is exactly what just using `obs_error()` alone achieves.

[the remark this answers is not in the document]
shariarriday: `range(1, 0).concat(obs_error())` is not completely identical to `obs_error()` in terms of output. Though `range(1, 0)` calls on_complete immediately, `from_generator.hpp` calls `out_.on_next()` in it's `do_run` method which ultimately calls `from_steps_sub::on_next()`. If this function is not called, `from_steps_sub::do_run()` seems to return an empty array even though the `buf_` is not empty. I think that this maybe a bug with how `fail` operator is created. Shall I attempt to fix this in the PR as well?

libcaf_core/caf/flow/observable.hpp
```diff
@@ -330,6 +330,10 @@ class observable_def {
     return add_step(step::on_error_complete<output_type>{});
   }
 
+  auto on_error_return_item(output_type f) {
+    return add_step(step::on_error_return_item<output_type>{std::move(f)});
```
Neverlord, line 334:
```suggestion
  auto on_error_return_item(output_type item) {
    return add_step(step::on_error_return_item<output_type>{std::move(item)});
```

libcaf_core/caf/flow/observable.hpp
```diff
@@ -635,6 +639,12 @@ transformation<step::on_error_complete<T>> observable<T>::on_error_complete() {
   return transform(step::on_error_complete<T>{});
 }
 
+template <class T>
+transformation<step::on_error_return_item<T>>
+observable<T>::on_error_return_item(T t) {
+  return transform(step::on_error_return_item<T>{std::move(t)});
```
Neverlord, line 645:
```suggestion
observable<T>::on_error_return_item(T item) {
  return transform(step::on_error_return_item<T>{std::move(item)});
```

libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -112,6 +112,9 @@ class observable {
   /// Recovers from errors by converting `on_error` to `on_complete` events.
   transformation<step::on_error_complete<T>> on_error_complete();
 
+  /// Recovers from errors by returning an item.
+  transformation<step::on_error_return_item<T>> on_error_return_item(T t);
```
Neverlord, line 116:
```suggestion
  transformation<step::on_error_return_item<T>> on_error_return_item(T item);
```

Neverlord (changes requested):
Just some nitpick on argument names, looks good otherwise.

Neverlord (approved):

PR #1591, opened 2023-10-07
Fix heap-use-after-free with the testing mailbox

### LINKED ISSUE #1589: heap-use-after-free under ASAN with the new test fixture
While putting together a regression test for https://github.com/actor-framework/actor-framework/issues/1584, I came across this lovely bug. It boils down to CAF trying to clean up the same actor twice.

Here's a minimal example:

```c++
TEST("WTF") {
  auto dummy1 = sys.spawn([] { return behavior{[](int) {}}; });
  auto dummy2 = sys.spawn(
    [](event_based_actor* self, actor buddy) {
      self->request(buddy, infinite, 42).await([](int) {});
      return behavior{[](int) {}};
    },
    dummy1);
  inject().with(dummy2.address()).to(dummy2);
}
```

Only happens with the new test fixture. ASAN didn't complain when using this test with the old deterministic setup.

Neverlord:
That was nasty. The issue here was that we have unstashed skipped items by putting them back into the mailbox. Our new mailbox from the deterministic mailbox will then store this in the global list of messages alongside a strong pointer to the receiver. If the receiver is being cleaned up as a result of becoming unreachable, we increment the strong ref count from 0 back to 1... The default mailbox is not affected by this.

Closes #1589.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 
 - The class `caf::test::outline` is now properly exported from the test module.
   This fixes builds with dynamic linking against `libcaf_test`.
+- Fix a crash with the new deterministic test fixture when cleaning up actors
+  with stashed messages (#1589).
 
 ### Deprecated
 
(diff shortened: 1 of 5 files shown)


Neverlord:
Also added new assertions that would have save me a couple hours debugging this... Obviously we won't get coverage on these lines, so we can just ignore the complaints from Codecov here. 🙂

shariarriday (approved):

PR #1598, opened 2023-10-10
Fix leak when using spawn_inactive

### LINKED ISSUE #1597: Leak when using spawn_inactive
When using `spawn_inactive` and *not* calling the launcher explicitly, CAF leaks the actor handle.

Neverlord:
Closes #1597.

### DIFF
--- a/libcaf_core/caf/actor_system.test.cpp
+++ b/libcaf_core/caf/actor_system.test.cpp
@@ -50,10 +50,11 @@ TEST("spawn_inactive creates an actor without launching it") {
       check(!*flag);
       launch();
       check(*flag);
-      print_debug("calling launch() twice is a no-op");
-      *flag = false;
-      launch();
-      check(!*flag);
+      SECTION("calling launch() twice is a no-op") {
+        *flag = false;
+        launch();
+        check(!*flag);
+      }
     }
     self->wait_for(hdl);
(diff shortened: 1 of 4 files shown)


libcaf_core/caf/actor_system.test.cpp
```diff
@@ -0,0 +1,57 @@
...
+    check(*flag);
+  }
+  SECTION("the actor launches automatically at scope exit") {
+    {
+      auto [self, launch] = sys.spawn_inactive<dummy_actor>(flag);
+      check(!*flag);
+    }
+    check(*flag);
+  }
+}
```
riemass, line 55: This test verifies that the actor has launched (called `fn()` in both cases). However the ticket reports leaking an actor handle, which is fixed by adding the `fn.~F()` part. I think we should check the strong refs of self to fully test this fix, like we did in this test.
https://github.com/actor-framework/actor-framework/blob/1f3ee7facaaa53bbbc1b6af4474437f6df21b03e/libcaf_core/tests/legacy/mixin/requester.cpp#L383).

riemass (changes requested):
I have a small request for the unit test, aside from that, good catch :+1:.

riemass (changes requested):
Small nit.

riemass (approved):
:rocket:

PR #1607, opened 2023-10-15
Add new flow on_error_return with tests

shariarriday:

Adds new flow on_error_return for observables.

### DIFF
--- a/libcaf_core/caf/flow/step/on_error_return.hpp
+++ b/libcaf_core/caf/flow/step/on_error_return.hpp
@@ -14,10 +14,6 @@ namespace caf::flow::step {
 template <class ErrorHandler>
 class on_error_return {
 public:
-  using trait = detail::get_callable_trait_t<ErrorHandler>;
-
-  static_assert(trait::num_args == 1, "error_handler must take an argument");
-
   using handler_res
     = decltype(std::declval<ErrorHandler&>()(std::declval<const error&>()));
 
--- a/libcaf_core/caf/flow/step/on_error_return.hpp
+++ b/libcaf_core/caf/flow/step/on_error_return.hpp
@@ -1,0 +1,63 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/detail/type_traits.hpp"
+#include "caf/fwd.hpp"
+
+#include <utility>
+
+namespace caf::flow::step {
+
+template <class ErrorHandler>
+class on_error_return {
+public:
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/flow/step/on_error_return.hpp
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+#include "caf/detail/type_traits.hpp"
+#include "caf/fwd.hpp"
+
+#include <utility>
+
+namespace caf::flow::step {
+
+template <class T, class Predicate>
```
Neverlord, line 14:
```suggestion
template <class ErrorHandler>
```

A predicate is something that returns a boolean, so I think `ErrorHandler` is a more appropriate name here. Furthermore, we can deduce `T` from the signature of the error handler.

libcaf_core/caf/flow/step/on_error_return.hpp
```diff
@@ -0,0 +1,61 @@
...
+class on_error_return {
+public:
+  using trait = detail::get_callable_trait_t<Predicate>;
+
+  static_assert(std::is_same_v<typename trait::result_type, caf::expected<T>>,
+                "predicates must return a caf::expected type");
+
+  static_assert(trait::num_args == 0, "predicates must not take any argument");
```
Neverlord, line 22:
```suggestion
  using handler_res = decltype(std::declval<ErrorHandler&>()(std::declval<const error&>()));
  
  static_assert(detail::is_expected<handler_res>,
                "predicates must return a caf::expected type");

  static_assert(trait::num_args == 0, "predicates must not take any argument");
```

libcaf_core/caf/flow/step/on_error_return.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  using input_type = T;
+
+  using output_type = T;
```
Neverlord, line 26:
```suggestion
  using input_type = typename handler_res::value_type;

  using output_type = input_type;
```

libcaf_core/caf/flow/step/on_error_return.hpp
```diff
@@ -0,0 +1,63 @@
...
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
+    next.on_complete(steps...);
+  }
+
+  template <class Next, class... Steps>
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    if (auto result = fn_(what)) {
+      next.on_next(std::move(*result), steps...);
+      next.on_complete(steps...);
```
Neverlord, line 54:
```suggestion
      if (next.on_next(std::move(*result), steps...))
        next.on_complete(steps...);
```

Neverlord (changes requested):

Neverlord (approved):

Issue #1392, opened 2023-03-19
Add high-level APIs for sending HTTP requests


Neverlord:
Currently, we only include high-level APIs for spinning up HTTP servers. Adding functions for conveniently sending GET, POST, etc. is a natural next step.

Neverlord:
Pulling this into 0.19.5 since it's close to land. As discussed in the PR, we only need to support URI-based connect for now. That's what >90% of use cases will use anyways and we'll see whether there are feature requests for other ways to start an HTTP request.

Neverlord:
@riemass do you have a followup for the last PR in the pipeline or can we close this ticket?

riemass:
No followup, the last PR wraps things up nicely.

PR #1609, opened 2023-10-16
Creating a HTTP client class

riemass:
Relates #1392

### DIFF
--- a/libcaf_core/caf/span.hpp
+++ b/libcaf_core/caf/span.hpp
@@ -232,4 +232,10 @@ span<T> make_span(T* first, T* last) {
   return {first, last};
 }
 
+/// Convenience function to make using `caf::span` more convenient without the
+/// deduction guides.
+inline span<const std::string_view::value_type> make_span(std::string_view xs) {
+  return {xs.data(), xs.size()};
+}
+
 } // namespace caf
--- a/libcaf_core/caf/span.hpp
+++ b/libcaf_core/caf/span.hpp
@@ -207,7 +207,7 @@ span<std::byte> as_writable_bytes(span<T> xs) {
 /// Convenience function to make using `caf::span` more convenient without the
 /// deduction guides.
 template <class T>
-auto make_span(T& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
+auto make_span(T&& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
   return {xs.data(), xs.size()};
 }
 
@@ -230,12 +230,6 @@ span<T> make_span(T* first, size_t size) {
 template <class T>
 span<T> make_span(T* first, T* last) {
   return {first, last};
-}
-
-/// Convenience function to make using `caf::span` more convenient without the
-/// deduction guides.
-inline span<const std::string_view::value_type> make_span(std::string_view xs) {
-  return {xs.data(), xs.size()};
 }
 
 } // namespace caf
--- a/libcaf_core/caf/span.hpp
+++ b/libcaf_core/caf/span.hpp
@@ -207,7 +207,7 @@ span<std::byte> as_writable_bytes(span<T> xs) {
 /// Convenience function to make using `caf::span` more convenient without the
 /// deduction guides.
 template <class T>
-auto make_span(T&& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
+auto make_span(T& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
   return {xs.data(), xs.size()};
 }
 
--- a/libcaf_net/caf/net/http/client.cpp
+++ b/libcaf_net/caf/net/http/client.cpp
@@ -41,23 +41,19 @@ void client::suspend_reading() {
   down_->configure_read(receive_policy::stop());
 }
 
-// TODO
 void client::begin_header(http::method method, uri resource) {
-  // down_->begin_output();
-  // v1::begin_response_header(code, down_->output_buffer());
+  down_->begin_output();
+  v1::begin_request_header(method, resource, down_->output_buffer());
 }
 
-// TODO
 void client::add_header_field(std::string_view key, std::string_view val) {
   v1::add_header_field(key, val, down_->output_buffer());
 }
 
-// TODO
 bool client::end_header() {
   return v1::end_header(down_->output_buffer()) && down_->end_output();
 }
 
-// TODO
 bool client::send_payload(const_byte_span bytes) {
   down_->begin_output();
   auto& buf = down_->output_buffer();
@@ -66,7 +62,6 @@ bool client::send_payload(const_byte_span bytes) {
   return true;
 }
 
-// TODO
 bool client::send_chunk(const_byte_span bytes) {
   down_->begin_output();
   auto& buf = down_->output_buffer();
@@ -80,7 +75,6 @@ bool client::send_chunk(const_byte_span bytes) {
   return down_->end_output();
 }
 
-// TODO
 bool client::send_end_of_chunks() {
   std::string_view str = "0\r\n\r\n";
   auto bytes = as_bytes(make_span(str));
@@ -125,7 +119,6 @@ ptrdiff_t client::consume(byte_span input, byte_span) {
       }
       auto [hdr, remainder] = v1::split_header(input);
       // Wait for more data.
-      // TODO consumed or zero
       if (hdr.empty())
         return consumed;
       // Note: handle_header already calls up_->abort().
@@ -159,7 +152,6 @@ ptrdiff_t client::consume(byte_span input, byte_span) {
         if (!invoke_upper_layer(input.subspan(0, payload_len_)))
           return -1;
         consumed += static_cast<ptrdiff_t>(payload_len_);
-        // TODO: add line to server
         input.subspan(payload_len_);
         mode_ = mode::read_header;
       } else {
@@ -177,7 +169,6 @@ ptrdiff_t client::consume(byte_span input, byte_span) {
 
 // -- utility functions ------------------------------------------------------
 
-// TODO
 void client::write_response(status code, std::string_view content) {
   down_->begin_output();
   v1::write_response(code, "text/plain", content, down_->output_buffer());
--- a/libcaf_net/caf/net/http/client.cpp
+++ b/libcaf_net/caf/net/http/client.cpp
@@ -42,9 +42,9 @@ void client::suspend_reading() {
 }
 
 // TODO
-void client::begin_header(status code) {
-  down_->begin_output();
-  v1::begin_response_header(code, down_->output_buffer());
+void client::begin_header(http::method method, uri resource) {
+  // down_->begin_output();
+  // v1::begin_response_header(code, down_->output_buffer());
 }
 
 // TODO
--- a/libcaf_net/caf/net/http/client.cpp
+++ b/libcaf_net/caf/net/http/client.cpp
@@ -111,57 +111,58 @@ bool client::done_sending() {
 ptrdiff_t client::consume(byte_span input, byte_span) {
   CAF_LOG_TRACE(CAF_ARG2("bytes", input.size()));
   ptrdiff_t consumed = 0;
-  for (;;) {
-    if (mode_ == mode::read_header) {
-      if (input.size() >= max_response_size_) {
-        abort("Header exceeds maximum size.");
-        return -1;
-      }
-      auto [hdr, remainder] = v1::split_header(input);
-      // Wait for more data.
-      if (hdr.empty())
-        return consumed;
-      // Note: handle_header already calls up_->abort().
-      if (!handle_header(hdr))
-        return -1;
-      // Prepare for next loop iteration.
-      consumed += static_cast<ptrdiff_t>(hdr.size());
-      input = remainder;
-      // Transition to the next mode.
-      if (hdr_.chunked_transfer_encoding()) {
-        mode_ = mode::read_chunks;
-      } else if (auto len = hdr_.content_length()) {
-        // Protect against payloads that exceed the maximum size.
-        if (*len >= max_response_size_) {
-          abort("Payload exceeds maximum size.");
-          return -1;
-        }
-        // Transition to read_payload mode and continue.
-        payload_len_ = *len;
-        mode_ = mode::read_payload;
-      } else {
-        // TODO: we may *still* have a payload since HTTP can omit the
-        //       Content-Length field and simply close the connection
-        //       after the payload.
-        if (!invoke_upper_layer(const_byte_span{}))
-          return -1;
-      }
-    } else if (mode_ == mode::read_payload) {
-      if (input.size() >= payload_len_) {
-        if (!invoke_upper_layer(input.subspan(0, payload_len_)))
-          return -1;
-        consumed += static_cast<ptrdiff_t>(payload_len_);
-        input.subspan(payload_len_);
-        mode_ = mode::read_header;
-      } else {
-        // Wait for more data.
-        return consumed;
-      }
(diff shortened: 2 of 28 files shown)


libcaf_net/caf/net/http/v1.cpp
```diff
@@ -62,12 +88,18 @@ void write_header(status code, span<const string_view_pair> fields,
...
+void begin_response_header(status code, byte_buffer& buf) {
   writer out{&buf};
   out << "HTTP/1.1 "sv << std::to_string(static_cast<int>(code)) << ' '
       << phrase(code) << "\r\n"sv;
 }
 
+void begin_request_header(http::method method, std::string_view path,
+                          byte_buffer& buf) {
+  writer out{&buf};
+  out << format(method) << " "sv << path << " HTTP/1.1\r\n"sv;
```
Neverlord, line 100:
```suggestion
  out << format(method) << ' ' << path << " HTTP/1.1\r\n"sv;
```

libcaf_net/caf/net/http/lower_layer.hpp
```diff
@@ -46,6 +47,20 @@ class CAF_NET_EXPORT lower_layer : public generic_lower_layer {
...
   virtual bool send_end_of_chunks() = 0;
 
+  /// Asks the stream to swap the HTTP layer with `next` after returning from
+  /// `consume`.
+  /// @note may only be called from the upper layer in `consume`.
+  virtual void switch_protocol(std::unique_ptr<octet_stream::upper_layer> next)
+    = 0;
+};
+
+class lower_layer::server : public lower_layer {
```
Neverlord, line 57: Needs `CAF_NET_EXPORT`. Same for the client implementation.

Neverlord (changes requested):
Looks good so far.

There's a bit of code duplication that you've mentioned already. Namely `send_header`, `send_chunk`, `mpx`, etc. However, I don't mind them too much. Not enough to throw CRTP base types at it.

riemass:
> However, I don't mind them too much. Not enough to throw CRTP base types at it.

OK, thanks for the feedback. I'll fix according to the review.

libcaf_core/caf/span.hpp
```diff
@@ -232,4 +232,10 @@ span<T> make_span(T* first, T* last) {
   return {first, last};
 }
 
+/// Convenience function to make using `caf::span` more convenient without the
+/// deduction guides.
+inline span<const std::string_view::value_type> make_span(std::string_view xs) {
```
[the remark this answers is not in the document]
riemass: I missed that. It only works with an lvalue reference though, temporary `string_view` doesn't work.

libcaf_net/caf/net/http/client.hpp
```diff
@@ -0,0 +1,154 @@
...
+    read_payload,
+    read_chunks,
+  };
+
+  using upper_layer_ptr = std::unique_ptr<http::upper_layer::client>;
+
+  // -- constants --------------------------------------------------------------
+
+  /// Default maximum size for incoming HTTP requests: 64KiB.
+  static constexpr uint32_t default_max_request_size = 65'536;
```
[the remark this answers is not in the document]
riemass: Increased it accordingly. Agree that the default size was too small.

riemass:
Would this be a good point to merge this PR, and leave the `client_factory` and `robot` test as a followup?

Neverlord:
> Would this be a good point to merge this PR, and leave the `client_factory` and `robot` test as a followup?

Yep, good thinking. Let's wrap this up and then look into the higher-level APIs.

libcaf_core/caf/span.hpp
```diff
@@ -207,7 +207,7 @@ span<std::byte> as_writable_bytes(span<T> xs) {
 /// Convenience function to make using `caf::span` more convenient without the
 /// deduction guides.
 template <class T>
-auto make_span(T& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
+auto make_span(T&& xs) -> span<std::remove_reference_t<decltype(xs[0])>> {
```
Neverlord, line 210: Please don't, this was on purpose to avoid having `make_span` take the address of a temporary. Allow rvalues here can lead to very nasty and hard to debug bugs.

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+      return -1;
+    }
+    auto [hdr, remainder] = v1::split_header(input);
+    // Wait for more data.
+    if (hdr.empty())
+      return 0;
+    // Note: handle_header already calls up_->abort().
+    if (!handle_header(hdr))
+      return -1;
+    // Prepare for next loop iteration.
```
Neverlord, line 126: There is no loop iteration.

```suggestion
    // Prepare for next call to consume.
```

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+  up_->prepare_send();
+}
+
+bool client::done_sending() {
+  return up_->done_sending();
+}
+
+ptrdiff_t client::consume(byte_span input, byte_span) {
+  CAF_LOG_TRACE(CAF_ARG2("bytes", input.size()));
+  ptrdiff_t consumed = 0;
```
Neverlord, line 113: Without a loop, we no longer need this accumulate over multiple iterations.

```suggestion
```

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+    }
+    auto [hdr, remainder] = v1::split_header(input);
+    // Wait for more data.
+    if (hdr.empty())
+      return 0;
+    // Note: handle_header already calls up_->abort().
+    if (!handle_header(hdr))
+      return -1;
+    // Prepare for next loop iteration.
+    consumed = static_cast<ptrdiff_t>(hdr.size());
```
Neverlord, line 127:
```suggestion
    auto consumed = static_cast<ptrdiff_t>(hdr.size());
```

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+      if (!invoke_upper_layer(const_byte_span{}))
+        return -1;
+      return consumed;
+    }
+  }
+  if (mode_ == mode::read_payload) {
+    if (input.size() >= payload_len_) {
+      if (!invoke_upper_layer(input.subspan(0, payload_len_)))
+        return -1;
+      consumed += static_cast<ptrdiff_t>(payload_len_);
```
Neverlord, line 154:
```suggestion
      auto consumed = static_cast<ptrdiff_t>(payload_len_);
```

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+      input.subspan(payload_len_);
+      mode_ = mode::read_header;
+      return consumed;
+    } else {
+      // Wait for more data.
+      return consumed;
+    }
```
Neverlord, line 161: I'd push this up to get rid of one layer of `if` nesting:

```
  if (input.size() >= payload_len_) {
    // Wait for more data.
    return 0;
  }
```

libcaf_net/caf/net/http/client.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+bool client::done_sending() {
+  return up_->done_sending();
+}
+
+ptrdiff_t client::consume(byte_span input, byte_span) {
+  CAF_LOG_TRACE(CAF_ARG2("bytes", input.size()));
+  ptrdiff_t consumed = 0;
+  if (mode_ == mode::read_header) {
```
Neverlord, line 114: Why not `switch (mode_)`?

libcaf_net/caf/net/http/client.test.cpp
```diff
@@ -0,0 +1,539 @@
...
+      client->begin_header(method, "/foo/bar"sv);
+      client->end_header();
+      THEN("the output contains the formatted request") {
+        check_eq(transport->output_as_str(), expected);
+      }
+    }
+  }
+  EXAMPLES = R"(
+    |  method  | enum value |
+    | GET      |     0      |
```
Neverlord, line 349: `method` has `to_string` and `from_string`, so we can actually use proper names here:

```suggestion
    | GET      |     get      |
```

and then just `block_parameters<net::http::method>()` to get rid of the static casts.

The enum values are the C++ identifiers from the source file: https://github.com/actor-framework/actor-framework/blob/master/libcaf_net/caf/net/http/method.hpp.

libcaf_net/caf/net/http/v1.cpp
```diff
@@ -30,10 +34,32 @@ writer& operator<<(writer& out, const std::string& str) {
   return out << std::string_view{str};
 }
 
+constexpr std::string_view format(http::method m) {
```
Neverlord, line 37: Why not use https://github.com/actor-framework/actor-framework/blob/master/libcaf_net/caf/net/http/method.hpp#L55 ?

Neverlord (changes requested):

libcaf_core/caf/span.hpp
riemass: Reverted back. 
Having a overload for temporary non-owning data would be beneficial though.

libcaf_net/caf/net/http/client.cpp
riemass: The construct here is that we do `read_header` first, and then fall through and `read_payload` or `read_chunks` in the same `consume` call.  This would be hard to do using switch.
I'll should probably refactor this with code clarity in mind.

libcaf_net/caf/net/http/client.cpp
riemass: We actually use the `consumed` outside the `if` case, so it's necessary having it in this scope. I'll refactor for clarity here (see other comment).

libcaf_net/caf/net/http/client.cpp
riemass: Added.

libcaf_net/caf/net/http/client.cpp
riemass: Explained in other comment.

libcaf_net/caf/net/http/client.cpp
riemass: Explained in other comment.

libcaf_net/caf/net/http/client.cpp
riemass: Done.

libcaf_net/caf/net/http/client.cpp
Neverlord: Ok, then I've just missed a piece. 👍

libcaf_net/caf/net/http/client.cpp
Neverlord: Ah, my assumption was that we remove the loop alongside any fall throughs. Don't worry about leaving it this way, though.

libcaf_net/caf/net/http/client.test.cpp
riemass: I tested the exact same setup before and got conversion errors when using `get` and `method::get`. After looking at the output from `to_string`, I got it working by using the full enum name, including the namespace. GET would be `caf::net::http::method::get`. Might be worth adding to the docs? Also might be worth looking into, an user would probably expect at least `method::get` to work.

Neverlord: Yeah, I've realized this as well. See https://github.com/actor-framework/actor-framework/pull/1621. We can pull that in first and then rebase this branch.

libcaf_net/caf/net/http/v1.cpp
riemass: How did I miss this one :dizzy_face: 
I would suggest changing `to_rfc_string` to return `string_view` instead of `string`?

Neverlord: I think I've used string for symmetry with `to_string`. However, no strong opinion either way. Feel free to change the return type.

libcaf_net/caf/net/http/client.test.cpp
riemass: Sounds good. Reviewed it.

libcaf_net/caf/net/http/client.test.cpp
```diff
@@ -0,0 +1,542 @@
...
+      client->begin_header(method, "/foo/bar"sv);
+      client->end_header();
+      THEN("the output contains the formatted request") {
+        check_eq(transport->output_as_str(), expected);
+      }
+    }
+  }
+  EXAMPLES = R"(
+    |  method  |            enum value             |
+    | GET      |  caf::net::http::method::get      |
```
Neverlord, line 352: This works without the fully qualified name now (and is much more readable this way).

```
    | GET      |  get      |
```

libcaf_net/caf/net/http/method.hpp
```diff
@@ -52,7 +52,7 @@ CAF_NET_EXPORT std::string to_string(method);
 
 /// Converts @p x to the RFC string representation, i.e., all-uppercase.
 /// @relates method
-CAF_NET_EXPORT std::string to_rfc_string(method x);
+CAF_NET_EXPORT std::string_view to_rfc_string(method x);
```
Neverlord, line 55: Can no longer throw since we return `string_view`:

```suggestion
CAF_NET_EXPORT std::string_view to_rfc_string(method x) noexcept;
```

libcaf_net/caf/net/http/v1.cpp
```diff
@@ -32,8 +36,8 @@ writer& operator<<(writer& out, const std::string& str) {
 
 } // namespace
 
-std::pair<std::string_view, byte_span> split_header(byte_span bytes) {
-  std::array<std::byte, 4> end_of_header{{
+std::pair<std::string_view, byte_span> split_header(const byte_span bytes) {
```
Neverlord, line 39: Adding `const` here makes sense to me. You can still override all of the bytes pointed to. This basically only prevents re-assigning the span variable itself.

```suggestion
std::pair<std::string_view, byte_span> split_header(byte_span bytes) {
```

libcaf_net/caf/net/http/v1.cpp
```diff
@@ -32,8 +36,8 @@ writer& operator<<(writer& out, const std::string& str) {
...
+  static constexpr std::array<std::byte, 4> end_of_header{{
```
Neverlord, line 40: Out of curiosity, what was your reasoning to make this static? Having these four bytes on the stack should pretty much always be faster than loading it from the data section. And we need to load it anyways. I can see global state useful for larger blocks of data, but a 4-byte array can just sit on the stack and is probably faster this way (if the compiler isn't optimizing it this way anyways).

```suggestion
  std::array<std::byte, 4> end_of_header{{
```

Neverlord (changes requested):
Just some small nitpicks this time, after this we can pull it in.

libcaf_net/caf/net/http/v1.cpp
riemass: The first question I got reading the name `split_header` was does the function modify the range pointed to by bytes, and the fastest way to check that was adding `const`. I left it there since I like the message it conveys through the function signature. However, I don't mind leaving the old signature.

libcaf_net/caf/net/http/v1.cpp
riemass: Good point. I added the `constexpr` to  be sure we optimize out the constructor call at compile time. Then my mind short circuited to `static constexpr` without giving it much thought, but full ack on your statement.

Neverlord (approved):

PR #1610, opened 2023-10-17
Switch back to single test binary

Neverlord:
Too many test binaries cause builds to fail because the runners run out of disk space. Hence, we switch back to a single binary per CAF module to avoid CI failures.

Also streamlined a couple of flow unit tests with our new fixture.


riemass (approved):
You've used the opportunity so add some additional changes and modernization :+1: . I think we lost a regression with the changes in `actor_system.test.cpp`, but it's not that important and it's a good rework. :fire:

PR #1614, opened 2023-10-21
Fix type-ids check in load_data for Message

allspark:
Hi,

i discovered that the `load_data` https://github.com/actor-framework/actor-framework/blob/0.19.4/libcaf_core/caf/message.cpp#L73 iterates over the placeholder for the actual-size of the `type_id_list`.

### DIFF
--- a/libcaf_core/caf/detail/type_id_list_builder.cpp
+++ b/libcaf_core/caf/detail/type_id_list_builder.cpp
@@ -102,7 +102,7 @@ void type_id_list_builder::reserve(size_t num_elements) {
 }
 
 void type_id_list_builder::push_back(type_id_t id) {
-  if ((size_ + 1) >= reserved_)
+  if (size_ >= reserved_)
     reserve(reserved_ + block_size);
   storage_[size_++] = id;
 }
--- a/libcaf_core/caf/detail/type_id_list_builder.hpp
+++ b/libcaf_core/caf/detail/type_id_list_builder.hpp
@@ -43,16 +43,6 @@ public:
   /// @pre `index < size()`
   type_id_t operator[](size_t index) const noexcept;
 
-  // -- iterator access --------------------------------------------------------
-
-  auto begin() const noexcept {
-    return storage_;
-  }
-
-  auto end() const noexcept {
-    return storage_ + size_;
-  }
-
   // -- conversions ------------------------------------------------------------
 
   /// Convertes the internal buffer to a @ref type_id_list and returns it.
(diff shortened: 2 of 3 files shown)


Neverlord:
Wow, thanks a lot for hunting this down!

> i discovered ...

Our of curiosity: how? 🙂

If I read this correctly, there are two bugfixes here:
1) CAF re-allocated memory for the type ID list too early (the off-by-one part)
2) CAF allocated too much memory for deserialized messages because it always added the padded size for whatever type has type ID 0

These are very subtle bugs, so how did you find them? Memory profiling?

allspark:
I explored how the `std::set<std::string>` (from `caf::actor_system::mpi`) for `message_types`  is used and how `void` is encoded as `type_id_t`. Then changed the `0` in  `CAF_BEGIN_TYPE_ID_BLOCK(core_module, 0)`  to `1` and got a segfault.

libcaf_core/caf/detail/type_id_list_builder.cpp
```diff
@@ -102,7 +102,7 @@ void type_id_list_builder::reserve(size_t num_elements) {
 }
 
 void type_id_list_builder::push_back(type_id_t id) {
-  if ((size_ + 1) >= reserved_)
+  if (size_ >= reserved_)
```
Neverlord, line 105: After reading this a couple of times and checking with `reserve`, I'm not sure whether this patch is correct.

This is the start of `reserve`:

```c++
void type_id_list_builder::reserve(size_t num_elements) {
  auto new_capacity = num_elements + 1; // One extra slot for the size prefix.
  if (reserved_ >= new_capacity)
    return;
  reserved_ = new_capacity;
  auto ptr = realloc(storage_, reserved_ * sizeof(type_id_t));
  ...
```

So `reserved_` is basically `max_size + 1`. I think that's the reason why this code checks for `if ((size_ + 1) >= reserved_)`: because `size_` is the number of elements (does not contain the dummy) but `reserved_` is the number of elements *plus the dummy*. I do agree that it's a bit confusing. If we'd go with your suggestion, we would also need to change how `reserved_` works (by setting it to `num_elements` instead of `new_capacity`). Something like this:

```c++
void type_id_list_builder::reserve(size_t num_elements) {
  if (reserved_ >= num_elements)
    return;
  reserved_ = num_elements;
  auto new_capacity = num_elements + 1; // One extra slot for the size prefix.
  auto ptr = realloc(storage_, new_capacity * sizeof(type_id_t));
  ...
```

But feel free to just drop this change from the commits instead and I'll clean that up at some later point.

Neverlord: Sorry, discard this comment. `size_` also includes the dummy.

allspark: I think `size_` contains the index for next element, because `size()` returns `size_ - 1`;

PR #1617, opened 2023-10-23
Add .readthedocs.yaml config file

### LINKED ISSUE #1611: Fix ReadTheDocs setup
Our manual for RTD broke after 0.19.3. Looks like they require a config file now: https://docs.readthedocs.io/en/stable/config-file/v2.html

riemass:
Closes #1611 

@Neverlord Can you trigger the readthedocs pipeline manually?

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,21 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+
+build:
+  os: ubuntu-22.04
+
+python: 
+  version: 3.12
+  install:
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - requirements: manual/requirements.txt
+
+# Build documentation in the 'manual' directory with Sphinx
+sphinx:
+  configuration: manual/conf.py
+
+# Build docs in additional formats: PDF and ePub
+formats:
+   - pdf
+   - epub
```
Neverlord, line 21: Currently, we don't offer those. We don't have any code to make PDF format nicely, so it's probably best if people just stick to the HTML version. They can download the HTML version for offline viewing, so let's just keep it that way.

PR #1619, opened 2023-10-24
Add scan flow and tests for reduce

shariarriday:

Adds a new flow "Scan" and tests for reduce

### DIFF
--- a/libcaf_core/caf/flow/observable.test.cpp
+++ b/libcaf_core/caf/flow/observable.test.cpp
@@ -68,102 +68,82 @@ TEST("element_at(n) only takes the element with index n") {
   }
 }
 
-TEST("reduce(z, f) applies a function to each item, sequentially, and emits "
-     "the final value") {
-  SECTION("reduce(z, f) emits final value for a range of size m") {
+TEST("reduce(init, fn) combines all items into one final value using fn") {
+  std::plus<int> adder;
+  SECTION("reduce(init, fn) emits final value for a range of size m") {
     SECTION("blueprint") {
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -68,6 +68,106 @@ TEST("element_at(n) only takes the element with index n") {
   }
 }
 
+TEST("reduce(z, f) applies a function to each item, sequentially, and emits "
+     "the final value") {
```
Neverlord, line 72:
```suggestion
TEST("reduce(init, fn) combines all items into one final value using fn") {
  std::plus<int> adder;
```

(and then pass `adder` everywhere below instead of constructing a `std::plus` object each time.

libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -68,6 +68,106 @@ TEST("element_at(n) only takes the element with index n") {
...
+          mat(range(1, 3).concat(obs_error())).reduce(0, std::plus<int>())),
+        sec::runtime_error);
+      check_eq(collect(mat(obs_error()).reduce(0, std::plus<int>())),
+               sec::runtime_error);
+    }
+  }
+}
+
+TEST("scan(z, f) applies a function to each item sequentially, and emits each "
+     "successive value") {
```
Neverlord, line 124:
```suggestion
TEST("scan(init, fn) creates successive reduced values using fn") {
```

(plus using the `adder` as mentioned before to reduce verbosity)

Neverlord (changes requested):
Already looks solid, just some nitpicks to streamline the tests.

PR #1621, opened 2023-10-24
Use short names for CAF enums

Neverlord:
I recently suggested using enums in outlines, but then I realized that the enum rendering in CAF is quite verbose at the moment. So this is bascially a quality of life change.


riemass (approved):
You where one step ahead with this PR :D

PR #1622, opened 2023-10-25
Improve test coverage and remove dead code

Neverlord:
Made a pass over the I/O tests to see what legacy tests we can migrate or kick out. Added some new tests for low-level BASP functionality, I'll probably replace a bunch of the current legacy tests with Robot tests instead.


libcaf_io/caf/io/basp/connection_state.test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "caf/io/basp/connection_state.hpp"
+
+#include "caf/test/outline.hpp"
+
+using namespace caf::io::basp;
+
+namespace {
+
+OUTLINE("critical connection states require a connection shutdown") {
```
shariarriday, line 13: Shouldn't we use the newer `TEST` structure instead?

Neverlord: This is from the new testing framework. 🙂

See https://github.com/actor-framework/actor-framework/tree/master/libcaf_test#scenario-outlines.

shariarriday: Thanks for the clarification.

shariarriday (approved):

PR #1625, opened 2023-10-26
Restrict scope guard to noexcept functions

### LINKED ISSUE #1550: Restrict scope guard to accept only noexcept functions
A clang-tidy check is complaining that calling an arbitrary function in the destructor of the scope guard can throw an exception. We should restrict scope guard to only accept `noexcept` invokable types.
The suggestion here is to add a `static_assert` in the class to check if invoking the object might throw, and then go through the existing uses of scope guard and add `noexcept` declarations. After looking at the code, we should be able to declare most of the scope guard lambdas `noexcept` without any further code changes, but some cases might need other workarounds.

riemass:
Like discussed before, this PR adds a static assert to the scope guard making sure the provided function is noexcept. 
I've also went and added `noexcept` to all lambdas so the solution compiles without issues, however clang tidy does complain a lot about some cases where the lambdas might throw under rare circumstances. 

Closes #1550.

### DIFF
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -71,7 +71,8 @@ void ChatWidget::init(actor_system& system, const std::string& name,
 }
 
 void ChatWidget::sendChatMessage() {
-  auto cleanup = detail::make_scope_guard([=] { input()->setText(QString()); });
+  auto cleanup
+    = detail::make_scope_guard([=]() noexcept { input()->setText(QString()); });
   QString line = input()->text();
   if (line.isEmpty()) {
     // Ignore empty lines.
--- a/examples/qtsupport/chatwidget.cpp
+++ b/examples/qtsupport/chatwidget.cpp
@@ -71,8 +71,6 @@ void ChatWidget::init(actor_system& system, const std::string& name,
 }
 
 void ChatWidget::sendChatMessage() {
-  auto cleanup
-    = detail::make_scope_guard([=]() noexcept { input()->setText(QString()); });
   QString line = input()->text();
   if (line.isEmpty()) {
     // Ignore empty lines.
@@ -89,7 +87,6 @@ void ChatWidget::sendChatMessage() {
     } else {
       print("*** list of commands:\n"
             "/setName <new name>\n");
-      return;
     }
   } else {
     auto msg = name_;
@@ -100,6 +97,7 @@ void ChatWidget::sendChatMessage() {
       publisher_->push(msg);
     }
   }
+  input()->setText(QString());
 }
 
 void ChatWidget::changeName() {
--- a/libcaf_core/caf/detail/format.cpp
+++ b/libcaf_core/caf/detail/format.cpp
@@ -10,7 +10,6 @@
 #include "caf/detail/parser/is_char.hpp"
 #include "caf/detail/parser/read_unsigned_integer.hpp"
 #include "caf/detail/print.hpp"
-#include "caf/detail/scope_guard.hpp"
 #include "caf/parser_state.hpp"
 #include "caf/pec.hpp"
 #include "caf/raise_error.hpp"
@@ -309,10 +308,6 @@ void copy_verbatim(ParserState& ps, copy_state& cs);
 template <class ParserState>
 void copy_formatted(ParserState& ps, copy_state& cs) {
   cs.reset();
-  auto guard = make_scope_guard([&]() noexcept {
-    if (ps.code <= pec::trailing_character)
-      cs.fn = copy_verbatim<string_parser_state>;
-  });
   size_t tmp = 0;
   auto make_tmp_consumer = [&tmp] {
     tmp = 0;
@@ -407,6 +402,8 @@ void copy_formatted(ParserState& ps, copy_state& cs) {
   }
   fin();
   // clang-format on
+  if (ps.code <= pec::trailing_character)
+    cs.fn = copy_verbatim<string_parser_state>;
 }
 
 // Copies the input verbatim to the output buffer until finding a format string.
--- a/libcaf_core/caf/detail/format.cpp
+++ b/libcaf_core/caf/detail/format.cpp
@@ -309,7 +309,7 @@ void copy_verbatim(ParserState& ps, copy_state& cs);
 template <class ParserState>
 void copy_formatted(ParserState& ps, copy_state& cs) {
   cs.reset();
-  auto guard = make_scope_guard([&] {
+  auto guard = make_scope_guard([&]() noexcept {
     if (ps.code <= pec::trailing_character)
       cs.fn = copy_verbatim<string_parser_state>;
   });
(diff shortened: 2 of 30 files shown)


examples/qtsupport/chatwidget.cpp
```diff
@@ -71,7 +71,8 @@ void ChatWidget::init(actor_system& system, const std::string& name,
 }
 
 void ChatWidget::sendChatMessage() {
-  auto cleanup = detail::make_scope_guard([=] { input()->setText(QString()); });
+  auto cleanup
+    = detail::make_scope_guard([=]() noexcept { input()->setText(QString()); });
```
Neverlord, line 75: This can just move to the end of the function. There is one `return` in the `if-else` tree that we can just remove because it does nothing.

libcaf_core/caf/detail/parser/read_config.hpp
```diff
@@ -192,7 +192,7 @@ void read_config_map(State& ps, Consumer&& consumer) {
 template <class State, class Consumer>
 void read_config_uri(State& ps, Consumer&& consumer) {
   uri_builder builder;
-  auto g = make_scope_guard([&] {
+  auto g = make_scope_guard([&]() noexcept {
     if (ps.code <= pec::trailing_character)
       consumer.value(builder.make());
```
Neverlord, line 197: Can throw, because it allocates memory. Maybe we can just add something to the testing DSL for this instead of using scope guards. Applies to a lot of other places as well, so let me think a bit about this.

libcaf_core/caf/logger.hpp
```diff
@@ -353,8 +353,10 @@ class CAF_CORE_EXPORT logger {
 #define CAF_PUSH_AID(aarg)                                                     \
   caf::actor_id CAF_PP_UNIFYN(caf_aid_tmp)                                     \
     = caf::logger::thread_local_aid(aarg);                                     \
-  auto CAF_PP_UNIFYN(caf_aid_tmp_guard) = caf::detail::make_scope_guard(       \
-    [=] { caf::logger::thread_local_aid(CAF_PP_UNIFYN(caf_aid_tmp)); })
+  auto CAF_PP_UNIFYN(caf_aid_tmp_guard)                                        \
+    = caf::detail::scope_guard([=]() noexcept {                                \
+        caf::logger::thread_local_aid(CAF_PP_UNIFYN(caf_aid_tmp));             \
+      })
```
Neverlord, line 359: If we declare this lambda `noexcept`, we'll also need to declare `thread_local_aid` `noexcept`.

libcaf_core/tests/legacy/dsl.cpp
```diff
@@ -252,7 +252,7 @@ SCENARIO("passing requirements increment the 'good' counter") {
       auto silent_expr = [&] {                                                 \
         auto bk = caf::test::logger::instance().make_quiet();                  \
         auto guard = caf::detail::make_scope_guard(                            \
-          [bk] { caf::test::logger::instance().levels(bk); });                 \
+          [bk]() noexcept { caf::test::logger::instance().levels(bk); });      \
```
Neverlord, line 255: Adding `noexcept` here requires that we also add `noexcept` to `levels` in `libcaf_test/caf/test/unit_test.hpp`.

libcaf_io/tests/legacy/io/unpublish.cpp
```diff
@@ -68,7 +68,7 @@ BEGIN_FIXTURE_SCOPE(fixture)
 
 CAF_TEST(actors can become unpublished) {
   auto testee = mars.sys.spawn<dummy>(ssp);
-  auto guard = detail::make_scope_guard([&] {
+  auto guard = detail::make_scope_guard([&]() noexcept {
     // The MM holds a reference to this actor publishing it, so we need to kill
     // it manually.
     anon_send_exit(testee, exit_reason::user_shutdown);
```
Neverlord, line 74: Just put this at the end instead of using a scope guard.

Neverlord (changes requested):

Neverlord:
Please rebase to pass CI again.

libcaf_core/caf/detail/parser/read_config.hpp
riemass: In the case we don't have enough memory here, we're having an problem anyways, whether we throw in the scope guard destructor or in an noexcept function. Not sure how to handle out-of-memory conditions, it falls back to the custom allocator and system resources.

libcaf_core/caf/logger.hpp
riemass: I will add it to logger. Just a note here, I tried following the chain and adding `noexcept` everywhere and failed. As you noted in the previous comment as well, there are places where it's just not possible.

libcaf_core/caf/detail/format.cpp
```diff
@@ -309,7 +309,7 @@ void copy_verbatim(ParserState& ps, copy_state& cs);
 template <class ParserState>
 void copy_formatted(ParserState& ps, copy_state& cs) {
   cs.reset();
-  auto guard = make_scope_guard([&] {
+  auto guard = make_scope_guard([&]() noexcept {
     if (ps.code <= pec::trailing_character)
       cs.fn = copy_verbatim<string_parser_state>;
```
Neverlord, line 314: Looks like I didn't spot this last time. Please move below `fin()` to be consistent with the other recent changes and drop the include for `scope_guard`.

libcaf_core/caf/detail/json.cpp
```diff
@@ -215,7 +215,7 @@ template <class ParserState, class Consumer>
 void read_json_null_or_nan(ParserState& ps, Consumer consumer) {
   enum { nil, is_null, is_nan };
   auto res_type = nil;
-  auto g = make_scope_guard([&] {
+  auto g = make_scope_guard([&]() noexcept {
```
Neverlord, line 218: Same, move below `fin()` and drop scope guards.

libcaf_core/caf/scheduled_actor.cpp
```diff
@@ -280,7 +280,7 @@ resumable::resume_result scheduled_actor::resume(execution_unit* ctx,
   if (!activate(ctx))
     return resumable::done;
   size_t consumed = 0;
-  auto guard = detail::make_scope_guard([this, &consumed] {
+  auto guard = detail::make_scope_guard([this, &consumed]() noexcept {
     CAF_LOG_DEBUG("resume consumed" << consumed << "messages");
```
Neverlord, line 284: Logging might throw.

```suggestion
```

libcaf_test/caf/test/outline.cpp
```diff
@@ -107,7 +107,7 @@ void outline::run() {
     }
   }
   ctx_->parameters = ctx_->example_parameters[ctx_->example_id];
-  auto guard = detail::make_scope_guard([this] {
+  auto guard = detail::make_scope_guard([this]() noexcept {
     auto& ptr = ctx_->steps[std::make_pair(0, ctx_->example_id)];
```
Neverlord, line 111: This can technically throw, so clang-tidy might complain later about it. We can just move it out of the lambda and capture it instead.

Neverlord (changes requested):
Small nits, otherwise looks good.

Neverlord (approved):

PR #1628, opened 2023-10-27
Fix subscribe implementations that emit errors

Neverlord:
Observables must call on_subscribe before calling any other member function on an observer.

@shariarriday thanks for reporting this the other day. I finally found the time to look into this. I've also replaced several places in the `observables` test where we've concatenated an empty range with the `fail` operator to work around this bug previously. I've also renamed `mat` to `build` in the fixture. We'll probably implement https://reactivex.io/documentation/operators/materialize-dematerialize.html at some point, so we should not use "materialize" for something else. I'll also send a followup to rename the materializers in the flow API to builders.

### DIFF
--- a/libcaf_core/caf/flow/observable.test.cpp
+++ b/libcaf_core/caf/flow/observable.test.cpp
@@ -29,9 +29,9 @@ TEST("element_at(n) only takes the element with index n") {
       check_eq(collect(range(1, 3).element_at(0)), vector{1});
     }
     SECTION("observable") {
-      check_eq(collect(mat(range(1, 1)).element_at(0)), vector{1});
-      check_eq(collect(mat(range(1, 2)).element_at(0)), vector{1});
-      check_eq(collect(mat(range(1, 3)).element_at(0)), vector{1});
+      check_eq(collect(build(range(1, 1)).element_at(0)), vector{1});
+      check_eq(collect(build(range(1, 2)).element_at(0)), vector{1});
+      check_eq(collect(build(range(1, 3)).element_at(0)), vector{1});
     }
   }
   SECTION("element_at(1) picks the second element") {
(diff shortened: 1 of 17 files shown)


libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -438,50 +421,46 @@ TEST("on_error_return() replaces an error with a value") {
   SECTION("on_error_return() replaces an error with a value from the handler") {
     auto return_42 = [](const error&) { return expected<int>{42}; };
     SECTION("blueprint") {
-      check_eq(collect(range(1, 0)
-                         .concat(obs_error(sec::runtime_error))
-                         .on_error_return(return_42)),
+      check_eq(collect(
+                 range(1, 0).concat(obs_error()).on_error_return(return_42)),
```
shariarriday, line 425: I think that we can refactor it to as we don't need to use range(1, 0) to create an empty observable anymore.
```suggestion
                 check_eq(collect(obs_error().on_error_return(return_42)), vector{42});
```

libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -438,50 +421,46 @@ TEST("on_error_return() replaces an error with a value") {
...
-      check_eq(collect(mat(range(1, 0).concat(obs_error(sec::runtime_error)))
-                         .on_error_return(return_42)),
-               vector{42});
-      check_eq(collect(mat(range(1, 2)
-                             .concat(obs_error(sec::runtime_error))
-                             .concat(range(1, 2)))
-                         .on_error_return(return_42)),
+      check_eq(
+        collect(
+          build(range(1, 0).concat(obs_error())).on_error_return(return_42)),
```
shariarriday, line 449: Similarly this can be refactored to
```suggestion
          check_eq(collect(build(obs_error().on_error_return(return_42))), vector{42});
```

shariarriday (approved):

PR #1634, opened 2023-10-31
Add start_with step for observables

shariarriday:
@Neverlord you had mentioned that this can be implemented as `just(x).concat(*this)`. But, upon reading the documentation for this step it looks that the only difference between `just()` and `start_with()` is that `start_with` expects a function that outputs `x`.

### DIFF
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -377,9 +377,9 @@ public:
   }
 
   /// @copydoc observable::start_with
-  template <class F>
-  auto start_with(F fn) && {
-    return materialize().start_with(std::forward<F>(fn));
+  template <class Input>
+  auto start_with(Input&& value) && {
+    return materialize().start_with(std::forward<Input>(value));
   }
 
   /// @copydoc observable::flat_map
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -376,6 +376,12 @@ public:
     return materialize().concat(std::forward<Inputs>(xs)...);
   }
 
+  /// @copydoc observable::start_with
+  template <class F>
+  auto start_with(F fn) && {
+    return materialize().start_with(std::forward<F>(fn));
+  }
+
   /// @copydoc observable::flat_map
   template <class F>
   auto flat_map(F f) && {
--- a/libcaf_core/caf/flow/observable.test.cpp
+++ b/libcaf_core/caf/flow/observable.test.cpp
@@ -543,29 +543,58 @@ TEST("on_error_return() replaces an error with a value") {
   }
 }
 
-TEST("start_with(fn) makes an observable that emits output of fn()") {
-  const auto fn = [] { return 1; };
-  SECTION("start_with(fn) adds fn() as first element") {
+TEST("start_with(value) builds observable that emits value first") {
+  const auto value = 1;
+  SECTION("start_with(value) adds value as first element") {
     SECTION("blueprint") {
-      check_eq(collect(range(2, 3).start_with(fn)), vector{1, 2, 3, 4});
-      check_eq(collect(range(2, 3).start_with(fn).take(3)), vector{1, 2, 3});
-      check_eq(collect(range(2, 3).start_with(fn).first()), vector{1});
-      check_eq(collect(range(2, 3).start_with(fn).last()), vector{4});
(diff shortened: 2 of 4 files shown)


Neverlord:
I don't think this belongs to the observable builder. I read the [ReactiveX docs](https://reactivex.io/documentation/operators/startwith.html) such that `x.startWith(y)` is equivalent to `just(y).concat(x)`.

Here's an [example for RxJs](https://www.learnrxjs.io/learn-rxjs/operators/combination/startwith):

```js
// RxJS v6+
import { startWith } from 'rxjs/operators';
import { of } from 'rxjs';

//emit (1,2,3)
const source = of(1, 2, 3);
//start with 0
const example = source.pipe(startWith(0));
//output: 0,1,2,3
const subscribe = example.subscribe(val => console.log(val));
```

Translated to CAF, this could read as:

```cpp
// crates the range 0, 1, 2, 3, 4
make_observable().range(1, 3).start_with(0).for_each(...);
```

RxJava also offers an overload for `start_with` that takes an observable (or iterable).

libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -173,6 +173,12 @@ class observable {
     return reduce(T{}, [](T x, T y) { return x + y; });
   }
 
+  /// Adds the output of fn() to the beginning of current observable.
+  template <class F>
+  auto start_with(F fn) {
+    return ctx()->make_observable().just(fn()).concat(*this);
```
Neverlord, line 179: I don't see much upside for taking a function object as argument. We call it immediately anyways, so I'd rather take the value directly. Furthermore, if we take a value, we can dispatch on `if constexpr (is_observable_v<T>)` to allow users to call the function with a value or another observable. In the later case, we can just call `std::move(value).concat(*this)`.

Neverlord (changes requested):

libcaf_core/caf/flow/observable.hpp
```diff
@@ -370,6 +370,12 @@ class observable_def {
     return materialize().concat(std::forward<Inputs>(xs)...);
   }
 
+  /// @copydoc observable::start_with
+  template <class Input>
+  auto start_with(Input value) && {
```
Neverlord, line 375:
```suggestion
  auto start_with(Input&& value) && {
```

libcaf_core/caf/flow/observable.test.cpp
```diff
@@ -543,6 +543,40 @@ TEST("on_error_return() replaces an error with a value") {
   }
 }
 
+TEST("start_with(value) builds observable that emits value first") {
+  const auto value = 1;
+  SECTION("start_with(value) adds value as first element") {
```
Neverlord, line 548: Please also add a test for passing an observable to `start_with`.

Neverlord (changes requested):

libcaf_core/caf/flow/observable.test.cpp
shariarriday: I have already added the following test. Shall I move it to a different section?

```suggestion
  check_eq(collect(build(range(2, 3).start_with(range(1, 2)))),
               vector{1, 2, 2, 3, 4});
    }
```

libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -173,6 +173,15 @@ class observable {
     return reduce(T{}, [](T x, T y) { return x + y; });
   }
 
+  /// Adds the output of fn() to the beginning of current observable.
+  template <class Input>
+  auto start_with(Input value) {
+    if constexpr (is_observable_v<Input>)
+      return std::move(value).merge(*this);
```
Neverlord, line 180: Please read up on the semantics of the operators. Merge will not make sure that the result will start with the inputs. It'll *merge* the things, not concatenate them. The tests only pass because the data sets are so small that the merge operator just forwards everything in one shot.

```c++
      return std::move(value).concat(*this);
```

Neverlord (changes requested):

libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -178,6 +178,15 @@ class observable {
     return reduce(T{}, [](T x, T y) { return x + y; });
   }
 
+  /// Adds the output of fn() to the beginning of current observable.
```
Neverlord, line 181: The documentation string does not reflect the actual implementation.

PR #1637, opened 2023-11-01
Allow checking for custom logger factories

Neverlord:
Currently, it's not possible to check whether a custom logger factory has been installed. This prevents use cases where a module (or library on top of CAF) wants to provide it's own default logger factory that should be installed only if the user didn't already install a custom logger.

### DIFF
--- a/libcaf_core/caf/actor_system_config.cpp
+++ b/libcaf_core/caf/actor_system_config.cpp
@@ -33,6 +33,8 @@ constexpr std::string_view default_config_file = "caf-application.conf";
 
 } // namespace
 
+// -- constructors, destructors, and assignment operators ----------------------
+
 actor_system_config::~actor_system_config() {
   // nop
 }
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/actor_system_config.cpp
```diff
@@ -488,4 +492,13 @@ const settings& content(const actor_system_config& cfg) {
   return cfg.content;
 }
 
+// -- factories ----------------------------------------------------------------
+
+intrusive_ptr<logger>
+actor_system_config::make_logger(actor_system& sys) const {
+  if (logger_factory_)
```
shariarriday, line 499: Would it be better if we use `has_logger_factory` here instead?

Neverlord: We call `logger_factory_(sys)` on the next line, so I think using the member function here would do more obfuscation than good.

shariarriday (approved):

PR #1638, opened 2023-11-01
Make exec_main utils more generic

Neverlord:
Small tweak to our `exec_main_helper` to make it re-usable with subtypes of `actor_system`.


riemass (approved):
LGTM :+1:

PR #1640, opened 2023-11-01
Move legacy test code out of the core module

Neverlord:
Push code from the core module that only exists for (legacy) testing to `dsl.hpp`.


libcaf_test/caf/test/dsl.hpp
```diff
@@ -689,16 +718,361 @@ class test_coordinator_fixture {
...
+        trigger_timeouts();
+    }
+
+    void enqueue(caf::resumable* ptr) override {
+      jobs.push_back(ptr);
+      if (after_next_enqueue_ != nullptr) {
+        CAF_LOG_DEBUG("inline this enqueue");
+        std::function<void()> f;
+        f.swap(after_next_enqueue_);
+        f();
```
riemass, line 1047: We could spare a few bytes here. 
```suggestion
        after_next_enqueue_();
        after_next_enqueue_ = std::function<void()>{};
```

riemass (changes requested):
This and the previous PR look good, kudos for separating the concerns of testing.
We lost mailbox factory tests in this PR, is that fine? They do seem like test setup checks. There are some nits and discussion questions, otherwise looks good.

libcaf_test/caf/test/dsl.hpp
```diff
@@ -689,16 +718,361 @@ class test_coordinator_fixture {
...
+        return false;
+      if (jobs.size() >= 2)
+        std::rotate(jobs.rbegin(), jobs.rbegin() + 1, jobs.rend());
+      return try_run_once();
+    }
+
+    /// Executes a single event in FIFO order or fails if no event is available.
+    void run_once() override {
+      if (jobs.empty())
+        CAF_RAISE_ERROR("No job to run available.");
```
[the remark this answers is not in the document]
Neverlord: - No, `run_once` has the semantics of "there must be an event or its a logic error".
- No, `CAF_RAISE_ERROR` will call `abort` if there aren't exceptions available.

Either way, if `run_once` fails, you have to go fix your code (or test).

libcaf_test/caf/test/dsl.hpp
Neverlord: No, we can't do that. The callback might call `after_next_enqueue` again. If we don't swap out before calling, this will result in UB (and a very nasty debugging session).

Neverlord:
> We lost mailbox factory tests in this PR, is that fine? They do seem like test setup checks.

That's fine. As you've mentioned, this old test was basically a stepping stone for our new deterministic test fixture in the new framework. So we don't lose anything by removing it.

riemass (approved):
Looks good. Thanks for the discussion feedback.

PR #1641, opened 2023-11-02
Get rid of make_scope_guard

Neverlord:
C++17 made `make_scope_guard` superfluous by introducing class template argument deduction.

@riemass just a quick followup on your recent PR to modernize the scope guard implementation while we're at it. 🙂


riemass (approved):
Thanks for picking this up, looks good.

PR #1642, opened 2023-11-03
Remove caf.middleman.manual-multiplexing option

### LINKED ISSUE #1450: Get rid of CAF options for testing only
From the discussion in #1447. There are a couple of options that we use for setting up deterministic testing:

- [x] `caf.middleman.manual-multiplexing`
- [x] `caf.logger.inline-output`
- [x] setting `caf.scheduler.policy` to `"testing"`

However, users can actually define these things from a config file (and break the application). Our setup for deterministic testing should use a separate mechanism to make sure the configuration cannot put CAF into test mode.

Neverlord:
I know that this removes the OpenSSL tests entirely at the moment. I've already prepared Robot tests to replace them in `topic/openssl-robot-tests`. That's going to be a followup to this PR.

Closes #1450.


riemass (changes requested):
Looks good. Just to note, grepping the code base still shows uses of "manual-multiplexing":
```
libcaf_net/caf/net/middleman.cpp
119:    .add<bool>("manual-multiplexing",

examples/caf-application.conf
51:    manual-multiplexing = false

libcaf_io/caf/io/middleman.cpp
206:    .add<bool>("manual-multiplexing",
492:  if (!get_or(config(), "caf.middleman.manual-multiplexing", false)) {
516:      .set("caf.middleman.manual-multiplexing", true);
```
I think at least the first one is redundant, for the `application.conf` file, we can probably remove it there as well, and for the `io::middleman`, I skimmed the remaining tests and it seems it's not used, and it's combined with another config parameter, `network-backend`. 
Could you go through the list and see if it needs any more refactoring? Otherwise, this is good.

riemass (approved):

PR #1647, opened 2023-11-06
Add HTTP client robot tests

riemass:
Implement HTTP client robot tests for the simple case of manually sending a request. 
Relates #1392

### DIFF
--- a/robot/http/client_driver.cpp
+++ b/robot/http/client_driver.cpp
@@ -62,11 +62,12 @@ public:
   }
 
   virtual void abort(const caf::error& reason) override {
-    printf("%s\n", to_string(reason).c_str());
+    std::cerr << "*** abort called with error - " << to_string(reason)
+              << std::endl;
+    exit(EXIT_FAILURE);
   }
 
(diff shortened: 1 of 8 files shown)


robot/http/client_driver.cpp
Neverlord, line 1: We use hyphenated names for the other binaries: `client-driver.cpp` to stay consistent.

robot/http/client_driver.cpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  std::shared_ptr<detail::latch> latch() const noexcept {
+    return latch_;
+  }
+
+private:
+  http::lower_layer::client* down = nullptr;
+  http::method method_;
+  std::string resource_;
```
Neverlord, line 105: Got a bit confused about this member. The `resource` parameter for the program is an URI, but this captures only the path (plus query). So I think naming it `path_` would be clearer:

```suggestion
  std::string path_;
```

Neverlord (changes requested):
Looks good, just two small change requests.

Neverlord (approved):

PR #1648, opened 2023-11-06
Fix parsing of configs with dot-separated keys

Neverlord:
Parsing `foo.bar = 42` in a config file should be equivalent to `foo { bar = 42 }` in order to stay consistent with the syntax that we use on the CLI.


riemass (approved):
LGTM :+1:

PR #1649, opened 2023-11-06
Run robot tests in pipeline

riemass:


riemass:
The hard part of this PR was making the file `robot/dependencies.txt` visible inside the `docker.build` command. Please merge after [this PR](https://github.com/actor-framework/continuous-integration/pull/2).  
In that PR I've also made the CMake `set` use `FORCE` by default so we can override build flags, which is particularly useful for the `autobahn` testsuite. Looking at the build flags, I guess we supposed this works, but didn't check manually until now.

After merging CI, I'll remove the branch specifier in the Jenkinsfile library and push again to make sure everything passes.

Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 
-@Library('caf-continuous-integration@topic/neverlord/extra-scripts') _
+@Library('caf-continuous-integration@topic/riemass/docker-workdir') _
```
Neverlord, line 3: Upstream PR is merged:

```suggestion
@Library('caf-continuous-integration') _
```

.ci/almalinux-8/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM almalinux:8
 
+ADD ./robot/dependencies.txt .
```
Neverlord, line 3: `ADD` may pull in URL, please use `COPY` when dealing with local files.

Neverlord (changes requested):

Neverlord:
> In that PR I've also made the CMake `set` use `FORCE` by default so we can override build flags, which is particularly useful for the `autobahn` testsuite. Looking at the build flags, I guess we supposed this works, but didn't check manually until now.

Yep, that's how it was supposed to work. 👍

PR #1650, opened 2023-11-06
Make term inspectable and usable with FILE handles

Neverlord:
Make `caf::term` inspectable plus prep work for having the colors from `caf::term` work with `stdout` and `stderr`. Not really something that we can test automatically, unfortunately. If we run tests through Robot, it won't be a TTY. Tested manually, FWIW.

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -62,6 +62,7 @@ caf_add_component(
     message_priority
     pec
     sec
+    term
     thread_owner
   HEADERS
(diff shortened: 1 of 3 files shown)


Neverlord:
> Looks good to me. I'm kind of surprised we had to define STDIN_FILENO to 1 and STDERR_FILENO to 2, thought these where already defined somewhere. Did you test it only on Mac, or you've tried Windows and Linux as well?

I've tested the new code on macOS, but the term stuff was tested on all platforms by now. I don't remember details on STDIN_FILENO, though.

PR #1655, opened 2023-11-09
Limit request size in HTTP high level API

riemass:
Followup from our meeting, and I wanted to read the server factory code a bit more. 
We could make this more generic, however I'd wait for the refactoring we plan, since `lp` has `max_message_length`, which is the length excluding the length prefix, and `websocket` has `max_frame_size` which applies to both server and client.

### DIFF
--- a/examples/http/rest.cpp
+++ b/examples/http/rest.cpp
@@ -27,13 +27,17 @@ static constexpr uint16_t default_port = 8080;
 
 static constexpr size_t default_max_connections = 128;
 
+static constexpr size_t default_max_request_size
+  = caf::net::http::server::default_max_request_size;
+
 // -- configuration ------------------------------------------------------------
 
 struct config : caf::actor_system_config {
   config() {
     opt_group{custom_options_, "global"} //
       .add<uint16_t>("port,p", "port to listen for incoming connections")
-      .add<size_t>("max-connections,m", "limit for concurrent clients");
+      .add<size_t>("max-connections,m", "limit for concurrent clients")
+      .add<size_t>("max-request-size,r", "limit for single request size");
     opt_group{custom_options_, "tls"} //
       .add<std::string>("key-file,k", "path to the private key file")
       .add<std::string>("cert-file,c", "path to the certificate file");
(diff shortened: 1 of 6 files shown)


libcaf_net/caf/net/http/config.hpp
```diff
@@ -44,6 +44,9 @@ class CAF_NET_EXPORT server_config : public dsl::server_config_value {
 
   /// Store actors that the server should monitor.
   std::vector<strong_actor_ptr> monitored_actors;
+
+  /// Store the maximum request size.
```
Neverlord, line 48:
```suggestion
  /// Store the maximum request size with 0 meaning "default".
```

examples/http/rest.cpp
```diff
@@ -27,13 +27,17 @@ static constexpr uint16_t default_port = 8080;
 
 static constexpr size_t default_max_connections = 128;
 
+static constexpr size_t default_max_request_size
+  = caf::net::http::server::default_max_request_size;
```
Neverlord, line 31: If we want to expose this to users, it should go to `defaults.hpp` as `caf::defaults::net::http_max_request_size`.

libcaf_net/caf/net/http/server_factory.hpp
```diff
@@ -73,16 +73,19 @@ class http_conn_factory
...
+      max_request_size_(max_request_size) {
     // nop
   }
 
   net::socket_manager_ptr make(net::multiplexer* mpx,
                                connection_handle conn) override {
     auto app = net::http::router::make(routes_);
     auto serv = net::http::server::make(std::move(app));
+    if (max_request_size_ > 0)
+      serv->max_request_size(max_request_size_);
```
Neverlord, line 88: I would pass the argument through without checking and instead do the checking in the server, since the server should sanity-check/verify its arguments.

Neverlord (changes requested):
Thanks for the quick followup. 👍

Just some small change requests.

Neverlord (approved):

PR #1658, opened 2023-11-10
Fix shutdown for actors with open streams

### LINKED ISSUE #1657: Stream are not cleaned up during actor shutdown
This was reported in chat (thanks @allspark!): sending an exit message to a stream source does not terminate the actor, because the stream state is not cleaned up properly and the actor remains alive.

Neverlord:
Closes #1657.

There are actually a few smaller fixes/changes in there as well:

- On the actor clock: calling `dispose()` on a scheduled action will now call `dispose()` on the decorator to clean up the actor reference early.
- The `interval_action` looked rather suspicious. We will pass it to the actor clock, so it must be thread safe. Probably leftover from the early flow prototypes. Just uses the regular single-shot action implementation now.
- The deterministic fixture now disposes all pending timeouts on shutdown. This caused my test to hang since actors weren't cleaned up properly.

Also ported the `stream` test to the new framework.


libcaf_core/caf/send.hpp
```diff
@@ -54,6 +54,16 @@ void unsafe_send_as(Source* src, const Dest& dest, Ts&&... xs) {
                                                std::forward<Ts>(xs)...);
 }
 
+template <message_priority P = message_priority::normal, class Dest,
+          class... Ts>
+void unsafe_anon_send(const Dest& dest, Ts&&... xs) {
+  static_assert(sizeof...(Ts) > 0, "no message to send");
```
riemass, line 60: Just to check: this is unsafe because we don't check if the receiver accepts given message?

libcaf_core/caf/flow/op/interval.cpp
```diff
@@ -87,12 +43,16 @@ class interval_sub : public interval_sub_base {
...
-      pending_ = ctx_->delay_until(last_,
-                                   action{make_counted<interval_action>(this)});
+      schedule_next(last_);
     }
   }
 
-  void fire(interval_action* act) override {
+  void schedule_next(coordinator::steady_time_point timeout) {
+    auto act = make_single_shot_action([this] { fire(); });
+    pending_ = ctx_->delay_until(timeout, std::move(act));
```
riemass, line 52: I believe one of the guidelines was to avoid variables where possible?
```suggestion
    pending_ = ctx_->delay_until(timeout,
                                 make_single_shot_action([this] { fire(); }));
```

libcaf_core/caf/scheduled_actor.cpp
```diff
@@ -231,15 +231,27 @@ bool scheduled_actor::cleanup(error&& fail_state, execution_unit* host) {
   // Clear state for open requests.
   awaited_responses_.clear();
   multiplexed_responses_.clear();
-  // Cancel any active flow.
+  // Cancel flows and streams.
+  stream_sources_.clear();
+  if (!stream_subs_.empty()) {
+    decltype(stream_subs_) subs;
+    subs.swap(stream_subs_);
```
riemass, line 238: To better understand this: why did you swap out here instead of clearing in-place? Can `ptr->cancel()` index into `stream_subs_`? Same question for below code with bridges and disposables.

libcaf_core/caf/scheduled_actor.cpp
```diff
@@ -359,13 +371,27 @@ void scheduled_actor::quit(error x) {
...
   while (!watched_disposables_.empty()) {
-    for (auto& ptr : watched_disposables_)
+    decltype(watched_disposables_) disposables;
+    disposables.swap(watched_disposables_);
+    for (auto& ptr : disposables)
       ptr.dispose();
-    watched_disposables_.clear();
-    run_actions();
   }
+  run_actions();
```
riemass, line 394: Since this is repeated, it would look good factored out into its own function, IMHO.

riemass (changes requested):
Looks good. I have a few question to comprehend the changes.

libcaf_core/caf/send.hpp
Neverlord: Yes, you can pass in `strong_actor_ptr` and it will just send the message even without knowing whether it's a typed actor or not.

libcaf_core/caf/scheduled_actor.cpp
Neverlord: Exactly. Calling `cancel`/`drop` might try to erase itself from the map. Hence, we must make sure we're not changing the map while iterating it.

riemass (approved):

PR #1659, opened 2023-11-10
Add fix to dispose subscription on error for merge

shariarriday:
Previously the subscription was not disposed for `on_error` / `failed_subscription` of an input observable. It has been added with a check in `on_next` to ensure that further input observables are not added if subscription fails for any earlier input.

### DIFF
--- a/libcaf_core/caf/flow/op/merge.hpp
+++ b/libcaf_core/caf/flow/op/merge.hpp
@@ -87,6 +87,8 @@ public:
 
   void on_next(const observable<T>& what) override {
     CAF_ASSERT(what);
+    if (!sub_)
+      return;
     auto key = next_key_++;
     inputs_.emplace(key, merge_input<T>{});
     using fwd_impl = forwarder<T, merge_sub, size_t>;
@@ -306,6 +308,7 @@ private:
       else
         ++i;
     }
+    sub_.dispose();
   }
 
   bool done() const noexcept {
--- a/libcaf_core/caf/flow/op/merge.hpp
+++ b/libcaf_core/caf/flow/op/merge.hpp
@@ -100,7 +100,7 @@ public:
     sub_ = nullptr;
     err_ = what;
     stop_inputs();
-    if (inputs_.empty()) {
+    if (out_ && inputs_.empty()) {
       auto out = std::move(out_);
       out.on_error(what);
     }
@@ -308,7 +308,8 @@ private:
       else
         ++i;
     }
-    sub_.dispose();
+    auto sub = std::move(sub_);
+    sub.dispose();
   }
 
   bool done() const noexcept {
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/flow/op/merge.hpp
```diff
@@ -306,6 +308,7 @@ class merge_sub : public subscription::impl_base,
       else
         ++i;
     }
+    sub_.dispose();
```
Neverlord, line 311: Calling `dispose()` might call `on_complete` or `on_error`, so we must always first reset `sub_` and then call `dispose()`. This is the general pattern we use:

```suggestion
    auto sub = std::move(sub_);
    sub.dispose();
```

libcaf_core/tests/legacy/flow/op/merge.cpp
```diff
@@ -86,6 +86,36 @@ SCENARIO("the merge operator combines inputs") {
...
+      THEN("the observer receives the error of first observable") {
+        using ivec = std::vector<int>;
+        auto snk = flow::make_auto_observer<int>();
+        make_observable()
+          .fail<int>(sec::runtime_error)
+          .merge(make_observable().fail<int>(sec::end_of_stream))
+          .subscribe(snk->as_observer());
+        ctx->run();
+        CHECK_EQ(snk->state, flow::observer_state::aborted);
+        CHECK_EQ(snk->err, sec::runtime_error);
```
Neverlord, line 113: It's undefined which error will be observed. The current implementation may always yield `runtime_error`, but that's not required.

```suggestion
```

libcaf_core/caf/flow/op/merge.hpp
```diff
@@ -87,6 +87,8 @@ class merge_sub : public subscription::impl_base,
 
   void on_next(const observable<T>& what) override {
     CAF_ASSERT(what);
+    if (!sub_)
+      return;
```
Neverlord, line 91: Shouldn't we add the same guard to `on_complete` and `on_error`?

Neverlord (changes requested):
Glad to see this sorted out!

libcaf_core/caf/flow/op/merge.hpp
shariarriday: I have not completed the testing yet and there are some scenarios I am trying to check. But, both of `on_complete` and `on_error` resets sub_ to nullptr. I am not sure if it is necessary yet.

shariarriday: I have added a check for `out_` in `on_error`. But, as we are not doing any operation in sub_ and setting it to nullptr, I don't think it is required to add `if (!sub_)` check in `on_error` and `on_complete`.

Neverlord (approved):

PR #1661, opened 2023-11-11
Minor tweaks

Neverlord:
Just two small tweaks for things I've noticed.

### DIFF
--- a/robot/http/client.robot
+++ b/robot/http/client.robot
@@ -9,7 +9,7 @@ Test Teardown       Terminate HTTP Server
 
 
 *** Variables ***
-${HTTP_URL}         http://localhost:8000
+${HTTP_URL}         http://localhost:55516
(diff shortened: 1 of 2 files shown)


robot/http/client.robot
```diff
@@ -9,7 +9,7 @@ Test Teardown       Terminate HTTP Server
 
 
 *** Variables ***
-${HTTP_URL}         http://localhost:8000
+${HTTP_URL}         http://localhost:55516
```
riemass, line 12: Are you changing the port number to be consistent with the rest.robot?

riemass (approved):
Looking good.

robot/http/client.robot
Neverlord: Consistency, yes, but the test also failed on me because something else was using 8000.

PR #1663, opened 2023-11-14
Implement new release semantics for flow handles

Neverlord:
The subscription and the observer hold mutual references, which makes it very easy to accidentally destroy the current object when disposing objects. The `release_later` functions delay releasing the reference count of discarded objects until all actions have completed. This makes it safe to discard a subscription or observer, even if both objects end up discard each other by delaying the destruction of the objects.

### DIFF
--- a/libcaf_core/caf/async/blocking_consumer.hpp
+++ b/libcaf_core/caf/async/blocking_consumer.hpp
@@ -69,6 +69,7 @@ public:
     }
 
     void on_complete() {
+      // nop
     }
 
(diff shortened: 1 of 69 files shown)


Neverlord:
@shariarriday no idea why `clang-format` suddenly started complaining. The errors are in unrelated files, so I've opened #1664 to fix that (and update GH action versions while we're at it). We'll just ignore the `check-formatting` error in this PR for the time being.

shariarriday (approved):

PR #1670, opened 2023-11-17
Add HTTP Client DSL

riemass:
Adds high level DSL for creating HTTP requests, yielding a `async::future` response. 
Relates #1392


examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+  auto& auth = resource.authority();
+  auto port = auth.port != 0 ? auth.port : defatul_port(resource.scheme());
+  auto conn
+    = http::with(sys)
+        // Optionally enable TLS.
+        .context(ssl::context::enable(use_ssl)
+                   .and_then(ssl::emplace_client(ssl::tls::v1_2))
+                   .and_then(ssl::load_verify_file_if(ca_file)))
+        // Connect to "$host:$port".
+        .connect(std::get<0>(auth.host), port)
```
Neverlord, line 92: Disassembling the URI should not be necessary. This makes it very brittle to use and hard to understand. This should simply be:

```c++
connect(resource)
```

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+    std::cerr << "*** missing mandatory option --resource" << std::endl;
+    return EXIT_FAILURE;
+  }
+  auto resource = std::move(*maybe_resource);
+  auto use_ssl = caf::get_or(cfg, "tls.enable", false);
+  auto ca_file = caf::get_as<std::string>(cfg, "tls.ca-file");
+  if (resource.scheme() != "http" && resource.scheme() != "https") {
+    std::cerr << "*** only HTTP is supported at the moment" << std::endl;
+    return EXIT_FAILURE;
+  }
```
Neverlord, line 80: Should be checked by `connect` further down, not by the user.

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+      .add<std::string>("payload,p", "Optional payload to send");
+    opt_group{custom_options_, "tls"} //
+      .add<bool>("enable", "enables encryption via TLS")
+      .add<std::string>("ca-file", "CA file for trusted servers");
+  }
+
+  caf::settings dump_content() const override {
+    auto result = actor_system_config::dump_content();
+    caf::put_missing(result, "method", default_method);
+    caf::put_missing(result, "payload", "");
```
Neverlord, line 54: I wouldn't include it if it has no default. Just like you've omitted `resource`.

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+// static constexpr std::string_view default_host = "localhost";
+
+constexpr auto default_method = http::method::get;
+
+// -- configuration setup ------------------------------------------------------
+struct config : actor_system_config {
+  config() {
+    opt_group{custom_options_, "global"}
+      .add<http::method>("method,m", "HTTP method to use")
+      .add<uri>("resource,r", "Requested resource (URL)")
```
Neverlord, line 44: I think using a positional argument would be more natural, i.e., `client http://my-host/my-path` (as opposed to `client -r http://my-host/my-path`.

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+        .context(ssl::context::enable(use_ssl)
+                   .and_then(ssl::emplace_client(ssl::tls::v1_2))
+                   .and_then(ssl::load_verify_file_if(ca_file)))
+        // Connect to "$host:$port".
+        .connect(std::get<0>(auth.host), port)
+        // If we don't succeed at first, try up to 5 times with 1s delay.
+        .retry_delay(1s)
+        .max_retry_count(5)
+        // After connecting, send a get request.
+        .request(method, resource.path_query_fragment(), payload);
```
Neverlord, line 97: Again, I'd just pass the URI along. The user shouldn't care about the nitty-gritty details. Ideally, the `request` should be overloaded to allow both a path (as string) and an URI.

```suggestion
        .request(method, resource, payload);
```

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+        std::cout << std::string_view{reinterpret_cast<const char*>(
+                                        r.body().data()),
+                                      r.body().size()}
+                  << std::endl;
+      },
+      [](const error& err) {
+        std::cerr << "*** Error received: " << to_string(err) << std::endl;
+      });
+  });
+  sys.await_all_actors_done();
```
Neverlord, line 121: Why did you put this here? CAF will wait for `testee` anyways, no?

libcaf_core/caf/detail/type_traits.hpp
```diff
@@ -1018,6 +1018,12 @@ struct has_init_host_system {
 template <class T>
 constexpr bool has_init_host_system_v = has_init_host_system<T>::value;
 
+/// Drop-in replacement for C++23's std::to_underlying.
+template <class Enum>
+constexpr std::underlying_type_t<Enum> to_underlying(Enum e) noexcept {
```
Neverlord, line 1023:
```suggestion
[[nodiscard]] constexpr auto to_underlying(Enum e) noexcept {
```

libcaf_net/caf/net/http/client_factory.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include <cstdint>
+#include <functional>
+#include <type_traits>
+#include <utility>
+#include <variant>
+
+namespace caf::net::http {
+
+/// Factory for the `with(...).connect(...).start(...)` DSL.
+class client_factory
```
Neverlord, line 35: Misses CAF_NET_EXPORT.

libcaf_net/caf/net/http/config.hpp
```diff
@@ -46,4 +47,27 @@ class CAF_NET_EXPORT server_config : public dsl::server_config_value {
   std::vector<strong_actor_ptr> monitored_actors;
 };
 
+struct client_config : public dsl::client_config_value {
```
Neverlord, line 50: Misses CAF_NET_EXPORT.

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+  auto ca_file = caf::get_as<std::string>(cfg, "tls.ca-file");
+  if (resource.scheme() != "http" && resource.scheme() != "https") {
+    std::cerr << "*** only HTTP is supported at the moment" << std::endl;
+    return EXIT_FAILURE;
+  }
+  auto method = caf::get_or(cfg, "method", default_method);
+  auto payload = caf::get_or(cfg, "payload", ""sv);
+  auto& auth = resource.authority();
+  auto port = auth.port != 0 ? auth.port : defatul_port(resource.scheme());
+  auto conn
```
Neverlord, line 85: This does not represent a connection. It's a future for the response plus the disposable for canceling the request. If nothing better comes to mind, `result` would still be a better name.

examples/http/client.cpp
```diff
@@ -0,0 +1,125 @@
...
+        .retry_delay(1s)
+        .max_retry_count(5)
+        // After connecting, send a get request.
+        .request(method, resource.path_query_fragment(), payload);
+  if (!conn) {
+    std::cerr << "*** Failed to initiate connection: "
+              << to_string(conn.error()) << std::endl;
+    return EXIT_FAILURE;
+  }
+  auto testee = sys.spawn([&conn](event_based_actor* self) {
```
Neverlord, line 103: Ignoring this bit. I assume we're going to use the blocking API here?

libcaf_net/caf/net/http/async_client.cpp
```diff
@@ -0,0 +1,66 @@
...
+void async_client::abort(const caf::error& reason) {
+  CAF_LOG_ERROR("Response abortet with: " << to_string(reason));
+  resp_.set_error(reason);
+}
+
+// -- http::client lower layer implementation --------------------------------
+
+caf::error async_client::start(http::lower_layer::client* ll) {
+  down = ll;
+  // Send request.
```
Neverlord, line 32: The comment is literally repeating the function name. ;)

```suggestion
```

libcaf_net/caf/net/http/async_client.cpp
```diff
@@ -0,0 +1,66 @@
...
+    fields.container().emplace_back(key, value);
+  });
+  http::response resp{static_cast<http::status>(hdr.status()),
+                      std::move(fields),
+                      byte_buffer{payload.begin(), payload.end()}};
+  resp_.set_value(std::move(resp));
+  return static_cast<ptrdiff_t>(payload.size());
+}
+
+// -- Utility functions ----
```
Neverlord, line 50: Widen to 80 characters and we never use uppercase in separators.

libcaf_net/caf/net/http/async_client.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+  ptrdiff_t consume(const http::response_header& hdr,
+                    caf::const_byte_span payload) override;
+
+  async::future<response> get_future() const {
+    return resp_.get_future();
+  }
+
+private:
+  // -- Utility functions ----
```
Neverlord, line 61: Either drop of rename the section to `member variables`.

libcaf_net/caf/net/http/client_factory.hpp
```diff
@@ -0,0 +1,122 @@
...
+  : public dsl::client_factory_base<client_config, client_factory> {
+public:
+  using super = dsl::client_factory_base<client_config, client_factory>;
+
+  using config_type = client_config;
+
+  using super::super;
+
+  /// Add header field to the request.
```
Neverlord, line 44:
```suggestion
  /// Add an additional HTTP header field to the request.
```

robot/CMakeLists.txt
```diff
@@ -145,19 +145,12 @@ if(TARGET CAF::net)
...
-    DEPENDENCIES
-      CAF::net)
-
   add_test(
     NAME "robot-http-client"
     COMMAND
       ${Python_EXECUTABLE}
         -m robot
-        --variable BINARY_PATH:$<TARGET_FILE:caf-robot-http-client-driver>
+        --variable BINARY_PATH:$<TARGET_FILE:client>
```
Neverlord, line 153: That target only exists when building the examples in the first place. I think the `if` statement above should read as `if(TARGET CAF::net AND CAF_ENABLE_EXAMPLES)`.

Not a request for this PR, more of a general observation. I think we should rename the example targets. A binary called `client` is too ambiguous.

Neverlord (changes requested):
Nice seeing this taking shape.

examples/http/client.cpp
riemass: That was my thought as well, but when I remove this line, I get a segfault with the location pointing to `future::then`. I will examine this to get more detail, maybe it was caused by something else.

examples/http/client.cpp
riemass: That's something that was on my mind. One thing confusing me here is that an URI contains everything needed to make the request, so having a code like this:
```
auto result = http::with(sys)
        .connect(resource)
        .get(resource);
```
I assume you suggest  `connect` only reading the address and port part of the resource, and `get` reading the path and query part? This would be easier to implement. 
Whats probably more intuitive is `connect` reading everything, like:
```
auto result = http::with(sys)
        .connect(resource)
        .get();
```
However this would require us overloading the DSL and splitting the API to different types when we connect to an URI vs when we connect to an host/port part.

Neverlord: Hm, this brings up a good point. The URI doesn't just configure host and port. It also tells whether to use TLS or not. You're currently using the `context` factories for that, but that shouldn't be necessary either.

```c++
auto result = http::with(sys)
        .connect(resource)
        .get();
```

I really like that. That would probably cover >99% of use cases anyways. I'd go as far as to *only* implement the URI interface. You can always just assemble an URI, so I don't see much point in making the implementation complicated if the user can just put together an URI for `connect`.

robot/CMakeLists.txt
riemass: Agree on this, we should consider prefixing the targets by path.

examples/http/client.cpp
riemass: Taking the capture by copy solved this.

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+        .retry_delay(1s)
+        .max_retry_count(5)
+        // After connecting, send a get request.
+        .request(method, payload);
+  if (!result) {
+    std::cerr << "*** Failed to initiate connection: "
+              << to_string(result.error()) << std::endl;
+    return EXIT_FAILURE;
+  }
+  sys.spawn([result](event_based_actor* self) {
```
Neverlord, line 87: We don't care for the disposable, so I'd capture only the future:

```suggestion
  sys.spawn([res = result->first](event_based_actor* self) {
```

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+        for (const auto& [key, value] : r.header_fields())
+          if (key != "Date")
+            std::cout << key << ':' << value << std::endl;
+        std::cout << std::string_view{reinterpret_cast<const char*>(
+                                        r.body().data()),
+                                      r.body().size()}
+                  << std::endl;
+      },
+      [](const error& err) {
+        std::cerr << "*** Error received: " << to_string(err) << std::endl;
```
Neverlord, line 102:
```suggestion
        std::cerr << "*** HTTP request failed: " << to_string(err) << std::endl;
```

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+        .request(method, payload);
+  if (!result) {
+    std::cerr << "*** Failed to initiate connection: "
+              << to_string(result.error()) << std::endl;
+    return EXIT_FAILURE;
+  }
+  sys.spawn([result](event_based_actor* self) {
+    result->first.bind_to(self).then(
+      [](const http::response& r) {
+        std::cout << "HTTP/1.0 " << detail::to_underlying(r.code()) << ' '
```
Neverlord, line 90:
```suggestion
        std::cout << "Server responded with HTTP " << detail::to_underlying(r.code()) << ": "
```

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+        // After connecting, send a get request.
+        .request(method, payload);
+  if (!result) {
+    std::cerr << "*** Failed to initiate connection: "
+              << to_string(result.error()) << std::endl;
+    return EXIT_FAILURE;
+  }
+  sys.spawn([result](event_based_actor* self) {
+    result->first.bind_to(self).then(
+      [](const http::response& r) {
```
Neverlord, line 89: I think the output is quite cryptic in this block. I'd like to have it a bite more structured, e.g. something like this:

```
Server responded with HTTP 200: OK
Header fields:
- Foo: Bar
- Something: Else
Payload:
Hello, world!
```

For the payload: we should first check whether the string actually contains a valid (UTF-8) string first. We already have the checks in `detail::rfc3629`, we could just add a function like `bool is_valid_utf8(const_byte_span bytes)` to the `byte_span.hpp` header and use it here.

If it's UTF-8: print to stdout.
Otherwise: print as hex dump. Something like this:

```cpp
// Print 8 bytes per row in hex.
while (!bytes.empty()) {
  auto row = bytes.subspan(0, 8);
  for (auto byte : bytes)
    printf("%02x", static_cast<int>(byte));
  std::cout << std::endl;
  bytes = bytes.subspan(8);
}
```

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+constexpr auto default_method = http::method::get;
+
+// -- configuration setup ------------------------------------------------------
+struct config : actor_system_config {
+  config() {
+    opt_group{custom_options_, "global"}
+      .add<http::method>("method,m", "HTTP method to use")
+      .add<std::string>("payload,p", "Optional payload to send");
+    opt_group{custom_options_, "tls"} //
+      .add<bool>("enable", "enables encryption via TLS")
```
Neverlord, line 43:
```suggestion
```

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto use_ssl = caf::get_or(cfg, "tls.enable", false);
+  auto ca_file = caf::get_as<std::string>(cfg, "tls.ca-file");
+  auto method = caf::get_or(cfg, "method", default_method);
+  auto payload = caf::get_or(cfg, "payload", ""sv);
+  auto result
+    = http::with(sys)
+        // Optionally enable TLS.
+        .context(ssl::context::enable(use_ssl)
+                   .and_then(ssl::emplace_client(ssl::tls::v1_2))
+                   .and_then(ssl::load_verify_file_if(ca_file)))
```
Neverlord, line 74: Not happy with this. The factory should know when to initialize TLS by itself (depending on whether the scheme is HTTP or HTTPS). As a user, I'd want something like this:

```
http::with(sys)
.tls_verify_file(caf::get_as<std::string>(cfg, "tls.ca-file"))
.connect(resource)
...
```

With `tls_verify_file` having no effect if:
- `caf::get_as` does not return a value
- the scheme is HTTP

examples/http/client.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+int caf_main(caf::actor_system& sys, const config& cfg) {
+  // Get URI from config (positional argument).
+  if (cfg.remainder.size() != 1) {
+    std::cerr << "*** expected mandatory positional argument: URL" << std::endl;
+    return EXIT_FAILURE;
+  }
+  uri resource;
+  if (auto err = parse(cfg.remainder[0], resource)) {
+    std::cerr << "*** failed to parse URI" << std::endl;
```
Neverlord, line 62:
```suggestion
    std::cerr << "*** failed to parse URI: " << to_string(err) << std::endl;
```

Neverlord (changes requested):

examples/http/client.cpp
```diff
@@ -0,0 +1,123 @@
...
+          auto bytes = r.body();
+          while (!bytes.empty()) {
+            auto row = bytes.subspan(0, 8);
+            for (auto byte : row)
+              printf("%02x", std::to_integer<int>(byte));
+            std::cout << std::endl;
+            if (bytes.size() < 8)
+              bytes = bytes.subspan(bytes.size());
+            else
+              bytes = bytes.subspan(8);
```
Neverlord, line 112: I don't think this is correct. The first argument is the offset. To chop off the first 8 bytes, this should `bytes.subspan(8)`, shouldn't it? https://en.cppreference.com/w/cpp/container/span/subspan

Neverlord: Oh boy...

> The behavior is undefined if either Offset or Count is out of range.

What a dreadful design. Well, in this case:

```suggestion
            bytes = bytes.subspan(std::min(bytes.size(), 8u));
```

Neverlord: Sorry for comment-spamming. Since the API is so terrible, this means the `subspan(0, 8)` above is also unsafe.

Just split this into a utility lambda that returns the row (up to 8 bytes) and the remainder. Could use like this:

```c++
const_byte_span row;
std::tie(row, byte) = splitAt(bytes, 8);
```

riemass: No problem, I'm glad this is the only critique :smile:. I figured the first overload out, but yeah the first one also had the bug.

libcaf_net/caf/net/http/client_factory.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+namespace caf::net::http {
+
+/// Factory for the `with(...).connect(...).start(...)` DSL.
+class CAF_NET_EXPORT client_factory
+  : public dsl::client_factory_base<client_config, client_factory> {
+public:
+  using super = dsl::client_factory_base<client_config, client_factory>;
+
+  using config_type = client_config;
```
Neverlord, line 40: The `client_config` is built with an eager setup in mind, where the user can provide a socket, ready-to-use connection or only the credentials. I simply wouldn't use it here.

We'll have either a lazy config (URI + optional context factory) or an error. Here's an outline for what I mean:

```c++
struct client_config {
  struct lazy { // Note: *not* deriving `has_ctx` here.
    /// The locator for the remote resource (HTTP or HTTPS).
    uri locator;

    /// Factory function for creating an SSL/TLS context when using HTTPS.
    std::function<...> ctx_factory;

    /// The delay between connection attempts.
    timespan retry_delay = std::chrono::seconds{1};

    /// The timeout when trying to connect.
    timespan connection_timeout = infinite;

    /// The maximum amount of retries.
    size_t max_retry_count = 0;
  };

  ...
};

/// inside client_factory

  /// Sets the optional SSL context factory.
  /// @param ctx_fn The function creating the SSL context for encryption.
  /// @returns a reference to `*this`.
  template <class Factory>
  client_factory& context_factory(Factory&& factory) {
    if (auto* lazy = get_if<client_config::lazy>(config().data))
      lazy->ctx_factory = std::move(factory);
    return *this;
  }
```

Probably means we need to copy the few functions from `caf::net::dsl::client_factory_base` over (`retry_delay` etc.) but I wouldn't mind that too much. I also not super happy with the entire DSL stuff myself, btw. At some point we need to make a pass over it and simplify it.

libcaf_net/caf/net/http/client_factory.hpp
```diff
@@ -0,0 +1,174 @@
...
+    auto ret = app_t->get_future();
+    auto http_client = http::client::make(std::move(app_t));
+    auto transport = transport_t::make(std::move(conn), std::move(http_client));
+    transport->active_policy().connect();
+    auto ptr = net::socket_manager::make(cfg.mpx, std::move(transport));
+    cfg.mpx->start(ptr);
+    return std::pair{std::move(ret), disposable{std::move(ptr)}};
+  }
+
+  return_t do_start(config_type& cfg, dsl::client_config::lazy& data) {
```
Neverlord, line 131: There's quite a bit of code in there. We're not in a template here, so why not put this into a `.cpp` file? Probably also applies to `request`.

examples/http/client.cpp
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  caf::settings dump_content() const override {
+    auto result = actor_system_config::dump_content();
+    caf::put_missing(result, "method", default_method);
+    return result;
+  }
+};
+
+std::pair<const_byte_span, const_byte_span> split_at(const_byte_span bytes,
```
Neverlord, line 53:
```suggestion
auto split_at(const_byte_span bytes,
```

Also, this is used exactly in one location. So I'd put it as a lambda with no capture to where we actually use it in order to keep related code close together.

Neverlord (changes requested):

examples/http/client.cpp
Neverlord: > I'm glad this is the only critique

Erm... this was just a comment on something that I've noted before starting the actual review. 😅

libcaf_net/caf/net/http/client_factory.hpp
riemass: I'm kind of against specializing this feature only to HTTP. 
- It wouldn't really play nice with the idea of the client being able to use both interfaces (`.context` and `.context_factory`). 
- If we don't inherit `has_ctx` then the client won't have the `shared_ptr<ssl::context` ctx`, and special functions that come from it like `connection_with_ctx`, so the code duplication would be considerable.
- We would have to do this setup again if we wanted to introduce `has_lazy_context` to websocket (`ws` vs `wss`).
Could we think of any other way to refactor this? I could factor out the factory stuff from `has_ctx` into `has_ctx_factory` and include it everywhere? All in all, I agree that the DSL stuff could use some refactoring for clarity.

Neverlord: > It wouldn't really play nice with the idea of the client being able to use both interfaces (`.context` and `.context_factory`).

My mental model of these two is that one is eager and one is lazy. You cannot be both, which is why I'm unhappy with `has_ctx` trying to be both. These two are opposites, pick one or the other.

If do want to keep things uniformly, then we could make `has_ctx` lazy instead and not have an eager version at all. Then, `has_ctx` would only have one member: `make_ctx`. The eager `.context()` version would then simply set `make_ctx` to return the eagerly constructed object while the `.context_factory()` version would override the `make_ctx` member with whatever factory the user provides. A default-constructed `make_ctx` member would then mean "no SSL".

> If we don't inherit `has_ctx` then the client won't have the `shared_ptr<ssl::context> ctx`, and special functions that come from it like `connection_with_ctx`, so the code duplication would be considerable.

We could always just do something like `has_ctx{std::make_shared<ssl::context>(factory())}.connection_with_ctx(...)` if we need to at the point where we lazily spin up everything. However, I think there's not much code duplication, because `connection_with_ctx` really just wraps `ctx->new_connection(fd)` with a few if-else branches to check whether we so SSL or not plus error handling.

> We would have to do this setup again if we wanted to introduce has_lazy_context to websocket (`ws` vs `wss`).

Wouldn't the `client_config` we implement here be generic enough to push it to the `dsl` namespace as a followup?

I think either that or make `has_ctx` be always lazy. Both works for me.

riemass: I was also drawing the parallel from the `base_config` used by the `with_t` - alias from `generic_config_value`, which has `has_ctx` as well and transfers the information to the client or server config. In that way it should have both versions. I agree it can't be both at the same time, but needs to support both.  
After some tryouts, what I'd find best would be to displace the `has_ctx` from lazy, server, conn, ... variant types to the config directly.

riemass: Done. Reworked the lazy parts.

libcaf_net/caf/net/http/client_factory.hpp
riemass: Separated.

riemass (commented):
Thanks for the added value. Some of the points you addressed I've expected as review feedback, like where to put the `ssl` helper functions, or the fact that some of the code is repeated.
There are a few places I'm not entirely happy with, like the `shared_ptr` to `ssl::context`, or that we have 6 `ssl` helper functions, but I don't see a value in overthinking this anymore, and OFC I approve the changes.
I'll have to modify the payload to accept binary data, I've kind of forgot about that review item. I'll do that one next.

libcaf_net/caf/net/dsl/has_context.hpp
```diff
@@ -20,15 +20,20 @@ class has_context : public Base {
...
     auto& dref = static_cast<Subtype&>(*this);
     auto& cfg = dref.config();
-    if (auto* ptr = cfg.as_has_ctx())
-      ptr->ctx = std::make_shared<ssl::context>(std::move(ctx));
-    else if (cfg)
+    if (auto* ptr = cfg.as_has_make_ctx()) {
+      auto ctx_ptr = std::make_shared<ssl::context>(std::move(ctx));
+      ptr->make_ctx([ctx_ptr]() -> expected<ssl::context> { //
+        return std::move(*ctx_ptr);
+      });
```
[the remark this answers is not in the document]
Neverlord: ACK, changed to `shared_ptr` again.

libcaf_net/caf/net/dsl/has_make_ctx.hpp
```diff
@@ -0,0 +1,71 @@
...
+    return std::visit(get_ptr, data);
+  }
+
+  template <typename F>
+  void make_ctx(F&& factory) noexcept {
+    make_ctx_ = std::forward<F>(factory);
+  }
+
+  auto& make_ctx() noexcept {
+    return make_ctx_;
```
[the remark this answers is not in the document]
Neverlord: Having trivial getters and setters is nonsensical anyways. I've simply made the member public. We still need to check it before calling, though. A default factory won't do, because not setting a factory at all is an error when starting an HTTPS server but we can provide a default when starting an HTTPS server. So we need that distinction.

Neverlord (approved):

PR #1672, opened 2023-11-19
Translate legacy tests for buffer operator

shariarriday:
Updates the legacy tests for merge operator to use the new test framework and deterministic framework

### DIFF
--- a/libcaf_core/caf/flow/op/buffer.test.cpp
+++ b/libcaf_core/caf/flow/op/buffer.test.cpp
@@ -43,25 +43,6 @@ struct noskip_trait {
 };
 
 struct fixture : test::fixture::deterministic, test::fixture::flow {
-  template <class Impl>
-  void add_subs(intrusive_ptr<Impl> uut) {
-    auto data_sub
-      = make_counted<test::fixture::flow::passive_subscription_impl>(
-        coordinator());
-    uut->fwd_on_subscribe(fwd_data,
-                          caf::flow::subscription{std::move(data_sub)});
-    auto ctrl_sub
-      = make_counted<test::fixture::flow::passive_subscription_impl>(
-        coordinator());
-    uut->fwd_on_subscribe(fwd_ctrl,
-                          caf::flow::subscription{std::move(ctrl_sub)});
-  }
-
-  template <class T>
-  auto trivial_obs() {
-    return test::fixture::flow::make_trivial_observable<T>(coordinator());
-  }
-
   // Similar to buffer::subscribe, but returns a buffer_sub pointer instead of
   //  type-erasing it into a disposable.
   template <class Trait = noskip_trait>
@@ -198,16 +179,11 @@ SCENARIO("the buffer operator forwards errors") {
   }
 }
 
-} // WITH_FIXTURE(fixture)
-
-WITH_FIXTURE(fixture) {
-
 SCENARIO("buffers dispose unexpected subscriptions") {
   GIVEN("an initialized buffer operator") {
     WHEN("calling on_subscribe with unexpected subscriptions") {
       THEN("the buffer disposes them immediately") {
         auto snk = flow::make_passive_observer<cow_vector<int>>();
-        auto grd = make_unsubscribe_guard(snk);
         auto uut = raw_sub(3, make_observable().never<int>(),
                            make_observable().never<int64_t>(),
                            snk->as_observer());
@@ -222,6 +198,9 @@ SCENARIO("buffers dispose unexpected subscriptions") {
         check(!uut->disposed());
         check(data_sub->disposed());
         check(ctrl_sub->disposed());
+        uut->dispose();
+        run_flows();
+        check(uut->disposed());
       }
     }
   }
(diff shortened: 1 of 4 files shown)


libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,577 @@
...
+    auto ctrl_sub = make_counted<passive_subscription_impl>(coordinator());
+    uut->fwd_on_subscribe(fwd_ctrl,
+                          caf::flow::subscription{std::move(ctrl_sub)});
+  }
+
+  template <class T>
+  auto trivial_obs() {
+    return flow::make_trivial_observable<T>(coordinator());
+  }
+};
```
Neverlord, line 97: Why do we need two separate fixtures? Did you try something like this instead?

```c++
struct fixture : test::fixture::deterministic, test::fixture::flow {};
```

libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,577 @@
...
+          cow_vector<int>{64, 128},
+        };
+        ctx->make_observable()
+          .from_container(inputs) //
+          .buffer(3)
+          .for_each([&outputs](const cow_vector<int>& xs) {
+            outputs.emplace_back(xs);
+          });
+        ctx->run();
+        check_eq(outputs, expected);
```
Neverlord, line 119: We have `collect()` now. Why didn't you use it here?

libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,577 @@
...
+          pub.as_observable()
+            .observe_on(self) //
+            .buffer(3, 1s)
+            .do_on_complete([closed] { *closed = true; })
+            .for_each([outputs](const cow_vector<int>& xs) {
+              outputs->emplace_back(xs);
+            });
+        });
+        dispatch_messages();
+        print_info("emit the first six items");
```
Neverlord, line 147: I think `debug` level would be more appropriate.

Neverlord (changes requested):
Just a first pass, I'll go through the remaining tests after the first round of changes.

libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,499 @@
...
+      = make_counted<test::fixture::flow::passive_subscription_impl>(
+        coordinator());
+    uut->fwd_on_subscribe(fwd_ctrl,
+                          caf::flow::subscription{std::move(ctrl_sub)});
+  }
+
+  template <class T>
+  auto trivial_obs() {
+    return test::fixture::flow::make_trivial_observable<T>(coordinator());
+  }
```
Neverlord, line 63: I wonder if we really need this. I think the tests would work just as well with `never` and wouldn't require us to hand-roll a new observable implementation for testing.

libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,499 @@
...
+      = make_counted<test::fixture::flow::passive_subscription_impl>(
+        coordinator());
+    uut->fwd_on_subscribe(fwd_data,
+                          caf::flow::subscription{std::move(data_sub)});
+    auto ctrl_sub
+      = make_counted<test::fixture::flow::passive_subscription_impl>(
+        coordinator());
+    uut->fwd_on_subscribe(fwd_ctrl,
+                          caf::flow::subscription{std::move(ctrl_sub)});
+  }
```
Neverlord, line 58: Did you check whether this still actually does anything after getting rid of `nil_observable`? I believe we can just rip this out as well.

libcaf_core/caf/flow/op/buffer.test.cpp
```diff
@@ -0,0 +1,499 @@
...
+      THEN("the observer receives an on_complete event") {
+        auto snk = flow::make_passive_observer<cow_vector<int>>();
+        auto uut = raw_sub<skip_trait>(3, trivial_obs<int>(),
+                                       trivial_obs<int64_t>(),
+                                       snk->as_observer());
+        add_subs(uut);
+        snk->request(42);
+        run_flows();
+        uut->dispose();
+        run_flows();
```
Neverlord, line 460: I think this is the better (and safer) option to clean up tests. Please drop the unsubscribe guard and use this instead. Can be a convenience function in the fixture or just add these two lines after each test setup.

With this gone as well, I believe the fixture should remain unchanged. Which is a good thing, we had quite a bit of unnecessary complexity in there that we should actively strive to get rid of.

Neverlord (changes requested):

libcaf_core/caf/flow/op/buffer.test.cpp
shariarriday: We still have `passive_subscription_impl` that was added to this fixture. Should I look into removing it from the fixture as well?

Neverlord: Yep, give it a shot. Less is more.

Neverlord (approved):

Issue #1646, opened 2023-11-05
Port flow tests from the legacy DSL to the new framework


Neverlord:
The directory `libcaf_core/tests/legacy/flow` still has unit tests for the flow API written in the old style. They need to be migrated to the new framework with our new `flow` fixture.

@shariarriday I've attached the Python script that I've used to help modernize old tests. It's just a bunch of text replacement to fix includes and transform things like `CHECK_EQ` to `check_eq`, but it speeds things up a bit. Use with `python3 modernize.py <path-to-file>` (had to add `.txt` for GitHub, just drop the extra file ending after download).

[modernize.py.txt](https://github.com/actor-framework/actor-framework/files/13258856/modernize.py.txt)

PR #1676, opened 2023-11-26
Port flow tests to the updated tests framework

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/1646

Updates the tests for following files

- mixed
- multicaster
- observe_on
- publish
- single
- for_each
- flat_map
- concat_map
- generation


libcaf_core/caf/flow/for_each.test.cpp
```diff
@@ -58,29 +46,31 @@ SCENARIO("for_each iterates all values in a stream") {
...
           auto outputs = std::vector<int>{};
-          ctx->make_observable()
+          make_observable()
             .repeat(7) //
             .as_observable()
             .take(7)
             .map([](int x) { return x * 3; })
-            .do_on_error([](const error& err) { FAIL("on_error: " << err); })
+            .do_on_error([this](const error& err) {
+              fail(std::string_view("on_error: " + caf::to_string(err)));
```
Neverlord, line 67:
```suggestion
              fail("on_error: " + caf::to_string(err));
```

libcaf_core/caf/flow/multicaster.test.cpp
```diff
@@ -0,0 +1,130 @@
...
+        auto print_stats = [&] {
+          const auto& states = uut.impl().observers();
+          for (size_t index = 0; index < states.size(); ++index) {
+            printf("observer %d: buf size = %d; demand = %d\n", (int) index,
+                   (int) states[index]->buf.size(),
+                   (int) states[index]->demand);
+          }
+        };
+        print_stats();
+
```
Neverlord, line 48: Either drop or use `print_debug`.

```suggestion
```

Neverlord (changes requested):

libcaf_core/caf/flow/for_each.test.cpp
shariarriday: @Neverlord We can not use `fail("on_error: " + caf::to_string(err));` as we will face the following error `cannot convert from 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>' to 'caf::format_string_with_location'`.

Neverlord: Right, this takes a format string. Then there's no need at all to build a string in the first place: `fail("on_error: {}", caf::to_string(err))`. 🙂

Neverlord (approved):

PR #1677, opened 2023-11-28
Implement caf::chunk

riemass:
This PR implements `caf::chunk` as a standalone type.
`lp::frame` becomes an alias for this type, and uses only the binary interface.
Meanwhile `webSocket::frame` has the same internal type as the chunk and exposes both, binary and textual chunk interface.

EDIT: Added a simple binary file IO interface. Feel free to comment on the API. Since we use blocking OS calls, we read exactly `num` bytes every time, except on the last call. Otherwise, I'd go with a policy based read options, like we use in the `caf::net::transport`. 

Relates #1573.


libcaf_core/caf/async/file.test.cpp
```diff
@@ -115,4 +133,51 @@ TEST("async file I/O") {
   }
 }
 
+TEST("async binary file I/O") {
+  actor_system_config cfg;
+  actor_system sys{cfg};
+  auto prom = std::make_shared<std::promise<expected<byte_buffer>>>();
+  auto res = prom->get_future();
+  SECTION("read bytes from file") {
+    auto pub = async::file(sys, quotes_file).read_bytes().run([](auto&& src) {
```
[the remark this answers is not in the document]
riemass: I taught I could get away with the text file, but you're right. I've added a binary file and rebased to get the coverage up. You can review the last commit.

libcaf_core/caf/async/file.hpp
```diff
@@ -13,9 +13,11 @@
 namespace caf::detail {
 
 /// A generator that emits characters from a file.
+template <typename T>
```
Neverlord, line 16: Style: we always use `class` to refer to template parameters.

```suggestion
template <class T>
```

libcaf_core/caf/async/file.hpp
```diff
@@ -52,7 +56,10 @@ class file_reader {
   template <class Step, class... Steps>
   void pull(size_t n, Step& step, Steps&... steps) {
     if (file_ == nullptr) {
-      file_ = fopen(path_.c_str(), "r");
+      if constexpr (std::is_same_v<T, std::byte>)
+        file_ = fopen(path_.c_str(), "rb");
+      else
+        file_ = fopen(path_.c_str(), "r");
```
Neverlord, line 62:
```suggestion
      file_ = fopen(path_.c_str(), mode());
```

With `mode` being a private member function, e.g.:

```c++
  static constexpr const char* mode() noexcept {
      if constexpr (std::is_same_v<T, std::byte>)
        return "rb";
      else
        return "r";
  }
```

libcaf_core/caf/async/file.test.cpp
```diff
@@ -39,7 +39,7 @@ constexpr std::string_view quotes_numbered_lines = R"_(1:
 7:
 )_";
 
-TEST("async file I/O") {
+TEST("async textual file I/O") {
```
Neverlord, line 42:
```suggestion
TEST("async text file I/O") {
```

libcaf_core/caf/async/file.test.cpp
```diff
@@ -115,4 +115,55 @@ TEST("async file I/O") {
...
+    bytes.push_back(static_cast<std::byte>(i));
+  actor_system_config cfg;
+  actor_system sys{cfg};
+  auto prom = std::make_shared<std::promise<expected<byte_buffer>>>();
+  auto res = prom->get_future();
+  SECTION("read bytes from file") {
+    auto pub
+      = async::file(sys, byte_range_file).read_bytes().run([](auto&& src) {
+          return std::forward<decltype(src)>(src)
+            .map([](std::byte b) { return b ^ std::byte{0xFF}; })
```
Neverlord, line 132: Why did you put that `b ^ 0xFF` no-op here?

libcaf_core/caf/chunk.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  chunk(chunk&&) = default;
+
+  chunk(const chunk&) = default;
+
+  chunk& operator=(chunk&&) = default;
+
+  chunk& operator=(const chunk&) = default;
+
+  explicit chunk(const_byte_span buf);
```
Neverlord, line 115: Please add an `explicit` constructor from `intrusive_ptr<data>` (to implement `as_chunk()` in other types).

libcaf_core/caf/chunk.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+  void swap(chunk& other) {
+    data_.swap(other.data_);
+  }
+
+  const_byte_span bytes() const noexcept {
+    return data_ ? const_byte_span{data_->storage(), data_->size()}
+                 : const_byte_span{};
+  }
+
```
Neverlord, line 148: Please add these:

```
const intrusive_ptr<data>& get_data() const&
intrusive_ptr<data>&& get_data() &&
```

libcaf_core/caf/flow/string.hpp
```diff
@@ -85,6 +86,48 @@ class to_lines_step {
   std::vector<char> buf_;
 };
 
+/// Turns a sequence of bytes into a sequence of chunks.
+class to_chunks_step {
```
Neverlord, line 90: This has no connections to strings. Move to e.g. `libcaf_core/caf/flow/chunk.hpp`?

libcaf_net/caf/net/web_socket/frame.hpp
```diff
@@ -17,93 +18,12 @@
...
-#elif defined(CAF_MSVC)
-#  pragma warning(push)
-#  pragma warning(disable : 4200)
-#endif
-
 namespace caf::net::web_socket {
 
 /// An implicitly shared type for passing along WebSocket data frames, i.e.,
 /// text or binary frames.
 class CAF_NET_EXPORT frame {
```
Neverlord, line 25: Add `as_chunk()` functions and `static ... from_chunk(...)`.

Neverlord (changes requested):

libcaf_core/caf/flow/string.hpp
riemass: What do you think of renaming the file instead ? From 'flow/string.hpp' to something like 'flow/transformations.hpp'?

libcaf_core/caf/flow/string.hpp
```diff
@@ -85,6 +86,48 @@ class to_lines_step {
...
+public:
+  using input_type = std::byte;
+
+  using output_type = chunk;
+
+  to_chunks_step(size_t num) : num_{num} {
+  }
```
Neverlord, line 97:
```suggestion
  explicit to_chunks_step(size_t num) : num_(num) {
    // nop
  }
```

CHANGELOG.md
```diff
@@ -84,6 +84,8 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 - The implementation of `operator<` for `caf::unordered_flat_map` was broken and
   relied on the order of elements in the map. We have removed it, since it has
   never worked correctly and a correct implementation would be too expensive.
+- We removed the already deprecated file `caf/byte.hpp`, since we use the c++ 17
+  standard.
```
Neverlord, line 88: We can't do that in a dot-release. The next window to remove APIs is 1.0.

libcaf_core/caf/chunk.cpp
```diff
@@ -0,0 +1,78 @@
...
+    this->~data();
+    free(this);
+  }
+}
+
+chunk::chunk(const_byte_span buf) {
+  init(buf.size(), buf);
+}
+
+chunk::chunk(intrusive_ptr<data> data) : data_{std::move(data)} {
```
Neverlord, line 64: This is trivial ctor, so I'd put it to the header to give the compiler more leeway in inlining it. Also: `data_()` for consistency.

libcaf_core/caf/chunk.hpp
```diff
@@ -0,0 +1,176 @@
...
+  chunk() = default;
+
+  chunk(chunk&&) = default;
+
+  chunk(const chunk&) = default;
+
+  chunk& operator=(chunk&&) = default;
+
+  chunk& operator=(const chunk&) = default;
```
Neverlord, line 113: I believe all of these can be marked as `noexcept`, no? It's just bumping a ref count (or adopting a ref).

libcaf_core/caf/chunk.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+  explicit chunk(const_byte_span buf);
+
+  explicit chunk(intrusive_ptr<data> data);
```
Neverlord, line 117:
```suggestion
  explicit chunk(intrusive_ptr<data> data) noexcept : data_(std::move(data)) {
    // nop
  }
```

libcaf_core/caf/flow/byte.hpp
```diff
@@ -0,0 +1,66 @@
...
+} // namespace caf::detail
+
+namespace caf::flow {
+
+/// Provides transformations for sequences of bytes.
+class byte {
+public:
+  /// Returns a transformation step that converts a sequence of bytes into
+  /// a sequence of chunks.
+  static detail::to_chunks_step to_chunks(size_t num) {
```
Neverlord, line 61:
```suggestion
  static auto to_chunks(size_t num) {
```

libcaf_net/caf/net/lp/frame.hpp
```diff
@@ -4,75 +4,12 @@
...
-private:
-  explicit frame(caf::span<const const_byte_span> bufs);
-
-  template <class... Args>
-  void init(size_t payload_size, Args&&... arg);
-
-  intrusive_ptr<web_socket::frame::data> data_;
-};
+// We use caf::chunk as an lp frame.
+// TODO: consider removing the alias and use caf::chunk directly in v1.0.
```
Neverlord, line 12:
```suggestion
/// An implicitly shared type for binary data frames.
// TODO: deprecate with 1.0
```

libcaf_net/caf/net/web_socket/frame.cpp
```diff
@@ -20,36 +20,6 @@ size_t calc_total_size(span<const Buffer> buffers) {
...
-    if (!buf.empty()) {
-      memcpy(pos, buf.data(), buf.size());
-      pos += buf.size();
-    }
-  }
-}
-
 frame::frame(const_byte_span bytes) {
   init(bytes.size(), bytes);
 }
```
Neverlord, line 25: Can't we implement this in terms of `chunk`? This ought to work, right?

```c++
frame::frame(const_byte_span bytes) : data_(chunk{bytes}.get_data()) {
  // nop
}
```

Neverlord (changes requested):
Just some minor improvement suggestions.

libcaf_core/caf/chunk.hpp
riemass: Now that I look at it, this could be the case for rule of zero and we should delete them all? Except the default one, since that one wouldn't be defaulted but deleted.

PR #1685, opened 2023-12-09
Add sample operator for observables

shariarriday:
relates https://github.com/actor-framework/actor-framework/issues/1389

Adds a new sample operator modeled after the buffer operator.


libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,319 @@
...
+  using select_token_type = int64_t;
+
+  output_type operator()(const std::optional<input_type>& xs) {
+    return output_type{xs};
+  }
+};
+
+///
+template <class Trait>
+class sample_sub : public subscription::impl_base {
```
Neverlord, line 36: Looks like you've simply left the trait in here because you've based this on `buffer`. However, this trait does not configure anything now and should be dropped. Also, output type should be `T`, not `optional<T>`. If the last interval had no value, the `sample` operator simply emits nothing.

```suggestion
template <class T>
class sample_sub : public subscription::impl_base {
```

Neverlord (changes requested):

libcaf_core/caf/flow/op/sample.test.cpp
```diff
@@ -0,0 +1,355 @@
...
+}
+
+SCENARIO("the sample operator forwards errors") {
+  GIVEN("an observable that produces some values followed by an error") {
+    WHEN("calling .sample() on it") {
+      THEN("the observer receives the values and then the error") {
+        auto outputs = std::make_shared<std::vector<int>>();
+        auto err = std::make_shared<error>();
+        auto pub = caf::flow::multicaster<int>{coordinator()};
+        sys.spawn([&pub, outputs, err, this](caf::event_based_actor* self) {
```
Neverlord, line 98: Why do we capture `this`?

libcaf_core/caf/flow/op/sample.test.cpp
```diff
@@ -0,0 +1,355 @@
...
+                           make_observable().never<int64_t>(),
+                           snk->as_observer());
+        snk->request(42);
+        run_flows();
+        uut->fwd_on_next(fwd_data, 1);
+        uut->fwd_on_next(fwd_data, 2);
+        uut->fwd_on_next(fwd_data, 3);
+        check_eq(uut->pending(), true);
+        uut->fwd_on_error(fwd_data, sec::runtime_error);
+        check_eq(snk->buf, std::vector{3});
```
Neverlord, line 191: I don't think the operator should have produced a value here. The case is different for `buffer`, because it uses the timeout merely as a "maximum delay". However, `sample` is supposed to emit values at a fixed interval. There was no tick, hence there should be no value.

We can argue whether to still produce that final value or not, but the operator should wait for the next tick before shutting down.

Same for `on_complete` behavior. Basically: if the input flow stops, we still wait for the next tick. The tick might either emit a value or not (depending on whether `buf_` holds a value) and then checks if the operator needs to shut down.

Neverlord (changes requested):

libcaf_core/caf/flow/op/sample.test.cpp
```diff
@@ -0,0 +1,355 @@
...
+            .sample(1s)
+            .do_on_complete([closed] { *closed = true; })
+            .for_each([outputs](const int& xs) { outputs->emplace_back(xs); });
+        });
+        dispatch_messages();
+        print_debug("emit the first six items");
+        pub.push({1, 2, 4, 8, 16, 32});
+        run_flows();
+        dispatch_messages();
+        print_debug("force an empty sample");
```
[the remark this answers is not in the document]
shariarriday: Yes, this is an artifact of old tests where std::nullopt was being emitted. This needs to be removed.

libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,311 @@
...
+  const error& err() const noexcept {
+    return err_;
+  }
+
+  bool pending() const noexcept {
+    return buf_.has_value();
+  }
+
+  bool can_emit() const noexcept {
+    return buf_.has_value() || has_shut_down(state_);
```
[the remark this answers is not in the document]
shariarriday: I am a little bit confused about it to and I forgot to discuss regarding this. `buffer` has a similar function which probably needs to be renamed or the return value should be inverted.

https://github.com/actor-framework/actor-framework/blob/2caef0f0023279ef19a269ec449b90f676bc55d8/libcaf_core/caf/flow/op/buffer.hpp#L92C1-L94C4

libcaf_core/caf/flow/op/sample.test.cpp
shariarriday: `this` was captured to call `make_observable`. This has been fixed by calling `self->make_observable()`.

libcaf_core/caf/flow/op/sample.hpp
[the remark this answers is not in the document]
Neverlord: The story is a bit different for `buffer`, because it has a bit more going on. But for `sample`, we should stick to the simple model of emitting only on a tick. That being said, maybe there's room to simplify the `buffer` as well at some later time.

libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,299 @@
...
+      sub.cancel();
+      return;
+    }
+    value_sub_ = std::move(sub);
+    value_sub_.request(1);
+  }
+
+  void fwd_on_complete(sample_input_t) {
+    value_sub_.release_later();
+    shutdown();
```
Neverlord, line 99: I don't think this is a robust solution to call `shutdown` here. The whole thing is driven by the ticks, so why not drop the `shutdown` call here and in `fwd_on_error` below and then check in `fwd_on_next(sample_emit_t, ...)` whether `value_sub_` is invalid? This gives the tick an opportunity to emit one last item before calling it a day and the implementation of `shutdown`state_` can become much simpler as a result as well. It doesn't need to switch on the state.

Neverlord (changes requested):

libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,273 @@
...
+    if (demand_ == 0)
+      return;
+    --demand_;
+    auto sampled = buf_.has_value();
+    if (sampled) {
+      out_.on_next(*buf_);
+      buf_.reset();
+    }
+    if (sampled)
+      value_sub_.request(1);
```
Neverlord, line 144: This will give infinite credit to the source over time. We always call `value_sub_.request(1)` whenever receiving an item. Giving credit here as well will accumulate over time.

```suggestion
    if (buf_) {
      out_.on_next(*buf_);
      buf_.reset();
    }
```

libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+  void request(size_t n) override {
+    CAF_ASSERT(out_.valid());
+    demand_ += n;
+    // If we can ship an item, schedule an event to do so.
+    if (demand_ == n && can_emit()) {
+      parent_->delay_fn([strong_this = intrusive_ptr<sample_sub>{this}] {
+        strong_this->on_request();
+      });
+    }
```
Neverlord, line 162: Seems to me this shouldn't be here. After the recent changeset, we only emit from `on_next` for ticks. Likewise, `can_emit` seems superfluous as well.

Neverlord (changes requested):

libcaf_core/caf/flow/op/sample.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+  // -- callbacks for the forwarders -------------------------------------------
+
+  void fwd_on_subscribe(sample_input_t, subscription sub) {
+    if (!running() || value_sub_ || !out_) {
+      sub.cancel();
+      return;
+    }
+    value_sub_ = std::move(sub);
+    value_sub_.request(1);
```
Neverlord, line 88: We can give more credit here. From the perspective of the source, this operator has basically infinite capacity.

```suggestion
    value_sub_.request(defaults::flow::buffer_size);
```

Giving some more credit here means that the source can call `on_next` multiple times consecutively. By giving just 1 credit, the source might have to break up `on_next` calls.

Neverlord (approved):

Neverlord:
@shariarriday didn't survive the rebase. Can you please update the code? Thx!

shariarriday:
> @shariarriday didn't survive the rebase. Can you please update the code? Thx!

@Neverlord I have updated the PR, I have also fixed the includes for `sample.test.cpp`. Please review it before merging into master.

Issue #1652, opened 2023-11-08
Enable logging by default and re-structure log components


Neverlord:
Currently, CAF builds without any logging by default. This means users need to build a version of CAF with logging enabled for troubleshooting. Even critical system messages are not shown in the default configuration.

We have recently opened up our logging API to make it easy to install custom loggers. So I think we should use this opportunity to re-think our logging approach more broadly. I'm thinking of changing the default log level to `INFO` and always build CAF with logging enabled. We also don't make good use of the log components, so here's how I would like to structure them going forward:

- `caf.alert` will only ever log messages of severity `warning` and above that will (or may) affect the system as a whole; this is stuff like failing to initialize SSL
- `caf.system` logs messages that are currently logged as `caf_flow`, i.e., actor spawn events, messages, actor shutdown, etc.
- `caf.core` logs messages from the core module (events on the registry, on the scheduler, etc)
- `caf.io` logs messages from the I/O module (BASP and things that are related to `publish` / `remote_actor`).
- `caf.net` logs messages from the NET module (connections and protocol-specific events)

The general idea here is that `caf.alert` should basically always be on, even in production. Messages from this component basically tell you that your application might start doing funny things or stop working as expected. The recommendation for `caf.system` would be to turn it on during development and/or for troubleshooting. The last three are mostly useful when working on CAF itself but users might want to turn them on when troubleshooting specific issues or during development to get a better understanding of the overall system.

The new default value for `caf.logger.(file|console).excluded-components` would be `["caf.system", "caf.core", "caf.io", "caf.net"]`, i.e., only `caf.alert` would be enabled by default.

The only logging that I would still keep behind a CMake build option is trace logging. Trace logs are expensive and traces into internal CAF functions are (with some very rare exceptions) only useful when working on CAF itself. I think it's perfectly fine to require custom builds for this use case.

Neverlord:
CAF 0.19.5 introduced a new set of logging functions. However, CAF itself still uses the old, macro-based, API.

The first step would be moving to the new functions. Everything under `caf_core` should use `log::core::(debug|info|...)`, everything under `caf_io` should use `log::io::(debug|info|...)`, and so on. Some places also use `CAF_ARG`. This should simply be replaced by the name. For example:

```
  // old
  CAF_LOG_DEBUG(CAF_ARG(foo));

  // new
  log::debug("foo = {}", foo);
```

[...]

PR #1688, opened 2023-12-12
Add new chunked_string type

Neverlord:
This is prep work for #1652.

The idea here is to have log events use the monotonic buffer resource to build up the message and (optional) fields. However, we don't want `std::string` or `std::vector<char>` inside of a buffer resource, because it would produce gaps (by re-allocating the internal buffer when pushing to it). The chunked strings allow us to store the result from `detail::format` (or `std::format`) without wasting memory blocks and it also avoids unnecessary copies when growing the buffer. The chunked strings are ideal for generating log events, because we build up a message and/or field values and then consume the whole thing once (for generating file and/or console output) before releasing the memory resources again. The chunked string is a bit more inconvenient to use then "regular" strings, but that'll be well worth it.

This PR only adds the new type `caf::linked_string_chunk`. Integrating it into the logger will be a followup.


libcaf_core/caf/linked_string_chunk.hpp
```diff
@@ -0,0 +1,150 @@
...
+    return *this;
+  }
+
+  linked_string_chunk_builder_output_iterator& operator++() {
+    return *this;
+  }
+
+  linked_string_chunk_builder_output_iterator operator++(int) {
+    return *this;
+  }
```
riemass, line 106: First time reading the implementation of a pure output iterator. It's a bit wired, but should be correct, at least from what I see on cppreference.

riemass (approved):
The idea is great and we have something similar with our `json` implementation.

Initially reviewing this, I think the class names and use cases are hard to comprehend.
Regarding class names, I would really like to see the change from `linked_string_chunk` to `linked_string_node`, since that is a bit more common name. 

The use cases are hard to comprehend because It's hard to reason what function operates on the whole linked list, and when only on the pointed to chunk. For example reading the function signature:
```
void render_val(const linked_string_chunk* val) {
```
knowing that the data structure is actually a single link node, and seeing that we have a `const*` I would assume this function renders the whole list.
And when reading the signature:
```
std::string to_string(const linked_string_chunk& head) {
```
since I see `const&` I would assume only one chunk gets converted to the string, however, here again we iterate till the end and convert everything.

I would like to, at least, change the signatures to consistently use `const*` and leave our references. Sadly, it would be hard having a `linked_string_list` type and be clear about copy semantics (or the lack of). If the API will be used extensively in our new logger, it might be worth giving it a taught. 

Since this is more of a discussion item, I still approve the PR.

Neverlord:
Thanks for the feedback! I've re-organized the thing a bit. There's not `chunked_string`, which is what the user usually would care about. This type has the member functions that operate on the whole thing now. The linked list underneath uses `chunked_string::node`. I think this makes the API a lot easier to navigate now. Thoughts?

riemass:
The structure looks great now. We will get used to the shallow copy semantics.

riemass (approved):
Gave it another look. This is

PR #1689, opened 2023-12-13
Increase Prometheus coverage

riemass:
It seems we only use Prometheus in `robot/middleman/driver.cpp`. 

The reason we have so low coverage for Prometheus is that the `robot` folder is in the ignore list. Bad news is that removing the ignore on this folder this will also expect the drivers to be covered, so it's best to let the pipeline pass once and see the result.

### DIFF
--- a/robot/http/rest.robot
+++ b/robot/http/rest.robot
@@ -14,6 +14,11 @@ ${BINARY_PATH}          /path/to/the/server
 ${SSL_PATH}             /path/to/the/pem/files
 ${MAX_REQUEST_SIZE}     2048
 
+# Prometheus variables
+${PR_SERVER_PORT}       55516
+${PR_HTTP_URL}          http://localhost:55516/metrics
+${PR_HTTP_BAD_URL}      http://localhost:55516/foobar
+
 *** Test Cases ***
 HTTP Test Add Key Value Pair
     [Tags]    POST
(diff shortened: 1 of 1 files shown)


Neverlord:
Even if the `robot` folder itself is ignored, the coverage from the other folders will still be counted. So this change will only add noise.

Are you maybe confusing the `caf.io` and the `caf.net` Prometheus implementation? The test you mention will cover the Prometheus Broker and has nothing to do with the implementation in `caf.net` https://app.codecov.io/gh/actor-framework/actor-framework/blob/master/libcaf_io%2Fcaf%2Fdetail%2Fprometheus_broker.cpp.

We simply don't have tests yet that execute the `caf.net` Prometheus code.

riemass:
You're right, I got the `net` and `io` drivers mixed up. I've enabled Prometheus in our `caf::net` robot tests to see the result.

riemass:
The `codecov` shows 100% on `prometheus.hpp/cpp` now. @Neverlord can I leave the test in `rest.robot`, or should I factor it out to a new file?

robot/http/rest.robot
```diff
@@ -14,6 +14,11 @@ ${BINARY_PATH}          /path/to/the/server
 ${SSL_PATH}             /path/to/the/pem/files
 ${MAX_REQUEST_SIZE}     2048
 
+# Prometheus variables
+${PR_SERVER_PORT}       55515
+${PR_HTTP_URL}          http://localhost:55515/metrics
+${PR_HTTP_BAD_URL}      http://localhost:55515/foobar
```
Neverlord, line 20: Please use port 55516. This will collide when running test suites in parallel.

Neverlord (changes requested):

Neverlord:
> can I leave the test in `rest.robot`, or should I factor it out to a new file?

Can stay, just use a different port. Also, please squash the commits since the changes of the first commit have been reverted anyways.

riemass:
Rebased and squashed.

Neverlord (approved):

Issue #1692, opened 2023-12-14
Move detail::json::intrusive_list to another namespace


Neverlord:
The type `caf::detail::json::intrusive_list` has nothing to with JSON. It's just a generic linked-list type with a custom allocator. We have uses for this type in another places, too. Hence, we should either move this type to a different namespace or consider merging it with `intrusive::linked_list`.

PR #1693, opened 2023-12-14
Simplify implementation of chunked_string

Neverlord:
I was about to implement something similar to the chunked string builder again. So I looked for ways to make parts of the implementation re-usable or re-use something else that already exists. Turns out we already have a `linked_list` with a custom allocator: `detail::json::linked_list`. We'll definitely have to move this somewhere else (see #1692), but it's still the right tool for the job.

### DIFF
--- a/libcaf_core/caf/chunked_string.hpp
+++ b/libcaf_core/caf/chunked_string.hpp
@@ -6,6 +6,7 @@
 
 #include "caf/detail/build_config.hpp"
 #include "caf/detail/core_export.hpp"
+#include "caf/detail/json.hpp"
 #include "caf/detail/print.hpp"
 #include "caf/fwd.hpp"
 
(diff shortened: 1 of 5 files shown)


libcaf_core/caf/chunked_string.hpp
```diff
@@ -17,25 +18,38 @@
...
 namespace caf {
 
-/// Represents a chunked string that as a linked list of string views.
-struct chunked_string {
-  /// A single node in the linked list.
-  struct node {
-    /// The characters of this chunk.
-    std::string_view content;
+/// Represents a chunked string as a linked list of string views.
+class CAF_CORE_EXPORT chunked_string {
```
riemass, line 22: Whats keeping us from doing an alias here?
```
using chunked_string = detail::json::linked_list<std::string_view> chunks_;
```

riemass (approved):
Looking good and DRY.

libcaf_core/caf/chunked_string.hpp
Neverlord: Two points:

- users mostly care about calling `copy_to` or `copy_quoted_to`, moving these to free functions would make the API more unwieldy
- this type is supposed to be immutable, we don't want to expose `push_back` & friends
- objects of this type are supposed to super small: they are basically the size of single pointer now; a linked list has a lot more going on

PR #1694, opened 2023-12-14
Implement new log event type for logging

Neverlord:
More prep work for https://github.com/actor-framework/actor-framework/issues/1652. The new `log_event` is going to be used for the `logger` once this has landed.

### DIFF
--- a/libcaf_core/caf/log_event.cpp
+++ b/libcaf_core/caf/log_event.cpp
@@ -1,0 +1,60 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/log_event.hpp"
+
+#include "caf/make_counted.hpp"
+
+#include <new>
+
+namespace caf {
+
+namespace {
+
+template <class T>
+using allocator_t = detail::monotonic_buffer_resource::allocator<T>;
+
+std::string_view deep_copy_impl(detail::monotonic_buffer_resource* resource,
+                                std::string_view str) {
+  auto* buf = allocator_t<char>{resource}.allocate(str.size());
+  memcpy(buf, str.data(), str.size());
+  return std::string_view{buf, str.size()};
+}
+
+} // namespace
+
+log_event_ptr log_event::make(unsigned level, std::string_view component,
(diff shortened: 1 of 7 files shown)


libcaf_core/caf/log_event.hpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  [[nodiscard]] std::string_view deep_copy(std::string_view str);
+
+  union { // Suppress the dtor call for fields_.
+    list_type fields_;
+  };
+};
+
+/// Builds a chunked string by allocating each chunk on a monotonic buffer.
```
riemass, line 258: Copy-paste probably.
```suggestion
/// Builds a chunked log_event by allocating each field on a monotonic buffer.
```

libcaf_core/caf/log_event.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+log_event_fields_builder::log_event_fields_builder(
+  resource_type* resource) noexcept {
+  new (&fields_) list_type(resource);
+}
+
+std::string_view log_event_fields_builder::deep_copy(std::string_view str) {
+  auto* buf = allocator_t<char>{resource()}.allocate(str.size());
+  memcpy(buf, str.data(), str.size());
+  return std::string_view{buf, str.size()};
```
riemass, line 52: Exactly the same code as `deep_copy_impl`. We could delegate the call to keep it a bit more DRY.

riemass (changes requested):
Found a few nits here and there, some questions, look good otherwise.

libcaf_core/caf/log_event.hpp
```diff
@@ -0,0 +1,335 @@
...
+  const char* file_name_;
+
+  /// The name of the function in which the event was generated.
+  const char* function_name_;
+
+  /// The user-defined message of the event.
+  chunked_string message_;
+
+  /// Pointer to the first user-defined field of the event.
+  const field_node* first_field_ = nullptr;
```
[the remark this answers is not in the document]
Neverlord: That, but also because we don't need the full list. The pointer is really all we need and this minimizes memory footprint.

libcaf_core/caf/log_event.cpp
Neverlord: Uh, good catch. Forgot to remove the body here after adding `deep_copy_impl`.

PR #1696, opened 2023-12-16
Port remaining op flow tests to new framework

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/1646

Ports the remaining flow op tests to the updated framework


libcaf_test/caf/test/fixture/flow.hpp
```diff
@@ -144,6 +144,81 @@ class CAF_TEST_EXPORT flow {
     caf::flow::coordinator* parent_;
   };
 
+  /// A subscription implementation without internal logic.
+  class passive_subscription_impl final
```
Neverlord, line 148: Didn't we have something like this recently and used `never` instead?

libcaf_test/caf/test/fixture/flow.hpp
```diff
@@ -144,6 +144,81 @@ class CAF_TEST_EXPORT flow {
...
+        this->sub = std::move(sub);
+        return;
+      }
+      sub.cancel();
+    }
+
+    int on_next_calls = 0;
+    int on_error_calls = 0;
+    int on_complete_calls = 0;
+    bool accept_next = false;
```
Neverlord, line 215: I think this would be a better name:

```suggestion
    bool accept_subscription = false;
```

libcaf_test/caf/test/fixture/flow.hpp
```diff
@@ -203,6 +278,12 @@ class CAF_TEST_EXPORT flow {
     return coordinator()->add_child(std::in_place_type<auto_observer<T>>);
   }
 
+  /// Returns a new canceling observer.
+  template <class T>
+  auto make_canceling_observer(bool accept_first = false) {
```
Neverlord, line 283: It's not super obvious what this will do, so I'd like to see some documentation on this. something like this:

```
  /// Returns a new canceling observer. The subscriber will either call `cancel` on
  /// its subscription immediately in `on_subscribe` or wait until the first
  /// call to `on_next` when setting `accept_first_subscription` to `true`.
  template <class T>
  auto make_canceling_observer(bool accept_first_subscription = false) {
```

Neverlord (changes requested):

libcaf_test/caf/test/fixture/flow.hpp
shariarriday: Yes, I considered using never_sub, but as it does not have the `demand` property used in https://github.com/actor-framework/actor-framework/pull/1696/files#diff-43095856bdeef9fbcd0ae3a4b69dc300d8420f533176e1c93c0733fcc9be9febR220-R221 I had to add it in the flow.

libcaf_core/caf/flow/op/cell.test.cpp
```diff
@@ -2,71 +2,67 @@
...
 #include "caf/flow/op/cell.hpp"
 
-#include "caf/flow/observable.hpp"
-#include "caf/flow/scoped_coordinator.hpp"
+#include "caf/test/fixture/deterministic.hpp"
+#include "caf/test/fixture/flow.hpp"
+#include "caf/test/scenario.hpp"
 
-#include "core-test.hpp"
+#include "caf/all.hpp"
```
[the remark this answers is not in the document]
shariarriday: I have noticed that same header has been used in https://github.com/actor-framework/actor-framework/blob/67670b688f55298aa35c673e26dc7fdc4be2fb06/libcaf_core/caf/flow/op/buffer.test.cpp#L12 I will update that header too.

libcaf_test/caf/test/fixture/flow.hpp
Neverlord: Is that the only two places where do that check? The entire test is weird, since it says something about "the observer receives the item" but then the only inputs are `never`. Just for this one (weird) test, I wouldn't add this class.

Neverlord (approved):

PR #1697, opened 2023-12-19
Allow passing inspectable types to detail::format

Neverlord:
In order to make the logging API more useful, `detail::format` now supports any inspectable type. This allows users to pass a variables as-is to logging functions and CAF only does the string conversion if the log statement has an effect (i.e., the log event is accepted).

### DIFF
--- a/libcaf_core/caf/detail/format.hpp
+++ b/libcaf_core/caf/detail/format.hpp
@@ -11,8 +11,11 @@
 //       The wrappers should also add support for types that provide an
 //       `inspect` overload.
 
+#include "caf/deep_to_string.hpp"
 #include "caf/detail/build_config.hpp"
+#include "caf/detail/is_complete.hpp"
 #include "caf/detail/source_location.hpp"
+#include "caf/detail/type_traits.hpp"
 
 #include <string>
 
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/detail/format.hpp
```diff
@@ -22,21 +25,59 @@
...
+
+template <class T>
+inline constexpr bool fmt_needs_to_string
+  = !is_formattable<T> && has_inspect_overload_v<stringification_inspector, T>;
+
+template <class T>
+constexpr T&&
+fmt_fwd(std::remove_reference_t<T>& arg,
+        std::enable_if_t<is_formattable<std::decay_t<T>>>* = nullptr) noexcept {
+  return static_cast<T&&>(arg);
```
riemass, line 39: Isn't this unsafe? You cast an lvalue reverence to an rvalue reference?

Neverlord: No, that's pretty much the canonical implementation of `forward`.

Say you have the expression, `type&&`. If `type` is `foo`, this will deduce to `foo&&`. However, if `type` is `foo&`, it will deduce to just `foo&`. This is called "reference collapsing" (e.g., see https://en.cppreference.com/w/cpp/language/reference).

riemass: Oops, I forgot that template references work that way. Thanks for explaining it.

riemass (approved):
Good to go. :+1:

Issue #1066, opened 2020-03-13
(Almost) no way to make use of fine grained log levels


tobim:
CAF defines the integer representation for the log levels with a step size of 3. This suggests that a downstream application can choose intermediate levels to allow for a finer grained control of log output printing. Using a log macro like that works just fine:
``` c++
#define LOG_LEVEL_CHATTY 10
#define LOG_CHATTY(output) CAF_LOG_IMPL("dummy", LOG_LEVEL_CHATTY, output)
```

But it doesn't add value, because there is no way to teach the logger about the `"chatty"` level and use it as the maximum level. If one sets `logger.verbosity` to `"chatty"` in the `actor_system_config`, it falls back to `"quiet"` and doesn't even fully initialize.

It would be helpful to have an access function so the level can be changed after the initialization: `config& cfg() { return cfg_; }`.

It would also be helpful to just define more levels in CAF itself, how about:
```
#define CAF_LOG_LEVEL_QUIET 0
#define CAF_LOG_LEVEL_FATAL 1
#define CAF_LOG_LEVEL_CRITICAL 2
#define CAF_LOG_LEVEL_ERROR 3
#define CAF_LOG_LEVEL_WARNING 6
#define CAF_LOG_LEVEL_UNEXPECTED 7
#define CAF_LOG_LEVEL_INFO 9
#define CAF_LOG_LEVEL_CHATTY 10
#define CAF_LOG_LEVEL_VERBOSE 11
#define CAF_LOG_LEVEL_DEBUG 12
#define CAF_LOG_LEVEL_DEBUG2 13
#define CAF_LOG_LEVEL_DEBUG3 14
#define CAF_LOG_LEVEL_TRACE 15
```

PS: Have a look at [here](https://github.com/tenzir/vast/blob/d6e41e9a794596b339ebbe86ffc4c7244344cf65/libvast/src/logger.cpp) to learn how we circumvent that problem for now.

Neverlord:
The natural customization point in CAF is `actor_system_config`. Wouldn't it solve the issue if the logger would ask the system config for log level names that aren't hard-coded? Users then could simply provide a mapping function.

tobim:
That would work as well.

Neverlord:
I'd like to add a bit more context to this for future reference.

> CAF defines the integer representation for the log levels with a step size of 3. This suggests that a downstream application can choose intermediate levels

[...]

PR #1698, opened 2023-12-20
Restructure the logging API

Neverlord:
- Rename `caf::{log_event => log::event}`
- Add new `caf::log::level` class with properly typed constants for the existing log level macros.
- Remove member functions such as `logger::debug` that require a component argument.
- Implement new logging API under `caf::log` with one header (and namespace) per component: `system`, `core`, `io`, `net`, `openssl`, and `test`.
- Remove the `print_<level>` functions from the test reporter and instead simply use the new `log::test::<level>` functions.
- Drop automatic logging from `CAF_RAISE_ERROR`, because it cannot log with the appropriate component.
- Use the new `log::system` logging API throughout CAF wherever an event might impact the operation of the entire system.
- Implement log component filtering on the unit test binaries. This enables CAF developers to print log messages as part of the test execution when using the deterministic fixture.

### DIFF
--- a/libcaf_test/caf/test/runner.cpp
+++ b/libcaf_test/caf/test/runner.cpp
@@ -14,6 +14,7 @@
 #include "caf/config_option_set.hpp"
 #include "caf/detail/log_level.hpp"
 #include "caf/detail/set_thread_name.hpp"
+#include "caf/log/level.hpp"
 
 #include <condition_variable>
 #include <iostream>
@@ -79,7 +80,8 @@ config_option_set make_option_set() {
     .add<std::string>("suites,s", "regex for selecting suites")
     .add<std::string>("tests,t", "regex for selecting tests")
     .add<std::string>("available-tests,A", "print tests for a suite")
-    .add<std::string>("verbosity,v", "set verbosity level of the reporter");
+    .add<std::string>("verbosity,v", "set verbosity level of the reporter")
+    .add<std::vector<std::string>>("log-component-filter,l", "set log filter");
   return result;
 }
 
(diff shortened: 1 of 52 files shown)


Neverlord:
@riemass I've made a pass over our logging API in preparation for 0.19.5, also laying some ground work for #1652 and #1066. I think the free function interface is a much better API than trying to shove everything into the `logger` interface. The `log.hpp.in` is also meant as a blueprint for any user-defined API on top of CAF and will make it easy to integrate custom log levels, I hope. What do you think?

robot/logging/driver.cpp
```diff
@@ -14,8 +15,68 @@ CAF_BEGIN_TYPE_ID_BLOCK(driver, caf::first_custom_type_id)
 
 CAF_END_TYPE_ID_BLOCK(driver)
 
+namespace app {
+
 constexpr std::string_view component = "app";
 
+/// Logs a message for the `caf.core` component with `debug` severity.
```
riemass, line 22: Copy paste remnants, also below.
```suggestion
constexpr std::string_view component = "app";

/// Logs a message for the `app` component with `debug` severity.
```

libcaf_test/caf/test/runner.cpp
```diff
@@ -121,6 +123,18 @@ std::optional<std::regex> to_regex(std::string_view regex_string) {
...
 
+auto default_log_component_filter() {
+  using namespace std::literals;
+  return std::vector{
+    "caf"s,        // legacy CAF logging
+    "caf_flow"s,   // legacy CAF logging
+    "caf.core"s,   // events from log::core
+    "caf.io"s,     // events from log::io
+    "caf.net"s,    // events from log::net
+    "caf.openssl"s // events from log::openssl
```
riemass, line 134: Should we add the additional `system` component here?

riemass:
> I think the free function interface is a much better API than trying to shove everything into the `logger` interface. 

Full ACK. Simple functions are convenient and spare lines of code, meaning that we will be willing to write and leave logs in our system. 
For one part, I'd like to have both. This is a compile time resolved implementation. What I'd like to see is some sort of *runtime* logs. While working on my example using the new logger, at one moment I'd like to be able to just write like this:
```
auto logger = component_logger(my_name);
logger->debug("Hello, world");
```
However, this is something we probably won't need at our side, so it's up to discussion should we implement it or leave for our users to implement. 

> The `log.hpp.in` is also meant as a blueprint for any user-defined API on top of CAF and will make it easy to integrate custom log levels, I hope. 

Users will probably just copy-paste their logger, to skip bothering with `cmake`. Which means we probably won't have to worry about backward compatibility, and can do our thing there and keep the code DRY.

Aside from that, stringification of `message` was one thing that was missing. And I like the fact that we automated the component. Big :+1: for this PR.

riemass (changes requested):

libcaf_test/caf/test/runner.cpp
Neverlord: No, `caf.system` is never suppressed by default anywhere. That component is only used if something goes really wrong, so we want to have that information available.

robot/logging/driver.cpp
Neverlord: Yeah, forgot to remove the doxygen docu. Makes no sense to have them in a driver in the first place.

riemass (approved):
Good to go.

PR #1699, opened 2023-12-21
Add datagram transport with udp example

riemass:
My initial idea was to implement an example using `udp_datagram_socket`, and it kind of turned into adding a `datagram_transport` layer that we where missing. Since I didn't like the idea of modeling an upper layer without prior consulting with you, the implementation uses an actor shell to communicate with the rest of the system. 

What's basically hindering us from refactoring the current transport layer is the fact that an UDP socket is not connected and can receive data from multiple endpoints, and send data to multiple endpoints. We can add an abstraction for connected UDP sockets, there is OS supports that, but again, I don't want this PR to drift too much without feedback.


examples/udp/transport.cpp
```diff
@@ -0,0 +1,180 @@
...
+behavior printer(event_based_actor* self) {
+  return {[](ip_endpoint ep, std::string message) {
+            app::info("caf", "received {} bytes from {}:{}", message.size(),
+                      ep.address(), ep.port());
+            std::cout << message << std::endl;
+          },
+          [self](const message&) {
+            app::error("caf", "received unexpected message");
+            self->quit();
+          }};
```
Neverlord, line 89: We always use a trailing comma. For better Git diffs, but also to get `clang-format` to do a better job formatting.

```suggestion
          },};
```

examples/udp/transport.cpp
Neverlord, line 1: This is a very low-level API. I'm not sure if we really want to expose this at all. Interacting with socket managers and caf-net protocol stacks is a quite advanced topic. I think putting this out as an example could lead people astray as to what is actually the intended APIs for developing CAF applications (as opposed to extending CAF capabilities).

examples/udp/transport.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Send payload over udp.
+error send_payload(actor_system& sys, std::string addr, uint16_t port) {
+  auto tg = logger::trace("caf", "Trying to send data to {}:{}", addr, port);
```
Neverlord, line 94: That's inconsistent. You have wrapped the logging API with the `app` namespace. Why not provide a `trace` function there?

examples/udp/transport.cpp
```diff
@@ -0,0 +1,180 @@
...
+  auto print_actor = sys.spawn(printer);
+  // Simplified case, we use only the first resolved address.
+  auto addrs = ip::resolve(addr);
+  if (addrs.empty())
+    return make_error(sec::runtime_error, "No resolved ip address");
+  auto dest = ip_endpoint{addrs[0], port};
+  auto local = ip_endpoint{make_ipv4_address(127, 0, 0, 1), 0};
+  auto sock = make_udp_datagram_socket(local);
+  if (!sock) {
+    std::cerr << "*** failed to open udp with: " << std::endl;
```
Neverlord, line 104: This is also inconsistent. Either we always use `stderr` or we use `app::error`. I'd be in favor of the latter.

libcaf_net/caf/net/datagram_transport.cpp
```diff
@@ -0,0 +1,97 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/net/datagram_transport.hpp"
+
+namespace {
+
+// Name of the component for logging.
+static constexpr std::string_view component = "datagram_transport";
```
Neverlord, line 10: That's not quite how the components are meant to be used. The component should be the module, not the class. We already have `log::net::(debug|...)` for that purpose. The functions taking a `component` shouldn't be used directly.

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -0,0 +1,90 @@
...
+  byte_buffer read_buf_;
+
+  /// Caches outgoing data.
+  byte_buffer write_buf_;
+
+  /// Stores the max number of bytes to receive.
+  size_t max_read_size_ = max_datagram_size;
+
+  /// Destination for the outgoing datagrams.
+  ip_endpoint dest_;
```
Neverlord, line 81: That's a very limiting design. It's very common to serve multiple clients from a single socket. This is also how you would implement an HTTP 3 server (QUIC sits on top of UDP).

libcaf_net/caf/net/datagram_transport.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  /// Worker actor for processing incoming datagrams.
+  actor worker_ = {};
+
+  // Actor shell representing this app.
+  net::actor_shell_ptr self;
```
Neverlord, line 87: This is also very limiting and not at all how the layering API should operate. The datagram transport should be much closer in design to the octet transport, i.e., communicate via interfaces to the next layer. Running an actor shell in the transport does not compose.

Neverlord (changes requested):
Closing this PR. Not because I think the direction isn't worth pursuing at all, but a new transport would have to introduce a new set of interfaces for communication to the next layer and right now there is no one really demanding high-level UDP APIs in CAF, neither do we have immediate use cases ourselves. I believe our priorities should be elsewhere at the moment.

We do need to add coverage for the UDP sockets, but that can be a simple driver on top of the low-level API plus a Robot test.

PR #1701, opened 2023-12-26
Increase coverage of TCP connect 

riemass:
DRAFT to see get new coverage report.


libcaf_net/caf/net/tcp_accept_socket.cpp
```diff
@@ -124,12 +124,8 @@ make_tcp_accept_socket(const uri::authority_type& node, bool reuse_addr) {
...
                       to_string(node));
   // Prefer ipv6 addresses.
-  std::stable_sort(std::begin(addrs), std::end(addrs),
-                   [](const ip_address& lhs, const ip_address& rhs) {
-                     if (lhs.embeds_v4())
-                       return rhs.embeds_v4() ? lhs < rhs : false;
-                     return rhs.embeds_v4() ? true : lhs < rhs;
-                   });
+  std::stable_partition(std::begin(addrs), std::end(addrs),
+                        [](const ip_address& ip) { return !ip.embeds_v4(); });
```
Neverlord, line 128: Yeah, that's definitely a better way to do it. Good catch! 🙂

Neverlord (changes requested):
Testing socket I/O in a unit test means we're basically testing CAF with itself. I think this would be better suited to write Robot tests for. Doesn't even have to do binary, we could implement a trivial telnet driver and talk to it over TCP in Robot. Wouldn't you agree?

libcaf_net/caf/net/tcp_accept_socket.cpp
riemass: There's one more place where we reorder for ipv6. I'll push the changes.

riemass:
> Testing socket I/O in a unit test means we're basically testing CAF with itself. I think this would be better suited to write Robot tests for. Doesn't even have to do binary, we could implement a trivial telnet driver and talk to it over TCP in Robot. Wouldn't you agree?

I think we have more fine-grained control when configuring the CAF application and inspecting the test state when doing internal unit tests. For robot tests, telnet can be one side, but the other one is still our application that will be CLI driven which can be messy sometimes. Otherwise I agree, when we have a clear workflow - what can go to robot should be in robot.

Neverlord (approved):
Hm, I'd still prefer the cases that actually do connection to have Robot tests instead, but I'm ok with keeping it this way for now and moving on.

Issue #1702, opened 2023-12-27
Deprecate class-based `spawn` version


Neverlord:
Currently, we have two separate APIs in the `actor_system`: one for spawning "function-based actors" and one for spawning "class-based actors". We should get rid of the latter for several reasons:

- It's easy to get class-based actors wrong, e.g., by not properly shielding against cyclic references that can be broken with stateful actors.
- Class-based actors use a different syntax than function-based actors, which makes it harder for newcomers to learn when to use what.
- We already recommend users to not subtype actors directly but instead implement a state class and then spawn a `stateful_actor<State>`.

Since we already recommend stateful actors when using classes, we basically only need a replacement for this particular use case. I think of something like this:

```c++
struct cell_state {
  cell_state() = default;

  explicit cell_state(int init) : value(init) {
    // nop
  }

  behavior make_behavior() {
    return {
      [this](get_atom) { return value; },
      [this](put_atom, int new_value) { value = new_value; },
    };
  }

  int value = 0;
};

// Function object; spawn via sys.spawn(cell_impl, ctor-args...);
constexpr auto cell_impl = caf::stateful<cell_state>();
```

By introducing a simple way to turn a state into a "function-based actor", we would get rid of the last use case where we recommend using class-based actors and could deprecate the class-based version of `spawn`.

Neverlord:
The new building block is `actor_from_state`. 🙂

Neverlord:
This is already taken care of and documented in the changelog.

PR #1706, opened 2024-01-02
Add utils for spawning actors from state classes

Neverlord:
Prep work for #1702. Adding this feature now allows us to update the best practices with the next release and then deprecate the class-based `spawn` versions with 1.0.

### DIFF
--- a/libcaf_core/caf/actor_from_state.hpp
+++ b/libcaf_core/caf/actor_from_state.hpp
@@ -1,0 +1,101 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/detail/type_traits.hpp"
+#include "caf/event_based_actor.hpp"
+#include "caf/fwd.hpp"
+#include "caf/typed_event_based_actor.hpp"
+
+#include <new>
+#include <type_traits>
+#include <utility>
+
+namespace caf::detail {
+
+/// An event-based actor with managed state. The state is constructed separately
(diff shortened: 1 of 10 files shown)


libcaf_core/caf/actor_from_state.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+/// Helper class for automating the creation of an event-based actor with
+/// managed state.
+template <class State>
+struct actor_from_state_t {
+  /// The behavior type of the actor.
+  using behavior_type = decltype(std::declval<State*>()->make_behavior());
+
+  static_assert(detail::is_behavior_v<behavior_type>,
+                "State::make_behavior() must return a behavior");
```
riemass, line 69: I'm supposing we don't support `make_behavior` returning `void` and calling `self->become` internally for class based actors?

riemass (approved):
Looking good. Two little questions just for my better understanding.

libcaf_core/caf/behavior.hpp
```diff
@@ -112,10 +112,14 @@ class CAF_CORE_EXPORT behavior {
...
 
-  behavior& unbox() {
+  behavior& unbox() & {
     return *this;
   }
 
+  behavior&& unbox() && {
+    return std::move(*this);
+  }
+
```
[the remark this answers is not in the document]
Neverlord: I wrote `make_behavior().unbox()` somewhere, so I checked whether this will actually move the behavior.

libcaf_core/caf/actor_from_state.hpp
Neverlord: That's not the use case for this utility. We need the behavior type to deduce the proper handle for the actor and this strikes me as the natural place to do so.

PR #1707, opened 2024-01-03
CI: drop FreeBSD 12 (EOL), switch to FreeBSD 14

Neverlord:
Cirrus is failing all FreeBSD 12 builds, so we need to switch to the latest version.


riemass (approved):
Nice :+1:

PR #1708, opened 2024-01-03
CI: drop fedora-37, switch to fedora-39

riemass:
Fedora-37 reached EOL. 
Also switched the sanitizer builds from fedora-38 to fedora-39.


Neverlord (approved):
Thanks for taking care of it! 👍

PR #1709, opened 2024-01-04
Update best practices on spawning actors

Neverlord:
Consistently use the "function-based API" for spawning actors in all examples and `actor_from_state` to convert state classes into function objects. This makes the API surface of CAF smaller, since we only recommend one API for spawning actors. Ultimately making CAF easier to teach and use.

I've also deleted a few examples that overlapped with other examples. Overall, I think this new approach to state-based actors is a good improvement and nicely avoids the sharp edges of the previous approaches (e.g., using `stateful_actor<...>` or directly subtyping `event_based_actor`). Further, I've also used this opportunity to remove `all.hpp` includes from our examples to reflect best practices (i.e., include-what-you-use).


examples/dynamic_behavior/skip_messages.cpp
```diff
@@ -31,12 +31,12 @@ void caf_main(actor_system& system) {
   scoped_actor self{system};
   self->request(serv, std::chrono::seconds(10), ping_atom_v)
     .receive(
-      [&](pong_atom) {
+      [&self, worker](pong_atom) {
```
riemass, line 34:
```suggestion
      [self, worker](pong_atom) {
```

examples/dynamic_behavior/skip_messages.cpp
```diff
@@ -31,12 +31,12 @@ void caf_main(actor_system& system) {
...
         aout(self) << "received response from "
                    << (self->current_sender() == worker ? "worker\n"
                                                         : "server\n");
       },
-      [&](error& err) {
+      [&self, worker](error& err) {
```
riemass, line 39: Not sure if taking self by reference is a best practice.
```suggestion
      [self, worker](error& err) {
```

riemass (changes requested):

examples/dynamic_behavior/skip_messages.cpp
Neverlord: This is a `scoped_actor`. You actually *have* to capture this by reference, since you cannot copy it.

examples/dynamic_behavior/skip_messages.cpp
Neverlord: This is also a `scoped_actor`.

riemass (approved):

PR #1710, opened 2024-01-06
Add new actor_ostream::println utility function

Neverlord:
The new `println` integrates our new formatting backend into the `actor_ostream` class. I'm not super keen on `actor_ostream`, so maybe we will end up removing it eventually. However, we do use it in several examples and it should be at least nice to use and properly integrated while we have it.


riemass (approved):
LGTM :+1:

PR #1712, opened 2024-01-07
Fix behavior of caf::net::make_pipe on Windows

Neverlord:
Fixes a bug when using `make_pipe` on Windows where the OS would close the pipe after two minutes. Since there is no `pipe()` equivalent on Windows, we use a connected socket pair (of TCP sockets). The function used to close the read/write ends to turn the sockets into a unidirectional communication mode. However, closing the write channel on a TCP socket sets the socket into the state FIN_WAIT, which waits two minutes before closing the socket. So we can't use it that way.

Original bug report / context: https://github.com/zeek/broker/issues/316.


riemass (approved):
Patch looks good. Appreciate the explanation.

PR #1713, opened 2024-01-08
Remove deprecated functions and classes

Neverlord:
First chunk of removing deprecated API pieces. Removing the deprecated groups is going to be separate PR, because it's not as trivial.


riemass (approved):
LGTM. Nice to see things go. :rocket:

PR #1716, opened 2024-01-11
Rename CLI for caf-net Prometheus and enable SSL

### LINKED ISSUE #1690: Rename Prometheus CLI arguments and add SSL support
Currently the `caf::net` module peeks into the flags of `caf::io` module to see if we need to set up the Prometheus endpoint, mostly for historic reasons.
We should update the CLI flags for `caf::net`, and include the option for configuring the endpoint to use `SSL`.

riemass:
Closes #1690 

With this PR, the CLI args for `caf::net` and `caf::io` prometheus servers are separates. Additionaly, the `caf::net` implementation can use TLS. 

Also some minor fixes: removed unneeded headers and bumped a port number in `robot` since it was used elsewhere.


Neverlord:
@riemass what has the forward_list_node change to do with Prometheus?

riemass:
> @riemass what has the forward_list_node change to do with Prometheus?

I successfully pushed to the wrong branch.

riemass:
The PR has the correct implementation again. Sorry for the inconvenience.

libcaf_net/caf/net/middleman.cpp
```diff
@@ -7,50 +7,67 @@
...
+bool inspect(Inspector& f, prom_tls_config& x) {
+  return f.object(x).fields(f.field("key-file", x.key_file),
+                            f.field("cert-file", x.cert_file));
+}
+
 struct prom_config {
   uint16_t port;
   std::string address = "0.0.0.0";
   bool reuse_address = true;
+  prom_tls_config tls;
```
Neverlord, line 40:
```suggestion
  std::optional<prom_tls_config> tls;
```

libcaf_net/caf/net/middleman.cpp
```diff
@@ -7,50 +7,67 @@
...
 
 template <class Inspector>
 bool inspect(Inspector& f, prom_config& x) {
   return f.object(x).fields(
-    f.field("port", x.port), f.field("address", x.address).fallback("0.0.0.0"),
-    f.field("reuse-address", x.reuse_address).fallback(true));
+    f.field("port", x.port), //
+    f.field("address", x.address).fallback("0.0.0.0"),
+    f.field("reuse-address", x.reuse_address).fallback(true),
+    f.field("tls", x.tls).fallback(prom_tls_config{}));
```
Neverlord, line 49:
```suggestion
    f.field("tls", x.tls));
```

libcaf_net/caf/net/middleman.cpp
```diff
@@ -7,50 +7,67 @@
...
 void launch_prom_server(actor_system& sys, const prom_config& cfg) {
-  auto server = http::with(sys)
-                  .accept(cfg.port, cfg.address)
-                  .reuse_address(cfg.reuse_address)
-                  .route("/metrics", prometheus::scraper(sys))
-                  .start();
+  constexpr auto pem = ssl::format::pem;
+  auto server
+    = http::with(sys)
+        .context(ssl::context::enable(cfg.tls.valid())
```
Neverlord, line 56:
```suggestion
        .context(ssl::context::enable(cfg.tls.has_value())
```

Neverlord (changes requested):

libcaf_net/caf/net/middleman.cpp
```diff
@@ -7,50 +7,67 @@
...
+  constexpr auto pem = ssl::format::pem;
+  auto server
+    = http::with(sys)
+        .context(
+          ssl::context::enable(cfg.tls.has_value())
+            .and_then(ssl::emplace_server(ssl::tls::v1_2))
+            // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
+            .and_then(ssl::use_private_key_file(cfg.tls->key_file, pem))
+            // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
+            .and_then(ssl::use_certificate_file(cfg.tls->cert_file, pem)))
```
Neverlord, line 59: This is UB. Do something like this:

```c++
std::optional<std::string> key_file(const std::optional<prom_tls_config>& cfg) {
  if (cfg)
    return cfg->key_file;
  return {};
}
```

Same for `cert_file`. Basically lifting the inner member. Passing an `optional<string>` to the `use_....` functions will then do the right thing.

Neverlord (changes requested):

libcaf_net/caf/net/middleman.cpp
riemass: Oh snap, you're right, I made UB.

Neverlord (approved):

PR #1717, opened 2024-01-12
Remove legacy tests and update for new test framework (Part 4)

shariarriday:
Updated the following tests to new framework. Also introduced a test_type_id class to contain all of the test type id required by tests.
- ipv4_address
- ipv4_endpoint
- ipv4_subnet
- ipv6_address
- ipv6_endpoint
- ipv6_subnet
- json_array
- json_builder
- json_object
- json_reader
- json_value
- json_writer
- load_inspector
- mailbox_element
- message
- message_id
- select_all
- select_any


libcaf_test/caf/test/type_id.hpp
Neverlord, line 1: `libcaf_test` is a public library, moving internals from some tests to the lib is not something we should do. Please check if you can move stuff to `*.test.cpp` directly. If there a multiple tests depending on one the same types then drop it for now and we'll discuss where they should go separately.

Neverlord (changes requested):

libcaf_test/caf/test/type_id.hpp
Neverlord: @shariarriday did you see this comment? The PR still adds test internals to `caf_test`.

shariarriday: @Neverlord yes I am currently working on it. I was checking if it was possible to refactor the tests without using test internals from `caf_test`. But, I am finding there will be too much duplicate codes in that scenario. So, I am preparing a fix where these tests are removed for now. It should be prepared shortly.

Neverlord (approved):

PR #1718, opened 2024-01-14
Remove the obsolete path from mailbox elements

Neverlord:
The (long removed) legacy streaming API encoded forwarding paths directly on the mailbox elements. However, after removing the sequencer decorator, there is no longer any API piece in CAF using this feature. Hence, we can finally get rid of it.


riemass (approved):
LGTM. Never really noticed the stages part, but glad that they go.

PR #1719, opened 2024-01-15
Move linked_list from json to detail

### LINKED ISSUE #1692: Move detail::json::intrusive_list to another namespace

riemass:
Closes #1692.

Moves the linked list to detail namespace. I've named it `forward_list` to further differentiate it from `intrusive::linked_list`. 

Aside from that, while analyzing `intrusive::linked_list` I removed an unneeded initialization and some headers that where obsolete.

### DIFF
--- a/libcaf_core/caf/detail/forward_list.hpp
+++ b/libcaf_core/caf/detail/forward_list.hpp
@@ -94,9 +94,6 @@ operator!=(forward_list_iterator<T> lhs, forward_list_iterator<T> rhs) {
 //
 // The default-constructed list object is an empty list that does not allow
 // push_back.
-//
-// TODO: should not be part of the json namespace. Probably should be merged
-//       into intrusive::linked_list.
 template <class T>
 class forward_list {
 public:
--- a/libcaf_core/caf/detail/forward_list.hpp
+++ b/libcaf_core/caf/detail/forward_list.hpp
@@ -1,0 +1,273 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include <caf/detail/monotonic_buffer_resource.hpp>
+
+#include <cstddef>
+#include <iterator>
+
+namespace caf::detail {
+
+template <class T>
+struct forward_list_node {
+  T value;
+  forward_list_node* next;
+};
+
+template <class T>
+class forward_list_iterator {
+public:
+  using difference_type = ptrdiff_t;
+
+  using value_type = T;
+
+  using pointer = value_type*;
+
+  using reference = value_type&;
+
+  using iterator_category = std::forward_iterator_tag;
+
+  using node_pointer
+    = std::conditional_t<std::is_const_v<T>,
+                         const forward_list_node<std::remove_const_t<T>>*,
+                         forward_list_node<value_type>*>;
+
+  constexpr forward_list_iterator() noexcept = default;
+
+  constexpr explicit forward_list_iterator(node_pointer ptr) noexcept
+    : ptr_(ptr) {
+    // nop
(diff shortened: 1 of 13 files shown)


riemass:
We had a bug with postfix operator++ that showed up when iterating through the whole list in `deep_to_string`. 
Also, it's bogging me that we can default construct a `forward_list`, in which case it will work while empty, bug segfault on any insertion because it doesn't have a monotonic buffer resource. Removing the default ctor causes compile errors in `json`.

libcaf_core/caf/detail/forward_list.hpp
```diff
@@ -0,0 +1,275 @@
...
+
+// A minimal version of a linked list that has constexpr constructor and an
+// iterator type where the default-constructed iterator is the past-the-end
+// iterator. Properties that std::list unfortunately lacks.
+//
+// The default-constructed list object is an empty list that does not allow
+// push_back.
+//
+// TODO: should not be part of the json namespace. Probably should be merged
+//       into intrusive::linked_list.
```
Neverlord, line 99:
```suggestion
```

libcaf_core/caf/detail/forward_list.test.cpp
```diff
@@ -0,0 +1,92 @@
...
+  check_eq(deep_to_string(uut), "[]");
+  fill(uut, 1, 2, 3, 4);
+  check_eq(uut.size(), 4ul);
+  check_eq(deep_to_string(uut), "[1, 2, 3, 4]");
+}
+
+TEST("push_back adds elements to the back of the list") {
+  detail::monotonic_buffer_resource resource;
+  list_type uut{&resource};
+  // list_type uut;
```
Neverlord, line 41:
```suggestion
```

libcaf_core/caf/intrusive/linked_list.hpp
```diff
@@ -50,9 +50,7 @@ class linked_list {
   linked_list() noexcept = default;
 
   linked_list(linked_list&& other) noexcept {
-    if (other.empty()) {
-      init();
-    } else {
+    if (!other.empty()) {
```
Neverlord, line 53: If you do `my_list = linked_list{}`, then `my_list` should be empty. This implementation turns it into a no-op.

riemass: Not sure I understand this fully. 
`my_list = linked_list{}` calls the `opterator=` for `my_list`, which will call `init` since the right side is empty.
The `linked_list{}` call is a default ctor. 
If we would write `auto my_list = std::move(some_empty_list);` since we don't have an initializer list, the values for head and tail will be default initialized like in the class body to the addresses of one another, like they would be when calling `init`, and the ctor body won't touch them afterwards. 
That would be my understanding, could you point me in the right direction if I got that wrong?

Neverlord: Sorry, I've confused the move-constructor with the copy-assignment. :)

libcaf_core/caf/detail/mbr_list.hpp
```diff
@@ -0,0 +1,266 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include <caf/detail/monotonic_buffer_resource.hpp>
+
+#include <cstddef>
+#include <iterator>
```
Neverlord, line 10: Should include `<new>` for placement new and `<utility>` for `swap`.

Neverlord (changes requested):

Neverlord (approved):

Issue #1715, opened 2024-01-09
Version 0.19.5 shared build fails on MinGW-w64


brechtsanders:
When building version 0.19.5 on Windows under MSYS2 shell with MinGW-w64 GCC 13.2.0 with the following CMake flags
```raw
-DCMAKE_BUILD_TYPE:STRING=Release -DBUILD_SHARED_LIBS:BOOL=ON -DCAF_ENABLE_EXAMPLES:BOOL=OFF -DCAF_ENABLE_TESTING:BOOL=OFF -DCMAKE_CXX_FLAGS:STRING="-fcommon -D_POSIX_THREAD_SAFE_FUNCTIONS"
```
I get the following output which seems to be releated to `dllexport`/`dllimport` issues.:
[ninja_output.txt](https://github.com/actor-framework/actor-framework/files/13878062/ninja_output.txt)

Note that version 0.19.4 was building just fine.

Neverlord:
Thanks for opening an issue! Looks like we're missing an export declaration somewhere at a first glance. We'll look into it.

Neverlord:
I did get linker errors for a DLL build with MSVC as well. I'm not sure if it's the same issue with MinGW. We should also start using DLL builds on CI, I'm already working on that.

Neverlord:
@brechtsanders I've backported the patch to the 0.19 series in the branch `issue/1715-backport`. Would you mind double-checking that it's working for you as well before we cut an 0.19.6 release?

brechtsanders:
Unfortunately this is the result of a shared build of branch `issue/1715-backport`:
```raw
FAILED: libcaf_test.dll libcaf_test.dll.a
[...]

PR #1721, opened 2024-01-17
Fix DLL builds with MSVC

Neverlord:
Fixes a linker issue when building DLLs on Windows and also builds CAF with shared libs on Windows CI.

Relates #1715.


Neverlord:
This turned out a bit frustrating. It's running on my local Windows box, not quite sure how to reproduce the CI error yet.

shariarriday (approved):

shariarriday:
Did we need the additional export macros to build successfully? Maybe we can skip them if they are not required.

Neverlord:
> Did we need the additional export macros to build successfully? Maybe we can skip them if they are not required.

They are still required. Even if CAF itself would compile without them, it would cause a linker error for users that try to use any of the classes that we don't export.

PR #1722, opened 2024-01-17
Remove abstract_channel

Neverlord:
The abstract channel interface has been the common ancestor for groups and actors. However, after removing groups, we no longer have any need for this interface.


riemass (approved):
LGTM.

PR #1723, opened 2024-01-19
Remove the class monitorable_actor

Neverlord:
All actor types in CAF inherit from `monitorable_actor`. Hence, the class really serves no purpose, since we can simply merge it into `abstract_actor`.


riemass (approved):
LGTM. Nice to see stuff get simplified. There's also [this UML diagram](https://actor-framework.readthedocs.io/en/stable/Actors.html#common-actor-types) in the manual that we need to update.

Neverlord:
> There's also [this UML diagram](https://actor-framework.readthedocs.io/en/stable/Actors.html#common-actor-types) in the manual that we need to update.

Noted. I think we'll have to do a full pass over the manual soon, some things are already outdated or will become outdated with 1.0.

PR #1724, opened 2024-01-20
Allow actors to send messages during cleanup

Neverlord:
In the past, users had to very careful about writing cleanup code for actors. If an actor became unreachable, it would run its cleanup code with a reference count of 0 and usually call the destructor after running the cleanup logic (except if there are other weak references). This meant that sending a message from cleanup code would cause UB by creating a strong reference to a destroyed actor.

This set of changes changes how CAF cleans up unreachable actors. If a reference count drops to 0, CAF will check whether the actor has properly terminated. If not, CAF will now *first* bump it back up to 1 and *then* call the cleanup code (which will set the terminated flag to make sure the cleanup code runs exactly once). After running the cleanup code, CAF decrements the reference count again (and proceeds with the regular logic if it drops back to 0 again).

This removes one of the sharp edges in CAF that made it really easy to cut yourself. I've also streamlined the flags and cleanup logic in general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   setting the configuration option `caf.net.prometheus-http`. The option has the
   following fields: `port`, `address`, `tls.key-file`, and `tls.cert-file`. When
   setting the TLS options, the server will use HTTPS instead of HTTP.
+- Sending messages from cleanup code (e.g., the destructor of a state class) is
+  now safe. Previously, doing so could cause undefined behavior by forming a
+  strong reference to a destroyed actor.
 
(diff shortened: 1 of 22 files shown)


Neverlord:
`dynamic_spawn` failed on one build. Not sure if it's related or not quite yet.

Neverlord:
Yeah, this is more complicated. If we go back from 0 to 1, it means that any weak reference lurking around can now produce a strong ref again concurrently, while we're trying to clean up the actor. I need to figure out of how to block/close the mailbox before going back to a strong reference count of 1 so that a concurrent `weak_actor_ptr::lock` cannot produce a strong ref for sending messages to.

riemass (approved):
Good catch of the bug. LGTM.

Issue #1725, opened 2024-01-25
Remove all mixins


Neverlord:
CAF uses mixins to "insert" common functions like `send` and `become` to various actor types. Mixins are great for prototyping, however, they are not a first-class primitive in C++ and our implementation relies on CRTP to simulate them. This obfuscate the code for developers that aren't used to mixins from other languages, makes the code less tool-friendly and introduces templates where we don't really need them.

Removing the mixins will introduce a bit of code duplication. However, these functions have been stable for a long time and the only duplication will happen in the brokers (that we'll remove eventually). So I think the bit of code duplication is well worth it to make the code less reliant on templates and easier to read.

Neverlord:
Looks like I got a bit ahead of myself. Simplifying a bit here, our inheritance graph looks a somewhat like this:

```
.
└── abstract_actor
    ├── blocking_actor              <- sender (dynamically typed)
    ├── scheduled_actor
    │   ├── event_based_actor       <- sender (dynamically typed)
    │   └── typed_event_based_actor <- sender (statically typed)
    ├── abstract_actor_shell
    │   ├── actor_shell             <- sender (dynamically typed)
    │   └── typed_actor_shell       <- sender (statically typed)
    └── abstract_broker
        ├── broker                  <- sender (dynamically typed)
        └── typed_broker            <- sender (statically typed)
```

The `sender` mixin is needed to add the proper `send` overloads at the leaf type. We only know whether we'll have dynamically or statically typed actor at the leaf, so we can't pull that up. The same goes for the `requester`.

If we get rid of the `sender` mixin, we'll have to duplicate quite a bit of code. And if we would want to add some new `send` variation, we'd have it to add in lots of places.

PR #1726, opened 2024-01-25
Remove the behavior_changer mixin

Neverlord:
Using mixin-based composition is great for prototyping. However, mixins are not a first-class thing in C++ and the template-based mixins in CAF make it harder to read the code for newcomers (and tools). Not using the mixin for inserting `become()` and `unbecome()` introduces a bit of code duplication, but overall makes the code base easier to follow.

Relates #1725.


riemass (approved):
Fully agree on removing this mixin, the resulting code size really isn't big. Hope the other mixins turn out simple as this one.

PR #1727, opened 2024-01-26
Simplify and streamline existing send functions

Neverlord:
- Add a new function to perform type-checks for send functions
- Remove the `eq_impl` helper; consistently use `enqueue`
- Move business logic for send functions to `local_actor`
- Simplify `sender` mixin by delegating to the new functions

@riemass not quite removing the `sender`, but at least making it more lightweight (I hope).


riemass (approved):
LGTM and eases things up :+1:

PR #1729, opened 2024-01-27
Use more concise C++17 namespace syntax

Neverlord:
Just a small nitpick I've found.


riemass (approved):
Just as a note, there's a clang-tidy check for this that we could integrate. #1435

Neverlord:
> Just as a note, there's a clang-tidy check for this that we could integrate. #1435

I've played around with `clang-tidy` yesterday, that's how I found it.

PR #1732, opened 2024-01-27
Add missing unit tests to the CMake file

Neverlord:
@shariarriday adding test code without actually running it on CI shouldn't have happened. I only found them today by accident.

Please make sure to actually run the tests you're working on locally. You can run `<binary> -s <name>` to run a single suite. For example, running `caf-core-test -s metaprogramming` would have shown you `Suites: 0 / 0`, clearly indicating that the test wasn't actually added to the build. At least the tests seem to work after enabling them.


shariarriday (approved):
Thanks, I think the tests were removed while resolving conflicts. I will ensure that the tests are properly added before finalizing PR.

PR #1733, opened 2024-01-27
Allow setting name when using `actor_from_state`

dominiklohmann:
This adds the missing support for setting the actor name when using the new `actor_from_state` utility.

Note that there's a related bug in `detail::has_name<T>`, which doesn't compile when `T` is marked `final`. However, that is likely best fixed in a separate PR.

### DIFF
--- a/libcaf_core/caf/actor_from_state.hpp
+++ b/libcaf_core/caf/actor_from_state.hpp
@@ -37,6 +37,18 @@ public:
       state.~State();
   }
 
+  const char* name() const override {
+    if constexpr (detail::has_name<State>::value) {
+      if constexpr (!std::is_member_pointer<decltype(&State::name)>::value) {
+        if constexpr (std::is_convertible<decltype(State::name),
+                                          const char*>::value) {
+          return State::name;
+        }
+      }
+    }
+    return super::name();
+  }
+
   union {
     State state;
   };


libcaf_core/caf/actor_from_state.hpp
```diff
@@ -37,6 +37,18 @@ class actor_from_state_impl : public Base {
       state.~State();
   }
 
+  const char* name() const override {
+    if constexpr (detail::has_name<State>::value) {
+      if constexpr (!std::is_member_pointer<decltype(&State::name)>::value) {
+        if constexpr (std::is_convertible<decltype(State::name),
+                                          const char*>::value) {
```
dominiklohmann, line 44: There are I think two further problems with this, but it also affects all other stateful actors:
1. This function should really return a `std::string_view` and not a `const char*`. Adding a non-static `name` member or a `static constexpr std::string_view` just silently fails right now.
2. Unfortunately, this name is not overridable at runtime. If I have a number of instances of this actor, they all share the same name in logs. That's really unfortunate for debugging issues in production, where a customer can sometimes only send over log files.

Neverlord: The reason that `abstract_actor::name()` returns `const char*` is mostly historical at that point and something that bugs me as well. Version 1.0 is a good opportunity to revise that. So instead of using a (static) member variable here, we could just check if the state provides a `std::string_view name() const` member function instead. That leaves it to the user whether the state class returns a constant from that function or has a `std::string` member. The name really shouldn't change after constructing the state, but I don't think there's a way to enforce that.

If we want to return the name to be something other than `const char*`, it's not going to be trivial change, though. Not hard either, but it involves touching `abstract_actor` and most (all?) classes that inherit from it. You can give it a shot if you have the cycles or else just open a ticket instead.

Neverlord: Uh, wait. I think I made it `static` in the `state_based_actor`, because the actor might log something before the state is constructed. That would mean calling `name()` would lead to UB if it would access a state member variable. I'd have to double-check whether that's (still) the case.

Neverlord: Yeah, `make_actor` logs the name. In fact, I think that's the only time an actor logs its name. That's before a state-based actor will construct it's state. So if we want to change the `name` so that it may return a state member, we'd have to change things around more than just changing function signatures.

dominiklohmann: I think that'd be a good breaking change to make.

> I think I made it `static` in the `state_based_actor`, because the actor might log something before the state is constructed.

Given that these logs are not valuable to the user, but rather only to the CAF developer, I'd argue that these should not include the actor name.

> You can give it a shot if you have the cycles or else just open a ticket instead.

I think the ticket is preferable; I did have some time to hack around this weekend but I think you know the ins and outs of this part of the code much better than I do, so it doesn't strike me as something where my time is well spent. I'll write one after we resolved this thread.

Should we merge the PR as-is then?

PR #1735, opened 2024-01-28
Fix build on MSVC

Neverlord:
I have no idea why it didn't fail in the PR but only after merging #1727 to the main branch.


riemass (approved):
:clap:

PR #1737, opened 2024-01-29
Port legacy tests to new test framework (Part 5)

shariarriday:
- Updates some legacy tests to new framework. 
- Includes have been updated for some of the previous tests.


libcaf_core/caf/const_typed_message_view.test.cpp
```diff
@@ -0,0 +1,39 @@
...
+TEST("const message views allow access via get") {
+  auto msg = make_message(1, 2, 3, "four");
+  require((msg.match_elements<int, int, int, std::string>()));
+  const_typed_message_view<int, int, int, std::string> view{msg};
+  check_eq(get<0>(view), 1);
+  check_eq(get<1>(view), 2);
+  check_eq(get<2>(view), 3);
+  check_eq(get<3>(view), "four");
+}
+
```
Neverlord, line 36: We currently have no test for `to_tuple`. I think we should add one while we're at it:

```
TEST("to_tuple can convert messages to tuples") {
  auto msg = make_message(1, 2, 3, "four");
  SECTION("to_tuple returns nullopt if the types mismatch") {
    ...
  }
  SECTION("to_tuple returns the values if the types match") {
    ...
  }
```

libcaf_core/caf/cow_tuple.test.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+using namespace caf;
+
+namespace {
+
+struct fixture : test::fixture::deterministic {
+  template <class... Ts>
+  caf::byte_buffer serialize(const Ts&... xs) {
+    caf::byte_buffer buf;
+    caf::binary_serializer sink{sys, buf};
```
Neverlord, line 26: We don't really need the fixture here. You can turn these three functions into free functions and pass `nullptr` instead of `sys`.

libcaf_core/caf/node_id.test.cpp
```diff
@@ -49,8 +49,6 @@ T unbox(std::optional<T> x) {
   return std::move(*x);
 }
 
-} // namespace
-
 #define CHECK_PARSE_OK(str, ...)                                               \
```
Neverlord, line 52: Defines are an old C construct. We never put macros into namespaces by convention, because macros are not affected by namespaces in the first place.

Neverlord (changes requested):
Another dozen down! 🙂

PR #1738, opened 2024-01-30
Remove macro based logging from libcaf_core

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/1652

### DIFF
--- a/libcaf_core/caf/abstract_actor.cpp
+++ b/libcaf_core/caf/abstract_actor.cpp
@@ -161,8 +161,8 @@ bool abstract_actor::cleanup(error&& reason, execution_unit* host) {
   });
   if (!set_fail_state)
     return false;
-  log::core::debug("cleanup {} {} {}", CAF_ARG(id()), CAF_ARG(node()),
-                   CAF_ARG(fail_state_));
+  log::core::debug("cleanup id = {} node = {} fail-state = {}", id(), node(),
+                   fail_state_);
   // send exit messages
   for (attachable* i = head.get(); i != nullptr; i = i->next.get())
     i->actor_exited(fail_state_, host);
--- a/libcaf_core/caf/abstract_actor.cpp
+++ b/libcaf_core/caf/abstract_actor.cpp
@@ -12,7 +12,7 @@
 #include "caf/config.hpp"
 #include "caf/default_attachable.hpp"
 #include "caf/execution_unit.hpp"
-#include "caf/logger.hpp"
+#include "caf/log/core.hpp"
 #include "caf/mailbox_element.hpp"
 #include "caf/system_messages.hpp"
 
@@ -46,7 +46,7 @@ void abstract_actor::attach(attachable_ptr ptr) {
     return true;
   });
   if (!attached) {
-    CAF_LOG_DEBUG(
+    log::core::debug(
       "cannot attach functor to terminated actor: call immediately");
     ptr->actor_exited(fail_state, nullptr);
   }
@@ -72,7 +72,7 @@ size_t abstract_actor::detach_impl(const attachable::token& what,
     if ((*i)->matches(what)) {
       ++count;
       if (!dry_run) {
-        CAF_LOG_DEBUG("removed element");
+        log::core::debug("removed element");
         attachable_ptr next;
         next.swap((*i)->next);
         (*i).swap(next);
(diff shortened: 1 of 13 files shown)


Neverlord (changes requested):

libcaf_core/caf/abstract_actor.cpp
```diff
@@ -165,8 +165,8 @@ bool abstract_actor::cleanup(error&& reason, execution_unit* host) {
   });
   if (!set_fail_state)
     return false;
-  CAF_LOG_DEBUG("cleanup" << CAF_ARG(id()) << CAF_ARG(node())
-                          << CAF_ARG(fail_state_));
+  log::core::debug("cleanup id = {} node = {} fail-state = {}", id(), node(),
```
Neverlord, line 168: To make this a bit more readable:

```suggestion
  log::core::debug("cleanup: id = {}, node = {}, fail-state = {}", id(), node(),
```

libcaf_core/caf/abstract_actor.cpp
```diff
@@ -189,15 +189,15 @@ void abstract_actor::register_at_system() {
     return;
   setf(is_registered_flag);
   [[maybe_unused]] auto count = home_system().registry().inc_running();
-  CAF_LOG_DEBUG("actor" << id() << "increased running count to" << count);
+  log::core::debug("actor {} increased running count to {}", id(), count);
```
Neverlord, line 192: I think this line should go to the `system` category (needs `log/system.hpp`), where we log events for spawning / terminating actors.

```suggestion
  log::system::debug("actor {} increased running count to {}", id(), count);
```

libcaf_core/caf/abstract_actor.cpp
```diff
@@ -189,15 +189,15 @@ void abstract_actor::register_at_system() {
...
 }
 
 void abstract_actor::unregister_from_system() {
   if (!getf(is_registered_flag))
     return;
   unsetf(is_registered_flag);
   [[maybe_unused]] auto count = home_system().registry().dec_running();
-  CAF_LOG_DEBUG("actor" << id() << "decreased running count to" << count);
+  log::core::debug("actor {} decreased running count to {}", count);
```
Neverlord, line 200:
```suggestion
  log::system::debug("actor {} decreased running count to {}", count);
```

libcaf_core/caf/blocking_actor.cpp
```diff
@@ -140,8 +140,8 @@ class blocking_actor_runner : public resumable {
     sys.release_private_thread(thread_);
     if (!hidden_) {
       [[maybe_unused]] auto count = sys.registry().dec_running();
-      CAF_LOG_DEBUG("actor" << self_->id() << "decreased running count to"
-                            << count);
+      log::core::debug("actor {} decreased running count to {}", self_->id(),
```
Neverlord, line 143:
```suggestion
      log::system::debug("actor {} decreased running count to {}", self_->id(),
```

libcaf_core/caf/blocking_actor.cpp
```diff
@@ -174,7 +174,7 @@ void blocking_actor::launch(execution_unit*, bool, bool hide) {
   // decrementing the count before releasing the thread.
   if (!hide) {
     [[maybe_unused]] auto count = sys.registry().inc_running();
-    CAF_LOG_DEBUG("actor" << id() << "increased running count to" << count);
+    log::core::debug("actor {} increased running count to {}", id(), count);
```
Neverlord, line 177:
```suggestion
    log::system::debug("actor {} increased running count to {}", id(), count);
```

Neverlord (changes requested):
Please also take care of the `..._IF` macros like `CAF_LOG_DEBUG_IF`.

`CAF_LOG_DEBUG_IF(cond, ...)` becomes `if (cond)log::core::debug(...)`.

libcaf_core/caf/abstract_actor.cpp
shariarriday: I have added the following in the robot tests to ensure that the build is successful after these changes. [robot/logging/output.robot](https://github.com/actor-framework/actor-framework/pull/1738/commits/37b2a807f84b724bd46c530aecda5cde592e822c#diff-173a2923d29ecc6d8e5f62972ffedeb7f9e4bb71788871c923a68d6d742b734f)

Neverlord: We don't care about these. Instead of changing the baseline, please add `caf.system` to the ignored components for the test: https://github.com/actor-framework/actor-framework/blob/master/robot/logging/base.cfg

Neverlord:
Just an FYI. With this set of changes, we switch from building CAF without logging by default to building CAF with logging always built-in. This means there will be additional run-time checks now to check whether the output is suppressed by a filter (it is by default). I'll benchmark the changes you've made with our benchmarking suites to make sure the extra overhead is negligible, so the PRs might stay open for a while.

Neverlord (approved):

PR #1739, opened 2024-01-31
Re-implement actor_from_state using stateful_actor

Neverlord:
Re-using the `stateful_actor` to implement `actor_from_state` reduces code duplication and using `actor_from_state` no longer delays constructing of the state. Previously, the state would only be constructed when calling `make_behavior`. This means accessing the state prior would be undefined behavior. This change makes it now safe the access the state as soon as the actor has been fully constructed.

Notably, this will allow us to have `stateful_actor::name()` return a member of the `State` class. Up until now, users had to provide a `static const char* name` member in the state class. Now, we can think about how to allow users to have a simple `std::string name` member or `std::string_view name()` getter.


riemass (approved):
LGTM :+1:.
The lifetime management from the explanation is the strongest argument for deprecating class based actors.

PR #1742, opened 2024-02-02
Refactor the scheduler interface

### LINKED ISSUE #1674: Refactor the scheduler interface
Currently, we have a namespace `caf::scheduler`. However, a single interface `caf::scheduler` would suffice. The two implementations (work stealing and work sharing) should be implementation details, not implemented in the public API.

riemass:
Closes #1674.

Hiding the `coordinator` in cpp pulled a couple of other classes into the cpp, since they are in a `circular` template dependency (`circular` for the lack of a better name). I've also cleaned up headers on all the files I've touched and removed a lot of legacy files. 
@Neverlord  What do you think if we should loose the `policy` template specialization and implement `work_sharing` and `work_stealing` as complete classes?


libcaf_core/caf/local_actor.hpp
Neverlord, line 1: None of these changes have anything to do with the PR. I suspect these are tool-found? Mind filing these findings separately? Same for the other files that were only touched to re-organize includes.

libcaf_core/caf/scheduler.cpp
```diff
@@ -0,0 +1,773 @@
...
+
+#include <condition_variable>
+#include <fstream>
+#include <ios>
+#include <iostream>
+#include <memory>
+#include <random>
+#include <thread>
+
+#define CONFIG(str_name, var_name)                                             \
```
Neverlord, line 29: There are multiple implementations, but this seems to be for the work-stealing implementation specifically. In any case, all calls of this macro are in a single function, so a lambda would be better.

libcaf_core/caf/scheduler.cpp
```diff
@@ -0,0 +1,773 @@
...
+  size_t id_;
+  // pointer to central coordinator
+  scheduler_impl_ptr parent_;
+  // policy-specific data
+  policy_data data_;
+  // instance of our policy object
+  Policy policy_;
+};
+
+/// Policy-based implementation of the abstract coordinator base class.
```
Neverlord, line 310: There no longer is an abstract coordinator.

libcaf_core/caf/scheduler.hpp
```diff
@@ -4,6 +4,85 @@
...
+/// A coordinator creates the workers, manages delayed sends and
+/// the central printer instance for {@link aout}. It also forwards
+/// sends from detached workers or non-actor threads to randomly
+/// chosen workers.
+class CAF_CORE_EXPORT scheduler : public actor_system::module {
+public:
+  // -- factory functions ------------------------------------------------------
+  static scheduler* make_work_stealing(actor_system& sys);
+
+  static scheduler* make_work_sharing(actor_system& sys);
```
Neverlord, line 26: Should not return raw pointers (or not be public).

libcaf_core/caf/scheduler.hpp
```diff
@@ -4,6 +4,85 @@
...
+
+  static void cleanup_and_release(resumable*);
+
+  virtual detail::actor_local_printer_ptr printer_for(local_actor* self);
+
+  // -- properties -------------------------------------------------------------
+
+  static size_t default_thread_count() noexcept {
+    return std::max(std::thread::hardware_concurrency(), 4u);
+  }
```
Neverlord, line 55: I think this should move to the `.cpp` file, even if it's a small function. If we ever change the default around, we want linked applications to use the new default.

libcaf_core/tests/legacy/thread_hook.cpp
```diff
@@ -99,22 +101,24 @@ BEGIN_FIXTURE_SCOPE(fixture<counting_thread_hook>)
...
-  auto fallback = scheduler::abstract_coordinator::default_thread_count();
+  auto fallback = scheduler::default_thread_count();
   assumed_thread_count = get_or(cfg, "caf.scheduler.max-threads", fallback)
                          + 2; // clock and private thread pool
-  auto& sched = sys.scheduler();
-  if (sched.detaches_utility_actors())
-    assumed_thread_count += sched.num_utility_actors();
+                              // auto& sched = sys.scheduler();
+                              // if (sched.detaches_utility_actors())
+  //   assumed_thread_count += sched.num_utility_actors();
```
Neverlord, line 109: Did you mean to remove this?

libcaf_core/tests/legacy/thread_hook.cpp
```diff
@@ -99,22 +101,24 @@ BEGIN_FIXTURE_SCOPE(fixture<counting_thread_hook>)
...
-  auto fallback = scheduler::abstract_coordinator::default_thread_count();
+  auto fallback = scheduler::default_thread_count();
   assumed_thread_count = get_or(cfg, "caf.scheduler.max-threads", fallback)
                          + 3; // clock, private thread pool, and  detached actor
-  auto& sched = sys.scheduler();
-  if (sched.detaches_utility_actors())
-    assumed_thread_count += sched.num_utility_actors();
+                              // auto& sched = sys.scheduler();
+                              // if (sched.detaches_utility_actors())
+  // assumed_thread_count += sched.num_utility_actors();
```
Neverlord, line 120: Did you mean to remove this?

libcaf_io/caf/io/basp/worker.hpp
```diff
@@ -35,7 +33,7 @@ class CAF_IO_EXPORT worker : public detail::abstract_worker,
 
   using super = detail::abstract_worker;
 
-  using scheduler_type = scheduler::abstract_coordinator;
+  using scheduler_type = scheduler;
```
Neverlord, line 36: The alias is longer than the actual type. 😉

I'd drop it.

Neverlord (approved):
Just some nitpicks.

libcaf_core/caf/scheduler.cpp
riemass: Calls are in the initializer list so we cant make a lambda here, but I'll inline the calls, doesn't look bad.

libcaf_core/caf/scheduler.hpp
riemass: This was in a cpp but I inlined it thinking it's a short function. Moving back.

libcaf_core/tests/legacy/thread_hook.cpp
riemass: Yes, tested and forgot to remove.

libcaf_core/caf/local_actor.hpp
riemass: Cleaned up 11 files, will follow up.

libcaf_io/caf/io/basp/worker.hpp
riemass: Seems it wasn't used.

Neverlord:
> @Neverlord What do you think if we should loose the policy template specialization and implement work_sharing and work_stealing as complete classes?

@riemass all for it, but let's do that as a followup.

Neverlord (approved):

PR #1744, opened 2024-02-05
Consolidate serializer tests

shariarriday:


libcaf_core/caf/serialization.test.cpp
```diff
@@ -0,0 +1,767 @@
...
+  }
+
+private:
+  std::vector<std::string> parse_value(const std::string& list_val,
+                                       const std::string& delimiter) {
+    const std::regex delimiter_regex{delimiter};
+    return std::vector<std::string>{
+      std::sregex_token_iterator(list_val.cbegin(), list_val.cend(),
+                                 delimiter_regex, -1),
+      {}};
```
Neverlord, line 633: I think using a `regex` here is overkill. Isn't this simply a `split`? https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/string_algorithms.hpp#L35

Neverlord (changes requested):
Just one nitpick.

libcaf_core/caf/serialization.test.cpp
shariarriday: Thanks, that is helpful. I was not aware of the existing split functions. I am working on updating the PR accordingly.

Neverlord (approved):
Thanks for wrapping this up! 👍

Issue #1745, opened 2024-02-06
Implement new mail API


Neverlord:
Straw man API examples:

```
self->mail(1, 2, 3).delay(1s, strong_ref).send(buddy);
self->mail(1, 2, 3).schedule(now() + 5s, weak_ref).request(buddy, 1s).then(...);
self->mail(1, 2, 3).delegate(buddy);
anon_mail(1, 2, 3).urgent().send(buddy);
```

Context: #1227.

Neverlord:
Nothing left to do, except documentation updates and deprecation of the old API.

PR #1748, opened 2024-02-09
Implement new mail DSL for anonymous messages

Neverlord:
First chunk for the new `mail` API. The anonymous messages are the simplest DSL, but the other pieces will follow the same template.

Relates #1745.

### DIFF
--- a/libcaf_core/caf/anon_mail.test.cpp
+++ b/libcaf_core/caf/anon_mail.test.cpp
@@ -1,0 +1,153 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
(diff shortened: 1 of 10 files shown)


libcaf_core/caf/anon_mail.test.cpp
```diff
@@ -0,0 +1,153 @@
...
+      .from(nullptr)
+      .to(dummy);
+  }
+  SECTION("invalid receiver") {
+    anon_mail("hello world").urgent().send(actor{});
+    check_eq(mail_count(), 0u);
+  }
+}
+
+TEST("send delayed message") {
```
riemass, line 50: We have good coverage because `delay` is implemented via `schedule`, but I would still like seeing a `.schedule(...)` call in the tests, at least for exemplary purposes.

riemass (changes requested):
Looks good, just two nits from me.

riemass (approved):

PR #1751, opened 2024-02-12
Auto-move values from messages whenever possible

### LINKED ISSUE #1740: Auto-move values from messages whenever possible
Currently, using a message handler such as `[](std::string str) { ... }` will always lead to a copy of the string. Users can ask CAF for a mutable reference into the message via `[](std::string& str) { ... }`. Due to the copy-on-move semantics, this will cause the message to detach its contents as necessary.

We can simplify this by having CAF to always move message elements by default if the reference count is exactly 1. So basically, the workflow would be:

- If any message handler takes a mutable ref (1) *or* the reference count is exactly 1:
  - Get an unshared reference to the message values
  - Call the matching handler as follows:
    - `T&` and `const T&` -> pass the reference
    - `T` pass `std::move(value)`
- Otherwise:
  - Call the matching handler as follows:
    - `const T&` -> pass the (`const`) reference
    - `T` pass `value` (make a copy of the value, but not the entire message)

In practice, this means that users can skip using mutable reference and just use the regular value semantics. Users should still use `const T&` for non-trivial types such as `std::string` whenever they don't need to modify/move the input to the message handler, but that's just the same as writing regular C++ functions.

(1) We'll keep support for `T&` in message handlers for backwards compatibility, but we can think about deprecating/removing it at some point.

This change has been championed by @dominiklohmann and I think it will definitely make CAF easier to teach and use.

Neverlord:
Closes #1740.

### DIFF
--- a/libcaf_core/caf/behavior.test.cpp
+++ b/libcaf_core/caf/behavior.test.cpp
@@ -6,7 +6,10 @@
 
 #include "caf/test/test.hpp"
 
+#include "caf/cow_string.hpp"
+
 using namespace caf;
+using namespace std::literals;
 
 namespace {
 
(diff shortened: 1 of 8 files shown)


riemass (approved):
LGTM. Using ref count to test this is a really good idea :+1: .

PR #1753, opened 2024-02-13
Implement new event_based_mail DSL

Neverlord:
The `event_based_mail` API supports asynchronous messages (by extending the `async_mail` DSL) and `request` for event-based actors.

The new DSL also implements a new `event_based_response_handle` type instead of re-using the existing response handle type. This is in preparation of discontinuing the previous response handle type, which we want to get rid of since it became too complex over the years.

The new response handle does not implement `to_observable()` yet, this is going to be a followup.

Relates #1745.

### DIFF
--- a/libcaf_core/caf/actor_system.hpp
+++ b/libcaf_core/caf/actor_system.hpp
@@ -610,6 +610,7 @@ public:
                   "only scheduled actors may get spawned inactively");
     CAF_SET_LOGGER_SYS(this);
     actor_config cfg{dummy_execution_unit(), nullptr};
+    cfg.flags = abstract_actor::is_inactive_flag;
     if constexpr (has_detach_flag(Os))
       cfg.flags |= abstract_actor::is_detached_flag;
     if constexpr (has_hide_flag(Os))
@@ -624,8 +625,9 @@ public:
     auto launch = [strong_ptr = std::move(res), host{cfg.host}] {
       // Note: we pass `res` to this lambda instead of `ptr` to keep a strong
       //       reference to the actor.
-      static_cast<Impl*>(actor_cast<abstract_actor*>(strong_ptr))
-        ->launch(host, has_lazy_init_flag(Os), has_hide_flag(Os));
+      auto dptr = static_cast<Impl*>(actor_cast<abstract_actor*>(strong_ptr));
+      dptr->unsetf(abstract_actor::is_inactive_flag);
+      dptr->launch(host, has_lazy_init_flag(Os), has_hide_flag(Os));
     };
     return std::make_tuple(ptr, launcher<decltype(launch)>(std::move(launch)));
   }
--- a/libcaf_core/caf/event_based_mail.hpp
+++ b/libcaf_core/caf/event_based_mail.hpp
@@ -1,0 +1,142 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/async_mail.hpp"
+#include "caf/event_based_response_handle.hpp"
+#include "caf/scheduled_actor.hpp"
+
+namespace caf {
+
+/// Provides a fluent interface for sending asynchronous messages to actors at a
+/// specific point in time.
+template <message_priority Priority, class Trait, class... Args>
+class event_based_scheduled_mail_t
+  : public async_scheduled_mail_t<Priority, Trait, Args...> {
+public:
+  using super = async_scheduled_mail_t<Priority, Trait, Args...>;
+
+  event_based_scheduled_mail_t(scheduled_actor* self, message&& content,
+                               actor_clock::time_point timeout)
+    : super(self, std::move(content), timeout) {
+    // nop
+  }
+
+  /// Sends the message to `receiver` as a request message and returns a handle
+  /// for processing the response.
+  /// @param receiver The actor that should receive the message.
+  /// @param ref_tag Either `strong_ref` or `weak_ref`. When passing
+  ///                `strong_ref`, the system will keep a strong reference to
+  ///                the receiver until the message has been delivered.
+  ///                Otherwise, the system will only keep a weak reference to
+  ///                the receiver and the message will be dropped if the
+  ///                receiver has been garbage collected in the meantime.
+  /// @param self_ref_tag Either `strong_self_ref` or `weak_self_ref`. When
+  ///                     passing `strong_self_ref`, the system will keep a
+  ///                     strong reference to the sender until the message has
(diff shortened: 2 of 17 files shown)


libcaf_core/caf/event_based_mail.test.cpp
```diff
@@ -0,0 +1,289 @@
...
+  using sender_actor = typed_actor<result<void>(int)>;
+  auto dummy = sys.spawn([]() -> dummy_behavior {
+    return {
+      [](int value) { return value * value; },
+    };
+  });
+  auto [self, launch] = sys.spawn_inactive<sender_actor::impl>();
+  auto self_hdl = actor_cast<actor>(self);
+  auto result = std::make_shared<int>(0);
+  self->mail(3).request(dummy).then([result](int x) { *result = x; });
```
dominiklohmann, line 272: If you're touching `request_handle` anyways, then I have one usability ask for it when the receiver is a typed actor:

```cpp
self->mail(3).request(...).then(
  [self](auto response) {
    // decltype(response) is expected<int>
  });
```

I know this may not be easy to implement with the current overloads available for `then(...)`, so maybe it just needs to be a separate function like `then_auto(...)`, but it would make code a lot easier to write and to maintain when using typed actors.

dominiklohmann: Ah, didn't know that you actually had this planned now. Must've missed that in our previous discussions on this. That makes me _really_ happy! ❤️

Neverlord:
@riemass regarding the second commit: I have no idea what change triggered that false positive in GCC. I've tried reproducing it in a Debian 11 Docker image (on Apple Silicon) and on a CentOS 7 (x64 hardware, but an older chip) without any luck. So it seems like it's not just compiler but also HW dependent somehow. For the record, this was the error we got on CI:

```
[ 35%] Building CXX object examples/CMakeFiles/delegating.dir/message_passing/delegating.cpp.o
In file included from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/intrusive_cow_ptr.hpp:7,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/message.hpp:13,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/error.hpp:12,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/attachable.hpp:8,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/abstract_actor.hpp:7,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/actor_system.hpp:7,
                 from /users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/examples/flow/observe-on.cpp:3:
/users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/intrusive_ptr.hpp: In function 'void {anonymous}::caf_main(caf::actor_system&, const {anonymous}::config&)':
/users/jenkins/workspace/work_issue_1745-event-based-mail@13/sources/libcaf_core/caf/intrusive_ptr.hpp:145:12: error: '*((void*)(&<structured bindings>)+8).caf::intrusive_ptr<caf::actor_control_block>::ptr_' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  145 |     return ptr_;
      |            ^~~~
cc1plus: all warnings being treated as errors
gmake[2]: *** [examples/CMakeFiles/observe-on.dir/build.make:82: examples/CMakeFiles/observe-on.dir/flow/observe-on.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1197: examples/CMakeFiles/observe-on.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```

That's a) seemingly unrelated to my additions and b) what-the-heck is that error, GCC?

Anyways, I don't think it's worth wasting time on.

libcaf_core/caf/event_based_mail.test.cpp
Neverlord: > Ah, didn't know that you actually had this planned now.

Sorry, that's my bad. I don't think I've ever mentioned it. 🙂

I thought about the response handle while working on this and ultimately decided splitting the response promise into two separate classes (event-based and blocking). I had that discussion from a while back still in mind and think its a worthwhile addition, but I wanted to have all the `mail` API pieces in place first.

riemass:
> That's a) seemingly unrelated to my additions and b) what-the-heck is that error, GCC?

Remember having something similar a while back when we enabled `-Wall` on the pipeline, and after I lost quite a bit of time on analysis, it turned out to be a compiler bug that was fixed in a later release. Centos7 is known for using older GCC releases, so I think you made a wise choice by not wasting more time on it. :smile:

libcaf_core/caf/event_based_mail.hpp
```diff
@@ -0,0 +1,142 @@
...
+      auto& clock = self()->clock();
+      in_flight = clock.schedule_message(actor_cast(super::self_, self_ref_tag),
+                                         actor_cast(receiver, ref_tag),
+                                         super::timeout_, mid,
+                                         std::move(super::content_));
+
+    } else {
+      self()->enqueue(make_mailbox_element(self()->ctrl(), mid.response_id(),
+                                           make_error(sec::invalid_request)),
+                      self()->context());
```
riemass, line 61: This should be simple enough to be covered by a unit test, delayed request to an invalid receiver.

riemass (changes requested):
Two small requests, looks good otherwise.

libcaf_core/caf/actor_system.hpp
```diff
@@ -624,8 +625,9 @@ class CAF_CORE_EXPORT actor_system {
     auto launch = [strong_ptr = std::move(res), host{cfg.host}] {
       // Note: we pass `res` to this lambda instead of `ptr` to keep a strong
       //       reference to the actor.
-      static_cast<Impl*>(actor_cast<abstract_actor*>(strong_ptr))
-        ->launch(host, has_lazy_init_flag(Os), has_hide_flag(Os));
+      auto dptr = static_cast<Impl*>(actor_cast<abstract_actor*>(strong_ptr));
+      dptr->unsetf(abstract_actor::is_inactive_flag);
```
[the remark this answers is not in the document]
Neverlord: I've needed this to get the deterministic fixture to work properly. If you send a message to an inactive actor, it'll be scheduled. With the deterministic fixture, we'll also run any initialization code immediately before the first message runs. Without this inactive flag, we will run the initialization code of an inactive actor. It'll shut down as a result (no behavior, no pending requests, etc) and consequently refuse any incoming message. It's a bit unfortunate that we need this flag on the actor itself, but I couldn't find another way to make the deterministic tests do the right thing.

riemass (approved):

PR #1756, opened 2024-02-14
Implement unit tests for scheduler implementations

### LINKED ISSUE #1674: Refactor the scheduler interface

riemass:
Closes #1674.

### DIFF
--- a/libcaf_core/caf/scheduler.test.cpp
+++ b/libcaf_core/caf/scheduler.test.cpp
@@ -1,0 +1,162 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/scheduler.hpp"
+
+#include "caf/test/outline.hpp"
+#include "caf/test/scenario.hpp"
+
+#include "caf/actor_system_config.hpp"
+#include "caf/async/file.hpp"
+#include "caf/detail/latch.hpp"
+
+#include <string>
+
+using namespace caf;
+using namespace std::literals;
+using detail::latch;
+
+namespace {
+
+struct testee : resumable, ref_counted {
+  testee(std::shared_ptr<latch> latch_handle)
+    : rendesvous(std::move(latch_handle)) {
+  }
+  ~testee() = default;
+  subtype_t subtype() const override {
+    return resumable::function_object;
+  }
+  resume_result resume(execution_unit*, size_t max_throughput) override {
+    if (++runs == 10) {
+      rendesvous->count_down();
+      received_throughput = max_throughput;
+      return resumable::done;
+    }
+    return resumable::resume_later;
+  }
+  void intrusive_ptr_add_ref_impl() override {
+    ref();
+  }
+  void intrusive_ptr_release_impl() override {
+    deref();
+  }
+  std::atomic<size_t> runs = 0;
+  size_t received_throughput = 0;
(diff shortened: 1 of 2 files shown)


Neverlord (changes requested):
These tests are not focused on the thing they are actually testing. The tests should test the scheduler at the proper level of abstraction (`resumable`), but it pulls in a lot of high-level API pieces that build on top of the scheduler API, even pulling in flows.

The tests should exercise only the units under test.

riemass:
> The tests should exercise only the units under test.

Noted. I'll rewrite the tests at the proper abstraction layer.

libcaf_core/caf/scheduler.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+using namespace std::literals;
+using detail::latch;
+
+namespace {
+
+struct testee : resumable, ref_counted {
+  testee(std::shared_ptr<latch> latch_handle)
+    : rendesvous(std::move(latch_handle)) {
+  }
+  ~testee() = default;
```
Neverlord, line 26:
```suggestion
```

libcaf_core/caf/scheduler.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+  size_t received_throughput = 0;
+  std::shared_ptr<latch> rendesvous;
+};
+
+OUTLINE("scheduling resumables") {
+  GIVEN("an actor system using the work <sched> scheduler") {
+    auto sched = block_parameters<std::string>();
+    actor_system_config cfg;
+    cfg.set("caf.scheduler.policy", sched);
+    cfg.set("caf.scheduler.max-throughput", 5);
```
Neverlord, line 54: Also, should set `cfg.set("caf.scheduler.max-threads", 2)` here to avoid spinning up dozens of unnecessary threads when running on decent hardware. Same for the other test.

libcaf_core/caf/scheduler.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    return resumable::resume_later;
+  }
+  void intrusive_ptr_add_ref_impl() override {
+    ref();
+  }
+  void intrusive_ptr_release_impl() override {
+    deref();
+  }
+  std::atomic<size_t> runs = 0;
+  size_t received_throughput = 0;
```
Neverlord, line 45:
```suggestion
  std::atomic<size_t> received_throughput = 0;
```

libcaf_core/caf/scheduler.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+        check_eq(worker->runs.load(), 10u);
+      }
+      AND_THEN("expect the correct max throughput") {
+        check_eq(worker->received_throughput, 5u);
+      }
+      AND_THEN("the scheduler releases the ref when done") {
+        check_eq(worker->get_reference_count(), 1u);
+      }
+    }
+    AND_WHEN("scheduling multiple resumables") {
```
Neverlord, line 73: Has nothing to do with the previous test, so it shouldn't continue it.

```suggestion
    WHEN("scheduling multiple resumables") {
```

riemass: Opened #1776 for this. I get an exception when I put two WHEN blocks inside an outline.

Neverlord (approved):

PR #1758, opened 2024-02-15
Fix request timeout handling for the new mail API

Neverlord:
I've forgotten to add the timeouts for `request` in the last PR.


riemass (approved):
LGTM. 

Unrelated to this PR, could we make a workaround on the `.first` call?
Instead of writing:
```
self->mail(3)
  .request(dummy, infinite)
  .first \\ disposable gets dropped here...
  .then(...);
```
could we wrap the API like:
```
auto disposable = self->mail(3)
  .request(dummy, infinite)
  .then(...);
```
It could be done with a custom implementation of `pair` that has a `then` member function that delegates the call to first, and returns the second.
Writing `first` is semantically confusing, `clang-format` works against the grain, and if you needed to store the disposable it would break the call chain.

Issue #1600, opened 2023-10-11
Remove signatures from typed actor handles


dominiklohmann:
Since 1.0 is coming up, I think it makes sense to revisit the idea of making a breaking change to remove signatures from typed actor handles.

From my experience working with CAF in a code base where we require every actor to be strongly typed, it is a major pain especially for newer developers to our code base to read compiler errors. I propose to add an indirection for this reason.

```cpp
// old
using calculator_handle = caf::typed_actor<
  auto (add_atom, int lhs, int rhs) -> caf::result<int>,
  auto (subtract_atom, int lhs, int rhs) -> caf::result<int>
>;

struct calculator_state {
  static constexpr auto name = "calculator";
};

auto calculator(calculator_handle::stateful_pointer<calculator_state> self)
    -> calculator_handle::behavior_type {
  // …
}
```

```cpp
// new
using calculator_handle = caf::typed_actor<"calculator">;

template <>
struct caf::typed_actor_traits<"calculator"> {
  using behavior_type = caf::typed_behavior<
    auto (add_atom, int lhs, int rhs) -> caf::result<int>,
    auto (subtract_atom, int lhs, int rhs) -> caf::result<int>
  >;
}

struct calculator_state {
};

auto calculator(calculator_handle::stateful_pointer<calculator_state> self)
    -> calculator_handle::behavior_type {
  // …
}
```

The advantage of this change is that error messages will reference `caf::typed_actor<"calculator">` rather than `caf::typed_actor<auto (add_atom, int, int) -> caf::result<int>, auto (subtract_atom, int, int)->caf::result<int>>` when referencing the actor handle in a compiler error. Especially with longer signatures this will be a huge improvement for the developer experience.

Neverlord:
> `caf::typed_actor<"calculator">`

Was there a recent C++ feature that allows using string literals in templates now?

I'm not so sure about the trait interface, though. What about this?

```c++
// new (2)
struct calculator_tag {
  using signatures = caf::type_list<
    auto (add_atom, int lhs, int rhs) -> caf::result<int>,
    auto (subtract_atom, int lhs, int rhs) -> caf::result<int>
  >;
}

using calculator_handle = caf::typed_actor<calculator_tag>;

struct calculator_state {
  // ...
};

auto calculator(calculator_handle::stateful_pointer<calculator_state> self)
    -> calculator_handle::behavior_type {
  // …
}
```

That might even be doable without introducing a breaking change.

Downsides:

[...]

PR #1761, opened 2024-02-19
Implement trait based type interfaces for actors

riemass:
Relates #1600 

Draft to check Windows/FreeBSD pipeline.

### DIFF
--- a/libcaf_core/caf/detail/typed_actor_util.hpp
+++ b/libcaf_core/caf/detail/typed_actor_util.hpp
@@ -8,6 +8,7 @@
 #include "caf/detail/type_list.hpp"
 #include "caf/fwd.hpp"
 #include "caf/response_promise.hpp"
+#include "caf/statically_typed.hpp"
 #include "caf/system_messages.hpp"
 #include "caf/typed_response_promise.hpp"
 
@@ -114,13 +115,13 @@ template <class... Ts>
 struct extend_with_helper;
 
 template <class... Xs>
-struct extend_with_helper<typed_actor<Xs...>> {
+struct extend_with_helper<type_list<Xs...>> {
   using type = typed_actor<Xs...>;
 };
 
 template <class... Xs, class... Ys, class... Ts>
-struct extend_with_helper<typed_actor<Xs...>, typed_actor<Ys...>, Ts...>
-  : extend_with_helper<typed_actor<Xs..., Ys...>, Ts...> {
+struct extend_with_helper<type_list<Xs...>, type_list<Ys...>, Ts...>
+  : extend_with_helper<type_list<Xs..., Ys...>, Ts...> {
   // nop
 };
 
@@ -145,5 +146,26 @@ struct is_normalized_signature<result<Out...>(In...)> {
 
 template <class F>
 constexpr bool is_normalized_signature_v = is_normalized_signature<F>::value;
+
+template <class SigsList>
+struct are_signatures_normalized;
+
+template <class... Sigs>
+struct are_signatures_normalized<type_list<Sigs...>> {
+  static constexpr bool value
+    = (detail::is_normalized_signature_v<Sigs> && ...);
+};
+
+template <class SigsList>
+inline constexpr bool are_signatures_normalized_v
+  = are_signatures_normalized<SigsList>::value;
+
+template <class... Sigs>
+struct broker_from_signatures;
+
+template <class... Sigs>
+struct broker_from_signatures<type_list<Sigs...>> {
+  using type = io::typed_broker<Sigs...>;
+};
 
 } // namespace caf::detail
--- a/libcaf_core/caf/fwd.hpp
+++ b/libcaf_core/caf/fwd.hpp
@@ -299,6 +299,8 @@ namespace io {
 class hook;
 class broker;
 class middleman;
+template <class...>
+class typed_broker;
 
 namespace basp {
 
--- a/libcaf_core/caf/statically_typed.hpp
+++ b/libcaf_core/caf/statically_typed.hpp
@@ -1,0 +1,19 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include <caf/type_list.hpp>
+
+namespace caf {
+
+/// A statically typed trait type wrapping the given parameter pack of
+/// signatures as a type list.
+/// @note used for backwards compatibility when declaring typed interfaces.
+template <class... Sigs>
+struct statically_typed {
+  using signatures = type_list<Sigs...>;
+};
+
+} // namespace caf
--- a/libcaf_core/caf/typed_actor.hpp
+++ b/libcaf_core/caf/typed_actor.hpp
@@ -8,31 +8,38 @@
 #include "caf/actor.hpp"
 #include "caf/actor_cast.hpp"
 #include "caf/actor_system.hpp"
+#include "caf/detail/to_statically_typed_trait.hpp"
+#include "caf/detail/type_list.hpp"
 #include "caf/fwd.hpp"
 #include "caf/intrusive_ptr.hpp"
-#include "caf/make_actor.hpp"
 #include "caf/stateful_actor.hpp"
 #include "caf/type_id_list.hpp"
 #include "caf/typed_actor_view_base.hpp"
 #include "caf/typed_behavior.hpp"
-#include "caf/typed_response_promise.hpp"
 
 #include <cstddef>
 #include <functional>
 
 namespace caf {
 
-/// Identifies a statically typed actor.
-/// @tparam Sigs Function signatures for all accepted messages.
-template <class... Sigs>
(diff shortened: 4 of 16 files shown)


libcaf_core/caf/detail/to_trait.hpp
```diff
@@ -0,0 +1,29 @@
...
+  using type = statically_typed<T>;
+};
+
+template <class T>
+struct to_trait_oracle<T, false> {
+  using type = T;
+};
+
+template <class T>
+using to_trait_t = typename to_trait_oracle<T, std::is_function_v<T>>::type;
```
Neverlord, line 27: `trait` is way too generic here as a name. Especially without context in its own file. `to_signatures_wrapper_t` maybe? Plus, please leave a documentation here what's is used for.

libcaf_core/caf/detail/typed_actor_util.hpp
```diff
@@ -114,16 +115,24 @@ template <class... Ts>
...
 };
 
+// template <class...>
+// struct typed_extend_helper;
+//
+// template <class... Ts, class... Ys, class... Es>
+// struct typed_extend_helper<type_list<Ts...>, type_list<Ys...>, Es...> {
+//   using type = typed_actor<Ts..., Es...>;
+// };
+
```
Neverlord, line 135:
```suggestion
```

libcaf_core/caf/detail/typed_actor_util.hpp
```diff
@@ -146,4 +155,25 @@ struct is_normalized_signature<result<Out...>(In...)> {
...
+template <class... Sigs>
+struct are_signatures_normalized<type_list<Sigs...>> {
+  static constexpr bool value
+    = (detail::is_normalized_signature_v<Sigs> && ...);
+};
+
+template <class... Sigs>
+constexpr bool are_signatures_normalized_v
+  = are_signatures_normalized<Sigs...>::value;
```
Neverlord, line 169: If we're passing a `type_list`, then then the class only has a single template parameter.

```suggestion
template <class SigsList>
struct are_signatures_normalized;

template <class... Sigs>
struct are_signatures_normalized<type_list<Sigs...>> {
  static constexpr bool value
    = (detail::is_normalized_signature_v<Sigs> && ...);
};

template <class SigsList>
inline constexpr bool are_signatures_normalized_v
  = are_signatures_normalized<SigsList>::value;
```

libcaf_core/caf/fwd.hpp
```diff
@@ -302,6 +300,7 @@ namespace io {
 class hook;
 class broker;
 class middleman;
+template <class...> class typed_broker;
```
Neverlord, line 303: As a one-off, this can just be clang-formatted as usualy (and push back the `clang-format on` where it used to be)

libcaf_core/caf/statically_typed.hpp
```diff
@@ -0,0 +1,16 @@
...
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include <caf/type_list.hpp>
+
+namespace caf {
+
+template <class... Sigs>
```
Neverlord, line 11: Please add documentation here.

libcaf_core/caf/typed_actor.hpp
```diff
@@ -55,30 +57,34 @@ class typed_actor : detail::comparable<typed_actor<Sigs...>>,
   static constexpr bool has_weak_ptr_semantics = false;
 
   /// Stores the template parameter pack.
-  using signatures = type_list<Sigs...>;
+  using trait = detail::to_trait_t<TraitOrSignature>;
+
+  /// Stores the template parameter pack.
+  using signatures = typename trait::signatures;
```
Neverlord, line 63: Push this above the static assert to avoid the repeated `to_trait_t<TraitOrSignature>`.

libcaf_core/caf/typed_actor.hpp
```diff
@@ -249,11 +260,59 @@ class typed_actor : detail::comparable<typed_actor<Sigs...>>,
...
+  using pointer_view = typed_actor_pointer<T1, T2, Ts...>;
+
+  using super::super;
+
+  typed_actor(const super& other) : super(other) {
+    // nop
+  }
+  typed_actor(super&& other) : super(std::move(other)) {
     // nop
   }
```
Neverlord, line 310:
```suggestion
  typed_actor(const super& other) = default;
  
  typed_actor(super&& other) = default;
```

libcaf_core/caf/typed_actor.hpp
```diff
@@ -249,11 +260,59 @@ class typed_actor : detail::comparable<typed_actor<Sigs...>>,
...
 
-  strong_actor_ptr ptr_;
+  typed_actor& operator=(std::nullptr_t) {
```
Neverlord, line 312: Shouldn't we also default `const typed_actor&` and `typed_actor&&` overloads?

libcaf_core/caf/typed_actor.test.cpp
Neverlord, line 1: There should be tests that check that assignment between these types works. Also, the static asserts can just be at file level (with the current `TEST` subscription as comment) since the tests are all empty.

libcaf_core/tests/legacy/typed_spawn.cpp
```diff
@@ -5,16 +5,16 @@
 #include "caf/config.hpp"
 
 // exclude this suite; seems to be too much to swallow for MSVC
-#ifndef CAF_WINDOWS
+// #ifndef CAF_WINDOWS
```
Neverlord, line 8: If MSVC is fine with the suite, just remove the lines instead of commenting them out.

Neverlord (changes requested):

libcaf_core/caf/typed_actor.hpp
riemass: Turns out you can't default these, since they are not special members.

Neverlord: Can't we just `using super::super;` here?

libcaf_core/caf/detail/to_trait.hpp
riemass: Will do. Not finding the right name here. Remember saying that `signatures_wrapper` isn't a good name, but it sounds better than `statically_typed_trait` though. Any opinions?

libcaf_core/caf/typed_actor.hpp
riemass: Actually, `using super::super` worked. I've had it like that in the first implementation and could have sworn that the compiler complained stating that you can't import the copy ctor.

libcaf_core/caf/typed_actor.hpp
riemass: I've added a static assert to test this. Not sure how it works, I think the implementation falls back to the base class, by binding a `const&` to it's `operator=`.

libcaf_core/caf/typed_actor.hpp
riemass: Seems like it can't use the copy ctor after all. When I removed the two ctor overloads with `super` from the `typed_behavior`, I got this error:
```
/home/samir/projects/actor-framework/libcaf_core/caf/typed_behavior.hpp:147:7: note: candidate constructor (the implicit move constructor) not viable: cannot bind base class object of type 'typed_behavior<statically_typed<result<void> (put_atom, int), result<int> (get_atom)>>' to derived class reference 'typed_behavior<result<void> (put_atom, int), result<int> (get_atom)> &&' for 1st argument
/home/samir/projects/actor-framework/libcaf_core/caf/typed_behavior.hpp:194:3: note: candidate inherited constructor not viable: no known conversion from 'typed_behavior<statically_typed<result<void> (put_atom, int), result<int> (get_atom)>>' to 'partial_behavior_init_t' for 1st argument
  typed_behavior(partial_behavior_init_t, Ts... xs)
  ^
/home/samir/projects/actor-framework/libcaf_core/caf/typed_behavior.hpp:277:16: note: constructor from base class 'typed_behavior<caf::statically_typed<caf::result<void> (caf::put_atom, int), caf::result<int> (caf::get_atom)>>' inherited here
  using super::super;
               ^
```
I get this error when I remove the `super` overload for typed behavior. Not sure why this doesn't affect

riemass:
I'm marking this ready for review. 

* Ported the tests from `typed_spawn` to the new trait based approach. I've had problems with the skipping behavior of an actor and the deterministic fixture - as soon as an message gets skipped, every next message that I want to `inject` will fail, because the nested `expect` will see the skipped message as the first in the mailbox and fail the expectation.
* The delegate test fails, and I can't get it to work. I've tried a few approaches, but in the end I'm blaming the `dispatch_message/s` method. Does anything come to mind when returning a `degegated<T>` in the behavior that could be odd in the testing setup?
* Will port the examples as a followup.

Neverlord:
17 commits is a bit much. Also, looks like there's a conflict now. Please rebase and squash before I'll start the review.

riemass:
> Please rebase and squash before I'll start the review

Done.

Neverlord (approved):

Issue #1752, opened 2024-02-13
 High-level caf-net primitive for custom binary protocols


Neverlord:
At the moment, there is no high-level API in place for operating on the byte input of a socket. We have added flow-based file processing in #1573. Hence, the utilities for operating on sequences of bytes or characters are already in place.

If we implement an interface that basically turns a TCP connection into a flow, we get
#1408 for free, since we already have the other building blocks.

PR #1762, opened 2024-02-20
High-level caf-net primitive for custom binary protocols

Neverlord:
Implements a new flow source that operates on octet streams.

@riemass this is only the client-side DSL for now. The server part is going to be a followup after agreeing on the implementation details. This setup is slightly different to what we use for WebSocket etc., because it hands a `publisher<T>` to the user for consuming the results. However, it basically gives you the full range of the flow API in a socket handler now. I quite like this approach and I'm thinking to use the same approach for the WebSocket and length-prefix protocols. Let me know what you think. 🙂

Relates #1752.

### DIFF
--- a/libcaf_net/caf/net/octet_stream/flow_bridge.test.cpp
+++ b/libcaf_net/caf/net/octet_stream/flow_bridge.test.cpp
@@ -1,0 +1,264 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#include "caf/net/octet_stream/flow_bridge.hpp"
+
+#include "caf/test/test.hpp"
+
+#include "caf/net/fwd.hpp"
+#include "caf/net/multiplexer.hpp"
+#include "caf/net/octet_stream/transport.hpp"
+#include "caf/net/socket_manager.hpp"
+#include "caf/net/stream_socket.hpp"
+
+#include "caf/detail/latch.hpp"
+#include "caf/flow/observable_builder.hpp"
+#include "caf/flow/scoped_coordinator.hpp"
+#include "caf/log/test.hpp"
+
+using namespace caf;
+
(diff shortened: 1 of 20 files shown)


libcaf_net/caf/net/octet_stream/with.test.cpp
```diff
@@ -0,0 +1,267 @@
...
+  rendezvous.count_down_and_wait();
+  auto want = std::vector<int>();
+  want.resize(5);
+  std::iota(want.begin(), want.end(), 1);
+  check_eq(*received, want);
+  // Cleanup.
+  ping_pong_thread.join();
+}
+
+TEST("passing never to a flow bridge omits outputs") {
```
riemass, line 208: Confusing. Description should probably go in the sense of "omitting send_from".

libcaf_net/caf/net/octet_stream/flow_bridge.test.cpp
```diff
@@ -0,0 +1,264 @@
...
+  auto bridge = net::octet_stream::flow_bridge::make(
+    16, 16,
+    [received_total, &rendezvous](flow::observable<std::byte> bytes) {
+      bytes.map([](std::byte item) { return std::to_integer<int>(item); })
+        .do_finally([&rendezvous] { rendezvous.count_down(); })
+        .for_each([received_total](int) { *received_total += 1; });
+    },
+    [](flow::coordinator* self) {
+      return self->make_observable().never<std::byte>();
+    });
```
riemass, line 215: I find the notation a bit confusing, specially with the phrase `... omits ...` in the test description. Could you just confirm my reasoning: The first lambda are `inputs` as in inputs from the flow to my application, and the second lambda refers to `outputs` as in my application outputs to the flow?

riemass (changes requested):
Sorry you had to wait a bit for the review. The tests have a nice demonstration of the API and I don't think I have anything to add/suggest. The function signatures really tell you everything you need to know. I was a bit confused on the input/output phrasing but I think I got it right. 
Just a few nits on variable names and wording, otherwise looks good.

riemass (approved):
:clap:

PR #1764, opened 2024-02-21
Remove obsolete experimental features

Neverlord:
Profiling and tracing are both interesting features to have, but they do overlap with metrics (at least the profiling) and were never fully fleshed out. I think we'll revisit both ideas at some point in the future, but the current implementations are too bare-bone.


riemass (changes requested):
I did a quick grep and the profiler is mentioned in `actor_system_config.cpp` and the `configure` script. Otherwise, looks good.

Neverlord:
> I did a quick grep and the profiler is mentioned in `actor_system_config.cpp` and the `configure` script. Otherwise, looks good.

Good catch, thanks for double checking!

riemass (approved):

PR #1765, opened 2024-02-22
Implement as_observable for event_based_mail

Neverlord:
Relates #1745.

### DIFF
--- a/libcaf_core/caf/event_based_mail.hpp
+++ b/libcaf_core/caf/event_based_mail.hpp
@@ -46,7 +46,7 @@ public:
   [[nodiscard]] auto request(const Handle& receiver, timespan relative_timeout,
                              RefTag ref_tag = {},
                              SelfRefTag self_ref_tag = {}) && {
-    detail::send_type_check<typename Trait::signatures, Handle, Args...>();
+    detail::send_type_check<none_t, Handle, Args...>();
     using response_type = response_type_t<typename Handle::signatures, Args...>;
     auto mid = self()->new_request_id(Priority);
     disposable in_flight_response;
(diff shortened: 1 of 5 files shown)


libcaf_core/caf/event_based_mail.test.cpp
```diff
@@ -243,6 +244,83 @@ TEST("send request message") {
...
+        return {
+          [](int value) { return value * value; },
+        };
+      });
+      auto err = std::make_shared<error>();
+      auto result = std::make_shared<int>(0);
+      SECTION("valid response") {
+        self->mail(3)
+          .request(dummy, infinite)
+          .as_observable<int>()
```
dominiklohmann, line 259: Can this infer the type for mail sent to a typed actor?

Neverlord: Sure, the test for that is below (line 306).

dominiklohmann: I missed that, thanks! 🙈

libcaf_core/caf/event_based_mail.test.cpp
```diff
@@ -243,6 +244,83 @@ TEST("send request message") {
       expect<exit_msg>().to(dummy);
     }
   }
+  SECTION("using .to_observable for the response") {
+    SECTION("valid response") {
```
riemass, line 248: We have "valid response" nested twice, maybe this description would be a better match with the below section.
```suggestion
    SECTION("response from dynamically typed receiver") {
```

riemass (changes requested):
Small nit, looks good otherwise :+1: .

libcaf_core/caf/event_based_mail.hpp
```diff
@@ -46,7 +46,7 @@ class event_based_scheduled_mail_t
   [[nodiscard]] auto request(const Handle& receiver, timespan relative_timeout,
                              RefTag ref_tag = {},
                              SelfRefTag self_ref_tag = {}) && {
-    detail::send_type_check<typename Trait::signatures, Handle, Args...>();
+    detail::send_type_check<none_t, Handle, Args...>();
```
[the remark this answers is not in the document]
Neverlord: The signatures here is used to check whether the sender can receive the response in its regular behavior. However, a `request` does not use the regular behavior. Hence, we must skip this check.

riemass (approved):
Thanks for the explanation.

PR #1766, opened 2024-02-22
Use the new mail syntax for asynchronous messages

Neverlord:
Refactoring for using the new `mail` API for sending asynchronous messages (requests will come later).


shariarriday (changes requested):
I think the following line should be updated as well to new mail API
```
self->send(read_atom_v, "http://www.example.com/index.html", uint64_t{0},
             uint64_t{4095})
      .send(parent);
```

https://github.com/actor-framework/actor-framework/pull/1766/files#diff-a1eaf58800cb0f87cc50d18a02e44aaa15cd1596c29651db02cbd262ca27796eR148-R149

shariarriday (approved):

PR #1773, opened 2024-02-25
Simplify function chaining on delayed requests

Neverlord:
@riemass here's the promised followup with the custom "pair" to make function chaining work similarly for regular and delayed requests. Calling `then` and `await` now simply returns the handle for the delayed request, so users that want to make use of it can simply obtain it as result of the entire chain. `as_observable` already has a return type and I'd rather leave it as-is. I think >90% of the time, users will drop the `disposable` anyways. If they do need it, though, the new type also gives access to its members and can be deconstructed like a regular `pair`.

Relates #1745.

### DIFF
--- a/libcaf_core/caf/blocking_mail.hpp
+++ b/libcaf_core/caf/blocking_mail.hpp
@@ -63,9 +63,9 @@ public:
                                            make_error(sec::invalid_request)),
                       self()->context());
     }
-    using hdl_t = blocking_response_handle<response_type>;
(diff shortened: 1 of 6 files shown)


riemass (approved):
Looks good, and lifts a lot of confusion with the `.first`. Leaving `as_observable` is fine, I guess. Do you plan modifying the blocking mail as well?

riemass (approved):
Great!

PR #1775, opened 2024-02-27
Switch to the new mail(...).request(...) syntax

Neverlord:
There's one actor type that currently can't use `mail(...).request(...)` but should: `actor_shell`. Enabling the new syntax for the shells is going to be a followup.


libcaf_core/caf/async_mail.test.cpp
```diff
@@ -142,108 +142,6 @@ TEST("delegate message") {
       }
     }
   }
-  SECTION("request message") {
```
riemass, line 145: Are these tests no longer relevant?

riemass (approved):
Looks good. Just wondering why you removed ~100 lines of tests?

libcaf_core/caf/async_mail.test.cpp
Neverlord: Hm, I've only seen `request message` here and removed it since those are tested in the `event_based_mail` test suite. However, I might've been a bit trigger heavy. The tests are actually about `delegate`. Let me think about where to move them.

PR #1778, opened 2024-02-28
Refactor CAF_LOG_TRACE statements

shariarriday:
closes https://github.com/actor-framework/actor-framework/issues/1652

### DIFF
--- a/libcaf_core/caf/flow/observer.hpp
+++ b/libcaf_core/caf/flow/observer.hpp
@@ -338,14 +338,14 @@ public:
   }
 
   void on_next(const value_type& item) override {
-    auto exit_guard = log::core::trace("item = {}",
-                                       const_cast<const value_type*>(&item));
+    auto lg = log::core::trace("item = {}",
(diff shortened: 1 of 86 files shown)


libcaf_core/caf/flow/observer.hpp
```diff
@@ -338,13 +338,14 @@ class buffer_writer_impl : public detail::atomic_ref_counted,
   }
 
   void on_next(const value_type& item) override {
-    CAF_LOG_TRACE(CAF_ARG(item));
+    auto exit_guard = log::core::trace("item = {}",
+                                       const_cast<const value_type*>(&item));
```
shariarriday, line 342: const_cast was introduced in this log statement to fix the error where the build fails with `error C2440: 'static_cast': cannot convert from 'const T' to 'const void *'` for some data types like `caf::chunk`. 
Please, let me know if you have suggestion for a better alternative here.

Neverlord (approved):
Thanks for taking care of refactoring! 👍

PR #1783, opened 2024-03-03
Update examples with trait based type signatures

riemass:
Relates #1600.


Neverlord (changes requested):
Some cases have a typo: triat => trait.

Neverlord (approved):

PR #1790, opened 2024-03-07
Fix unit test in typed event based actor

### LINKED ISSUE #1600: Remove signatures from typed actor handles

riemass:
This unit test was left in the #1761 as a review item that I couldn't figure out how to implement. 
Now that I've fixed it, this PR closes #1600.

### DIFF
--- a/libcaf_core/caf/typed_event_based_actor.test.cpp
+++ b/libcaf_core/caf/typed_event_based_actor.test.cpp
@@ -168,11 +168,11 @@ string_actor::behavior_type string_reverter() {
 
 // uses `return delegate(...)`
 string_actor::behavior_type string_delegator(string_actor::pointer self,
-                                             string_actor master) {
-  self->link_to(master);
+                                             string_actor next) {
+  self->link_to(next);
   return {
     [=](std::string& str) -> delegated<std::string> {
-      return self->delegate(master, std::move(str));
+      return self->delegate(next, std::move(str));
     },
   };
 }
@@ -294,8 +294,8 @@ TEST("starting a string delegator chain") {
     check_eq(aut->message_types(), iface);
   }
   inject().with("Hello World!"s).from(self).to(aut);
-  expect<std::string>().with(std::ignore).from(self).to(delegator);
-  expect<std::string>().with(std::ignore).from(self).to(reverter);
+  expect<std::string>().from(self).to(delegator);
+  expect<std::string>().from(self).to(reverter);
   expect<std::string>().with("!dlroW olleH"s).from(reverter).to(self);
   inject_exit(aut);
 }
--- a/libcaf_core/caf/typed_event_based_actor.test.cpp
+++ b/libcaf_core/caf/typed_event_based_actor.test.cpp
@@ -168,12 +168,11 @@ string_actor::behavior_type string_reverter() {
 
 // uses `return delegate(...)`
 string_actor::behavior_type string_delegator(string_actor::pointer self,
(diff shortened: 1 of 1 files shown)


libcaf_core/caf/typed_event_based_actor.test.cpp
```diff
@@ -287,14 +286,17 @@ TEST("chainging the behavior at runtime and skipping messages") {
...
   SECTION("message_types() returns an interface description") {
     std::set<std::string> iface{"(std::string) -> (std::string)"};
     check_eq(aut->message_types(), iface);
   }
-  // inject().with("Hello World!"s).from(self).to(aut);
-  // dispatch_message();
-  // expect<std::string>().with("!dlroW olleH"s).to(self);
+  inject().with("Hello World!"s).from(self).to(aut);
+  expect<std::string>().with(std::ignore).from(self).to(delegator);
+  expect<std::string>().with(std::ignore).from(self).to(reverter);
```
Neverlord, line 298: Don't add `with` if you don't care anyways.

```suggestion
  expect<std::string>().from(self).to(delegator);
  expect<std::string>().from(self).to(reverter);
```

libcaf_core/caf/typed_event_based_actor.test.cpp
```diff
@@ -168,12 +168,11 @@ string_actor::behavior_type string_reverter() {
 
 // uses `return delegate(...)`
 string_actor::behavior_type string_delegator(string_actor::pointer self,
-                                             string_actor master, bool leaf) {
-  auto next = leaf ? self->spawn(string_delegator, master, false) : master;
-  self->link_to(next);
+                                             string_actor master) {
```
Neverlord, line 171: The handle is just the next actor in the chain, so let's name it accordingly.

```suggestion
                                             string_actor next) {
```

Neverlord (changes requested):

Neverlord (approved):

PR #1792, opened 2024-03-08
Stabilize unit tests 

### LINKED ISSUE #1777: Test `caf.flow.generation` unstable on CI
I've seen this test fail on a couple of runs recently. We need to figure out whether it's an issue with the test or an actual bug in CAF.

riemass:
Closes #1777.
Stress tested the unit test. It seems like it fixes the issue. Please go over the explanation in the ticket, if you think there exists a better solution?

### DIFF
--- a/libcaf_core/caf/flow/generation.test.cpp
+++ b/libcaf_core/caf/flow/generation.test.cpp
@@ -242,8 +242,8 @@ SCENARIO("asynchronous buffers can generate flow items") {
         // also why we spin up the thread later: making sure we're hitting the
         // code paths we want to test here.
         run_flows();
-        auto bg_thread = std::thread{producer_impl, push};
         sub.dispose();
+        auto bg_thread = std::thread{producer_impl, push};
         run_flows();
         check(res.empty());
         bg_thread.join();


Neverlord:
Introducing random delays isn't a (stable) solution.

Is this the part that is failing?

```
    WHEN("canceling the subscription to the buffer") {
      THEN("the producer receives a cancel event") {
        auto [pull, push] = async::make_spsc_buffer_resource<int>();
        auto res = ivec{};
        auto sub = make_observable() //
                     .from_resource(pull)
                     .take(777)
                     .for_each([&res](int val) { res.push_back(val); });
        // Run initial actions to handle events from the initial request()
        // calls. Without this step, from_resource is in `running_` state and
        // we won't hit the code paths for disposing a "cold" object. This is
        // also why we spin up the thread later: making sure we're hitting the
        // code paths we want to test here.
        run_flows();
        auto bg_thread = std::thread{producer_impl, push};
        sub.dispose();
        run_flows();
        check(res.empty());
        bg_thread.join();
        check(cancelled);
      }
    }
```

If I read your description right, the issue is that the producer can sometimes be fast enough to add some data before the scheduled `do_dispose` lands? In that case, just switching these two lines should give us a deterministic order, no?

```
        sub.dispose();
        auto bg_thread = std::thread{producer_impl, push};
```

riemass:
> In that case, just switching these two lines should give us a deterministic order, no?

You're right. I didn't consider reordering it because I taught it would break the test invariant.

Neverlord (approved):

PR #1794, opened 2024-03-09
Wait for LP server to start in robot test

riemass:
I've run the `robot-lp-communication` test a dozen times locally and it seems that in some runs both clients fail to connect to the server. The cause is that we didn't wait for the server to start up. 
There is another, **very rare** test failure where both clients connect (`alice` and `bob`), but the messages from the second client never arrive to the first one. I couldn't discover the cause for this, so I hope the first issue was what triggered our pipeline to crash.


Neverlord (approved):
Thanks for tracking this down! 👍

PR #1802, opened 2024-03-13
Implement new mail cache type for manual stashing

Neverlord:
Adds a new `mail_cache` type and uses it in the dining philosophers example. By giving users an explicit way for skipping messages, we can go ahead and deprecate the old `skip` approach (as a followup).

Relates #1474.

### DIFF
--- a/libcaf_core/caf/mail_cache.hpp
+++ b/libcaf_core/caf/mail_cache.hpp
@@ -1,0 +1,64 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/detail/core_export.hpp"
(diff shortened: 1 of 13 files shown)


libcaf_core/caf/mail_cache.hpp
```diff
@@ -0,0 +1,64 @@
...
+  void stash(message msg);
+
+  /// Removes all elements from the cache and returns them to the mailbox.
+  void unstash();
+
+private:
+  local_actor* self_;
+  size_t max_size_ = 0;
+  size_t size_ = 0;
+  intrusive::stack<mailbox_element> elements_;
```
riemass, line 61: Wouldn't it make more sense to cache using a queue here?

Neverlord: Why? We only `push` and `pop` on it. The cache is never traversed otherwise.

riemass: I wonder where my thoughts where writing this :smile:.

riemass (approved):

Issue #1776, opened 2024-02-27
Exception when nesting same blocks inside BDD outlines


riemass:
The custom testing library supports multiple GIVEN, WHEN and THEN blocks when using [test scenarios](https://github.com/actor-framework/actor-framework/tree/master/libcaf_test#scenarios). However, there seems to be a bug when making multiple blocks of the same type in [scenario outlines](https://github.com/actor-framework/actor-framework/tree/master/libcaf_test#scenario-outlines). The code will compile without errors, but running the test executable will result with the following exceptions.
For multiple WHEN blocks we get:
```
Suite: caf.scheduler
  unhandled exception: abort test run
    loc: unknown
    msg: failed to select  the root block for the scenario
```
For multiple THEN blocks we get:
```
  unhandled exception: abort test run
    loc: /home/samir/projects/actor-framework/libcaf_core/caf/scheduler.test.cpp:65
    msg: too many THEN blocks in a WHEN block
```

PR #1803, opened 2024-03-14
Fix parsing of <arg> syntax in scenarios

### LINKED ISSUE #1799: Regular sections recognize <arg> syntax
```c++
SECTION("HTML uses <tag> syntax") {
  // crashes, because CAF tries to parse `<tag>` as-if using `OUTLINE`
}
```

This boils down to the parsing in https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/block.cpp#L98. This is currently unconditional. Instead, we need to introduce a new `block_type` for `outline` and only do the parsing for this type.

riemass:
Closes #1799.
Decided to do this first because having an dedicated outline block might be useful for #1776.


libcaf_test/caf/test/block.cpp
```diff
@@ -102,49 +102,55 @@ void block::lazy_init() {
...
-            state = cpy_state::verbatim;
-            break;
-          default:
-            parameter_name += c;
-            break;
-        }
-        break;
+  // Process <arg> syntax if the root block is an outline.
+  if (!ctx_->call_stack.empty()
+      && ctx_->call_stack.front()->type() == block_type::outline) {
```
Neverlord, line 107: Prefer guard style:

```suggestion
  if (ctx_->call_stack.empty()
      || ctx_->call_stack.front()->type() != block_type::outline) {
    description_ = raw_description_;
    return;
  }
```

libcaf_test/caf/test/runnable.cpp
```diff
@@ -29,6 +30,14 @@ void runnable::run() {
   auto guard
     = detail::scope_guard{[]() noexcept { current_runnable = nullptr; }};
   switch (root_type_) {
+    case block_type::outline:
+      if (auto guard = ctx_->get<outline>(0, description_, loc_)->commit()) {
+        do_run();
+        return;
+      }
+      CAF_RAISE_ERROR(std::logic_error,
+                      "failed to select  the root block for the scenario");
```
Neverlord, line 39:
```suggestion
                      "failed to select  the root block for the outline");
```

libcaf_test/caf/test/scenario.test.cpp
```diff
@@ -25,6 +25,23 @@ SCENARIO("a scenario may not contain a section") {
...
+  GIVEN("given a <description>") {
+    WHEN("when <parsing> the raw description") {
+      THEN("then we copy the <description> verbatim") {
+        auto i = ctx_->call_stack.begin();
+        check_eq((*i++)->description(),
+                 "scenario with <foo> syntax in description");
+        check_eq((*i++)->description(), "given a <description>");
+        check_eq((*i++)->description(), "when <parsing> the raw description");
+        check_eq((*i++)->description(),
+                 "then we copy the <description> verbatim");
```
Neverlord, line 40: Unsafe, and IMHO a bit hard to follow due to the iterator indirections. I'd prefer something like this:

```suggestion
        auto cs = ctx_->call_stack;
        require_eq(cs.size(), 4u);
        check_eq(cs[0]->description(),
                 "scenario with <foo> syntax in description");
        check_eq(cs[1]->description(), "given a <description>");
        check_eq(cs[2]->description(), "when <parsing> the raw description");
        check_eq(cs[3]->description(),
                 "then we copy the <description> verbatim");
```

Neverlord (changes requested):

libcaf_test/caf/test/outline.hpp
```diff
@@ -11,7 +11,7 @@
 
 namespace caf::test {
 
-class CAF_TEST_EXPORT outline : public runnable {
+class CAF_TEST_EXPORT outline_setter : public runnable {
```
Neverlord, line 14: This is not a setter. This is the object that will actually run the tests (examples).

I'd prefer a name like `runnable_with_examples` to have a better description of what it actually is, plus please move this to its own file.

riemass: Nice suggestion, will do.

Neverlord (approved):

PR #1804, opened 2024-03-14
Streamline the scheduler interface

Neverlord:
- Move the printer actor to the actor system.
- Move the clock to the actor system.
- Add new API for configuring custom printer and clock.

### DIFF
--- a/libcaf_test/caf/test/fixture/deterministic.cpp
+++ b/libcaf_test/caf/test/fixture/deterministic.cpp
@@ -10,6 +10,7 @@
 #include "caf/actor_system.hpp"
 #include "caf/actor_system_config.hpp"
 #include "caf/detail/actor_local_printer.hpp"
+#include "caf/detail/actor_system_access.hpp"
 #include "caf/detail/mailbox_factory.hpp"
 #include "caf/detail/print.hpp"
 #include "caf/detail/source_location.hpp"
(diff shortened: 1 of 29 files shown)


libcaf_test/caf/test/fixture/deterministic.cpp
```diff
@@ -514,6 +502,32 @@ detail::mailbox_factory* deterministic::config::mailbox_factory() {
...
+  auto& ptr = printers_[self->id()];
+  if (!ptr)
+    ptr = make_counted<actor_local_printer_impl>(self);
+  return ptr;
+}
+
+void deterministic::system_impl::custom_setup(actor_system& sys,
+                                              actor_system_config& cfg,
+                                              void* custom_setup_data) {
+  auto* fix = static_cast<deterministic*>(custom_setup_data);
```
riemass, line 524: Bit curious why you needed to cast `void*` here? My guess is you didn't want to expose deterministic to `caf_core`?

riemass (approved):
:clap: :clap:  Since I've tried refactoring the same thing I know how tightly coupled some components where. Approved.

libcaf_test/caf/test/fixture/deterministic.cpp
Neverlord: It's just the most generic interface that allows us to hook basically any initialization code into the system. Right now, we just have the deterministic fixture here, but I'm certain we'll have other use cases at some point.

PR #1805, opened 2024-03-14
Scheduler unification step 1

Neverlord:
We currently have multiple scheduler interfaces:
- `caf::scheduler`
- `caf::execution_unit`
- `caf::async::execution_context`
- `caf::flow::coordinator` (derives from `caf::async::execution_context`)

The `scheduler` cares about `resumable`, while the latter two operate on `action`. However, there's no real reason why the latter two shouldn't simply extend our `scheduler` interface. The `execution_context` is a weird one, it basically introduces a coupling between the `proxy_registry` and scheduling, which we should get rid of.

This PR is step 1 in consolidating the various interfaces. It refactors `action::impl` to derive from `resumable` and breaks the coupling of the `proxy_registry` to the scheduling. These two refactoring steps will then allow us to get rid of the execution unit entirely and to have `async::execution_context` inherit from `scheduler`.

### DIFF
--- a/libcaf_core/caf/action.cpp
+++ b/libcaf_core/caf/action.cpp
@@ -8,6 +8,18 @@
 
 namespace caf {
 
+resumable::subtype_t action::impl::subtype() const noexcept {
+  return resumable::function_object;
+}
+
+void action::impl::ref_resumable() const noexcept {
+  ref_disposable();
+}
+
+void action::impl::deref_resumable() const noexcept {
+  deref_disposable();
+}
+
 action::action(impl_ptr ptr) noexcept : pimpl_(std::move(ptr)) {
   // nop
 }
--- a/libcaf_core/caf/action.hpp
+++ b/libcaf_core/caf/action.hpp
@@ -10,6 +10,7 @@
 #include "caf/detail/core_export.hpp"
 #include "caf/disposable.hpp"
 #include "caf/make_counted.hpp"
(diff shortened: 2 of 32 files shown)


Neverlord:
> These two refactoring steps will then allow us to get rid of the execution unit entirely and to have `async::execution_context` inherit from `scheduler`.

On second thought, I'm not quite sure whether we actually want to have the execution context inherit from `scheduler`. That would mean that an actor is a `scheduler`, which would technically allow you to run an actor on top of another actor. Maybe that would open up interesting design space, but I'm not sure about it quite yet. For now, let's just get rid of `execution_unit` and enable us to throw an `action` to the system scheduler. The latter will allow us to improve our `future` implementation such that we can do `.then` on a future without binding it to an actor first.

riemass:
> On second thought, I'm not quite sure whether we actually want to have the execution context inherit from `scheduler`. 

The `net::multiplexer` also comes to mind when you say `async::execution_context`. It's running actions in it's own thread with the `pollset_updater`. Is there a way we can go around having two schedulers instantiated, but run all actions on the core scheduler?

> For now, let's just get rid of `execution_unit` and enable us to throw an `action` to the system scheduler. The latter will allow us to improve our `future` implementation such that we can do `.then` on a future without binding it to an actor first.

All in for running a `future` without binding it first. The first set of changes is easy to follow, ofc. approving.

riemass (approved):

PR #1806, opened 2024-03-16
Implement new `set_idle_handler`

### LINKED ISSUE #1789: Implement new `set_idle_handler`
This came out of https://github.com/actor-framework/actor-framework/discussions/1590.

We currently have `after` and pass it to a behavior like this:

```c++
auto bhvr = behavior{
  [](int32_t value) {
    // ...
  },
  after(1s) >> [] {
    // ...
  }
};
```

There are several issues with this approach:
- It stops `behavior` from being composable via `or_else`. We have `message_handler` for the sole reason to have a `behavior` version without timeout. If we remove the timeout from `behavior`, we can get rid of `message_handler` entirely.
- It mixes separate concerns. The `after` makes sense for `receive` to pass a timeout to an otherwise blocking function. However, it's a bad fit for event-based actors.
- There's no control over the kind of reference that the system will hold onto for the actor while it has an idle handler defined (weak ref vs. strong ref).
- The idle handler will repeat indefinitely. A handler that should trigger only once requires users to call `become` with the same handlers again minus the `after` part.

To move forward, we should:
- Keep `after` for blocking APIs that use `receive`.
- Deprecate using `after` for constructing a `behavior`.
- Implement a new special handler:
  ```c++
    template <class RefType, class RepeatType>
    void set_idle_handler(timespan rel_timeout, RefType, RepeatType, Callback cb);
  ```

Wheres `RefType` is either `strong_ref` or `weak_ref` and `RepeatType` is either `repeat` or `once`. There can only be one idle handler at any point in time. To cancel an idle handler, users can call `set_idle_handler(nullptr)`.

Neverlord:
Closes #1789.


libcaf_core/caf/scheduled_actor.hpp
```diff
@@ -354,6 +361,31 @@ class CAF_CORE_EXPORT scheduled_actor : public abstract_scheduled_actor,
   }
 #endif // CAF_ENABLE_EXCEPTIONS
 
+  /// Sets a custom handler for timeouts that trigger after *not* receiving
+  /// a message for a certain amount of time.
+  template <class RefType, class RepeatType>
+  void set_idle_handler(timespan delay, RefType, RepeatType, idle_handler fun) {
+    if (delay == infinite)
+      return;
```
dominiklohmann, line 369: Technically speaking, with a strong reference, shouldn't the behavior be that the actor never dies?

I don't think it's practical to do that, but maybe this should be an assertion instead when the reference is strong and the timeout infinite.

riemass (approved):
LGTM. The new interface is easier to understand and use.

libcaf_core/caf/scheduled_actor.hpp
Neverlord: You're right, we shouldn't silently discard it. It's an error now. 👍

Issue #1800, opened 2024-03-12
Deprecate legacy APIs for 1.0


Neverlord:
Version 1.0 introduces new APIs that render older APIs obsolete. Namely:

- [x] `monitor(hdl)` -> `monitor(hdl, callback)`
- [x] `send(hdl, ...)` -> `mail(...).send(hdl)`
- [x] `request(hdl, timeout, ...)` -> `mail(...).request(hdl, timeout)`
- [x] `set_default_handler(...)` -> use `[](message msg) { ... }` in the `behavior` instead
- [x] `after(...) >> callback` -> use the new `set_idle_handler(...)` instead
- [x] `aout(...)` -> use `self->println(...)` instead
- [x] `function_view` -> use a `scoped_actor` with `mail(...).request(...).receive()`

This ticket depends on:

- #1789 (for `set_idle_handler`)
- #1745 (for the new `mail` DSL)
- #1474 (`stash` is required to replace `set_default_handler(skip)`

PR #1807, opened 2024-03-16
Add new println abstraction on the actor system

Neverlord:
The `aout` API relies on a printer actor that permanently runs in the background and adds additional overhead to printing due to the messaging involved. Binding the new printing API directly onto the actor system removes the messaging overhead and is also simpler to use.

Relates #1800 (for deprecating `aout`).

### DIFF
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -23,6 +23,82 @@
 #include <unordered_map>
 #include <unordered_set>
 
+// -- printer actor ------------------------------------------------------------
+
+namespace caf::detail {
+
+struct printer_actor_state {
+  struct actor_data {
+    std::string current_line;
+    actor_data() {
+      // nop
+    }
+  };
+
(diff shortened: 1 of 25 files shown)


libcaf_core/caf/actor_system.cpp
```diff
@@ -23,6 +23,82 @@
...
+        if (data_ptr != nullptr) {
+          flush(data_ptr, true);
+          data.erase(aid);
+        }
+      },
+      [](redirect_atom, const std::string&, int) {
+        // nop
+      },
+      [](redirect_atom, actor_id, const std::string&, int) {
+        // nop
```
riemass, line 92: Should the nop here stay, or is this a followup?

riemass (approved):
Approving. :+1:

libcaf_core/caf/actor_system.cpp
Neverlord: That handler is in there for backwards compatibility only. We will deprecate `aout`, so this entire printer actor thingy will be removed eventually.

PR #1808, opened 2024-03-16
Remove execution_unit from CAF

Neverlord:
The `execution_unit` class previously kept a pointer to the proxy registry. After removing that with the previous refactoring round, the `execution_unit` we take it one step further here by dropping this interface altogether, consistently using the `scheduler` interface instead.


riemass (approved):
LGTM :+1:

PR #1811, opened 2024-03-17
Fix two bugs with the new monitoring API

Neverlord:
- The new `monitor` overload with a callback would hold onto a strong reference to the monitored actor. By adding this permanent reference count, calling `monitor` once on an actor would prevent it from becoming unreachable. Furthermore, generating a `down_msg` is no longer necessary, since the callback should know which actor it belongs to. Hence, we can simplify the interface and only have an `error` argument.
- The new `monitor` overload would also store `this` as a raw pointer and later use that pointer to send a message back to `self`. This is fundamentally unsafe, because the raw pointer will be used by another actor at some point in the future. If `self` has been destroyed in the meantime, this would cause a heap-use-after-free.

### DIFF
--- a/libcaf_core/caf/detail/monitor_action.hpp
+++ b/libcaf_core/caf/detail/monitor_action.hpp
@@ -5,20 +5,20 @@
 #pragma once
 
 #include "caf/action.hpp"
-#include "caf/system_messages.hpp"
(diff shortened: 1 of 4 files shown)


riemass (approved):
Ok, noted and approved.

PR #1817, opened 2024-03-18
Parse username and password in URI

### LINKED ISSUE #1814: `uri::can_parse` rejects username and password
The function `uri::can_parse` rejects valid URLs that contain a username and a password before the hostname:
```
http://access:secret@node:80/file
```
Our implementation currently uses `@` as the boundary between userinfo and host, while an `:` character is used to break host and port. I've confirmed that the `:` causes the bug. Here is the relavant [source location](https://github.com/actor-framework/actor-framework/blob/bdd25138501aa0da56f269b32902d76f2be7fcd4/libcaf_core/caf/detail/parser/read_uri.hpp#L162).

riemass:
Closes #1814.
I've added an extra state to parse the content after ':', since it can be both, port or password. The content before ':' is stored temporarly to a memory variable. With the approach used I kept as much logic as possible in SM transitions. However, since port is parsed as a decimal number, and password as an string, I couldn't thik of a way to parse it in parallel. This is especially tricky because the deicmal parser keeps track of integer overflow. This ultimately caused a unit test to fail. The reason is that I don't know of a way to break out of the SM on EOF, like the transitions can. 
As a followup, I should probably introduce a new type for userinfo that stores the username/password separately so we can encode it properly afterward. 
@Neverlord If you could confirm the approach first and help me get the last test passing :smiley:.
BTW I found a stale state. Removed it.


Neverlord (commented):
- `std::stoi` is a no-go, because it can throw. If you need a conversion function: https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/detail/parse.hpp#L202
- Why do we need another string? We already have `str` to cache characters. I think the only thing we would actually need in addition is the position of `:` in `str`, am I wrong?

libcaf_core/caf/detail/parser/read_uri.hpp
```diff
@@ -99,13 +99,13 @@ template <class State, class Consumer>
...
   std::string str;
+  std::string memory;
   uint16_t port = 0;
   // Replaces `str` with a default constructed string to make sure we're never
   // operating on a moved-from string object.
   auto take_str = [&] {
-    using std::swap;
     std::string res;
-    swap(str, res);
+    std::swap(str, res);
```
Neverlord, line 108: Why did you change that? That's the canonical way to call `swap`. Since we are swapping strings, I'd also accept `str.swap(res)`.

libcaf_core/caf/uri.test.cpp
```diff
@@ -314,6 +314,14 @@ TEST("from string") {
   ROUNDTRIP("http://me@node:80?a=1&b=2");
   ROUNDTRIP("http://me@node:80#42");
   ROUNDTRIP("http://me@node:80?a=1&b=2#42");
+  ROUNDTRIP("http://user:pass@node");
```
Neverlord, line 317: That's not covering corner cases. At the very least, I'd expect:

```c++
  ROUNDTRIP("http://user:@node");
  ROUNDTRIP("http://user:p@node");
  ROUNDTRIP("http://user:pass@node");
```

libcaf_core/caf/detail/parser/read_uri.hpp
riemass: Basically, because I tought it was the same thing with one line less. I did my homework on the C++ reference manual, I will return it.

riemass:
I integrated the feedback. 
The thing with parsing after the ':' sign is that end of input signalizes the second part was indeed a port, and that's where I've had to validate the port number. I've tried setting the `read_host_or_port` as a regular state, and transitioning into the done state conditionally, but with little success because as soon as the SM sees the end it transitions into a fail state. So in the end I've added another term state overload, hope this doesn't break any CS theory.

libcaf_core/caf/uri.cpp
```diff
@@ -250,7 +250,14 @@ std::string to_string(const uri& x) {
...
   if (!x.userinfo.empty()) {
-    uri::encode(str, x.userinfo);
+    std::string_view userinfo = x.userinfo;
+    // TODO: with this approach, username can't contain a ':'.
+    auto pos = userinfo.find(':');
+    uri::encode(str, userinfo.substr(0, pos));
+    if (pos != std::string::npos) {
+      str += ':';
+      uri::encode(str, userinfo.substr(pos + 1));
+    }
```
Neverlord, line 260: Storing user and password in a single string makes it impossible to disambiguate what a `:` means. We need to change the `uri` class for this:

```c++
  /// Host subcomponent of the authority component. Either an IP address or
  /// an hostname as string.
  using host_type = std::variant<std::string, ip_address>;

  /// Userinfo subcomponent of the authority component. If present, contains the user name and optionally a password.
  struct userinfo_type {
    std::string name;
    std::optional<std::string> password;
  };

  /// Bundles the authority component of the URI, i.e., userinfo, host, and
  /// port.
  struct CAF_CORE_EXPORT authority_type {
    std::optional<userinfo_type> userinfo;
    host_type host;
    uint16_t port;
```

It's a bit awkward to have the double `optional`, but otherwise we can't distinguish between `foo://localhost` (no userinfo), `foo://@localhost` (empty username but no password), and `foo://:@localhost` (empty username and empty password).

libcaf_core/caf/uri.test.cpp
```diff
@@ -314,6 +314,19 @@ TEST("from string") {
   ROUNDTRIP("http://me@node:80?a=1&b=2");
   ROUNDTRIP("http://me@node:80#42");
   ROUNDTRIP("http://me@node:80?a=1&b=2#42");
+  ROUNDTRIP("http://user:@node");
```
Neverlord, line 317: These are also valid URIs (in case there's no equivalent yet):

```
  ROUNDTRIP("http://@node");
  ROUNDTRIP("http://:@node");
```

Neverlord (changes requested):

libcaf_core/caf/detail/parser/read_uri.hpp
```diff
@@ -116,6 +117,19 @@ void read_uri(State& ps, Consumer&& consumer) {
   auto set_path = [&] { consumer.path(take_str()); };
   auto set_host = [&] { consumer.host(take_str()); };
   auto set_userinfo = [&] { consumer.userinfo(take_str()); };
+  auto set_colon_position = [&] {
+    colon_position = str.size();
+    str.push_back(':');
+  };
+  auto set_host_and_port = [&]() -> bool {
+    auto str = take_str();
+    if (parse(std::string_view{str}.substr(colon_position + 1), port))
```
Neverlord, line 126: Instead of just picking the overflow error, it's probably better to construct a string parser state and call the other overload. Then forward the proper error code in case parsing fails.

libcaf_core/caf/detail/parser/read_uri.hpp
```diff
@@ -116,6 +117,19 @@ void read_uri(State& ps, Consumer&& consumer) {
   auto set_path = [&] { consumer.path(take_str()); };
   auto set_host = [&] { consumer.host(take_str()); };
   auto set_userinfo = [&] { consumer.userinfo(take_str()); };
+  auto set_colon_position = [&] {
+    colon_position = str.size();
+    str.push_back(':');
+  };
+  auto set_host_and_port = [&]() -> bool {
```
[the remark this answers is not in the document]
riemass: I didn't try that version, thanks for the idea. I'll give it a shot.

libcaf_core/caf/uri.cpp
riemass: I assumed we needed to split this, I've just wanted to get your feedback on how much refactoring is permissible here. What you're suggesting is also aligned with the [URI RFC](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1):

> The userinfo subcomponent may consist of a user name and, optionally, scheme-specific information about how to gain authorization to access the resource.

libcaf_core/caf/detail/parser/read_uri.hpp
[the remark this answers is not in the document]
riemass: The implementation of the parser will just `goto fsm_fin;` where it would be set to `pec::success`. I've had to add a line for error checking. Also, if I don't set it to `pec::success` prior calling the exit statement, it would break. I guess this isn't that scary since the default constructed state also holds a `pec::success`.

libcaf_core/caf/uri.test.cpp
riemass: Both of these failed. Fixed the implementaiton.

riemass:
The `robot-stream-remote` failed on two hosts initially. Reruning the pipeline solved it, and I couldn't reproduce the failure locally, so I hope this wont occur again.

libcaf_core/caf/uri_builder.cpp
```diff
@@ -18,7 +18,13 @@ uri_builder& uri_builder::scheme(std::string str) {
 }
 
 uri_builder& uri_builder::userinfo(std::string str) {
-  impl_->authority.userinfo = std::move(str);
+  return userinfo(std::move(str), std::nullopt);
+}
+
+uri_builder& uri_builder::userinfo(std::string str,
+                                   std::optional<std::string> password) {
```
Neverlord, line 25: Semantically, it makes no sense to have an overload with two parameters where the second can be disabled again, essentially giving you the same as the overload taking a single argument. If there's a second argument, it should be valid:

```suggestion
uri_builder& uri_builder::userinfo(std::string str, std::string password) {
```

libcaf_core/caf/detail/parser/fsm.hpp
```diff
@@ -84,8 +84,11 @@
     exit_statement;                                                            \
     goto fsm_after_fin;                                                        \
     s_##name : if (ch == '\0') {                                               \
+      ps.code = pec::success;                                                  \
       exit_statement;                                                          \
-      goto fsm_fin;                                                            \
+      if (ps.code == pec::success)                                             \
+        goto fsm_fin;                                                          \
+      goto fsm_after_fin;                                                      \
```
Neverlord, line 91: Not super happy about forcing a pessimization to all parsers that don't write to the code. What if, instead of writing to the state, we optionally pick up the return type, i.e., did you consider doing something like this?

```c++
{
  using exit_statement_res = decltype(exit_statement);
  if constexpr (std::is_same_v<exit_statement_res, pec>) {
    ps.code = (exit_statement);
    if (ps.code <= pec::trailing_character)
      goto fsm_fin; 
    goto fsm_after_fin;
  } else {
    exit_statement;
    goto fsm_fin; 
  }
}
```

(extra scope is for `exit_statement_res`)

Neverlord (changes requested):

libcaf_core/caf/uri_builder.cpp
riemass: Yeah, I was too fixated on differentiating between an empty password and no password that I missed the point here.

libcaf_core/caf/detail/parser/fsm.hpp
riemass: This is a good bridge between the first solution and the second one. Why didn't come up with this? :+1:

libcaf_core/caf/uri.cpp
```diff
@@ -135,6 +135,11 @@ class nop_builder {
     return *this;
   }
 
+  template <class T, class P>
+  nop_builder& userinfo(T&&, P&&) {
```
Neverlord, line 139: Can have one or two arguments, no?

```suggestion
  template <class... Ts>
  nop_builder& userinfo(Ts&&...) {
```

libcaf_core/caf/uri.cpp
```diff
@@ -249,8 +254,12 @@ std::string to_string(const uri& x) {
 
 std::string to_string(const uri::authority_type& x) {
   std::string str;
-  if (!x.userinfo.empty()) {
-    uri::encode(str, x.userinfo);
+  if (x.userinfo) {
+    uri::encode(str, x.userinfo->name);
+    if (x.userinfo->password) {
+      str += ':';
+      uri::encode(str, *x.userinfo->password);
```
Neverlord, line 261: Can be less repetitive:

```suggestion
    if (auto& pw = x.userinfo->password) {
      str += ':';
      uri::encode(str, *pw);
```

Issue #1816, opened 2024-03-18
Simplify the error class


Neverlord:
Currently, `caf::error` stores a `message`, which allows it to wrap pretty much everything. In practice, however, we only have use cases for two constructors:

- `error(Enum code)`
- `error(Enum code, std::string_view message)`

We should also make the former constructor `explicit` and drop `make_error`. This will drastically simplify the implementation and ultimately make the API easier to use as well.

Neverlord:
Yeah, we actually have a use for embedding a message in CAF itself. Let's just drop this for now.

PR #1820, opened 2024-03-19
Add new format_to_error utility function

Neverlord:
This was originally part of #1816, but I thought it's worth keeping even though we're not going through with the `error` redesign.


riemass (changes requested):
Found extra linke break. Looks good otherwise.

riemass (approved):

Issue #1798, opened 2024-03-12
Support splitting and UTF-8 conversions in flow::byte


Neverlord:
The use case for this is reading raw bytes, e.g., from the network, and then being able to split the input at some character / using some predicate and/or convert the raw bytes to UTF-8 (safely, i.e., doing a validity check).

We currently have this utility:

```c++
  /// Returns a transformation step that converts a sequence of bytes into
  /// a sequence of chunks.
  /// @param chunk_size The maximum number of bytes per chunk.
  static auto to_chunks(size_t chunk_size) noexcept {
    return detail::to_chunks_step{chunk_size};
  }
```

We'll add these ones:

```c++
  /// Returns a transformation step that converts a sequence of bytes into
  /// a sequence of chunks by splitting the input at a given separator.
  /// @param separator The separator to split at.
  static auto split_at(std::byte separator) noexcept {
    // ...
  }

  /// Returns a transformation step that converts a sequence of bytes into
  /// a sequence of (UTF-8) strings by splitting the input at a given separator.
  /// @param separator The separator to split at.
  static auto split_as_utf8_at(char separator) noexcept {
    // ...
  }
```

`split_at`: produces a sequence of `chunk`
`split_as_utf8_at`: produces a sequence of `cow_string` and checks each item for validity before emitting it using [`detail::rfc3629::valid`](https://github.com/actor-framework/actor-framework/blob/master/libcaf_core/caf/detail/rfc3629.hpp#L19).

PR #1821, opened 2024-03-20
Add support for splitting and utf-8 conversions

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/1798

This PR also adds some tests for `to_chunk` in `bytes.hpp`

### DIFF
--- a/libcaf_core/caf/chunk.hpp
+++ b/libcaf_core/caf/chunk.hpp
@@ -166,6 +166,10 @@ public:
     return std::move(data_);
   }
 
+  // -- comparison -------------------------------------------------------------
+
+  bool equal_to(const chunk& other) const noexcept;
+
 private:
   intrusive_ptr<data> data_;
 };
--- a/libcaf_core/caf/flow/byte.hpp
+++ b/libcaf_core/caf/flow/byte.hpp
@@ -3,6 +3,10 @@
 // https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
 
 #include "caf/chunk.hpp"
+#include "caf/cow_string.hpp"
+#include "caf/detail/rfc3629.hpp"
+#include "caf/error.hpp"
+#include "caf/sec.hpp"
 
 #include <cstddef>
 
@@ -54,6 +58,109 @@ private:
   byte_buffer buf_;
 };
 
+/// Turns a sequence of bytes into a sequence of chunks based on separator.
+class split_at_step {
+public:
+  using input_type = std::byte;
+
+  using output_type = chunk;
+
+  explicit split_at_step(std::byte separator) noexcept : separator_(separator) {
+    // nop
+  }
+
+  template <class Next, class... Steps>
+  bool on_next(const std::byte& b, Next& next, Steps&... steps) {
+    if (b == separator_)
+      return do_emit(next, steps...);
+    buf_.push_back(b);
+    return true;
+  }
+
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
+    if (!buf_.empty() && !do_emit(next, steps...))
+      return;
+    next.on_complete(steps...);
+  }
+
+  template <class Next, class... Steps>
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    if (!buf_.empty() && !do_emit(next, steps...))
+      return;
+    next.on_error(what, steps...);
+  }
+
+private:
+  template <class Next, class... Steps>
+  bool do_emit(Next& next, Steps&... steps) {
+    auto item = chunk{buf_};
+    buf_.clear();
+    return next.on_next(item, steps...);
+  }
+
+  std::byte separator_;
+  byte_buffer buf_;
+};
+
+/// Turns a sequence of bytes into a sequence of cow_string based on
+/// separator.
+class split_as_utf8_at_step {
+public:
+  using input_type = std::byte;
+
+  using output_type = cow_string;
+
+  explicit split_as_utf8_at_step(char separator) noexcept
+    : separator_(separator) {
+    // nop
+  }
+
+  template <class Next, class... Steps>
+  bool on_next(const std::byte& b, Next& next, Steps&... steps) {
+    if (static_cast<char>(b) == separator_)
+      return do_emit(next, steps...);
+    buf_.push_back(b);
+    return true;
+  }
+
+  template <class Next, class... Steps>
+  void on_complete(Next& next, Steps&... steps) {
+    if (!buf_.empty() && !do_emit(next, steps...))
+      return;
+    next.on_complete(steps...);
+  }
+
+  template <class Next, class... Steps>
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    if (!buf_.empty() && !do_emit(next, steps...))
+      return;
+    next.on_error(what, steps...);
+  }
+
+private:
+  template <class Next, class... Steps>
+  bool do_emit(Next& next, Steps&... steps) {
+    if (buf_.empty())
+      return next.on_next(cow_string{}, steps...);
+    if (!detail::rfc3629::valid(buf_)) {
+      buf_.clear();
+      next.on_error(make_error(sec::invalid_utf8), steps...);
+      return false;
+    }
+    std::string result;
+    result.reserve(buf_.size());
+    std::transform(buf_.begin(), buf_.end(), std::back_inserter(result),
+                   [](std::byte b) { return static_cast<char>(b); });
+    auto item = cow_string{std::move(result)};
+    buf_.clear();
+    return next.on_next(item, steps...);
+  }
+
+  char separator_;
+  byte_buffer buf_;
+};
+
 } // namespace caf::detail
 
 namespace caf::flow {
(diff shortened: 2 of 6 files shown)


libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
   byte_buffer buf_;
 };
 
+/// Turns a sequence of bytes into a sequence of chunks based on separator.
+class to_split_at {
```
Neverlord, line 60:
```suggestion
class split_at_step {
```

libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
...
+    buf_.clear();
+    return next.on_next(item, steps...);
+  }
+
+  std::byte separator_;
+  byte_buffer buf_;
+};
+
+/// Turns a sequence of bytes into a sequence of cow_string based on separator.
+class to_split_as_utf8_at {
```
Neverlord, line 105:
```suggestion
class split_as_utf8_at_step {
```

libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
...
+  template <class Next, class... Steps>
+  bool do_emit(Next& next, Steps&... steps) {
+    if (!detail::rfc3629::valid(buf_)) {
+      buf_.clear();
+      return false;
+    }
+    std::string result;
+    std::transform(buf_.begin(), buf_.end(), std::back_inserter(result),
+                   [](std::byte b) { return static_cast<char>(b); });
+    auto item = cow_string{result};
```
Neverlord, line 148:
```suggestion
    auto item = cow_string{std::move(result)};
```

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "caf/flow/multicaster.hpp"
+
+using namespace caf;
+
+namespace {
+
+/// Compares two chunks for equality.
+[[maybe_unused]] bool operator==(const caf::chunk& lhs, const caf::chunk& rhs) {
+  return std::equal(lhs.bytes().begin(), lhs.bytes().end(),
+                    rhs.bytes().begin());
```
Neverlord, line 22: Unsafe, should use the four-argument overload. However, please don't patch `chunk` at file scope. That makes it hard to understand what's going on. Better: give `chunk` an `equal_to` member function and use that for testing.

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  std::transform(data.begin(), data.end(), std::back_inserter(byte_arr),
+                 [](auto element) { return static_cast<std::byte>(element); });
+  return byte_arr;
+}
+
+WITH_FIXTURE(test::fixture::flow) {
+
+SCENARIO("to_chunks can create chunks from a byte array source") {
+  GIVEN("a container") {
+    auto input = convert_to_byte_arr(std::string{"Sample string"});
```
Neverlord, line 37:
```suggestion
    auto input = convert_to_byte_arr("Sample string"s);
```

(same for the other occurrences)

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+auto convert_to_byte_arr(const T& data) {
+  std::vector<std::byte> byte_arr;
+  std::transform(data.begin(), data.end(), std::back_inserter(byte_arr),
+                 [](auto element) { return static_cast<std::byte>(element); });
+  return byte_arr;
+}
+
+WITH_FIXTURE(test::fixture::flow) {
+
+SCENARIO("to_chunks can create chunks from a byte array source") {
```
Neverlord, line 35:
```suggestion
SCENARIO("to_chunks splits a sequence of bytes into chunks") {
```

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  GIVEN("a container") {
```
Neverlord, line 36: The container is only for building an `observable`:

```suggestion
  GIVEN("an observable<byte>") {
```

(same for the other `GIVEN`)

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+        std::vector<chunk> output{};
+        make_observable()
+          .from_container(input)
+          .transform(caf::flow::byte::to_chunks(5))
+          .for_each([&output](const chunk& x) { output.push_back(x); });
+        run_flows();
+        check_eq(output.size(), 3u);
+        check(output[0] == chunk{convert_to_byte_arr(std::string{"Sampl"})});
+        check(output[1] == chunk{convert_to_byte_arr(std::string{"e str"})});
+        check(output[2] == chunk{convert_to_byte_arr(std::string{"ing"})});
```
Neverlord, line 49:
```suggestion
        if (check_eq(output.size(), 3u)) {
          check(output[0] == chunk{convert_to_byte_arr("Sampl"sv)});
          check(output[1] == chunk{convert_to_byte_arr("e str"sv})});
          check(output[2] == chunk{convert_to_byte_arr("ing"sv})});
        }
```

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,220 @@
...
+namespace {
+
+/// Compares two chunks for equality.
+[[maybe_unused]] bool operator==(const caf::chunk& lhs, const caf::chunk& rhs) {
+  return std::equal(lhs.bytes().begin(), lhs.bytes().end(),
+                    rhs.bytes().begin());
+}
+
+template <class T>
+auto convert_to_byte_arr(const T& data) {
```
Neverlord, line 26: Does not return an array. Also a bit verbose, something like `to_bytes` would suffice.

Neverlord (changes requested):
I only tagged the first of the scenario blocks, but it applies to all scenarios in the file.

libcaf_core/caf/chunk.hpp
```diff
@@ -166,6 +166,13 @@ class CAF_CORE_EXPORT chunk {
     return std::move(data_);
   }
 
+  // -- comparison -------------------------------------------------------------
+
+  bool equal_to(const chunk& other) const noexcept {
```
Neverlord, line 171: This needs `#include <algorithm>`. Please push the implementation to the `.cpp` file to avoid including that header here.

libcaf_core/caf/chunk.hpp
```diff
@@ -166,6 +166,13 @@ class CAF_CORE_EXPORT chunk {
...
+    return std::equal(this->bytes().begin(), this->bytes().end(),
```
Neverlord, line 172: No need for `this->` in this context.

libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
...
+      return;
+    next.on_error(what, steps...);
+  }
+
+private:
+  template <class Next, class... Steps>
+  bool do_emit(Next& next, Steps&... steps) {
+    if (!detail::rfc3629::valid(buf_)) {
+      buf_.clear();
+      return false;
```
Neverlord, line 143: Before we return `false` here, we must call `on_error` on `next` first. We don't want to just discard invalid lines, this is an error that we need to propagate and then stop.

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+SCENARIO("split_as_utf8_at can discard invalid utf-8 string") {
+  GIVEN("an observable<byte>") {
+    auto input = to_bytes("Sample "s);
+    WHEN("transforming the input with split_as_utf8_at") {
+      THEN("only valid values from container is received") {
+        input.push_back(static_cast<std::byte>(0xc8));
+        auto input_end = to_bytes(" string"s);
+        input.insert(input.end(), input_end.begin(), input_end.end());
```
Neverlord, line 317:
```suggestion
        auto suffix = to_bytes(" string"s);
        input.insert(input.end(), suffix.begin(), suffix.end());
```

libcaf_core/caf/flow/byte.test.cpp
```diff
@@ -0,0 +1,335 @@
...
+        std::vector<cow_string> output{};
+        make_observable()
+          .from_container(input)
+          .transform(caf::flow::byte::split_as_utf8_at(' '))
+          .for_each([&output](const cow_string& x) { output.push_back(x); });
+        run_flows();
+        if (check_eq(output.size(), 2u)) {
+          check_eq(output[0], cow_string{"Sample"s});
+          check_eq(output[1], cow_string{"string"s});
+        }
```
Neverlord, line 327: The operator should emit `on_next("Sample")` followed by `on_error(...)`.

libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
...
+    }
+    std::string result;
```
Neverlord, line 145:
```suggestion
    std::string result;
    result.reserve(buf_.size());
```

libcaf_core/caf/flow/byte.hpp
```diff
@@ -54,6 +56,104 @@ class to_chunks_step {
...
+  void on_error(const error& what, Next& next, Steps&... steps) {
+    if (!buf_.empty() && !do_emit(next, steps...))
+      return;
+    next.on_error(what, steps...);
+  }
+
+private:
+  template <class Next, class... Steps>
+  bool do_emit(Next& next, Steps&... steps) {
+    if (!detail::rfc3629::valid(buf_)) {
```
Neverlord, line 141: Should shortcut empty items:

```c++
if (buf_.empty()) {
    return next.on_next(cow_string{} steps...);
}
```

I don't think I saw a test for that case.

Neverlord (changes requested):

libcaf_core/caf/flow/byte.hpp
shariarriday: I have added a new error `invalid_utf8` that will be passed to the `on_error` of `next` for this purpose.

libcaf_core/caf/flow/byte.hpp
shariarriday: I have added a new test which checks that empty `cow_string`s are passed to `on_next`.

Issue #1254, opened 2021-04-24
Stabilize ABI and set CMake package compatibility to `SameMajorVersion`


Neverlord:
We don't have an official versioning policy yet, but we generally follow the policy that patch-level changes shall be backwards compatible. Hence, applications that expect CAF version 0.18.1 should be able to build against 0.18.4 (but not 0.19.0).

The logical step would be to relax the version check of our CMake package. For CMake, this means just applying this patch:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 11dd1e96..628ccee9 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -450,7 +450,7 @@ install(EXPORT CAFTargets
 write_basic_package_version_file(
   "${CMAKE_CURRENT_BINARY_DIR}/CAFConfigVersion.cmake"
   VERSION ${CAF_VERSION}
-  COMPATIBILITY ExactVersion)
+  COMPATIBILITY SameMinorVersion)
 
 configure_package_config_file(
   "${CMAKE_CURRENT_SOURCE_DIR}/cmake/CAFConfig.cmake.in"
```

*However*, our CI currently doesn't check for ABI compatibility of changes. I'm not sure how CMake packages link targets. My assumption would be that this policy would also allow users to upgrade their local CAF installation without re-building applications. If that's the case then we need to figure out how to integrate ABI compatibility checks into our CI first (and whether we want to invest the time or just stick to the current yet restrictive `ExactVersion` policy).

Neverlord:
There's a tool for checking ABI compatibility: https://github.com/lvc/abi-compliance-checker.

Neverlord:
With 1.0 coming up, I thought about ABI compatibility again (yeah, I know, everyone's favorite topic). I think what we really want to support going forward is `SameMajorVersion`.

For this, I think we will need to introduce `inline namespace` for versioning into the code base. Basically, if we do any of that:

- Touch member variables of a class (add/remove/reorder)
- Touch virtual member functions of class (add/remove/reorder)
- Change (member) function parameters

We'll have to copy the thing into a new version namespace, do the changes there, and finally make the new version the `inline` (and remove `inline` from the old thing).

Here's a trivial example of what I mean:

```cpp
namespace caf {

inline namespace v1 {

struct foo {
  int8_t value1 = 0;
  int32_t value2 = 0;
  int8_t value2 = 0;
};

} // namespace v1

} // namespace caf
```

[...]

PR #1822, opened 2024-03-21
Hide most actor_system_config fields

Neverlord:
Moving the fields for the actor-system config into an opaque member will allow us adding new fields mid-release without breaking ABI.

Relates #1254.

### DIFF
--- a/libcaf_core/caf/fwd.hpp
+++ b/libcaf_core/caf/fwd.hpp
@@ -269,6 +269,7 @@ using int_gauge_family = metric_family_impl<int_gauge>;
 namespace detail {
 
 class actor_system_access;
+class actor_system_config_access;
 class mailbox_factory;
 class monotonic_buffer_resource;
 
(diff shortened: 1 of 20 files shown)


libcaf_core/caf/fwd.hpp
```diff
@@ -269,8 +269,10 @@ using int_gauge_family = metric_family_impl<int_gauge>;
 namespace detail {
 
 class actor_system_access;
+class actor_system_config_access;
 class mailbox_factory;
 class monotonic_buffer_resource;
+class opaque;
```
riemass, line 275: Where is this type defined? Could it be you refactored it?

riemass (approved):
Not really shure what you mean by `class opaque`, I grepped the code and couldn't find anything excpect the forward declaration. Otherwise, I understand the changes and approve the MR.

libcaf_core/caf/fwd.hpp
Neverlord: Yeah, that's a leftover from a previous experiment. Good catch!

PR #1824, opened 2024-03-22
Implementation hiding of `caf::net` protocol specific classes

riemass:
As agreed, the low hanging classes from `caf::net` are now interfaces, with imlementation hidden in the translation units. One commit per class and no functional changes. 

Open question to consider as a followup are:
- classes `octest_stream::transport` and `ssl::transport`, 
- `net::socket_manager` could probably be hidden as well with some effort. 

Relates #1254.

### DIFF
--- a/libcaf_net/caf/net/lp/framing.hpp
+++ b/libcaf_net/caf/net/lp/framing.hpp
@@ -4,24 +4,13 @@
 
 #pragma once
 
-#include "caf/net/lp/default_trait.hpp"
-#include "caf/net/lp/frame.hpp"
 #include "caf/net/lp/lower_layer.hpp"
 #include "caf/net/lp/upper_layer.hpp"
-#include "caf/net/middleman.hpp"
 #include "caf/net/octet_stream/upper_layer.hpp"
 
-#include "caf/actor_system.hpp"
-#include "caf/async/spsc_buffer.hpp"
-#include "caf/byte_span.hpp"
-#include "caf/cow_tuple.hpp"
 #include "caf/detail/net_export.hpp"
-#include "caf/disposable.hpp"
 
-#include <cstdint>
-#include <cstring>
 #include <memory>
-#include <type_traits>
 
 namespace caf::net::lp {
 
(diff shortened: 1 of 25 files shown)


libcaf_net/caf/net/lp/framing.hpp
Neverlord, line 1: Should add `~framing() override;`

libcaf_net/caf/net/web_socket/client.hpp
```diff
@@ -36,48 +22,13 @@ class CAF_NET_EXPORT client : public octet_stream::upper_layer {
 
   using upper_layer_ptr = std::unique_ptr<web_socket::upper_layer>;
```
Neverlord, line 24:
```suggestion

  // -- constructors, destructors, and assignment operators --------------------

  ~client() override;
```

libcaf_net/caf/net/web_socket/framing.hpp
```diff
@@ -30,177 +19,15 @@ class CAF_NET_EXPORT framing : public octet_stream::upper_layer,
 public:
   // -- member types -----------------------------------------------------------
 
-  using binary_buffer = std::vector<std::byte>;
-
   using upper_layer_ptr = std::unique_ptr<web_socket::upper_layer>;
```
Neverlord, line 23:
```suggestion

  // -- constructors, destructors, and assignment operators --------------------

  ~framing() override;
```

libcaf_net/caf/net/web_socket/server.hpp
```diff
@@ -42,42 +22,9 @@ class CAF_NET_EXPORT server : public octet_stream::upper_layer {
 
   using upper_layer_ptr = std::unique_ptr<web_socket::upper_layer::server>;
```
Neverlord, line 24:
```suggestion

 // -- constructors, destructors, and assignment operators --------------------
 
 ~server() override;
 
```

Neverlord (changes requested):
Just some nitpicks. Thanks for taking care of this!

libcaf_net/caf/net/lp/framing.hpp
riemass: Why? Framing only has a `make` function, otherwise empty. Or do you mean `~framing_impl`?

Neverlord: "otherwise empty" is precisely the reason why. `framing` inherits from two other classes and has virtual functions. So it needs a vtable. Since we don't implement any virtual function, the compiler has no place to put it (weak vtable). By overriding the destructor, we have at least one virtual member function.

PR #1826, opened 2024-03-23
Re-implement actor_system using the PIMPL idiom

Neverlord:
Relates #1254.


riemass (approved):
Looks good. Didn't go too much into the `impl` class, guess the functionality is the same as the old system.

PR #1830, opened 2024-03-25
Simplify the socket_manager virtual interface

riemass:
After giving the `socket_manager` another look, one of the `handle` functions wasn't used. We can also inline some functions to make it simpler.

### DIFF
--- a/libcaf_net/caf/net/socket_manager.hpp
+++ b/libcaf_net/caf/net/socket_manager.hpp
@@ -45,22 +45,13 @@ public:
   virtual socket handle() const = 0;
 
   /// Returns the owning @ref multiplexer instance.
-  multiplexer& mpx() {
-    CAF_ASSERT(mpx_ptr());
-    return *mpx_ptr();
-  }
-
-  /// Returns the owning @ref multiplexer instance.
-  const multiplexer& mpx() const {
+  multiplexer& mpx() const {
     CAF_ASSERT(mpx_ptr());
     return *mpx_ptr();
   }
 
   /// Returns a pointer to the owning @ref multiplexer instance.
-  virtual multiplexer* mpx_ptr() noexcept = 0;
-
-  /// Returns a pointer to the owning @ref multiplexer instance.
-  virtual const multiplexer* mpx_ptr() const noexcept = 0;
+  virtual multiplexer* mpx_ptr() const noexcept = 0;
 
   /// Queries whether the manager is registered for reading.
   virtual bool is_reading() const noexcept = 0;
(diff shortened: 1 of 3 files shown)


libcaf_net/caf/net/socket_manager.hpp
```diff
@@ -44,17 +44,17 @@ class CAF_NET_EXPORT socket_manager : public detail::atomic_ref_counted,
...
 
-  /// @private
-  virtual void handle(socket new_handle) = 0;
-
-  /// Returns a reference to the hosting @ref actor_system instance.
-  virtual actor_system& system() noexcept = 0;
-
   /// Returns the owning @ref multiplexer instance.
-  virtual multiplexer& mpx() noexcept = 0;
+  virtual multiplexer& mpx() {
```
Neverlord, line 48: Make up your mind. Either it's virtual and belongs to the `.cpp` file or it is inline and can be implemented here.

riemass: Sorry, I meant to remove `virtual`.

libcaf_net/caf/net/socket_manager.hpp
```diff
@@ -44,17 +44,17 @@ class CAF_NET_EXPORT socket_manager : public detail::atomic_ref_counted,
...
   /// Returns the owning @ref multiplexer instance.
-  virtual multiplexer& mpx() noexcept = 0;
+  multiplexer& mpx() {
+    CAF_ASSERT(mpx_ptr());
+    return *mpx_ptr();
+  }
 
   /// Returns the owning @ref multiplexer instance.
-  virtual const multiplexer& mpx() const noexcept = 0;
+  const multiplexer& mpx() const {
```
Neverlord, line 54: Actually, why not simplify this further? There's no reason for us to propagate `const` to the multiplexer in the first place:

```c++
  multiplexer& mpx() const;
  virtual multiplexer* mpx_ptr() const noexcept = 0;
```

riemass: Wouldn't this break encapsulation, you could do a mutable operation on the multiplexer from a const declared context?

Neverlord: The `multiplexer` is not part of the socket manager. The socket manager merely points to it. If you modify the `multiplexer`, it has no effect on the socket manager. No encapsulation broken, because the manager does not encapsulate a multiplexer. 🙂

riemass: That's a valid viewpoint. I'll rebase and push.

Neverlord (approved):

PR #1836, opened 2024-04-02
Update typos config

riemass:
This updates the extend list and solves issues that where detected with typos version 1.20.0.

### DIFF
--- a/.typos-config.toml
+++ b/.typos-config.toml
@@ -2,5 +2,8 @@
 extend-exclude = ["*.pem", "libcaf_net/test/pem.cpp", "Jenkinsfile"]
 
 [default.extend-words]
+PN = "PN"
+Yout = "Yout"
+aout = "aout"
 caf = "caf"
-CAF = "CAF"
+ws = "ws"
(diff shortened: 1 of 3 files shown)


.typos-config.toml
```diff
@@ -3,4 +3,7 @@ extend-exclude = ["*.pem", "libcaf_net/test/pem.cpp", "Jenkinsfile"]
 
 [default.extend-words]
 caf = "caf"
-CAF = "CAF"
+aout = "aout"
+ws = "ws"
+yout = "yout"
```
Neverlord, line 8: Typo? I couldn't find a match for `yout` (only for `layout`).

riemass: That's for the template parameter Yout, e.g..

```
error: `Yout` should be `Your`
  --> ./libcaf_core/caf/interface_mismatch.hpp:77:26
   |
77 |         type_list<Zs..., Yout(Yin...)>> {};
   |                          ^^^^
```

Neverlord: Ah, I didn't know it's matching case-insensitive.

riemass:
I've capitalized `Yout` and sorted the `extend-words`. Squashed because it's a small PR anyway, there's no need for two commits.

PR #1837, opened 2024-04-03
Remove logger factory from the config class

Neverlord:
The logger is a central component of the actor system. Allowing users to override this component should be carefully considered. For now, we should rather play it safe and not introduce a major feature by accident that will make it hard to change central interfaces such as the logger or the scheduler later on.


riemass (approved):
I respect the decision, we can always come back. :+1:

PR #1838, opened 2024-04-05
Add version header and ABI version token

Neverlord:
Relates #1254.

- shift the version information to CMake
- auto-generate CAF's version macros from CMake
- add a new version.hpp header to get versions from CAF at run-time
- add an ABI token to the `actor_system` constructor

The latter will force a call to `caf::version::make_abi_token()` for all instances where a user constructs an actor system. Since probably every CAF application (at least 99.999%) will construct an actor system at some point, this injects a call to `make_abi_token` into client code. The function lives in an inline namespace that will change with each major release. Hence, an application will fail to load a CAF library with the wrong ABI version. Additionally, the actor system also checks the value of the ABI token at run-time as an additional safeguard.

The new set of version functions in the namespace `caf::version` allows users to detect at run-time what version of CAF they have loaded. Users can also obtain the CAF version as a string now.

@riemass I thought about the inline namespaces a bit more. Introducing something like `caf::v1` is quite intrusive and would force us to use macros to open up the namespaces. So it would spread through the entire code base. This approach is much less intrusive. We only use an inline namespace in one place and basically make sure that CAF application will have to call this ABI-versioned function (by sneaking the call into client code when constructing an `actor_system`). That way, I think we still get the ABI "version check" to make sure applications can't accidentally load an incompatible CAF version. The only TODO item for this would be to add the ABI token calls when loading modules. Let me know what you think about this approach.

### DIFF
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -590,14 +590,18 @@ actor_system::networking_module::~networking_module() {
 
 namespace {} // namespace
 
-actor_system::actor_system(actor_system_config& cfg)
-  : actor_system(cfg, nullptr, nullptr) {
+actor_system::actor_system(actor_system_config& cfg, version::abi_token token)
+  : actor_system(cfg, nullptr, nullptr, token) {
   // nop
 }
 
 actor_system::actor_system(actor_system_config& cfg,
                            custom_setup_fn custom_setup,
-                           void* custom_setup_data) {
+                           void* custom_setup_data, version::abi_token token) {
+  // Make sure the ABI token matches the expected version.
+  if (static_cast<int>(token) != CAF_VERSION_MAJOR) {
+    CAF_CRITICAL("CAF ABI token mismatch");
+  }
   // This is a convoluted way to construct the implementation, but we cannot
   // just use new and delete here. The `custom_setup` function might call member
   // functions on this actor system when constructing `impl`, which requires us
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -588,8 +588,6 @@ actor_system::networking_module::~networking_module() {
   // nop
 }
 
-namespace {} // namespace
-
 actor_system::actor_system(actor_system_config& cfg, version::abi_token token)
   : actor_system(cfg, nullptr, nullptr, token) {
   // nop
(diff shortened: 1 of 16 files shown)


riemass:
What I liked about inline namespaces is that, if you need to break ABI, you can still introduce a `caf::experimental` namespace containing the new version, and do the transition on the next version bump.
Otherwise, this approach is much cleaner, which saves us a lot of hassle with versioning. Since we already discussed that we must follow the rules to avoid inline namespace explosion, we can implement this way and see how it goes.

libcaf_io/caf/io/middleman.hpp
```diff
@@ -248,9 +249,13 @@ class CAF_IO_EXPORT middleman : public actor_system::networking_module {
   /// Adds module-specific options to the config before loading the module.
   static void add_module_options(actor_system_config& cfg);
 
-  /// Returns a middleman using the default network backend.
+  /// Creates a new middleman instance.
   static actor_system_module* make(actor_system&);
 
+  /// Checks whether ghe ABI of the middleman is compatible with the CAF core.
```
riemass, line 255:
```suggestion
  /// Checks whether the ABI of the middleman is compatible with the CAF core.
```

libcaf_net/caf/net/middleman.hpp
```diff
@@ -58,11 +59,16 @@ class CAF_NET_EXPORT middleman : public actor_system_module {
...
 
-  static actor_system_module* make(actor_system& sys);
-
   /// Adds module-specific options to the config before loading the module.
   static void add_module_options(actor_system_config& cfg);
 
+  /// Creates a new middleman instance.
+  static actor_system_module* make(actor_system& sys);
+
+  /// Checks whether ghe ABI of the middleman is compatible with the CAF core.
```
riemass, line 68:
```suggestion
  /// Checks whether the ABI of the middleman is compatible with the CAF core.
```

libcaf_openssl/caf/openssl/manager.hpp
```diff
@@ -59,6 +60,10 @@ class CAF_OPENSSL_EXPORT manager : public actor_system_module {
   ///         default network backend.
   static actor_system_module* make(actor_system&);
 
+  /// Checks whether ghe ABI of the middleman is compatible with the CAF core.
```
riemass, line 63:
```suggestion
  /// Checks whether the ABI of the middleman is compatible with the CAF core.
```

libcaf_core/caf/actor_system.cpp
```diff
@@ -590,14 +590,18 @@ actor_system::networking_module::~networking_module() {
 
 namespace {} // namespace
```
riemass, line 591: Wait, what?

riemass (changes requested):
Found some typos. Otherwise looks good.

riemass (approved):

Issue #1832, opened 2024-03-27
Documentation update for 1.0


Neverlord:
Time to wrap things up for CAF 1.0.

The last remaining feature ticket we still have is #1752. We currently lack the server part for that ticket and maybe we make a pass to have a similar API for WebSocket. But that's basically it on new features. Some refactoring work remains (for example introducing the new ABI version with an inline namespace) and we need to bring our documentation up to date:

- [ ] add new manual section on the new mail API
- [ ] overhaul all manual sections on `spawn` and only use function-based spawn (and `actor_from_state`)
- [ ] add all of the new flow operators to the manual and mention them in the changelog
- [ ] update CAFcademy article on implementing actors (both parts)
- [ ] make a pass over all CAFcademy articles to exclusively use the new mail API
- [ ] add new new manual section on idle timeouts
- [ ] mention deprecation of idle timeouts via `after` in changelog
- [ ] mention the new best practice on message handlers (auto-move into values, cf. #1740)
- [ ] make a pass over the manual and all articles regarding the new `monitor` API

PR #1844, opened 2024-04-17
Deprecate the class-based spawn and actor pools

Neverlord:
The class-based spawn overload is deprecated and users should instead use the function-based and state-based spawn overloads.

The actor pool has a very cumbersome interface that is hard to use and comes with several limitations. Eventually, we want to replace this utility with shared mailboxes, rendering the pool-based approach obsolete.

Relates #1832 and relates #1800.


riemass (approved):
:+1:

PR #1848, opened 2024-05-03
Stabilize caf-net ABI by de-templatizing factories

Neverlord:
This turns factories for the `with`-DSL into regular classes to hide implementation details, in particular our "flow bridge" implementations. In places where we need templates, for example to have `web_socket::acceptor` pass extra values from the handshake to the accept event, I've put in new abstractions that only encapsulate this extra type information and provide regular interface classes that we can then pass to our non-template implementations.

The primary motivation for all of this is to allow us to change details on the implementations or to add new config parameters without breaking the ABI.

The PR mostly moves code around, so it's not as large as it seems.


libcaf_net/caf/net/web_socket/client_factory.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+dsl::client_config_value& client_factory::init_config(multiplexer* mpx) {
+  config_ = new config_impl(mpx);
+  return *config_;
+}
+
+expected<void> client_factory::sanity_check() {
+  if (config_->hs.has_mandatory_fields()) {
+    return {};
+  } else {
```
riemass, line 71: Unneeded else after a return statement.

riemass (changes requested):
Found some nits, looking good otherwise.

riemass (approved):

PR #1852, opened 2024-05-07
Better encapsulate purely internal code in caf-net

Neverlord:
This is wrapping up the remaining loose ends from #1848.

In order to leave us more room for future changes without worrying about breaking the ABI, we move code as much code as possible into a new `internal` namespace. No internal header may be included in a public header, only from `.cpp` files or other internal headers. Headers from the `internal` namespace are never installed. Nothing in `internal` may ever be exported.


libcaf_net/caf/internal/lp_flow_bridge.cpp
```diff
@@ -2,31 +2,31 @@
...
 
 #include "caf/async/blocking_producer.hpp"
 #include "caf/cow_tuple.hpp"
-#include "caf/detail/flow_bridge_base.hpp"
+#include "caf/internal/flow_bridge_base.hpp"
 
 #include <cstddef>
 
-namespace caf::detail {
+namespace caf::net::lp {
```
riemass, line 17: Any particular reason you didn't use `caf::internal` here?

libcaf_net/caf/net/web_socket/server_factory.cpp
```diff
@@ -6,68 +6,97 @@
...
+  // using impl_t = detail::accept_handler<Acceptor>;
+  // auto factory = std::make_unique<factory_t>(cfg.wca,
+  //                                            cfg.max_consecutive_reads);
+  // auto impl = impl_t::make(std::move(acc), std::move(factory),
+  //                          cfg.max_connections);
+  // auto impl_ptr = impl.get();
+  // auto ptr = net::socket_manager::make(cfg.mpx, std::move(impl));
+  // impl_ptr->self_ref(ptr->as_disposable());
+  // cfg.mpx->start(ptr);
+  // return expected<disposable>{disposable{std::move(ptr)}};
```
riemass, line 97: Old code.

riemass (changes requested):
Great work. Just some nits.

libcaf_net/caf/internal/lp_flow_bridge.cpp
Neverlord: Anything in this block goes into the anonymous namespace anyways, so I've picked the most convenient namespace for the implementation.

riemass (approved):

PR #1856, opened 2024-05-13
Add missing tests to improve coverage

shariarriday:
Tests have been added for the following
- beacon
- ipv4_subnet
- ipv6_subnet
- scoped_actor
- make_config_option

The following functions have been removed from settings as they have not been referenced in any public function
- `expected<std::string> get_or(const settings& xs, std::string_view name, const char* fallback)`
- `config_value& put_impl(settings& dict, const std::vector<std::string_view>& path, config_value& value)`

### DIFF
--- a/libcaf_core/caf/ipv4_subnet.test.cpp
+++ b/libcaf_core/caf/ipv4_subnet.test.cpp
@@ -4,7 +4,13 @@
 
 #include "caf/ipv4_subnet.hpp"
 
+#include "caf/test/fixture/deterministic.hpp"
 #include "caf/test/test.hpp"
+
+#include "caf/actor_system.hpp"
+#include "caf/actor_system_config.hpp"
+#include "caf/binary_deserializer.hpp"
+#include "caf/binary_serializer.hpp"
 
 using namespace caf;
 
@@ -50,5 +56,19 @@ TEST("ordering") {
   check_lt(addr(192, 168, 167, 0) / 24, addr(192, 168, 168, 0) / 24);
   check_lt(addr(192, 168, 168, 0) / 24, addr(192, 168, 168, 0) / 25);
 }
+
+WITH_FIXTURE(test::fixture::deterministic) {
+
+#define CHECK_SERIALIZATION(subn) check_eq(subn, serialization_roundtrip(subn))
+
+TEST("serialization") {
+  CHECK_SERIALIZATION(addr(192, 168, 168, 1) / 16);
+  CHECK_SERIALIZATION(addr(192, 168, 168, 1) / 24);
+  CHECK_SERIALIZATION(addr(192, 168, 168, 1) / 31);
+  CHECK_SERIALIZATION(addr(255, 255, 255, 1) / 8);
+  CHECK_SERIALIZATION(addr(127, 0, 0, 1) / 24);
+}
+
+} // WITH_FIXTURE(test::fixture::deterministic)
 
 } // namespace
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/config_option.hpp
```diff
@@ -105,6 +105,9 @@ class CAF_CORE_EXPORT config_option {
   /// - Store the converted value unless this option is stateless.
   error sync(config_value& x) const;
 
+  /// Tries to extract a value from the set location.
+  config_value get() const;
```
Neverlord, line 109: This is a very dangerous getter that we should remove again. The internal pointer may be null and if it isn't, it's pointing to something that the user has given us, so there's no need to ask the config option. What was the motivation to add this?

libcaf_core/caf/ipv4_subnet.test.cpp
```diff
@@ -16,6 +21,28 @@ ipv4_subnet operator/(ipv4_address addr, uint8_t prefix) {
   return {addr, prefix};
 }
 
+struct fixture {
```
Neverlord, line 24: Why aren't we just using the deterministic fixture here with [`serialization_roundtrip`](https://github.com/actor-framework/actor-framework/blob/master/libcaf_test/caf/test/fixture/deterministic.hpp#L634)?

libcaf_core/caf/unit.test.cpp
```diff
@@ -54,4 +54,11 @@ TEST("unit_results") {
   }
 }
 
+TEST("actor_address") {
+  actor_system_config cfg;
+  actor_system sys{cfg};
+  scoped_actor self{sys};
+  check(self.address().id());
```
Neverlord, line 61: This doesn't return a boolean.

```suggestion
  check_ne(self.address().id(), 0);
```

Neverlord (changes requested):

libcaf_core/caf/config_option.hpp
shariarriday: Currently, there is no public interface to use the `get_impl()` function and this getter has been added to use the implementation. I am removing the getter and the related tests.

Neverlord: What `get_impl`? Is it something `meta_->get(value_)` is calling internally? Then we can simply test it via the meta objects directly.

shariarriday: Yes `meta_->get(value_)` calls the `get_impl()` which is defined in `make_config_option.hpp` https://github.com/actor-framework/actor-framework/blob/e876df5f0797206bdb80b79f442db38285f16d56/libcaf_core/caf/make_config_option.hpp#L33-L38

I am trying to write tests with meta objects directly.

shariarriday: We can not use the meta objects directly as we need the object that is referenced by the following internal function https://github.com/actor-framework/actor-framework/blob/e876df5f0797206bdb80b79f442db38285f16d56/libcaf_core/caf/make_config_option.hpp#L40-L45

I think that the tests are not necessary here as the function pointer will be removed eventually
https://github.com/actor-framework/actor-framework/blob/e876df5f0797206bdb80b79f442db38285f16d56/libcaf_core/caf/config_option.hpp#L48-L50

Neverlord: Thanks for digging into this. I've removed the obsolete pieces. 👍

Neverlord (approved):

PR #1861, opened 2024-05-16
Fix math in chrono::datetime::force_utc

### LINKED ISSUE #1860: bug in chrono.cpp/ chrono.test.cpp?
### Discussed in https://github.com/actor-framework/actor-framework/discussions/1859

<div type='discussions-op-text'>

<sup>Originally posted by **xliao-neology** May 15, 2024</sup>

```cpp
auto str1 = chrono::to_string(make_timestamp());
std::cout << str1 << std::endl;

auto to_utc = [](std::string_view str) {
    return caf::chrono::datetime::from_string(str).and_then([](auto&& dt) {
      auto cpy = dt;
      cpy.force_utc();
      return expected<std::string>{to_string(cpy)};
    });
  };

auto str = chrono::to_string(std::chrono::system_clock::now()); 
std::cout<<*to_utc(str)<<std::endl;
```

```
2024-05-15T14:36:28.710562027-05:00
2024-05-15T09:36:28.710635744Z
```

I use central time which is 5hours behind UTC, the first output seems correct(14:36:28.710562027 is my local time), but when I call to_utc from chrono.test.cpp, the output is not the correct UTC time, it should be add the offset instead of of subtract it .

correct UTC time should be
2024-05-15T19:36:28.710635744Z

</div>

Neverlord:
It's a one character fix, but I've used the opportunity to improve our unit tests as well.

Fixes #1860.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   no user-info at all.
 - The parser for reading JSON and configuration files now properly handles
   Windows-style line endings (#1850).
+- Calling `force_utc` on a `caf::chrono::dateime` object now properly applies
+  the UTC offset. Previously, the function would shift the time into the wrong
+  direction (#1860).
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


riemass (approved):
:+1: 
I had to check this, 'timezone' library was standardized in C++20. While g++ 11 provides partial support for this feature, full support is only available in g++14.

PR #1863, opened 2024-05-21
Add tests to improve code coverage

shariarriday:
Tests have been added for json_builder.

Test has been added to validate that blocking actors are spawned. 

Some bugs have been fixed in json_builder related to serializing keys.

### DIFF
--- a/libcaf_core/caf/blocking_actor.test.cpp
+++ b/libcaf_core/caf/blocking_actor.test.cpp
@@ -14,6 +14,10 @@ using namespace caf;
 
 namespace {
 
+void testee(blocking_actor* self) {
+  self->receive([](int i) { return i; });
+}
+
 struct fixture : test::fixture::deterministic {
   scoped_actor self{sys};
 };
@@ -42,6 +46,18 @@ TEST("timeout_in_scoped_actor") {
   self->receive(after(std::chrono::milliseconds(20)) >>
                 [&] { timeout_called = true; });
   check(timeout_called);
+}
+
+TEST("spawn blocking actor") {
+  auto aut = sys.spawn(testee);
+  self->mail(42).send(aut);
+  dispatch_messages();
+  auto received = std::make_shared<bool>(false);
+  self->receive([this, received](int i) {
+    *received = true;
+    check_eq(i, 42);
+  });
+  check(*received);
 }
 
 } // WITH_FIXTURE(fixture)
--- a/libcaf_core/caf/json_builder.test.cpp
+++ b/libcaf_core/caf/json_builder.test.cpp
@@ -17,16 +17,6 @@
 using namespace caf;
 using namespace std::literals;
 
-namespace caf {
-
-template <class T>
-bool operator<(const span<T>& lhs, const span<T>& rhs) {
-  return std::lexicographical_compare(lhs.begin(), lhs.end(), rhs.begin(),
-                                      rhs.end());
-}
-
-} // namespace caf
-
 namespace {
 
 struct my_request;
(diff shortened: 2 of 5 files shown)


libcaf_core/caf/blocking_actor.test.cpp
```diff
@@ -44,6 +48,18 @@ TEST("timeout_in_scoped_actor") {
   check(timeout_called);
 }
 
+TEST("spawn blocking actor") {
+  auto blocking_actor = sys.spawn(testee);
```
Neverlord, line 52:
```suggestion
  auto aut = sys.spawn(testee);
```

libcaf_core/caf/blocking_actor.test.cpp
```diff
@@ -44,6 +48,18 @@ TEST("timeout_in_scoped_actor") {
...
+  self->mail(42).send(blocking_actor);
+  dispatch_messages();
+  auto received = std::make_shared<bool>(false);
+  self->receive([received](int i) {
+    *received = true;
+    test::runnable::current().check_eq(i, 42);
```
Neverlord, line 58: Should work if you just capture `this`

```suggestion
    check_eq(i, 42);
```

libcaf_core/caf/json_builder.test.cpp
```diff
@@ -93,32 +95,170 @@ TEST("empty JSON value") {
...
+  check(val.is_integer());
+  check_eq(val.to_integer(), 42);
+}
+
+TEST("int16") {
+  check(builder.value(int16_t{42}));
+  auto val = builder.seal();
+  check(val.is_integer());
+  check_eq(val.to_integer(), 42);
+}
```
Neverlord, line 117: There's quite a bit of repetition here that we can cut. The only real difference is the call to `value`:

```c++
TEST("integer") {
  SECTION("int8") {
    check(builder.value(int8_t{42}));
  }
  // ... sections for the other integer types ...
  auto val = builder.seal();
  check(val.is_integer());
  check_eq(val.to_integer(), 42);
}

File: libcaf_core/caf/span.hpp [hunk 4]
```diff
@@ -174,6 +174,12 @@ class span {
   size_t size_;
 };
 
+template <class T>
+bool operator<(const span<T>& lhs, const span<T>& rhs) {
+  return std::lexicographical_compare(lhs.begin(), lhs.end(), rhs.begin(),
+                                      rhs.end());
+}
+
```
Neverlord, line 182: If you need this in a test, please add a function for this in the test and remove this again. `std::span` doesn't support this, so neither do we.

libcaf_test/caf/test/runnable.hpp
```diff
@@ -65,7 +65,13 @@ class CAF_TEST_EXPORT runnable {
...
                 = detail::source_location::current()) {
     assert_save_comparison<T0, T1>();
-    if (lhs == rhs) {
+    if constexpr (std::is_floating_point_v<T0>
+                  && std::is_floating_point_v<T1>) {
+      if (lhs - rhs < std::numeric_limits<T0>::epsilon()) {
+        reporter::instance().pass(location);
+        return true;
+      }
+    } else if (lhs == rhs) {
```
Neverlord, line 74: I think the precision should not be hidden away in a check function. Instead, let's extend `assert_save_comparison` to prohibit floating points comparison using `operator==`.

I like the approach of doctest, so let's do something similar with an explicit `approx` class like this:

```
/// Allows comparing a value of type `T` with a configurable precision.
template <class T>
class approx {
public:
  explicit approx(T value) : value_(value) { }

  approx epsilon(T eps) const {
    auto copy = *this;
    copy.epsilon_ = eps;
    return copy;
  }

  friend class operator==(T lhs, approx<T> rhs) {
    return lhs >= min_value() && lhs <= max_value();
  }

  // ... operator==(approx<T>, T) ...
private:
  T min_value() { return value_ - epsilon_; }
  T max_value() { return value_ + epsilon_; }
  T value_;
  T epsilon_ = std::numeric_limit<T>::epsilon();
};
```

This will allow users to write `check_eq(my_value, approx{3.127})` to get the default precision or `check_eq(my_value, approx{3.127}.epsilon(0.001))` to do the comparison with a custom precision.

Neverlord (changes requested):
The commit message just says "Add tests for json_builder", but the commit adds other tests as well. So please amend the commit with a better commit message. Since you're force-pushing anyways, please rebase to latest head.

Neverlord:
It's also a bugfix now (see fixes in json_builder.cpp).

Neverlord (approved):

PR #1871, opened 2024-05-24
Change values for log levels

### LINKED ISSUE #1066: (Almost) no way to make use of fine grained log levels

Neverlord:
This sets the values for our log levels to the same values Log4j uses, giving users much more space for custom log levels. The logger already allows passing custom log levels to `log`, so this wraps up this topic (aside from documentation).

Closes #1066.

### DIFF
--- a/libcaf_core/caf/detail/log_level.hpp
+++ b/libcaf_core/caf/detail/log_level.hpp
@@ -8,16 +8,16 @@
 #define CAF_LOG_LEVEL_QUIET 0
 
 /// Integer value for the ERROR log level.
-#define CAF_LOG_LEVEL_ERROR 3
+#define CAF_LOG_LEVEL_ERROR 200
 
 /// Integer value for the WARNING log level.
-#define CAF_LOG_LEVEL_WARNING 6
+#define CAF_LOG_LEVEL_WARNING 300
 
 /// Integer value for the INFO log level.
-#define CAF_LOG_LEVEL_INFO 9
+#define CAF_LOG_LEVEL_INFO 400
 
 /// Integer value for the DEBUG log level.
-#define CAF_LOG_LEVEL_DEBUG 12
+#define CAF_LOG_LEVEL_DEBUG 500
 
 /// Integer value for the TRACE log level.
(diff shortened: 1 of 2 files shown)


shariarriday (approved):
From the description it looks like the documentation will be updated in a separate PR so it looks good to me currently.

PR #1876, opened 2024-05-26
Mention auto-move of values into message handlers

Neverlord:
Relates #1832.

### DIFF
--- a/examples/curl/curl_fuse.cpp
+++ b/examples/curl/curl_fuse.cpp
@@ -282,7 +282,7 @@ struct coordinator_state : base_state {
     self->println(color, "{}[{}]: spawned {} worker(s)", self->name(),
                   self->id(), idle.size());
     return {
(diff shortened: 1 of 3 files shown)


riemass (approved):
Approved. 
I don't know if we have an example of sending the same message to multiple destinations without copying. 
I guess something like:
```
auto msg = make_message(...);
for (const auto& receiver : receivers)
  self->mail(msg).send(receiver);
```
Might be worth adding to the manual/examples?

Neverlord:
> ```
> auto msg = make_message(...);
> for (const auto& receiver : receivers)
>   self->mail(msg).send(receiver);
> ```
> 
> Might be worth adding to the manual/examples?

Hm, not sure about a use case for this. If we have something that can be "subscribed to" by multiple receivers, we'd generally recommend streams. It'll also only work for dynamically typed actors.

PR #1877, opened 2024-05-26
Refactor floating point comparisons with approx

shariarriday:
Refactor tests to use `approx` class with adjustable epsilon to compare floating points, instead of comparing floating point with equal operator.

### DIFF
--- a/libcaf_core/caf/actor_system_config.test.cpp
+++ b/libcaf_core/caf/actor_system_config.test.cpp
@@ -4,6 +4,7 @@
 
 #include "caf/actor_system_config.hpp"
 
+#include "caf/test/approx.hpp"
 #include "caf/test/scenario.hpp"
 #include "caf/test/test.hpp"
 
@@ -28,9 +29,15 @@ using namespace std::literals;
   do {                                                                         \
     using ref_value_type = std::decay_t<decltype(var)>;                        \
     ref_value_type value{__VA_ARGS__};                                         \
-    check_eq(var, value);                                                      \
+    if constexpr (std::is_arithmetic_v<decltype(var)>)                         \
+      check_eq(var, test::approx{value});                                      \
+    else                                                                       \
+      check_eq(var, value);                                                    \
     if (auto maybe_val = get_as<decltype(var)>(cfg, #var)) {                   \
-      check_eq(*maybe_val, value);                                             \
+      if constexpr (std::is_arithmetic_v<decltype(var)>)                       \
+        check_eq(*maybe_val, test::approx{value});                             \
+      else                                                                     \
+        check_eq(*maybe_val, value);                                           \
     } else {                                                                   \
       auto cv = get_if(std::addressof(cfg.content), #var);                     \
       fail("expected type {}, got {}",                                         \
--- a/libcaf_core/caf/json_reader.test.cpp
+++ b/libcaf_core/caf/json_reader.test.cpp
@@ -4,6 +4,7 @@
 
 #include "caf/json_reader.hpp"
 
+#include "caf/test/approx.hpp"
 #include "caf/test/scenario.hpp"
 #include "caf/test/test.hpp"
 
@@ -222,6 +223,8 @@ struct fixture {
       if (res) {
         if constexpr (std::is_same_v<T, message>)
           res = this_test.check_eq(to_string(tmp), to_string(obj));
+        else if constexpr (std::is_arithmetic_v<T>)
+          res = this_test.check_eq(tmp, test::approx{obj});
         else
           res = this_test.check_eq(tmp, obj);
       }
--- a/libcaf_core/caf/serialization.test.cpp
+++ b/libcaf_core/caf/serialization.test.cpp
@@ -2,6 +2,7 @@
 // the main distribution directory for license terms and copyright or visit
 // https://github.com/actor-framework/actor-framework/blob/main/LICENSE.
 
+#include "caf/test/approx.hpp"
 #include "caf/test/fixture/deterministic.hpp"
 #include "caf/test/outline.hpp"
 #include "caf/test/runnable.hpp"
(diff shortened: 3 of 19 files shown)


libcaf_core/caf/actor_system_config.test.cpp
```diff
@@ -28,9 +29,15 @@ using namespace std::literals;
   do {                                                                         \
     using ref_value_type = std::decay_t<decltype(var)>;                        \
     ref_value_type value{__VA_ARGS__};                                         \
-    check_eq(var, value);                                                      \
+    if constexpr (std::is_floating_point_v<decltype(var)>)                     \
+      check_eq(var, test::approx{value});                                      \
+    else                                                                       \
+      check_eq(var, value);                                                    \
```
Neverlord, line 35: Why not just always compare against `approx{value}`? If it's an integer value, `epsilon` will be 0, so it should turn out the same.

libcaf_core/caf/json_reader.test.cpp
```diff
@@ -251,6 +252,22 @@ struct fixture {
     test_cases.emplace_back(std::move(f));
   }
 
+  // Specialization for `double` to add a test case for a given input.
+  void add_test_case(std::string_view input, double val) {
```
Neverlord, line 256: I think we can just use `approx` in the template version. Comparing integers with `approx` without setting an `epsilon` compares for equality (since the default `epsilon` is 0).

libcaf_core/caf/serialization.test.cpp
```diff
@@ -702,7 +703,11 @@ OUTLINE("serializing and then deserializing primitive values") {
             using lhs_t = std::decay_t<decltype(lhs)>;
             using rhs_t = std::decay_t<decltype(rhs)>;
             if constexpr (std::is_same_v<lhs_t, rhs_t>) {
-              check_eq(lhs, rhs);
+              if constexpr (std::is_floating_point_v<lhs_t>
+                            || std::is_floating_point_v<rhs_t>)
+                check_eq(lhs, test::approx{rhs});
+              else
+                check_eq(lhs, rhs);
```
Neverlord, line 710:
```suggestion
              check_eq(lhs, test::approx{rhs});
```

libcaf_test/caf/test/runnable.hpp
```diff
@@ -365,6 +365,9 @@ class CAF_TEST_EXPORT runnable {
       static_assert(std::is_signed_v<T0> == std::is_signed_v<T1>,
                     "comparing signed and unsigned integers is unsafe");
     }
+    static_assert(!(std::is_floating_point_v<T0>
+                    && std::is_floating_point_v<T1>),
+                  "comparing floating points should be done through approx");
```
Neverlord, line 370:
```suggestion
                  "comparing floating points using operator== is unsafe, use approx");
```

Neverlord (changes requested):

libcaf_core/caf/actor_system_config.test.cpp
shariarriday: If we directly compare against `approx{value}` then we will get compilation error for types where the used operators (>=, <=, + and - ) are not supported. Due to this, this floating check has been added. I have moved this floating point check inside `friend bool operator==(T lhs, approx<T> rhs)` to fix address the feedback.

libcaf_test/caf/test/approx.hpp
```diff
@@ -0,0 +1,54 @@
...
+    copy.epsilon_ = eps;
+    return copy;
+  }
+
+  friend bool operator==(T lhs, approx<T> rhs) {
+    if constexpr (std::is_floating_point_v<T>) {
+      return lhs >= rhs.min_value() && lhs <= rhs.max_value();
+    } else {
+      return lhs == rhs.value_;
+    }
```
Neverlord, line 30: > we will get compilation error for types where the used operators (>=, <=, + and - ) are not supported.

What type would we pass to `approx` where this is a problem? Using `==` for the `else` branch is incorrect: `check_eq(5, approx{4}.epsilon(1)` should pass.

libcaf_test/caf/test/approx.hpp
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  friend bool operator>(T lhs, approx<T> rhs) {
+    if constexpr (std::is_floating_point_v<T>) {
+      return lhs > rhs.max_value();
+    } else {
+      return lhs > rhs.value_;
+    }
+  }
```
Neverlord, line 39: Comparing against `approx` with `operator>` doesn't strike me as logically sound operation. Should be removed.

Neverlord (changes requested):

libcaf_test/caf/test/approx.hpp
Neverlord: I didn't check all places before where you've used `if constexpr` to check for floating points. I've assumed that we're dealing with arithmetic types, not with optionals and the like.

Ok, before we get lost in individual code pieces:

- `approx` should only work with types that are actual numbers, like integers and floating point values
- arrays, optionals, etc. should be dealt with at call side
- `approx` should only support `operator==`
- using `operator<` and `operator>` for floating points is fine, we only want to prohibit `operator==` for floating point numbers. The right place to put the `static_assert` might be just `check_eq`.

shariarriday: Thanks, this clarifies things up a lot. I am updating the PR to ensure that these are enforced properly.

libcaf_test/caf/test/runnable.hpp
```diff
@@ -65,6 +65,9 @@ class CAF_TEST_EXPORT runnable {
                 const detail::source_location& location
                 = detail::source_location::current()) {
     assert_save_comparison<T0, T1>();
+    static_assert(
+      !(std::is_floating_point_v<T0> && std::is_floating_point_v<T1>),
+      "comparing floating points using operator== is unsafe, use approx");
```
Neverlord, line 70: That same assert should go into `require_eq`.

Neverlord (changes requested):

libcaf_test/caf/test/runnable.hpp
shariarriday: As `require_eq` calls `check_eq` internally, the static_assert will be triggered when trying to compare floating point numbers with `require_eq(res, 1.0f)`. Should we duplicate the static_assert anyways?

Neverlord: Nah, sorry for not double-checking. 👍

Neverlord (approved):

PR #1880, opened 2024-05-28
Change handling of special messages

### LINKED ISSUE #1800: Deprecate legacy APIs for 1.0

Neverlord:
- Event-based actors can now handle types like `exit_msg` and `error` in their regular behavior.
- The `set_default_handler` member function on event-based actors is now deprecated. Instead, users should use a handler for `message` in their behavior as a catch-all. For skipping messages, CAF now includes a new `mail_cache` class that allows explicitly stashing messages for later processing.
- Special-purpose handlers for messages like `exit_msg` and `error` are now deprecated. Instead, users should handle these messages in their regular behavior.

Closes #1800.

### DIFF
--- a/libcaf_core/caf/detail/type_list.hpp
+++ b/libcaf_core/caf/detail/type_list.hpp
@@ -228,6 +228,35 @@ struct tl_at<type_list<E...>, N> {
 template <class List, size_t N>
 using tl_at_t = typename tl_at<List, N>::type;
 
+// list remove(list, value)
+
+template <class Input, class Value, class Output>
+struct tl_remove_impl;
+
+template <class Value, class... Out>
+struct tl_remove_impl<empty_type_list, Value, type_list<Out...>> {
+  using type = type_list<Out...>;
+};
+
+template <class T, class... Ts, class Value, class... Out>
+struct tl_remove_impl<type_list<T, Ts...>, Value, type_list<Out...>> {
+  using type = std::conditional_t<
+    std::is_same_v<T, Value>,
+    typename tl_remove_impl<type_list<Ts...>, Value, type_list<Out...>>::type,
+    typename tl_remove_impl<type_list<Ts...>, Value,
+                            type_list<Out..., T>>::type>;
+};
+
+template <class List, class T>
+struct tl_remove;
+
+template <class... Ts, class T>
+struct tl_remove<type_list<Ts...>, T>
+  : tl_remove_impl<type_list<Ts...>, T, type_list<>> {};
+
+template <class List, class T>
+using tl_remove_t = typename tl_remove<List, T>::type;
+
 // list filter(list, predicate)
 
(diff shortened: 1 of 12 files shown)


libcaf_core/caf/detail/type_list.hpp
```diff
@@ -228,6 +228,35 @@ struct tl_at<type_list<E...>, N> {
...
+};
+
+template <class T, class... Ts, class Value, class... Out>
+struct tl_remove_impl<type_list<T, Ts...>, Value, type_list<Out...>> {
+  using type = std::conditional_t<
+    std::is_same_v<T, Value>,
+    typename tl_remove_impl<type_list<Ts...>, Value, type_list<Out...>>::type,
+    typename tl_remove_impl<type_list<Ts...>, Value,
+                            type_list<Out..., T>>::type>;
+};
```
riemass, line 248: If I got this right, the implementation will reverse the items in the type list:
```
tl_remove<type_list<int, string, double, float>, float> => type_list<double, string, int>
```
I'd rather have the ordering preserved (with the appropriate item removed). This might not be important to the compiler, but I'd rather not confuse the users who read compiler logs more than they already are. 
```suggestion
template <class T, class... Ts, class Value, class... Out>
struct tl_remove_impl<type_list<T, Ts...>, Value, type_list<Out...>> {
  using type = std::conditional_t<
    std::is_same_v<T, Value>,
    typename tl_remove_impl<type_list<Ts...>, Value, type_list<Out...>>::type,
    typename tl_remove_impl<type_list<Ts...>, Value,
                            type_list<T, Out...>>::type>;
};
```

riemass (changes requested):
Kudos on remembering the typed actor interfaces, that one was bothering me for a while. 
I have a nit in one of the functions, comment below:

libcaf_core/caf/detail/type_list.hpp
riemass: Sorry, I made a mistake here, it's too late for recursive templates.

riemass (approved):

PR #1885, opened 2024-05-31
Exclude system messages from fallback handlers

### LINKED ISSUE #1884: Catch-all handlers apply to system messages
When defining a callback for `message`, this will currently include types such as `exit_msg` and `error` that have a special meaning. These are not "unexpected" messages, but system-level messages. The fallback handler should not apply to those kind of messages, since it will accidentally drop any `exit_msg` instead of handling them properly.

Neverlord:
Fixes #1884.


riemass (approved):
Good catch.

PR #1886, opened 2024-05-31
Add text-client example with robot test

shariarriday:
An example `text-client` that utilizes `octet_stream` and the `key-value-store` example to store/retrieve simple string in server.
Additionally, a robot test has been added to test the examples and the expected functionality.

### DIFF
--- a/examples/octet_stream/key-value-store.cpp
+++ b/examples/octet_stream/key-value-store.cpp
@@ -226,12 +226,14 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
           sys.spawn([events, kvs](caf::event_based_actor* self) {
             // For each buffer pair, we create a new flow ...
             events.observe_on(self).for_each([self, kvs](auto ev) {
+              self->println("*** accepted new connection ");
               auto [pull, push] = ev.data();
               pull //
                 .observe_on(self)
                 // ... that converts the lines to commands ...
                 .transform(caf::flow::byte::split_as_utf8_at('\n'))
-                .map([](const caf::cow_string& line) {
+                .map([self](const caf::cow_string& line) {
+                  self->println("received: {}", line.str());
                   caf::json_reader reader;
                   if (!reader.load(line.str()))
                     return std::shared_ptr<command>{}; // Invalid JSON.
@@ -269,6 +271,8 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
                 .on_backpressure_buffer(max_outstanding_messages)
                 // ... and pushes the results back to the client as bytes.
                 .transform(caf::flow::string::to_chars("\n"))
+                .do_finally(
+                  [self]() { self->println("*** lost connection to client"); })
                 .map([](char ch) { return static_cast<std::byte>(ch); })
                 .subscribe(push);
             });
--- a/examples/octet_stream/text-client.cpp
+++ b/examples/octet_stream/text-client.cpp
@@ -9,6 +9,7 @@
 #include "caf/async/spsc_buffer.hpp"
 #include "caf/caf_main.hpp"
 #include "caf/flow/byte.hpp"
+#include "caf/flow/string.hpp"
 #include "caf/json_reader.hpp"
 #include "caf/scheduled_actor/flow.hpp"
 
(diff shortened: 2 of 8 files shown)


examples/octet_stream/text-client.cpp
```diff
@@ -0,0 +1,144 @@
...
+              }
+              std::string json_line = R"_({"type":")_" + parsed_line[0]
+                                      + R"_(", "key":")_" + parsed_line[1]
+                                      + R"_(")_";
+              if (parsed_line.size() == 3) {
+                json_line += R"_(, "value":")_" + parsed_line[2] + R"_(")_";
+              }
+              json_line += "}\n";
+              lines->push(caf::as_bytes(caf::make_span(json_line)));
+              line.clear();
```
Neverlord, line 122: Whether the server speaks JSON or not should not be relevant to the example. We just expect the server to have some sort of text-based interface.

```suggestion
              lines->push(caf::as_bytes(caf::make_span(line)));
              line.clear();
```

Neverlord (changes requested):

examples/octet_stream/text-client.cpp
```diff
@@ -0,0 +1,144 @@
...
+  // Wait for CTRL+C or SIGTERM.
+  sys.println("*** Client is running, press CTRL+C to stop");
+  while (!shutdown_flag)
+    std::this_thread::sleep_for(250ms);
+  // Restore the default handlers.
+  signal(SIGTERM, default_term);
+  signal(SIGINT, default_sint);
+  // Shut down the client.
+  sys.println("*** shutting down");
+  return EXIT_SUCCESS;
```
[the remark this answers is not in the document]
Neverlord: Actually, the actor reading from `cin` can never terminate this way since it won't ever check it's mailbox. Pull out out the `cin` loop from the actor and do the loop in `main` instead, pushing data into the actor.

examples/octet_stream/key-value-store.cpp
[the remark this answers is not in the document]
shariarriday: `std::cout` has been used to ensure that the robot tests pass. When `sys.println()/self->println()` is used the robot tests fail as the print statements do not print to console immediately and they are buffered instead.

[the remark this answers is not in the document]
Neverlord: If buffering behavior is an issue, please add `fflush` to the default print function (https://github.com/actor-framework/actor-framework/blob/main/libcaf_core/caf/actor_system.cpp#L336) instead.

examples/octet_stream/text-client.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (!conn) {
+    sys.println("*** unable to run at port {}: {}", port, conn.error());
+    return EXIT_FAILURE;
+  }
+  // Wait for CTRL+C or SIGTERM.
+  sys.println("*** Client is running, press CTRL+C to stop");
+  auto line = std::string{};
+  while (std::getline(std::cin, line)) {
+    line = line + '\n';
+    caf::anon_mail(line).send(producer_actor);
```
Neverlord, line 125: This breaks backpressure. Use a blocking producer here. Then, we'll only need to spawn one worker actor that does both flows. Also means we no longer need to care about the actor handles, since the blocking producer will close the flow for the worker at scope exit.

robot/stream/octet.robot
```diff
@@ -0,0 +1,81 @@
...
+${CLIENT_PATH}      /path/to/the/client
+
+${FIRST_CLIENT_INPUT}         { "type": "put", "key": "A", "value": "26" }\n{ "type": "get", "key": "A" }\n{ "type": "del", "key": "A" }\n{ "type": "get", "key": "A" }     
+${FIRST_CLIENT_BASELINE}      reply: \nreply: 26\nreply: 26\nreply: {"error":"no_such_key"}
+${SECOND_CLIENT_INPUT}        { "type": "put", "key": "B", "value": "27" }\n{ "type": "get", "key": "B" }\n{ "type": "del", "key": "B" }\n{ "type": "get", "key": "B" }      
+${SECOND_CLIENT_BASELINE}     reply: \nreply: 27\nreply: 27\nreply: {"error":"no_such_key"}    
+
+
+*** Test Cases ***
+Send Some message
```
Neverlord, line 25: That's not a descriptive test case name. Also, there's no indication on what's the point of doing the same thing twice in the test?

robot/stream/octet.robot
```diff
@@ -0,0 +1,81 @@
...
+    ...  stderr=server.err
+    Wait For Server Startup
+
+Start Text Client
+    [Arguments]   ${stdin}
+    Start Process
+    ...  ${CLIENT_PATH}
+    ...  -p    ${SERVER_PORT}
+    ...  --caf.logger.file.path  text-store.log
+    ...  stdin=${stdin}
```
Neverlord, line 50: Is this closing `stdin` of the process after everything has been written? That would make the tests unstable, because our reader shuts down once `stdin` has been closed.

Neverlord (changes requested):

robot/stream/octet.robot
shariarriday: The process does not quit after `stdin` has completed writing. We have written a teardown function which closes the running processes after the test is completed.

Neverlord: Robot does close stdin of the process after writing. The example just didn't shut down as a result. I've restructured the code a bit and had to work around this behavior in the Robot tests. I've also simplified the example a bit, so please take a look at the changes I've made for future reference.

Neverlord (approved):

Issue #1889, opened 2024-06-08
json_builder issue


xliao-neology:
suppose I have struct as below.
////
struct result_j
{
  uint32_t a;
  uint32_t b;
  string c;
  vector<int> d;
  result_j()=default;
  result_j(uint32_t a, uint32_t b, string g, vector<int> d): a(a), b(b), c(c), d(d){}
};

template <class Inspector>
bool inspect(Inspector &f, struct result_j &x)
{
  return f.object(x).fields(f.field("a", x.a), f.field("b", x.b), f.field("c", x.c), f.field("d", x.d));
}

/// test code as below //////////////

    vector<int> d{1, 2, 3, 4};
    result_j resj{85, 95, "score", d};

     caf::json_builder builder;
     string result;
    auto res=builder.apply(resj);  
    if(res){
      auto jval = builder.seal();
      jval.print_to(result, 1);    
    }else{
      cout<<to_string(builder.get_error())<<endl;
    }
  std::cout<<result<<endl;

///above test code get error 
runtime_error("pop_if failed: expected member, found element")

//if do not check the return value of builder.apply,  it seems work. output as below.

     auto res=builder.apply(resj);  
      auto jval = builder.seal();
      jval.print_to(result, 1);  
      cout<<result<<endl;
//output
{
 "@type": "anonymous",
 "a": 85,
 "b": 95,
 "c": "score",
 "d": [
  1,
  2,
  3,
  4
 ]
}

it seems json_builder have some issues with nest case, but I am not sure why not checking the builder.apply return value, it works. Is this a bug?

Neverlord:
This is a bug. I've reproduced it on latest `main`:

```c++
#include "caf/actor_system.hpp"
#include "caf/caf_main.hpp"
#include "caf/json_builder.hpp"

#include <string>

using namespace std::literals;

struct abcd {
  uint32_t a;
  uint32_t b;
  std::string c;
  std::vector<int> d;
};

template <class Inspector>
bool inspect(Inspector& f, struct abcd& x) {
  return f.object(x).fields(f.field("a", x.a), f.field("b", x.b),
                            f.field("c", x.c), f.field("d", x.d));
}

void caf_main(caf::actor_system& sys) {
  abcd res{85, 95, "score", {1, 2, 3, 4}};
  caf::json_builder builder;
  if (builder.apply(res)) {
    std::string str;
    auto jval = builder.seal();
    jval.print_to(str, 1);
    sys.println("{}", str);
[...]

PR #1890, opened 2024-06-09
Fix begin_sequence to handle arrays inside struct

shariarriday:
begin_sequence previously pushed a new value to the stack which was causing a bug when attempting to pop member from the stack. This has been fixed by replacing the stack value instead of pushing new value.
Additionally, a test has been added to test this scenario.
Closes https://github.com/actor-framework/actor-framework/issues/1889


Neverlord (approved):
No complaints! 👍

PR #1891, opened 2024-06-16
Minor API tweaks

Neverlord:
Two small API changes to make live easier.

### DIFF
--- a/libcaf_core/caf/string_algorithms.cpp
+++ b/libcaf_core/caf/string_algorithms.cpp
@@ -54,9 +54,9 @@ void split(std::vector<std::string_view>& result, std::string_view str,
 std::string_view trim(std::string_view str) {
   auto non_whitespace = [](char c) { return !isspace(c); };
   if (std::any_of(str.begin(), str.end(), non_whitespace)) {
-    while (str.front() == ' ')
+    while (std::isspace(static_cast<unsigned char>(str.front())))
       str.remove_prefix(1);
-    while (str.back() == ' ')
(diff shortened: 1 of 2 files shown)


libcaf_core/caf/string_algorithms.cpp
```diff
@@ -54,9 +54,9 @@ void split(std::vector<std::string_view>& result, std::string_view str,
 std::string_view trim(std::string_view str) {
   auto non_whitespace = [](char c) { return !isspace(c); };
   if (std::any_of(str.begin(), str.end(), non_whitespace)) {
-    while (str.front() == ' ')
+    while (std::isspace(str.front()))
```
dominiklohmann, line 57: Nit: For all the functions from `<cctype>`, you technically need to cast to `unsigned char` [to avoid UB](https://en.cppreference.com/w/cpp/string/byte/isspace).

Neverlord: You are of course correct, thanks!

libcaf_test/caf/test/runnable.hpp
```diff
@@ -265,6 +265,10 @@ class CAF_TEST_EXPORT runnable {
       return unbox_all(vals, std::index_sequence_for<Ts...>{});
   }
 
+  size_t block_parameters_index() const {
```
shariarriday, line 268: This function is missing comment for documentation.
```suggestion
  /// Returns the `index` of block_parameter for the running test.
  size_t block_parameters_index() const {
```

shariarriday (approved):

PR #1892, opened 2024-06-17
Various fixes for 1.0

Neverlord:
This set of commits fixes various issues that I've encountered while porting a CAF-based project from CAF 0.19 to current `main` for catching breaking changes / bugs early (before cutting the actual release).


riemass (approved):
:+1: 
Great you did this, generally I'm surprised by the small, convenience stuff  we would've missed otherwise.

PR #1902, opened 2024-06-29
Fix ctor access when spawning io::broker

### LINKED ISSUE #1900: regression in caf::io::spawn_client
While we didn't really touch the I/O module recently, something broke `spawn_client`. Code to reproduce:

```
  sys.middleman().spawn_client([](caf::io::broker*) {}, "foo", 42);
```

### Discussed in https://github.com/actor-framework/actor-framework/discussions/1899

<div type='discussions-op-text'>

<sup>Originally posted by **xliao-neology** June 27, 2024</sup>
we been using spawn_client() perfectly in 0.19.6. However same code will have below error show up in 1.0.0.
I understand the trend is to use caf_net. But for now, any way that I can do to keep spawn_client() work with 1.0.0 release?

>>>>>
/usr/local/include/caf/actor_storage.hpp:30:5: error: ‘caf::io::abstract_broker::abstract_broker(caf::actor_config&)’ is protected within this context
   30 |     new (&data) T(std::forward<Us>(zs)...);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
</div>

riemass:
Closes #1900

### DIFF
--- a/libcaf_io/caf/io/middleman.test.cpp
+++ b/libcaf_io/caf/io/middleman.test.cpp
@@ -12,15 +12,18 @@ using namespace caf::io;
 
 namespace {
 
-// Regression test for GH-#1900.
-// Note: we can't reproduce this in compile time,
-// static_assert(std::is_same_v<..snippet.., expected<actor>>) won't work.
-TEST("spawn client works") {
-  caf::actor_system_config cfg;
-  cfg.load<io::middleman>();
-  caf::actor_system sys{cfg};
-  sys.middleman().spawn_client(
-    [](caf::io::broker*, caf::io::connection_handle) {}, "foo", 42);
+TEST("GH-1900 regression") {
+  // Note: we don't need to actually run `spawn_client`, we only check for a
+  // compiler error. We shouldn't actually run this code, since it will assume
+  // 42 is a socket descriptor.
+  auto fn = [] {
+    caf::actor_system_config cfg;
+    cfg.load<io::middleman>();
+    caf::actor_system sys{cfg};
+    sys.middleman().spawn_client(
+      [](caf::io::broker*, caf::io::connection_handle) {}, "foo", 42);
+  };
+  check(std::is_same_v<decltype(fn), decltype(fn)>);
 }
 
 } // namespace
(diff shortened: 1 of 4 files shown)


libcaf_io/caf/io/middleman.test.cpp
```diff
@@ -0,0 +1,26 @@
...
+// Regression test for GH-#1900.
+// Note: we can't reproduce this in compile time,
+// static_assert(std::is_same_v<..snippet.., expected<actor>>) won't work.
+TEST("spawn client works") {
+  caf::actor_system_config cfg;
+  cfg.load<io::middleman>();
+  caf::actor_system sys{cfg};
+  sys.middleman().spawn_client(
+    [](caf::io::broker*, caf::io::connection_handle) {}, "foo", 42);
+}
```
Neverlord, line 24: "spawn client works" is not a very descriptive test name. Just calling it regression test with the issue number is perfectly fine. The `connection_handle` will be interpreted as a socket internally, so just passing 42 means this code eventuall calls `closesocket(42)`, which I think we should avoid.

```
TEST("GH-1900 regression") {
  // Note: we don't need to actually run `spawn_client`, we only check for a compiler error. We shouldn't actually run this code, since it will assume 42 is a socket descriptor.
  auto fn = [] {
    caf::actor_system_config cfg;
    cfg.load<io::middleman>();
    caf::actor_system sys{cfg};
    sys.middleman().spawn_client(
      [](caf::io::broker*, caf::io::connection_handle) {}, "foo", 42);
  };
  check_eq(std::is_same_v<decltype(fn), void>);
}
```

Neverlord (changes requested):
Thanks for taking care of this. Adding the `friend` declaration probably is the best fix. 👍

libcaf_io/caf/io/middleman.test.cpp
riemass: Good feedback, full ack.  
What's the last line about? If it's just to satisfy the testing framework, did you mean to write something like: 
```
  check(!std::is_same_v<decltype(fn), void>);
```

Neverlord: Ah yeah, should be `check(std::is_same_v<decltype(fn), void>);`

This line is to have at least one check in the suite and to avoid a unused warning for `fn`.

riemass: I've also added the `!` symbol, because this test will fail like this.

Neverlord: Ah, I meant `decltype(fn())`, but doesn't really matter.

Neverlord (approved):

PR #1903, opened 2024-07-04
Fix errors when using std::format 

riemass:
This was reported in the community chatroom. 
We want to use the same C++ version when detecting std::format support as when compiling our targets.

### DIFF
--- a/libcaf_core/caf/detail/format.hpp
+++ b/libcaf_core/caf/detail/format.hpp
@@ -27,53 +27,40 @@
 namespace caf::detail {
 
 template <class T>
-inline constexpr bool is_formattable = is_complete<std::formatter<T, char>>;
-
-template <class T>
-inline constexpr bool fmt_needs_to_string
-  = !is_formattable<T>
-    && !std::is_same_v<
-      decltype(inspect_access_type<stringification_inspector, T>()),
-      inspector_access_type::none>;
-
-template <class T>
-constexpr T&&
-fmt_fwd(std::remove_reference_t<T>& arg,
-        std::enable_if_t<is_formattable<std::decay_t<T>>>* = nullptr) noexcept {
-  return static_cast<T&&>(arg);
-}
-
-template <class T>
-constexpr T&&
-fmt_fwd(std::remove_reference_t<T>&& arg,
-        std::enable_if_t<is_formattable<std::decay_t<T>>>* = nullptr) noexcept {
-  static_assert(!std::is_lvalue_reference_v<T>);
-  return static_cast<T&&>(arg);
-}
-
-template <class T>
-std::string
-fmt_fwd(std::remove_reference_t<T>& arg,
-        std::enable_if_t<fmt_needs_to_string<std::decay_t<T>>>* = nullptr) {
-  return deep_to_string(arg);
-}
-
-template <class T>
-std::string
-fmt_fwd(std::remove_reference_t<T>&& t,
-        std::enable_if_t<fmt_needs_to_string<std::decay_t<T>>>* = nullptr) {
-  static_assert(!std::is_lvalue_reference_v<T>);
-  return deep_to_string(arg);
+decltype(auto) fmt_fwd(T&& arg) {
+  using arg_t = std::decay_t<T>;
+  if constexpr (std::is_default_constructible_v<std::formatter<arg_t, char>>) {
+    return std::forward<T>(arg);
+  } else {
+    static_assert(
+      !std::is_same_v<
+        decltype(inspect_access_type<stringification_inspector, arg_t>()),
+        inspector_access_type::none>,
+      "stringification using std::formatter or caf::inspect not available");
+    return deep_to_string(arg);
+  }
 }
 
 template <class OutputIt, class... Args>
-auto format_to(OutputIt out, std::format_string<Args...> fstr, Args&&... args) {
-  return std::format_to(out, fstr, fmt_fwd<Args>(args)...);
+auto format_to(OutputIt out, std::string_view fstr, Args&&... args) {
+  // Note: make_format_args expects all args to be references, and since
+  // fmt_fwd returns by value in case of inspector stringification, we need this
+  // helper function to wrap the call.
+  auto format_to_helper = [&out, &fstr](const auto&... ts) {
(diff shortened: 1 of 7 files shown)


riemass:
Seems adding the library `caf_internal` to `CMAKE_REQUIRED_LIBRARIES` won't carry the selected cpp standard, so I had to add a compile flag. The second issue was `check_cxx_source_runs` actually setting the `resultVar` to either 1 for true, or 0/empty string for false.

Neverlord (changes requested):
That flag setup won't work on MSVC.

More importantly, we can't fix this now without braking the ABI. If user has built CAF 1.0.0 with C++20 and then builds the version after this fix, it'll have different symbols now.

So while it's a bit unfortunate, we need to make this an opt-in thing now. Remove the auto-detection entirely and turn `CAF_USE_STD_FORMAT` into an option that defaults to `OFF`. We can revisit this in CAF 2.0.

riemass:
In the end this PR had multiple things that I needed to mangle with:

- Had to ditch compile time checked `std::format_args<Args...>` to `std::string_view`. I wanted to keep an overload of the format functions using these features, but since `fmt_fwd` sometimes changes the type from `T` to `std::string`, that wouldn't be possible,
- `format` to `vformat` for run-time checks. `make_format_args` requires references as args, so worked around that,
- `fmt_fwd` needed to be rewritten,
- formatter for `chunked_string` had a nit error,
- the `difference_type` in `colorized_iterator` needs to behave like a signed integer, for the iterator to satisfy the `output_iterator` concept. 

Talk about metaprogramming and hard-to-read compile errors...

libcaf_core/caf/detail/format.hpp
```diff
@@ -27,53 +27,40 @@
...
-
-template <class T>
-std::string
-fmt_fwd(std::remove_reference_t<T>&& t,
-        std::enable_if_t<fmt_needs_to_string<std::decay_t<T>>>* = nullptr) {
-  static_assert(!std::is_lvalue_reference_v<T>);
-  return deep_to_string(arg);
+decltype(auto) fmt_fwd(T&& arg) {
+  if constexpr (std::is_default_constructible_v<
+                  std::formatter<std::decay_t<T>, char>>) {
```
Neverlord, line 32: Why did you switch from `is_complete<std::formatter<T, char>>` to `is_default_constructible`?

Don't repeat yourself. We use `std::decay_t<T>` twice, give it an alias:

```cpp
  using arg_t = std::decay_t<T>;
```

libcaf_core/caf/detail/format.hpp
```diff
@@ -27,53 +27,40 @@
...
 template <class OutputIt, class... Args>
-auto format_to(OutputIt out, std::format_string<Args...> fstr, Args&&... args) {
-  return std::format_to(out, fstr, fmt_fwd<Args>(args)...);
+auto format_to(OutputIt out, std::string_view fstr, Args&&... args) {
+  // Note: make_format_args expects args all args to be references, and since
+  // fmt_fwd returns by value in case of inspector stringification, we need this
+  // helper function to wrap the call.
+  auto format_to_helper = [&out, &fstr]<class... Ts>(Ts&&... ts) {
+    return std::vformat_to(out, fstr, std::make_format_args(ts...));
+  };
```
Neverlord, line 51: Why the explicit templating?

```suggestion
  auto format_to_helper = [&out, &fstr](const auto&... ts) {
    return std::vformat_to(out, fstr, std::make_format_args(ts...));
  };
```

libcaf_core/caf/detail/format.hpp
```diff
@@ -27,53 +27,40 @@
...
+      "stringification using std::formatter or caf::inspect not available");
+    return deep_to_string(arg);
+  }
 }
 
 template <class OutputIt, class... Args>
-auto format_to(OutputIt out, std::format_string<Args...> fstr, Args&&... args) {
-  return std::format_to(out, fstr, fmt_fwd<Args>(args)...);
+auto format_to(OutputIt out, std::string_view fstr, Args&&... args) {
+  // Note: make_format_args expects args all args to be references, and since
```
Neverlord, line 46:
```suggestion
  // Note: make_format_args expects all args to be references, and since
```

libcaf_core/caf/detail/format.hpp
```diff
@@ -27,53 +27,40 @@
...
+    return std::vformat_to(out, fstr, std::make_format_args(ts...));
+  };
+  return format_to_helper(fmt_fwd(args)...);
 }
 
 template <class... Args>
-std::string format(std::format_string<Args...> fstr, Args&&... args) {
-  return std::format(fstr, fmt_fwd<Args>(args)...);
+std::string format(std::string_view fstr, Args&&... args) {
+  // Note: make_format_args expects args all args to be references, and since
```
Neverlord, line 57:
```suggestion
  // Note: make_format_args expects all args to be references, and since
```

libcaf_core/caf/detail/format.hpp
```diff
@@ -27,53 +27,40 @@
...
+  // fmt_fwd returns by value in case of inspector stringification, we need this
+  // helper function to wrap the call.
+  auto format_helper = [&fstr]<class... Ts>(Ts&&... ts) {
```
Neverlord, line 60:
```suggestion
  auto format_helper = [&fstr](const auto&... ts) {
```

Neverlord (changes requested):

libcaf_core/caf/detail/format.hpp
riemass: I had `fmt_fwd<Ts>(ts...)` in there at one moment, so I needed the template. Good catch.

libcaf_core/caf/detail/format.hpp
riemass: `is_complete` failed to compile for me. The class has a default template implementation, with an `static_assert(is_default_constructible<...>)` inside the class. This means the class is well defined, but because of the assert fails to compile. That is at least what I suspect, I can share the build log. Since the class uses `is_default_constructible` anyways, I'd leave it like this.

Neverlord (approved):
Good job, thanks for taking care of this mess! 👍

PR #1910, opened 2024-07-10
Fix python version in FreeBSD

riemass:
Seems `pip39` is no longer available in pkg. This PR tries to fix the python version to 3.10.


riemass:
Seems 3.11 is the new default.

PR #1911, opened 2024-07-11
Add new module for background workers

Neverlord:
A new piece of infrastructure for running background tasks. We have a couple of those, like the Prometheus exporter, and use one-off solutions instead of having a dedicated API.

The class uses the PIMPL idiom to allow us to experiment with the implementation. I've also added a convenience function to `client_factory` which could technically go into a separate PR, but I happened to need that to test the new daemons module with some background task doing HTTP.

### DIFF
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -420,7 +420,7 @@ public:
       modules[mod_id].reset(mod_ptr);
     }
     // Let there be daemons.
-    modules[actor_system_module::daemons].reset(new detail::daemons);
+    modules[actor_system_module::daemons].reset(new detail::daemons(*parent));
     // Make sure meta objects are loaded.
     auto gmos = detail::global_meta_objects();
     if (gmos.size() < id_block::core_module::end
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -11,8 +11,10 @@
 #include "caf/actor_registry.hpp"
 #include "caf/actor_system_config.hpp"
 #include "caf/defaults.hpp"
+#include "caf/detail/actor_system_access.hpp"
 #include "caf/detail/assert.hpp"
 #include "caf/detail/critical.hpp"
+#include "caf/detail/daemons.hpp"
 #include "caf/detail/meta_object.hpp"
 #include "caf/detail/private_thread_pool.hpp"
 #include "caf/detail/test_coordinator.hpp"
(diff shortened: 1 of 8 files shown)


libcaf_core/caf/detail/daemons.cpp
```diff
@@ -0,0 +1,96 @@
...
+  // nop
+}
+
+daemons::~daemons() {
+  delete impl_;
+}
+
+void daemons::cleanup(int id) {
+  std::lock_guard guard{impl_->mtx};
+  auto i = impl_->threads.find(id);
```
riemass, line 47: What if the `id` isn't found?

libcaf_core/caf/actor_system.cpp
```diff
@@ -417,6 +419,8 @@ class actor_system::impl {
       auto mod_id = mod_ptr->id();
       modules[mod_id].reset(mod_ptr);
     }
+    // Let there be daemons.
+    modules[actor_system_module::daemons].reset(new detail::daemons);
```
riemass, line 423: Any reason this is separate from the `module_factories` approach used for the rest?

libcaf_core/caf/detail/actor_system_access.hpp
```diff
@@ -28,6 +30,8 @@ class CAF_CORE_EXPORT actor_system_access {
 
   detail::mailbox_factory* mailbox_factory();
 
+  detail::daemons* daemons();
+
```
riemass, line 34: Doesn't adding this method break ABI?

riemass (changes requested):
While I do get the idea of the daemons, couldn't we have implemented this in terms of detached/blocking actors, and thus leverage the recycling of spawned threads?

libcaf_core/caf/detail/daemons.cpp
Neverlord: That'd be bad... 🙈
I'll fix it.

libcaf_core/caf/detail/actor_system_access.hpp
Neverlord: Adding functions is fine - except when they're virtual. Generally, anything that breaks memory layout (touch member variables or virtual member functions) or changes existing functions is an issue. Adding more stuff is fine. Of course you can't build against, say 1.2, and then expect your binary to work with 1.1. Older version simply might not have certain symbols. But anything >= 1.2 should work (as long it's the same major version).

riemass: OK, got a bit mixed up here, didn't know that non-virtual member functions are fine. Thanks.

riemass (approved):
Approved, thanks for considering the ideas. :smile:

libcaf_core/caf/actor_system.cpp
Neverlord: Ah, didn't comment on this one. Basically, we'll always load the daemons module it's not something that we need to be configurable.

PR #1921, opened 2024-07-20
Fix crash in run_actions when calling quit

### LINKED ISSUE #1920: Crash in `run_actions` when calling `quit`
Reported by a user via email. If `run_actions` executes an action that calls `self->quit()`, it will invoke `run_actions` again, causing CAF to try dereference a `nullptr`.

Neverlord:
Fixes #1920.


riemass (approved):
Good find. This might be related to the bugs I was hunting, I will try out with the fix.

PR #1930, opened 2024-07-26
Minor API additions for use cases at Interance

Neverlord:
- `splice` for intrusive lists, similar to `std::list::splice`, except that it will always insert at the end
- `awaits_external_event` in combination with the `blocked` flag on the mailbox gives us a way to detect actors that are currently completely idle
- `too_many_worker_failures` is a new error code meant for supervisors that automatically restart workers but terminate once there are too many errors in a given time frame


riemass (approved):
:+1:

PR #1931, opened 2024-07-29
Generate Doxyfile from CMake

Neverlord:
- get the CAF version from CMake
- use absolute paths to the sources

With this change, the Doxyfile will be placed into the build directory. Hence, Doxygen must either run from the build directory or now or needs to receive the path to the `Doxyfile`, e.g., `doxygen build/Doxyfile`.


shariarriday (approved):
It looks like a great idea to configure these properties of documentation with cmake.

Issue #1908, opened 2024-07-10
"http/client" example does not dispose of the connections after completing request


shariarriday:
The example completes with the following log. The socket is not disposed before the shutdown is initiated.

> [caf.net:DEBUG] 2024-07-10T12:23:46.110735200+06:00 initiate shutdown
> [caf.core:DEBUG] 2024-07-10T12:23:46.110694700+06:00 initialized actor: name = caf.system.config-server
> [caf.net:DEBUG] 2024-07-10T12:23:46.110747600+06:00 apply 0 updates
> [caf.core:DEBUG] 2024-07-10T12:23:46.110786300+06:00 actor has no behavior and is ready for cleanup
> [caf.core:DEBUG] 2024-07-10T12:23:46.110842800+06:00 actor has no behavior and is ready for cleanup
> [caf.net:DEBUG] 2024-07-10T12:23:46.110811500+06:00 apply 0 updates
> [caf.core:DEBUG] 2024-07-10T12:23:46.110797800+06:00 cleanup: id = 2, node = invalid-node, fail-state = user_shutdown
> [caf.core:DEBUG] 2024-07-10T12:23:46.110926900+06:00 actor finalized
> [caf.core:DEBUG] 2024-07-10T12:23:46.110859200+06:00 cleanup: id = 3, node = invalid-node, fail-state = user_shutdown
> [caf.net:ERROR] 2024-07-10T12:23:46.110935900+06:00 Response abortet with: disposed
> [caf.core:DEBUG] 2024-07-10T12:23:46.110979800+06:00 actor finalized
> [caf.net:DEBUG] 2024-07-10T12:23:46.111029900+06:00 close socket = 460
> [caf.net:DEBUG] 2024-07-10T12:23:46.111159700+06:00 apply 0 updates
> [caf.net:DEBUG] 2024-07-10T12:23:46.111192200+06:00 apply 0 updates
> [caf.net:DEBUG] 2024-07-10T12:23:46.111217+06:00 close socket = 440
> [caf.core:DEBUG] 2024-07-10T12:23:46.111518800+06:00 make_behavior() did return a valid behavior
> [caf.core:DEBUG] 2024-07-10T12:23:46.111539600+06:00 initialized actor: name = printer_actor
> [caf.core:DEBUG] 2024-07-10T12:23:46.111623200+06:00 actor has no behavior and is ready for cleanup
[...]

PR #1932, opened 2024-07-29
Rename shutdown to abort_and_shutdown in http::router

riemass:
Other uses for `abort` and `shutdown` are OK, this is the only fix. 
Relates #1908.

### DIFF
--- a/libcaf_net/caf/net/http/router.cpp
+++ b/libcaf_net/caf/net/http/router.cpp
@@ -61,8 +61,7 @@ request router::lift(responder&& res) {
 }
 
 void router::shutdown(const error& err) {
-  abort(err);
-  down_->shutdown(err);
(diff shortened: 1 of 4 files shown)


libcaf_net/caf/net/http/router.cpp
```diff
@@ -65,6 +65,11 @@ void router::shutdown(const error& err) {
   down_->shutdown(err);
```
Neverlord, line 65: This code is not DRY. 😉
Please call the new function from the old one.

Neverlord (changes requested):

libcaf_net/caf/net/http/router.cpp
riemass: :speak_no_evil:

Neverlord (approved):

Issue #1929, opened 2024-07-26
More broken Doxygen output


Neverlord:
Even after fixing #1916, looks like some things are still broken in the docs:

- the code examples are not formatted properly here: https://www.actor-framework.org//static/doxygen/1.0.0/group___blocking_a_p_i.html
- looks like Doxygen got really confused here: https://www.actor-framework.org//static/doxygen/1.0.0/group___actor_creation.html
- more broken formatting: https://www.actor-framework.org//static/doxygen/1.0.0/group___implicit_conversion.html
- things are definitely missing here: https://www.actor-framework.org//static/doxygen/1.0.0/group___message_handling.html
- the first line is needlessly repeated in all examples (e.g. https://www.actor-framework.org//static/doxygen/1.0.0/hello_world_8cpp-example.html)
- this is not a namespace: https://www.actor-framework.org//static/doxygen/1.0.0/namespacecaf_1_1exit__reason.html, weirdly the `enum class` is properly documented separately

And I've just poked around a bit, so there's probably more stuff that needs fixing.

PR #1933, opened 2024-07-30
Fix deprecated and format errors in doxygen

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/1929

All of the reported issues have been resolved except https://www.actor-framework.org//static/doxygen/1.0.0/group___message_handling.html . It looks like there are no additional documentation for message handling.


Neverlord:
When navigating to a class it will show what header to include. For example, navigating to `abstract_actor` will show `#include <abstract_actor.hpp>`. However, the path should be `#include <caf/abstract_actor.hpp>`. Can we fix the include path? Otherwise, I'd rather disable this setting than showing wrong include directives.

shariarriday:
> When navigating to a class it will show what header to include. For example, navigating to `abstract_actor` will show `#include <abstract_actor.hpp>`. However, the path should be `#include <caf/abstract_actor.hpp>`. Can we fix the include path? Otherwise, I'd rather disable this setting than showing wrong include directives.

I am looking into this. If I can not find any acceptable solution then we may need to turn `SHOW_HEADERFILE` off.

Neverlord (approved):

PR #1941, opened 2024-09-01
Add retry flow operator

shariarriday:

Adds retry operator that retries only when the predicate returns true. The predicate should accept one argument of type caf::error.

### DIFF
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -31,6 +31,7 @@
 #include "caf/flow/op/on_backpressure_buffer.hpp"
 #include "caf/flow/op/prefix_and_tail.hpp"
 #include "caf/flow/op/publish.hpp"
+#include "caf/flow/op/retry.hpp"
 #include "caf/flow/op/sample.hpp"
 #include "caf/flow/op/zip_with.hpp"
 #include "caf/flow/step/all.hpp"
@@ -295,6 +296,12 @@ public:
     using val_t = output_type;
     static_assert(std::is_invocable_r_v<Init, Scanner, Init&&, const val_t&>);
     return add_step(step::scan<Scanner>{std::move(init), std::move(scanner)});
+  }
+
+  /// @copydoc observable::retry
+  template <class Predicate>
+  auto retry(Predicate predicate) {
+    return materialize().retry(predicate);
   }
 
   auto sum() && {
@@ -807,6 +814,13 @@ observable<T> observable<T>::sample(timespan period) {
   auto obs = pptr->add_child_hdl(std::in_place_type<op::interval>, period,
                                  period);
   return pptr->add_child_hdl(std::in_place_type<impl_t>, *this, std::move(obs));
+}
+
+template <class T>
+template <class Predicate>
+observable<T> observable<T>::retry(Predicate predicate) {
+  using impl_t = op::retry<T, Predicate>;
+  return parent()->add_child_hdl(std::in_place_type<impl_t>, *this, predicate);
 }
 
 // -- observable: combining ----------------------------------------------------
--- a/libcaf_core/caf/flow/observable_decl.hpp
+++ b/libcaf_core/caf/flow/observable_decl.hpp
@@ -215,6 +215,11 @@ public:
   /// Emits the most recent item of the input observable once per interval.
   observable<T> sample(timespan period);
 
+  /// Re-subscribes to the input observable on error for as long as the
+  /// predicate returns true.
+  template <class Predicate>
+  observable<T> retry(Predicate predicate);
+
   // -- combining --------------------------------------------------------------
 
   /// Combines the output of multiple @ref observable objects into one by
--- a/libcaf_core/caf/flow/op/retry.hpp
+++ b/libcaf_core/caf/flow/op/retry.hpp
@@ -6,12 +6,11 @@
 
 #include "caf/config.hpp"
 #include "caf/detail/assert.hpp"
-#include "caf/flow/backpressure_overflow_strategy.hpp"
 #include "caf/flow/observer.hpp"
 #include "caf/flow/op/hot.hpp"
 #include "caf/flow/subscription.hpp"
 
-#include <deque>
+#include <type_traits>
 #include <utility>
 
 namespace caf::flow::op {
@@ -21,8 +20,8 @@ class retry_sub : public subscription::impl_base, public observer_impl<T> {
 public:
   // -- constructors, destructors, and assignment operators --------------------
 
-  retry_sub(coordinator* parent, observer<T> out, Predicate predicate,
-            observable<T> in)
+  retry_sub(coordinator* parent, observer<T> out, observable<T> in,
+            Predicate predicate)
     : parent_(parent),
       out_(std::move(out)),
       in_(std::move(in)),
@@ -60,8 +59,8 @@ public:
 
   void on_subscribe(subscription sub) override {
     if (sub_) {
-      sub_.cancel();
-      sub_.release_later();
+      sub.cancel();
+      return;
     }
     sub_ = std::move(sub);
     if (demand_ > 0)
@@ -69,12 +68,10 @@ public:
   }
 
   void on_next(const T& item) override {
-    if (!out_)
+    if (!out_ || demand_ == 0)
       return;
-    if (demand_ > 0) {
-      --demand_;
-      out_.on_next(item);
-    }
+    --demand_;
+    out_.on_next(item);
   }
 
   void on_complete() override {
@@ -88,10 +85,11 @@ public:
     if (!out_)
       return;
     sub_.release_later();
-    if (predicate_(what))
-      parent()->delay_fn([this]() { in_.subscribe(this->as_observer()); });
-    else
+    if (predicate_(what)) {
+      parent_->delay_fn([sptr = intrusive_ptr{this}] { sptr->do_retry(); });
+    } else {
       out_.on_error(what);
+    }
   }
 
 private:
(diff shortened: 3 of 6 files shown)


libcaf_core/caf/flow/observable_decl.hpp
```diff
@@ -215,6 +215,10 @@ class observable {
   /// Emits the most recent item of the input observable once per interval.
   observable<T> sample(timespan period);
 
+  /// Retries input observable on error when predicate returns true.
```
Neverlord, line 218:
```suggestion
  /// Re-subscribes to the input observable on error for as long as the predicate returns true.
```

libcaf_core/caf/flow/op/retry.hpp
```diff
@@ -0,0 +1,172 @@
...
+    deref();
+  }
+
+  void on_subscribe(subscription sub) override {
+    if (sub_) {
+      sub_.cancel();
+      sub_.release_later();
+    }
+    sub_ = std::move(sub);
+    sub_.request(256);
```
Neverlord, line 65: The bookkeeping of the demand looks broken to me. Our subscriber communicates demand to use, that we should forward to `sub_`. If `sub_` goes down and re-subscribe, the existing demand should be passed over. There's also a `sub_.request(1);` further down below that I think should not exist.

The subscriber should drive demand. The `retry` operator should not hand out demand on its own, since it's not buffering items.

Neverlord (changes requested):

libcaf_core/caf/flow/observable.hpp
```diff
@@ -809,6 +816,14 @@ observable<T> observable<T>::sample(timespan period) {
   return pptr->add_child_hdl(std::in_place_type<impl_t>, *this, std::move(obs));
 }
 
+template <class T>
+template <class Predicate>
+observable<T> observable<T>::retry(Predicate predicate) {
+  using impl_t = op::retry<T, Predicate>;
+  auto* pptr = parent();
+  return pptr->add_child_hdl(std::in_place_type<impl_t>, *this, predicate);
```
Neverlord, line 824:
```suggestion
  return parent()->add_child_hdl(std::in_place_type<impl_t>, *this, std::move(predicate));
```

libcaf_core/caf/flow/op/retry.hpp
```diff
@@ -0,0 +1,171 @@
...
+    if (!out_)
+      return;
+    if (demand_ > 0) {
+      --demand_;
+      out_.on_next(item);
+      if (sub_)
+        sub_.request(1);
+      return;
+    }
+  }
```
Neverlord, line 78: The operator must not proactively hand out demand. The observer must drive demand, not this operator.

Calling `sub_.request` should (only) be done in two cases:
- `request` is called and `sub_` is valid
- `on_subscribe` is called and `demand_` is > 0

Neverlord (changes requested):

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+  auto retry_hdl = [&retry_count](const caf::error& what) {
+    if (what == sec::invalid_request && retry_count > 0) {
+      --retry_count;
+      return true;
+    } else
+      return false;
+  };
+  GIVEN("an observable that always fails") {
+    WHEN("calling retry") {
+      THEN("the observer receives error after retrying 3 times") {
```
Neverlord, line 43:
```suggestion
      THEN("the observer receives an error after retrying 3 times") {
```

More importantly, nothing in this test checks whether the retries actually happened or not.

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+                                   in);
+    out.on_subscribe(caf::flow::subscription{ptr});
+    return ptr;
+  }
+};
+
+WITH_FIXTURE(fixture) {
+
+SCENARIO("the retry operator retries observable multiple times") {
+  int retry_count = 3;
```
Neverlord, line 33: Counting down seems odd, since the variable doesn't actually count the retries (despite its name). I'd initialize with 0 and add a `max_retry_count` constant. By actually counting up, we can also use the `retry_count` for the tests below.

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        uut->on_complete();
+        run_flows();
+        check_eq(snk->buf, std::vector<int>{1, 2, 1, 1, 2});
+        check(snk->completed());
+        uut->on_next(3);
+        uut->on_next(4);
+        uut->on_complete();
+        run_flows();
+        check_eq(snk->buf, std::vector<int>{1, 2, 1, 1, 2});
+        check(snk->completed());
```
Neverlord, line 130: Calling `on_next` after `on_complete` has nothing to do with the retry logic we are testing here. Discard. Could be its own separate test, but does not belong here.

```suggestion
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        uut->on_next(1);
+        uut->on_complete();
+        run_flows();
+        check_eq(snk->buf, std::vector<int>{1, 2, 1});
+        check(snk->completed());
+      }
+    }
+    WHEN("calling on_error with runtime_error twice") {
+      THEN(
+        "flow is completed without forwarding error and emitting item stops") {
```
Neverlord, line 110:
```suggestion
      THEN("the flow completes without error") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        uut->on_next(3);
+        uut->on_next(4);
+        uut->on_complete();
+        run_flows();
+        check_eq(snk->buf, std::vector<int>{1, 2, 1, 1, 2});
+        check(snk->completed());
+      }
+    }
+    WHEN("calling on_error on operator with unexpected_message") {
+      THEN("flow is completed after forwarding error and emitting item stops") {
```
Neverlord, line 134:
```suggestion
      THEN("the flow aborts with an error") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+SCENARIO("disposing a retry operator completes the flow") {
+  auto retry_hdl = [](const caf::error& what) {
+    if (what == sec::runtime_error)
+      return true;
+    else
+      return false;
+  };
+  GIVEN("a retry operator") {
```
Neverlord, line 159:
```suggestion
  GIVEN("an observable that is retried on `runtime_error`") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+}
+
+SCENARIO("disposing a retry operator completes the flow") {
+  auto retry_hdl = [](const caf::error& what) {
+    if (what == sec::runtime_error)
+      return true;
+    else
+      return false;
```
Neverlord, line 157:
```suggestion
    return what == sec::runtime_error;
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        run_flows();
+        uut->dispose();
+        run_flows();
+        check(snk->aborted());
+      }
+    }
+  }
+}
+
+SCENARIO("retry operator stops emitting item when there is no demand") {
```
Neverlord, line 175:
```suggestion
SCENARIO("retry operators discard items when there is no demand") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        uut->on_error(sec::unexpected_message);
+        run_flows();
+        check_eq(snk->buf, std::vector<int>{1, 2});
+        check(snk->aborted());
+      }
+    }
+  }
+}
+
+SCENARIO("disposing a retry operator completes the flow") {
```
Neverlord, line 152:
```suggestion
SCENARIO("disposing a retry operator aborts the flow") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+        run_flows();
+        check_eq(outputs->size(), 4u);
+        check_eq(*err, error{}); // no error
+        check(*completed);
+      }
+    }
+  }
+}
+
+SCENARIO("retry can complete flow despite multiple errors") {
```
Neverlord, line 81:
```suggestion
SCENARIO("the retry operator can complete flows despite errors") {
```

libcaf_core/caf/flow/op/retry.test.cpp
```diff
@@ -0,0 +1,197 @@
...
+    auto ptr = make_counted<sub_t>(coordinator(), out, std::move(predicate),
+                                   in);
+    out.on_subscribe(caf::flow::subscription{ptr});
+    return ptr;
+  }
+};
+
+WITH_FIXTURE(fixture) {
+
+SCENARIO("the retry operator retries observable multiple times") {
```
Neverlord, line 32:
```suggestion
SCENARIO("the retry operator re-subscribes observables on error") {
```

libcaf_core/caf/flow/op/retry.test.cpp
Neverlord, line 1: Nothing seems to test the demand handling. At least, I'd expect a test that checks whether the retry operator forwards existing demand to the input observable on retry.

libcaf_core/caf/flow/op/retry.test.cpp
shariarriday: I have added a separate test to check that the items are not emitted after the flow is completed.

libcaf_core/caf/flow/op/retry.test.cpp
shariarriday: I had previously tested if the retries are happening by counting the elements in output. I have added more explicit tests to check for retries now.

libcaf_core/caf/flow/op/retry.hpp
```diff
@@ -0,0 +1,171 @@
...
+    if (!out_)
+      return;
+    sub_.release_later();
+    out_.on_complete();
+  }
+
+  void on_error(const error& what) override {
+    if (!out_)
+      return;
+    if (predicate(what))
```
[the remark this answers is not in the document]
shariarriday: Thanks for pointing it out.

Neverlord (approved):

PR #1942, opened 2024-09-06
Always install a default logger when running tests 

Neverlord:
When using the deterministic fixture, we install a logger that forwards all messages to the reporter. However, when not using the fixture, logs are silently discarded because no logger exists. Since we offer a log component specifically for testing, e.g. via `log::test::debug`, this is not what a user can reasonably expect.

This patch moves the logger implementation from the deterministic fixture to the reporter and always installs an instance of that logger as default.

### DIFF
--- a/libcaf_test/caf/test/reporter.cpp
+++ b/libcaf_test/caf/test/reporter.cpp
@@ -557,4 +557,84 @@ std::unique_ptr<reporter> reporter::make_default() {
   return std::make_unique<default_reporter>();
 }
 
+namespace {
+
+/// A logger implementation that delegates to the test reporter.
+class reporter_logger : public logger, public detail::atomic_ref_counted {
+public:
+  /// Increases the reference count of the coordinated.
+  void ref_logger() const noexcept final {
+    this->ref();
+  }
+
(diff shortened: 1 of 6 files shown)


libcaf_test/caf/test/reporter.cpp
```diff
@@ -557,4 +557,84 @@ std::unique_ptr<reporter> reporter::make_default() {
...
+
+  // -- logging ----------------------------------------------------------------
+
+  /// Writes an entry to the event-queue of the logger.
+  /// @thread-safe
+  void do_log(log::event_ptr&& event) override {
+    // We omit fields such as component and actor ID. When not filtering
+    // non-test log messages, we add these fields to the message in order to be
+    // able to distinguish between different actors and components.
+    if (event->component() != "caf.test") {
```
shariarriday, line 590: Shouldn't this check rely on the return data of  `accepts()`?

Neverlord: `accepts` is called prior to `do_log`, i.e., `do_log` is not called unless `accept` returns `true`.

shariarriday (approved):

Issue #1953, opened 2024-10-23
Strings are appended instead of replaced when deserializing. Is this intended?


odoriath-neo:
The easiest way to reproduce this is to add a default value to `std::string name` in examples/config/read-json.cpp:
```
struct user {
  uint32_t id;
  std::string name = "Paul ";
  std::optional<std::string> email;
};
```
Instead of replacing the provided name during apply(), it will append it and the result will look like this:
```
Entries loaded from file:
- user(1, "Paul John Doe", null)
- user(2, "Paul Jane Doe", *"jane@doe.com")
```
instead of the expected
```
Entries loaded from file:
- user(1, "John Doe", null)
- user(2, "Jane Doe", *"jane@doe.com")
```
Tested in main and 0.19.6

Neverlord:
That's not intended. The code simply assumes that the default constructor leaves all members "empty", but that is probably a naïve assumption to make.

Neverlord:
I can no longer reproduce this in current `main`. There has been one change to the `json_reader` in https://github.com/actor-framework/actor-framework/pull/1949 that might have fixed this issue as a side-effect (https://github.com/actor-framework/actor-framework/pull/1949/files#diff-c381c2891f6e22ea1661529de85a3164cd5ad9eb03faf2152bc106f00392ca29L649). So it should no longer happen as of 1.0.2. @odoriath-neo mind double-checking on your before I close the ticket as resolved?

Neverlord:
Has a regression test now and should be fixed as of 1.0.2.

PR #1959, opened 2024-11-07
Add regression test for #1953

Neverlord:


riemass (approved):
:+1: 
Nice. `print_unescaped` didn't clear the buffer, definitely the root cause. Hope no one depends on the old behavior :sweat_smile:.

PR #1962, opened 2024-11-19
Send close-frame on WebSocket abort

### LINKED ISSUE #1938: missing close-frame when shutting down websocket server
When sending a `kill SIGTERM` to the [websocket example](https://github.com/interance/warehouse-backend-example/tree/main) currently open websocket connections do not receive a close-frame.

Changing the [consume of frames](https://github.com/interance/warehouse-backend-example/blob/main/warehouse-backend-example/main.cpp#L80) to 

```
auto disposable_push = std::make_shared<caf::disposable>();
pull.observe_on(self)
            .do_finally([] { applog::info("WebSocket client disconnected"); })
            .for_each([disposable_push](const frame& in) mutable {
              applog::info("In: {}", in.as_text());
              disposable_push->dispose();
            });
```

and [line 87](https://github.com/interance/warehouse-backend-example/blob/main/warehouse-backend-example/main.cpp#L87) to
```
*disposable_push = events.observe_on(self)
```
then the connection receives a close-frame when sending some data. This seem to work only with the `producer`, 
doing the same with the consumer doesn't generate a close-frame

riemass:
Closes #1938. 
We didn't ship a close frame on `websocket::framing::abort` (we did on `shutdown` but that's another story). This PR adds the missing frame. 
[According to this section in the RFC](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.1)  the server closing the connection will send an close frame without a status code. The code will then be interpreted as 1005 by the other side. Our side can set another status code, but it's illegal to manually insert 1005 and 1006 according to [section 7.4](https://datatracker.ietf.org/doc/html/rfc6455#section-7.4). 
Also, as far as I've understood the RFC, the mask key should be present even if the payload is empty. Any message send by the client without the mask bit set will result in error, and the mask bit implies the mask key value in the frame. 

Also, there is no way to validate the closing frame in the `websocket robot` implementation. The autobahn testsuite doesn't check this, otherwise it would've shown up by now. I don't know if I should bring back the framing unit tests for this issue?


libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -116,6 +116,7 @@ class framing_impl : public framing {
   }
 
   void abort(const error& reason) override {
+    ship_closing_message();
```
Neverlord, line 119: You gave some really good explanation why we sent that here without an explicit code in the PR description. I'd love to see that explanation here as code documentation. 🙂

Neverlord (changes requested):

Neverlord:
> I don't know if I should bring back the framing unit tests for this issue?

Hm, lack of testing definitely is an issue. But I don't think unit tests are the right tool. A Robot test might be better suited if we can check whether a close frame was sent or not. Can be a followup, though.

riemass:
Seems like robot can't gracefully shut down the process under windows.

PR #1964, opened 2024-11-24
Avoid clang-tidy false positives for intrusive_ptr

Neverlord:
This is an interesting one that came out of a customer code base. `clang-tidy` would complain about a use-after-free in `intrusive_ptr` in some nested code constructs, claiming that `reset` would free a pointer only for the destructor to accessing the pointer again (passing it to `intrusive_ptr_access<T>::release`). That's of course not possible, because we re-assign `ptr_` in `release` after releasing the previous pointer. However, somehow `clang-tidy` did not make that connection. That was with `clang-tidy` from LLVM version 19.1.3. It also seemed to be a bit confused about `detach` in one instance.

Using the swap-with-temporary idiom in `reset` seems to be easier to swallow for `clang-tidy`. No functional change either way, just refactoring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 ### Fixed
 
 - Fix build error in `caf-net` when building with C++23 (#1919).
+- Restructure some implementation details of `intrsuive_ptr` (no functional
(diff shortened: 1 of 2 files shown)


riemass (approved):
That is a recent release, so it's good to get it out of the way with a rather simple fix. `clang-tidy` has a lot of context to track so false positives aren't surprising.

PR #1969, opened 2024-12-05
Support noncopyable functors in `scheduled_actor::run_*`

dominiklohmann:
This change makes it possible to use `scheduled_actor::run_delayed_weak` (and friends) with noncopyable callbacks.

### DIFF
--- a/libcaf_core/caf/scheduled_actor.hpp
+++ b/libcaf_core/caf/scheduled_actor.hpp
@@ -556,19 +556,22 @@ public:
   /// @returns A @ref disposable that allows the actor to cancel the action.
   template <class Duration, class F>
   disposable run_scheduled(
-    std::chrono::time_point<std::chrono::system_clock, Duration> when, F what) {
+    std::chrono::time_point<std::chrono::system_clock, Duration> when,
+    F&& what) {
     using std::chrono::time_point_cast;
-    return run_scheduled(time_point_cast<timespan>(when), make_action(what));
+    return run_scheduled(time_point_cast<timespan>(when),
+                         make_action(std::forward<F>(what)));
   }
 
   /// @copydoc run_scheduled
   template <class Duration, class F>
   disposable
   run_scheduled(std::chrono::time_point<actor_clock::clock_type, Duration> when,
-                F what) {
+                F&& what) {
     using std::chrono::time_point_cast;
     using duration_t = actor_clock::duration_type;
-    return run_scheduled(time_point_cast<duration_t>(when), make_action(what));
+    return run_scheduled(time_point_cast<duration_t>(when),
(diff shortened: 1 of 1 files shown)


Neverlord (changes requested):
Thanks for the patch!

As it stands, this patch changes function signatures, which in turn affects the ABI. I think using perfect forwarding is the right call in general, however, I would rather err on side of caution and use `std::move` internally instead of changing the signature. I think it's quite unlikely that we would actually break something here, but better safe than sorry.

The new functionality also needs tests (or at least a static assert in the existing tests) for move-only callback types. Not so much for testing this small change, but to have regression tests in place.

dominiklohmann:
I can make that change, but honestly I think that the policy you're setting is too strict, as it will require almost every change to wait on a major release. It's a template function in a header that in practice will always be inlined, and the underlying API that takes an action does not change in this PR—for that, the ABI change would be problematic.

Note that if I just move instead of forward I also need to add a `static_assert(!std::is_lvalue_reference_v<F> or std::is_const_v<F>)`. That's not an ABI break, but an API break, for the unlikely scenario that someone specified `F` as a non-const lvalue reference explicitly.

Anyways, just let me know which of the two you prefer and I'll push the change.

Neverlord:
> the policy you're setting is too strict

ABI certainly isn't a fun topic. FWIW, we've already revamped quite a bit of internal workings to give us more leeway for changes. In this particular case, I think value-then-move is just as valid as universal-ref-then-forward.

> static_assert(!std::is_lvalue_reference_v<F> or std::is_const_v<F>)

I get where you're coming from, but explicitly defining `F` is not the intended use. I wouldn't even have thought about shielding against misuse like that. The intended use is passing a value, so I think this can be just `static_assert(!is_reference_v<F>)`. Only breaks silly code to begin with with a trivial and obvious fix.

Neverlord:
> this will always get a different mangled name

Yes, that's precisely the ABI break because symbols that used to be there no longer exist and you need to recompile client code to use the new symbol names. It's probably very unlikely for this set of functions, but we can just use value-then-move parameters to not affect the ABI in the first place.

Neverlord:
Closing in favor of #1976.

dominiklohmann:
Thanks for taking this one over—I didn't yet have the time to get back to it, unfortunately.

Neverlord:
No worries, most of it was just writing some new unit tests and holiday season is coming up, so I didn't want to let this sit until next year. 🙂

dominiklohmann:
Precisely. The usual end of year rush to get things done before most people are off for a week or two in full effect for us, making it hard to find time for this in between.

Have a merry Christmas and a wonderful new year! 🎄🎆

PR #1979, opened 2024-12-29
Add tests for bool vector and fix implementation

shariarriday:
I have added a serialization/deserialization test for binary serializer. 

But, the previous implementation had a bug which caused the specialized implementation to not be executed. Due to the nature of the bug, I could not find any solution circumventing the ABI breaking changes. 
I think that we can plan to merge it in for 2.0. I am creating a draft pull request for this reason.

### DIFF
--- a/libcaf_core/caf/serialization.test.cpp
+++ b/libcaf_core/caf/serialization.test.cpp
@@ -465,13 +465,13 @@ struct fixture : caf::test::fixture::deterministic {
                       from);
   }
 
-  std::variant<int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t,
-               uint64_t, double, long double, std::string, std::u16string,
-               std::u32string, std::vector<int32_t>, std::list<int32_t>,
-               std::map<std::string, int32_t>,
-               std::unordered_map<std::string, int32_t>, std::set<int32_t>,
-               std::unordered_set<int32_t>, std::array<int32_t, 5>,
-               std::tuple<int32_t, std::string, int32_t>, c_array>
+  std::variant<
+    int8_t, int16_t, int32_t, int64_t, uint8_t, uint16_t, uint32_t, uint64_t,
+    double, long double, float, std::string, std::u16string, std::u32string,
+    std::vector<int32_t>, std::vector<bool>, std::list<int32_t>,
+    std::map<std::string, int32_t>, std::unordered_map<std::string, int32_t>,
+    std::set<int32_t>, std::unordered_set<int32_t>, std::array<int32_t, 5>,
+    std::tuple<int32_t, std::string, int32_t>, c_array>
   read_val(const std::string& type, const std::string& value) {
     str_list parsed_value;
     if (type == "i8") {
@@ -501,6 +501,9 @@ struct fixture : caf::test::fixture::deterministic {
     if (type == "ld") {
       return static_cast<long double>(std::stold(value));
     }
+    if (type == "f") {
+      return static_cast<float>(std::stof(value));
+    }
     if (type == "real") {
       return std::stod(value);
     }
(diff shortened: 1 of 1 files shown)


libcaf_core/caf/serialization.test.cpp
```diff
@@ -729,11 +746,13 @@ OUTLINE("serializing and then deserializing primitive values") {
     | binary_serializer   | u32    | 123456        |
     | binary_serializer   | u64    | 123456789     |
     | binary_serializer   | ld     | 123.5         |
+    | binary_serializer   | f      | 3.14          |
     | binary_serializer   | real   | 12.5          |
     | binary_serializer   | string | Hello, world! |
     | binary_serializer   | u16str | Hello, world! |
     | binary_serializer   | u32str | Hello, world! |
     | binary_serializer   | vector | 1, 42, -31    |
+    | binary_serializer   | vector_bool | true,true,false,false,true,true,true,false |
```
Neverlord, line 755: That's a bit unwindy, abbreviating it indents nicer and is less verbose:

```suggestion
    | binary_serializer   | vbool | true,true,false,false,true,true,true,false |
```

Is there a particular reason why this has 8 values? Could just have two to avoid this overly long line.

Neverlord (changes requested):
It's only an ABI break because it removes the `value(std::vector<bool>& x)` overload. Bring that back and simply call that from the new `list(value(std::vector<bool>& x)` overload. Then we can merge it right away.

libcaf_core/caf/serialization.test.cpp
shariarriday: The serializer and deserializer use a switch case to serialize/deserialize an array into a single byte. I have added 8 entries to ensure that all conditions are being handled properly. Please let me know if it will be okay to skip it.

Neverlord (changes requested):
Sorry, I've realized that we cannot pull this in now. It's not an ABI break, but it changes the wire format. We don't want to change our data representation on the network in a minor release.

Please split this up into one PR for the `serialization.test.cpp` additions and one PR for the fix (that will go into a v2 branch).

shariarriday:
> Sorry, I've realized that we cannot pull this in now. It's not an ABI break, but it changes the wire format. We don't want to change our data representation on the network in a minor release.
> 
> Please split this up into one PR for the `serialization.test.cpp` additions and one PR for the fix (that will go into a v2 branch).

I am not sure splitting up the tests would be beneficial. The added test will not hit the specialization for serializing boolean vectors and it will just execute the code of serializing any vectors. Maybe it will be best to include the test and code changes in a single PR for 2.x . Please let me know if you think otherwise.

Neverlord:
I think having the tests separately pulled in now is still worth it. The non-binary inspectors do a have a separate code path for `vector<bool>` that currently isn't tested. Also:

> then add a separate test for the binary serialization that consumes bool vectors of various sizes (at the very least the edge cases, i.e., 0, 1, 8, 9).

This test should still go in. For now, this test can call the `value(...)` overload for `vector<bool>` to get the coverage on that code path, ideally with a TODO reminder to use `apply` instead once we pull in the other change.

libcaf_core/caf/serialization.test.cpp
```diff
@@ -861,6 +884,55 @@ OUTLINE("serializing a type that initializes members to a non-empty state") {
...
+SCENARIO("binary serializer and deserializer handle vectors of booleans") {
+  GIVEN("a binary serializer") {
+    auto sink = binary_serializer_wrapper{sys};
+    WHEN("serializing a vector of booleans with 8 values") {
+      auto val = std::vector<bool>{true,  false, true, true,
+                                   false, false, true, false};
+      check(sink.sink.value(val));
+      THEN("deserializing the result produces the value again") {
+        auto source = binary_deserializer{sys, sink.buffer};
+        std::vector<bool> copy = {};
```
Neverlord, line 897: For keeping a consistent style:

```suggestion
        auto copy = std::vector<bool>{};
```

Same in the occurrences below.

Neverlord (changes requested):
Just one style nitpick. 👍

Issue #1896, opened 2024-06-20
configuration options for length field of length prefix framing


weberdaniel:
Some length prefixed protocols don't write the payload size into the length field of a frame, but use the length field to denote the total size of the frame (either payload_size+size_of_length_field or just payload_size). For convenience the implementation of a configuration option for this might be considered. While  it is possible to solve this issue by implementing a custom framing class by e.g. using copy/paste, deriving from the framing class and then altering the constraints of the base class in the derived class, might result in a LSP violation according to "C++ Software Design" (Klaus Iglberger).

Neverlord:
Certainly makes sense to have this configurable. An option to configure the size of the length field might be a good idea as well while we're at it. Currently, we always assume a 32-bit field. Not sure if we can still squeeze this in for 1.0, but we'll have it on the radar.

weberdaniel:
The configuration option might be convenient for people who try it out for the first time. Most space communication protocols are standardized by CCSDS (https://public.ccsds.org/). Examples would be the TC Space Data Link Protocol (CCSDS 232.0-B-4) or TM Space Data Link Protocol (CCSDS 132.0-B-3 ). Most have a fixed size header with the length field on an arbitrary position within the header. Some also support additional secondary headers to be activated or not (for example in telemetry). In such specialized cases customizing/adopting the length-prefix framing case is probably a necessity anyway. That being said, sometimes it may be necessary to convert data units from TCP/IP to UDP/IP especially for robotic real-time applications.

dominiklohmann:
I tried using the length prefix framing for size-prefixed FlatBuffers, but ran into the issue that the length is currently assumed to always be 32 bit instead of 64 bit. Are there any plans to make this configurable?

Neverlord:
> Are there any plans to make this configurable?

Yes. However, making the size configurable will break the ABI because we need to change `config_type` (and thus the type of the member variable) in `caf/net/lp/with.hpp`. So the next window for this feature is 2.0.

weberdaniel:
thanks for the help, i actually wanted to share some positive experiences with CAF 0.19.3 on a conference, but workload is too high.

PR #1983, opened 2025-01-13
Update lp framing to support max message size

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/1896

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ is based on [Keep a Changelog](https://keepachangelog.com).
   `caf.metrics.filters`. This change was made to make the configuration options
   more consistent.
 - The default maximum message size for the length-prefix framing has been
-  reduced to 60 MB. This change was made to improve security by default.
+  reduced to 64 MB. This change was made to improve security by default.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,15 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 - The configuration options under `caf.metrics-filters` have been moved to
   `caf.metrics.filters`. This change was made to make the configuration options
   more consistent.
+- The default maximum message size for the length-prefix framing has been
+  reduced to 60 MB. This change was made to improve security by default.
 
 ### Added
 
 - Added `monitor` API to WebSocket and HTTP servers in the `with` DSL (#2026).
+- When starting a server or client using length-prefix framing, users can now
+  specify the maximum message size via `max_message_size` and the number of
+  bytes for the length prefix via `size_field`.
 
 ### Fixed
 
--- a/examples/length_prefix_framing/chat-client.cpp
+++ b/examples/length_prefix_framing/chat-client.cpp
@@ -27,6 +27,9 @@ namespace ssl = caf::net::ssl;
 
 static constexpr uint16_t default_port = 7788;
 
+static constexpr caf::net::lp::size_field_type default_size
+  = caf::net::lp::size_field_type::u4;
+
 static constexpr std::string_view default_host = "localhost";
 
 static constexpr std::string_view default_name = "";
@@ -37,8 +40,11 @@ struct config : caf::actor_system_config {
   config() {
     opt_group{custom_options_, "global"} //
       .add<uint16_t>("port,p", "port of the server")
+      .add<caf::net::lp::size_field_type>("size,s",
+                                          "length prefix size of the server")
       .add<std::string>("host,H", "host of the server")
-      .add<std::string>("name,n", "set name");
+      .add<std::string>("name,n", "set name")
+      .add<std::string>("size,s", "set size of lp framing");
     opt_group{custom_options_, "tls"} //
       .add<bool>("enable", "enables encryption via TLS")
       .add<std::string>("ca-file", "CA file for trusted servers");
@@ -49,6 +55,7 @@ struct config : caf::actor_system_config {
     caf::put_missing(result, "port", default_port);
     caf::put_missing(result, "host", default_host);
     caf::put_missing(result, "name", default_name);
+    caf::put_missing(result, "size", default_size);
     return result;
   }
 };
@@ -58,6 +65,7 @@ struct config : caf::actor_system_config {
 int caf_main(caf::actor_system& sys, const config& cfg) {
   // Read the configuration.
   auto port = caf::get_or(cfg, "port", default_port);
+  auto size = caf::get_or(cfg, "size", default_size);
   auto host = caf::get_or(cfg, "host", default_host);
   auto name = caf::get_or(cfg, "name", default_name);
   auto use_ssl = caf::get_or(cfg, "tls.enable", false);
@@ -76,7 +84,7 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
                    .and_then(ssl::emplace_client(ssl::tls::v1_2))
                    .and_then(ssl::load_verify_file_if(ca_file)))
         // Connect to "$host:$port".
-        .connect(host, port)
+        .connect(host, port, size)
         // If we don't succeed at first, try up to 10 times with 1s delay.
         .retry_delay(1s)
         .max_retry_count(9)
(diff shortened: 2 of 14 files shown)


examples/length_prefix_framing/chat-client.cpp
```diff
@@ -69,7 +69,7 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
   auto [line_producer, line_pull]
     = caf::async::make_blocking_producer<caf::chunk>();
   auto conn
-    = caf::net::lp::with(sys)
+    = caf::net::lp::with(sys, INT32_MAX)
```
Neverlord, line 72: 32-bit length should be default, i.e., no changes should be necessary.

libcaf_net/caf/net/dsl/lp_custom_config.hpp
Neverlord, line 1: This header needs to go and the config type should be opaque. Please look at the WebSocket implementation as an example (e.g., `web_socket/client_factory.(hc)pp`)

libcaf_net/caf/net/lp/client_factory.cpp
```diff
@@ -33,6 +33,13 @@ class client_factory::config_impl : public dsl::client_config_value {
   using super = dsl::client_config_value;
 
   using super::super;
+
+  explicit config_impl(multiplexer* mpx, size_t max_message_length)
```
Neverlord, line 37: Rather than configuring the length, the ticket was about selecting the integer type on the wire, i.e., 1, 2, 4 or 8 byte.

This also shouldn't be an integer but an enum type in order to make it clear what options are available to the user, e.g.:

```c++
/// Configures the integer type used as the length prefix.
enum class size_field_type {
  /// Configures a length prefix with a single byte.
  u1,
  // Configures a length prefix with two bytes.
  u2,
  // Configures a length prefix with four bytes.
  u4,
  // Configures a length prefix with eight bytes.
  u8,
};
```

Neverlord (changes requested):

libcaf_net/caf/net/dsl/client_config.hpp
```diff
@@ -39,6 +39,18 @@ struct server_address {
   uint16_t port;
 };
 
+/// Configures the integer type used as the length prefix.
+enum class size_field_t {
```
Neverlord, line 43: We use the `_t` suffix mainly for STL-style type traits, not for proper class/enum names.

```suggestion
enum class size_field_type {
```

libcaf_net/caf/net/lp/framing.cpp
```diff
@@ -42,12 +42,27 @@ class framing_impl : public framing {
 
   static constexpr size_t hdr_size = sizeof(uint32_t);
 
-  static constexpr size_t max_message_length = INT32_MAX - sizeof(uint32_t);
-
   // -- constructors, destructors, and assignment operators --------------------
 
-  explicit framing_impl(upper_layer_ptr up) : up_(std::move(up)) {
-    // nop
+  explicit framing_impl(upper_layer_ptr up, dsl::size_field_t lp_size)
```
Neverlord, line 47: `explicit` only makes sense on single argument constructors (or constructors with default parameters that may be called with one argument).

```suggestion
  framing_impl(upper_layer_ptr up, dsl::size_field_t lp_size)
```

libcaf_net/caf/net/lp/framing.cpp
```diff
@@ -186,14 +201,17 @@ class framing_impl : public framing {
   upper_layer_ptr up_;
 
   size_t message_offset_ = 0;
+
+  size_t max_message_length_ = 0;
```
Neverlord, line 205: Again, this is not about a length check. This is about configuring how many bytes we are reading from / writing to the wire! This version still uses 4 bytes unconditionally for the length.

Neverlord (changes requested):

libcaf_net/caf/net/lp/framing.cpp
```diff
@@ -40,14 +45,26 @@ class framing_impl : public framing {
 
   // -- constants --------------------------------------------------------------
 
-  static constexpr size_t hdr_size = sizeof(uint32_t);
-
-  static constexpr size_t max_message_length = INT32_MAX - sizeof(uint32_t);
+  static constexpr size_t max_message_length = INT64_MAX - sizeof(uint64_t);
```
shariarriday, line 48: Should the `max_message_length` be calculated based on selected length instead?

Neverlord: Correct.

examples/length_prefix_framing/chat-client.cpp
```diff
@@ -36,6 +38,7 @@ struct config : caf::actor_system_config {
   config() {
     opt_group{custom_options_, "global"} //
       .add<uint16_t>("port,p", "port of the server")
+      .add<uint16_t>("size,s", "length prefix size of the server")
```
Neverlord, line 41: I like the idea of making this configurable. We shouldn't go through an integer though, since it requires manual conversions.

Move `size_field_type` to its own header under `caf/net/lp` and use our usual enum setup (declaring the three conversion functions plus `inspect` and then tell CMake to generate the scaffolding). You can look at `sec.hpp` for the setup (except not using `CAF_ERROR_CODE_ENUM`).

Once we have the proper setup, we can use the enum type directly here.

Neverlord (changes requested):
The `size_field_type` setup is something we can do now, but this PR is touching the DSL stuff. Since we're going to refactor that, this PR will have to wait a bit until the `with`-DSL stuff has been sorted out.

riemass:
> The `size_field_type` setup is something we can do now, but this PR is touching the DSL stuff. Since we're going to refactor that, this PR will have to wait a bit until the `with`-DSL stuff has been sorted out.

`with`-DSL is sorted out :zap:.

libcaf_net/caf/net/lp/framing.cpp
Neverlord: This makes no sense to me. 

```suggestion
```

This can no longer be a constant, since we need to consult `std::numeric_limit<T>::max()`, depending on which type to use for the size.

Neverlord: Forget that, using it this way makes even less sense since it's just the maximum of our integer type anyways.

I think this ended up the maximum size because `recv` won't let you read more than INT32_MAX. However, I think we should actually use this opportunity to make it configurable like we already do in the HTTP API.

shariarriday: I have added a `max_message_length` configuration similar to the `max_request_size` configuration in http API.

CHANGELOG.md
```diff
@@ -22,10 +22,15 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 - The configuration options under `caf.metrics-filters` have been moved to
   `caf.metrics.filters`. This change was made to make the configuration options
   more consistent.
+- The default maximum message size for the length-prefix framing has been
+  reduced to 60 MB. This change was made to improve security by default.
```
riemass:
```suggestion
  reduced to 64 MB. This change was made to improve security by default.
```

libcaf_net/caf/internal/net_config.hpp
```diff
@@ -35,8 +36,12 @@ class net_config {
   public:
     class lazy {
     public:
-      lazy(uint16_t port, std::string bind_address, bool reuse)
-        : port(port), bind_address(std::move(bind_address)), reuse_addr(reuse) {
+      lazy(uint16_t port, std::string bind_address, bool reuse,
+           net::lp::size_field_type lp_size = net::lp::size_field_type::u4)
```
riemass: Since the `size_field_type` is specific only to LP, and the `lazy`/`socket` are universal, is there a way to refactor this? I see that we also have these members in the `lp::with_t::config_impl`, shouldn't that be enough?

riemass (changes requested):
PR seems fine, got a question in the comment.

libcaf_net/caf/internal/net_config.hpp
Neverlord: Yeah, those were redundant and most likely a leftover from earlier iterations. Good think you've caught it!

PR #1985, opened 2025-01-17
Fix warnings building CAF with LLVM 19 on macOS

dominiklohmann:
This fixes all compiler warnings when building CAF on macOS with LLVM 19.1.6.

This fixes three kinds of bugs:
- `-Wdocumentation`: `@throws` now requires a description.
- `-Winvalid-offsetof`: This was already suppressed for GCC, but Clang now detects this as well.
- `-Wmissing-field-initializers`: For fields that were not initialized.

For completeness, I've set my build up like this:

```bash
llvm_prefix="$(brew --prefix llvm)"
export PATH="$(brew --prefix python)/libexec/bin:${PATH}"
export PATH="${llvm_prefix}/bin:${PATH}"
export CC="${llvm_prefix}/bin/clang"
export CXX="${llvm_prefix}/bin/clang++"
export LDFLAGS="-Wl,-rpath,${llvm_prefix} ${LDFLAGS}"
export CPPFLAGS="-isystem ${llvm_prefix}/include ${CPPFLAGS}"
export CXXFLAGS="-isystem ${llvm_prefix}/include/c++/v1 ${CXXFLAGS}"
```

I've then configured my CAF build like this:

```
cmake -B build
```

And ran the build using:

```
cmake --build build
```


libcaf_core/caf/detail/stringification_inspector.test.cpp
```diff
@@ -16,9 +16,9 @@ using namespace std::literals;
 namespace {
 
 struct dummy_user {
-  std::string first_name;
-  std::optional<std::string> last_name;
-  std::optional<std::string> nickname;
+  std::string first_name = {};
+  std::optional<std::string> last_name = {};
+  std::optional<std::string> nickname = {};
```
Neverlord, line 21: These changes make no sense to me. A struct default-initializes its fields and all fields here have default constructors. When/why would the compiler warn here? We have many structs like this and never initialize fields explicitly in this way.

dominiklohmann: In the test below you do aggregate initialization and only specify some of the fields, leaving the other ones uninitialized. That's where the warning is coming from. There are generally two ways to fix this: Either specify all fields in the aggregate initialization, or specify them as I did here.

dominiklohmann: I also understand that this makes no difference in practice here, but `-Wmissing-field-initializers` is pretty strict about this unfortunately.

Neverlord: Got it, thanks. 👍

CI isn't happy about about the changes, though.

dominiklohmann: Should be fixed in a7bd07562113a996902b19f11fb57e867243b342

dominiklohmann: … and 1487b5ef4 for the MSVC build.

Neverlord (approved):

PR #1990, opened 2025-01-24
Allow custom actor serialization

Neverlord:
- Turn actor handles into a first-class type for serializers.
- Add new base type for serializers that operate on byte sequences.

(note: merges to the `release/2` branch, not `main`)

### DIFF
--- a/libcaf_core/caf/serializer.cpp
+++ b/libcaf_core/caf/serializer.cpp
@@ -28,6 +28,32 @@ bool serializer::end_associative_array() {
   return end_sequence();
 }
 
+bool serializer::value(const strong_actor_ptr& ptr) {
+  auto aid = actor_id{0};
+  auto nid = node_id{};
+  if (ptr != nullptr) {
+    aid = ptr->id();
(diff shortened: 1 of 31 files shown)


libcaf_core/caf/serializer.cpp
```diff
@@ -28,6 +28,32 @@ bool serializer::end_associative_array() {
   return end_sequence();
 }
 
+bool serializer::value(const strong_actor_ptr& ptr) {
+  auto aid = actor_id{0};
+  auto nid = node_id{};
```
shariarriday, line 33: Can we consider adding some tests for these functions? I can work on a PR for these if required.

shariarriday (approved):
Some of the updated serialization code lacks tests. It would be great if they were added in this or in a future PR.

libcaf_core/caf/serializer.cpp
Neverlord: They certainly need better coverage, we'll tackle that as a followup.

Issue #1968, opened 2024-12-04
`observable::from_{callable,container}` do not work well with generator-like APIs


dominiklohmann:
We make a lot of use of generators with an API similar to C++23's `std::generator`. I tried to adapt them into a flow, and ran into a few limitations that I think can and should be resolved.

First, I tried using `self->make_observable().from_container(make_gen())`. Unfortunately, `from_container` makes some assumptions about containers that are incorrect for generators:
1. Generators are not copyable
2. Generators have no `const_iterator` member type
3. The return types of `begin()` and `end()` for generators are not the same
4. Const generators cannot be iterated over

Second, I tried using `self->make_observable().from_callable(make_gen().to_callable())`, where `generator::to_callable()` returns a functor that wraps the optional with the behavior that `from_callable` expects. However, this also made the assumption that the callable is copyable that it can be called on a const object.

I was able to work around the restrictions for `from_callable` by changing the callable to wrap the generator in a shared pointer, but that does feel like something I should not need to be doing.

dominiklohmann:
@Neverlord I took a stab at this in https://github.com/actor-framework/actor-framework/pull/2024, but am not sure if you'll accept that approach, as it breaks the interaction between `from_container` and `retry` that is relied upon in at least one unit test.

PR #1992, opened 2025-01-26
Add on_error_resume_next flow operator

shariarriday:

### DIFF
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -50,6 +50,7 @@
 #include <functional>
 #include <numeric>
 #include <type_traits>
+#include <utility>
 #include <vector>
 
 namespace caf::flow {
@@ -272,8 +273,9 @@ public:
 
   /// @copydoc observable::on_error_resume_next
   template <class Predicate, class Fallback>
-  auto on_error_resume_next(Predicate predicate, Fallback fallback) {
-    return materialize().on_error_resume_next(predicate, std::move(fallback));
+  auto on_error_resume_next(Predicate&& predicate, Fallback&& fallback) {
+    return materialize().on_error_resume_next(
+      std::forward<Predicate>(predicate), std::forward<Fallback>(fallback));
   }
 
   /// @copydoc observable::sample
@@ -837,12 +839,17 @@ observable<T> observable<T>::retry(Predicate predicate) {
 }
 
 template <class T>
-template <class Predicate>
-observable<T> observable<T>::on_error_resume_next(Predicate predicate,
-                                                  observable<T> fallback) {
-  using impl_t = op::on_error_resume_next<T, Predicate>;
-  return parent()->add_child_hdl(std::in_place_type<impl_t>, *this, predicate,
-                                 std::move(fallback));
+template <class Predicate, class Fallback>
+observable<T> observable<T>::on_error_resume_next(Predicate&& predicate,
+                                                  Fallback&& fallback) {
+  using fallback_t = std::decay_t<Fallback>;
+  static_assert(is_observable_v<fallback_t>, "Fallback must be an observable");
+  static_assert(std::is_same_v<typename fallback_t::output_type, T>,
+                "Fallback observable must emit the same type as the source");
+  using impl_t = op::on_error_resume_next<T, std::decay_t<Predicate>>;
+  return parent()->add_child_hdl(
+    std::in_place_type<impl_t>, *this, std::forward<Predicate>(predicate),
+    std::forward<Fallback>(fallback).as_observable());
 }
 
 // -- observable: combining ----------------------------------------------------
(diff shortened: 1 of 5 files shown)


libcaf_core/caf/flow/observable.hpp
```diff
@@ -269,6 +270,12 @@ class observable_def {
     return materialize().buffer(count, period);
   }
 
+  /// @copydoc observable::catch
+  template <class Predicate, class Fallback>
+  auto catch_and_resume(Predicate predicate, Fallback fallback) {
+    return materialize().catch_and_resume(predicate, std::move(fallback));
```
dominiklohmann, line 276:
```suggestion
  template <class Predicate, class Fallback>
  auto catch_and_resume(Predicate&& predicate, Fallback&& fallback) {
    return materialize().catch_and_resume(std::forward<Predicate>(predicate),
                                          std::forward<Fallback>(fallback));
```

Copying lambdas will not work when the lambda captures a non-copyable type, e.g., because it captured a unique pointer.

This also applies to other places throughout the code in this PR, and also many other flow operators in general (which I hope can be fixed for CAF 2.x).

See also:
- https://github.com/actor-framework/actor-framework/issues/1968

Neverlord (changes requested):
We never used `catch` for operator names. Mainly because that implies exceptions. Didn't you mean `on_error_resume_next` here? There's an open todo for that: https://github.com/actor-framework/actor-framework/issues/1389#issuecomment-1724998598.

Neverlord (changes requested):

libcaf_core/caf/flow/op/on_error_resume_next.test.cpp
```diff
@@ -0,0 +1,283 @@
...
+    auto ptr = make_counted<sub_t>(coordinator(), out, in, std::move(predicate),
+                                   std::move(fallback));
+    out.on_subscribe(caf::flow::subscription{ptr});
+    return ptr;
+  }
+};
+
+WITH_FIXTURE(fixture) {
+SCENARIO("the on_error_resume_next operator resumes observables on error") {
+  int retry_count = 0;
```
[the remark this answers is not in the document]
shariarriday: Yes, I confused the semantics a little. My current implementation does not switch over once and instead retries. I am updating the tests and flow to ensure that the observable switches over once.

libcaf_core/caf/flow/op/on_error_resume_next.test.cpp
```diff
@@ -0,0 +1,203 @@
...
+        run_flows();
+        check_eq(outputs->size(), 4u);
+        check_eq(*err, error{}); // no error
+        check(*completed);
+      }
+    }
+  }
+}
+
+SCENARIO("on_error_resume_next discards items when there is no demand") {
```
Neverlord, line 107: This test doesn't do what it claims to do. A `multicaster` will buffer items internally if there is no demand. So we will never actually call `on_next` on the operator without demand. To test that, you'll have to cast `uut` back to the actual type and call `on_next` manually.

You could use a `never` as source to make sure there is no interference.

Neverlord (changes requested):

libcaf_core/caf/flow/op/on_error_resume_next.test.cpp
```diff
@@ -14,7 +14,6 @@ using namespace caf;
 namespace {
 
 WITH_FIXTURE(test::fixture::flow) {
-
 SCENARIO("on_error_resume_next switches to a fallback observable on error") {
```
Neverlord, line 17: Did you remove that on purpose? We separate the fixture scope from the tests/scenarios with a vertical space.

libcaf_core/caf/flow/op/on_error_resume_next.test.cpp
```diff
@@ -108,25 +107,33 @@ SCENARIO("on_error_resume_next discards items when there is no demand") {
...
+        using impl_t = caf::flow::op::on_error_resume_next_sub<
+          int, decltype(on_error_resume_next_predicate)>;
+        auto on_error_resume = caf::flow::op::on_error_resume_next{
+          coordinator(),
+          make_observable().never<int>(),
+          std::move(on_error_resume_next_predicate),
+          make_observable().never<int>(),
+        };
+        auto uut = intrusive_ptr<impl_t>{static_cast<impl_t*>(
+          on_error_resume.subscribe(snk->as_observer()).ptr())};
```
Neverlord, line 123: Do we really need to use the operator manually? I think something like this should work?

```cpp
        auto sub = src1.as_observable()
                     .on_error_resume_next(
                       on_error_resume_next_predicate,
                       make_observable().iota(1).take(2).as_observable())
                     .subscribe(snk->as_observer());
        auto* uut = dynamic_cast<impl_t*>(sub.ptr());
        require(uut != nullptr);
```

Neverlord (changes requested):

Neverlord (approved):

PR #1993, opened 2025-01-27
Mark mail types as `[[nodiscard]]`

dominiklohmann:
This makes it harder to incorrectly use `self->mail(…)` and `anon_mail(…)`; it's easy to accidentally forget to send mail or delegate it or send a request with it.


Neverlord (approved):
Definitely a good change, thanks! 🙏

PR #1995, opened 2025-01-29
Update pipeline distros

riemass:
Update Feodra, other images are still maintained.
Just as a note, debian 13 and almalinux 10 are scheduled to be released in 2025.
We build Alpinelinux 3.18 which is the oldest still supported release. The distribution has versions 3.19, 3.20 and 3.21 released. I don't think there is a need to build on all of them, since it's a small distro with frequent updates.


Neverlord (approved):
Thanks for keeping our CI up to date. 👍

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -420,8 +420,11 @@ class framing_impl : public framing {
     auto code_val = static_cast<uint16_t>(code);
     uint32_t mask_key = 0;
     byte_buffer payload;
+    // GCC version 14.1.1 raising -Wfree-nonheap-object error
+    CAF_PUSH_WARNINGS
     payload.reserve(msg.size() + 2);
-    payload.push_back(static_cast<std::byte>((code_val >> 8) & 0x00FF));
+    CAF_POP_WARNINGS
+    payload.push_back(static_cast<std::byte>((code_val & 0xFF00) >> 8));
```
Neverlord, line 427: Did you mean to push that change? First shifting and masking what you want to keep seems more intuitive to me.

riemass: This is actually the original version that I reset, but I can return the shift then mask version when I rebase. Seems more natural to me as well.

Neverlord (approved):

PR #1996, opened 2025-01-29
Remove remnants of caf profiler

riemass:
Noticed we had some stuff left form the CAF profiler which doesn't really exist since 0.18.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,6 @@ config = [
     ],
     // Default CMake flags for the builds.
     buildFlags: [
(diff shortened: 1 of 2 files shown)


Neverlord (changes requested):
`caf-prof` is actually not related to the old profiler API. The input to that script is the output of a process profiler. I'm not sure if someone is actually still using that script (or whether it's still functional), but removing that script is a topic for a separate discussion.

riemass:
OK, that was an misinterpretation. I've fished out the script back and force pushed.

Neverlord (approved):

PR #2003, opened 2025-02-04
Add a `filter_map` step to flows

dominiklohmann:
Working with flows, I found myself very frequently writing code like this:

```cpp
my_observable
  .map([](foo_t foo) { return cond(foo) ? to_optional(move(foo)) : nullopt; })
  .filter([](const optional<foo_t>& foo) { return foo.has_value(); })
  .map([](optional<foo_t> foo) { return move(*foo); })
```

There's a less noisy way to write this with `flat_map`, but that's arguably not intuitive:

```cpp
my_observable
  .flat_map([self](auto x) {
    return cond(x)
      ? self->make_observable().just(move(x))
      : self->make_observable().empty();
  })
```

The change from this PR introduces `filter_map`, which shortens the entire thing to just this:

```cpp
my_observable
  .filter_map([](foo_t foo) { return cond(foo) ? to_optional(move(foo)) : nullopt; })
```

### DIFF
--- a/libcaf_core/CMakeLists.txt
+++ b/libcaf_core/CMakeLists.txt
@@ -260,6 +260,7 @@ caf_add_component(
     caf/flow/op/zip_with.test.cpp
     caf/flow/scoped_coordinator.cpp
     caf/flow/single.test.cpp
+    caf/flow/step/filter_map.test.cpp
     caf/flow/step/ignore_elements.test.cpp
     caf/flow/step/skip_last.test.cpp
     caf/flow/step/take_last.test.cpp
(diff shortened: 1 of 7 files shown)


Neverlord:
Our `flat_map` operator also recognizes iterables and optionals. So you should already be able to write this:

```c++
my_observable
  .flat_map([](foo_t foo) { return cond(foo) ? to_optional(move(foo)) : nullopt; })
```

And it will do exactly what you would do by hand: `map(...).filter(...).map(...)`: https://github.com/actor-framework/actor-framework/blob/9ac0faf/libcaf_core/caf/flow/observable.hpp#L875

dominiklohmann:
That behavior of `flat_map` really should be documented.

Neverlord:
> That behavior of `flat_map` really should be documented.

Definitely, we'll update our docs.

PR #2004, opened 2025-02-04
Add combine_latest flow operator

shariarriday:

### DIFF
--- a/libcaf_core/caf/flow/op/combine_latest.hpp
+++ b/libcaf_core/caf/flow/op/combine_latest.hpp
@@ -1,0 +1,302 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/master/LICENSE.
+
+#pragma once
+
+#include "caf/flow/observable_decl.hpp"
+#include "caf/flow/observer.hpp"
+#include "caf/flow/op/cold.hpp"
+#include "caf/flow/subscription.hpp"
+#include "caf/log/core.hpp"
+#include "caf/make_counted.hpp"
+
+#include <algorithm>
+#include <deque>
+#include <memory>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace caf::flow::op {
+
+template <class F, class... Ts>
+struct combine_latest_oracle {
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/flow/op/combine_latest.hpp
```diff
@@ -0,0 +1,302 @@
...
+    return !out_;
+  }
+
+  void request(size_t n) override {
+    if (out_) {
+      demand_ += n;
+      for_each_input([n](auto, auto& input) {
+        if (input.sub)
+          input.sub.request(n);
+      });
```
Neverlord, line 96: Demand handling is completely busted and gives out infinite credit to the inputs. Consider two inputs A and B, alternatingly calling `on_next`. Each time, the observer calls `request(1)`. Each time A *and* B receive one credit in this implementation which means A gets its credit back when it calls `on_next` and then gets another credit if B calls `on_next`.

`demand_--` in `push()` isn't checking if there's still pending demand from the output, so this implementation will eventually underflow `demand_` and violate the protocol with the observer by calling `on_next` without having credit to do so.

This implementation took the scaffolding from `zip_with`, which is fine for the utility functions and item-joining aspect. However, the flow of credit through this operator is completely different, much more akin to `merge`.

Let me think where to go from here. The operator would need to be redesigned from the ground up, so I'm not sure if it's worth salvaging.

shariarriday: > Each time A and B receive one credit in this implementation which means A gets its credit back when it calls on_next and then gets another credit if B calls on_next.

I have noticed that `zip_with` follows the same pattern. I have continued with that, because from what I understand both A and B should receive credit when `request` is called in the example. 
Are you suggesting that the demand should be handled somewhat dynamically? In that case, can you provide any idea of how the credits should be allocated among inputs? One idea could be to give credit to the latest/oldest input.

Neverlord: The `zip_with` operator emits one item after A *and* B have produced an item. Hence, the `zip_with` operator can assign the same credit to each of its inputs. This operator emits one item after A *or* B have produced an item.

The `combine_latest` operator is basically a special case of `merge`. I'll give it some more thought, maybe there's a simpler approach to this.

Neverlord:
Closing in favor of #2016.

PR #2006, opened 2025-02-05
Remove Almalinux 8

riemass:
Alma 8 reached EOL.
Also, we copied the security support dates instead of active  support dates [[1]](https://endoflife.date/almalinux). Fixed it for Alma 9. 
Alma 10 is in beta now, we'll add it once it gets to stable.


Neverlord (approved):
Thanks for taking care of it! 🙏

PR #2008, opened 2025-02-06
Add new uri::with_userinfo function

Neverlord:
Just a small addition for something that came up recently.

### DIFF
--- a/libcaf_core/caf/uri.cpp
+++ b/libcaf_core/caf/uri.cpp
@@ -119,6 +119,43 @@ std::optional<uri> uri::authority_only() const {
   return uri{std::move(result)};
 }
 
+namespace {
+
+template <class Password>
+uri::impl_ptr with_userinfo_impl(const uri::impl_ptr& src, std::string&& name,
+                                 Password&& password) {
+  uri::userinfo_type userinfo{std::move(name),
+                              std::forward<Password>(password)};
+  auto result = make_counted<uri::impl_type>();
+  result->scheme = src->scheme;
+  result->authority = src->authority;
+  result->authority.userinfo.emplace(std::move(userinfo));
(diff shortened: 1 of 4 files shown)


libcaf_core/caf/uri.cpp
```diff
@@ -119,6 +119,43 @@ std::optional<uri> uri::authority_only() const {
...
+  uri::userinfo_type userinfo{std::move(name),
+                              std::forward<Password>(password)};
+  auto result = make_counted<uri::impl_type>();
+  result->scheme = src->scheme;
+  result->authority = src->authority;
+  result->authority.userinfo.emplace(std::move(userinfo));
+  result->path = src->path;
+  result->query = src->query;
+  result->fragment = src->fragment;
+  result->assemble_str();
```
riemass, line 136: If we ever add a field to `uri::impl_ptr`, we might forget to add it here. How about refactoring this as a `impl_ptr::deep_copy` member function? If the member function can be private that would be even better.

riemass (changes requested):
Minor code smell noted, but overall looks good!

libcaf_core/caf/uri.cpp
Neverlord: There's no copy constructor on `uri_impl` (a `deep_copy` function would basically be the same), because making a copy and then changing members would result in an inconsistent object. You need to set all the members and then call `assemble_str()`. Also, the copy would have to call `assemble_str()` initially, which would be wasteful. I've added a `copy_members_from` function for this use case instead.

riemass: I understand there's no classic copy. The resulting function is what I was going at.

riemass (approved):
:+1:

Issue #2001, opened 2025-02-02
Use fast pimpl idiom for our serialization classes


Neverlord:
We are already using the pimpl idiom in a few places to give us more leeway for changing the implementation without breaking the ABI and to reduce compile times. However, there's a lot more classes where it would make sense to push implementation details out of the header. Of course forcing heap allocations onto all of these would be decremental to the performance. Hence, fast pimpl idiom. Isn't quite as flexible because the size of the implementation object is fixed but avoids the extra heap allocations.

The first set of classes where deploying the fast pimpl idiom makes sense are our serialization classes.

Neverlord:
- [ ] config_value_writer
- [ ] config_value_reader
- [ ] json_builder
- [ ] json_reader
- [ ] json_writer
- [ ] binary_serializer
- [ ] binary_deserializer

PR #2010, opened 2025-02-08
Turn actor_registry into an interface

Neverlord:
Part of our ongoing effort to push implementation details out of header files (cf. #2001).


shariarriday (approved):
Is there any reason that we are not creating the `actor_registry_impl` class inside `actor_registry.cpp` and creating the object through a static function?

Neverlord:
> Is there any reason that we are not creating the `actor_registry_impl` class inside `actor_registry.cpp` and creating the object through a static function?

Yes: that would force us to go though a heap allocation. This way, the actor system has the `actor_registry_impl` as a member and we can also completely hide the `start/stop` functions that only exist for the `actor_system`.

PR #2014, opened 2025-02-10
Allow setting max_concurrent for merge operators

### LINKED ISSUE #2012: The `merge` flow operator can have at most 8 inputs
I recently stumbled[^1] upon the limit of 8 concurrently active inputs in the `merge` flow operator.

Unfortunately, there is no easy way to tell as an input, so you're likely stalling indefinitely in this case.

Two thoughts on this:
1. Any hardcoded limit for `merge` is not practical for inputs that may run indefinitely or for a very long time, but rather only for those that have a very short lifetime.
2. _Iff_ such a limit has to exist, then inputs must be able to tell when they aren't getting any demand because of it, for example by failing.

[^1]: https://matrix.to/#/!qNeqsDdXWDBZExYTpF:gitter.im/$igRDKqsyoPJ6iLu4dqNjt2fNpHnbRuCvRaeoEmoCiyg?via=gitter.im&via=matrix.org&via=the-gdn.net

Neverlord:
Closes #2012.

### DIFF
--- a/libcaf_core/caf/flow/op/merge.hpp
+++ b/libcaf_core/caf/flow/op/merge.hpp
@@ -382,9 +382,9 @@ public:
   // -- constructors, destructors, and assignment operators --------------------
 
   template <class... Inputs>
-  merge(coordinator* parent, observable<T> input0, observable<T> input1,
-        Inputs... inputs)
-    : super(parent) {
+  merge(coordinator* parent, size_t max_concurrent, observable<T> input0,
+        observable<T> input1, Inputs... inputs)
+    : super(parent), max_concurrent_(max_concurrent) {
     static_assert((std::is_same_v<observable<T>, Inputs> && ...));
     using vector_t = std::vector<observable<T>>;
     using gen_t = gen::from_container<vector_t>;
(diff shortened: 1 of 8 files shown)


libcaf_core/caf/flow/op/merge.hpp
```diff
@@ -382,9 +382,9 @@ class merge : public cold<T> {
   // -- constructors, destructors, and assignment operators --------------------
 
   template <class... Inputs>
-  merge(coordinator* parent, observable<T> input0, observable<T> input1,
-        Inputs... inputs)
-    : super(parent) {
+  merge(coordinator* parent, size_t max_concurrent, observable<T> input0,
```
dominiklohmann, line 385: I think a comment here explaining the scaling behavior of `max_concurrent` would be nice to have. What's the downside of just setting this to `std::numeric_limits<size_t>::max()`?

What's a reasonable default if my streams are finite, and what if they are not? If I do set a limit and an input runs into it, how do they notice?

Neverlord: Restricting the number of concurrent subscriptions is a safety measure that is common in ReactiveX implementations. If you have an `observable<observable<T>>`, CAF simply doesn't know how many inner observables there are. Subscribing to an observable might open files or network connections, so blindly opening all of them might be a bad idea. Setting the default quite low is a conscious choice.

However, I do apologize for the lack of documentation. It's not reasonable to assume that users of CAF have used other rx-APIs before. I've added the rational behind this behavior to the manual section for the `merge` operator.

I've also tweaked the implementation to always set the number to the number of observables when merging a pre-defined number of observables and use the default maximum of 8 only when merging an `observable<observable<T>>`.

PR #2019, opened 2025-02-13
Make it harder to misuse concurrency limits in `merge`

dominiklohmann:
Without this change it's easy to accidentally set the new concurrency limit option for the `merge` operator to something like `std::numeric_limits<T>::max()`, which would silently overflow on platforms where `sizeof(size_t) < sizeof(T)`.

### DIFF
--- a/libcaf_core/caf/flow/observable.hpp
+++ b/libcaf_core/caf/flow/observable.hpp
@@ -861,7 +861,7 @@ struct has_max_concurrent_arg : std::false_type {};
 
 template <class T, class... Ts>
 struct has_max_concurrent_arg<T, Ts...> {
-  static constexpr bool value = std::is_unsigned_v<std::decay_t<T>>;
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/flow/observable.hpp
```diff
@@ -861,7 +861,7 @@ struct has_max_concurrent_arg : std::false_type {};
 
 template <class T, class... Ts>
 struct has_max_concurrent_arg<T, Ts...> {
-  static constexpr bool value = std::is_unsigned_v<std::decay_t<T>>;
+  static constexpr bool value = std::is_same_v<std::remove_cvref_t<T>, size_t>;
```
Neverlord, line 864: `remove_cvref_t` is C++20. `decay_t` would work.

dominiklohmann: Ah, didn't think about that too deeply, it's become an instinct now to use that over decay. Will fix.

Neverlord (approved):

PR #2020, opened 2025-02-15
Refactor serialization to use fast_pimpl idiom

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/2001

### DIFF
--- a/libcaf_core/caf/internal/json_node.hpp
+++ b/libcaf_core/caf/internal/json_node.hpp
@@ -1,0 +1,52 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/main/LICENSE.
+
+#pragma once
+
+#include <cstdint>
+#include <string_view>
+
+namespace caf::internal {
+
+/// Reflects the structure of JSON objects according to ECMA-404.
+enum class json_node : uint8_t {
+  element, /// Can morph into any other type except `member`.
+  object,  /// Contains any number of members.
+  member,  /// A single key-value pair.
+  key,     /// The key of a field.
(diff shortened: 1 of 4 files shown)


libcaf_core/caf/binary_serializer.cpp
```diff
@@ -17,300 +19,609 @@ namespace caf {
...
+    } else if (write_pos_ + x.size() <= buf_size) {
+      memcpy(buf_.data() + write_pos_, x.data(), x.size());
+    } else {
+      auto remaining = buf_size - write_pos_;
+      CAF_ASSERT(remaining < x.size());
+      auto first = x.begin();
+      auto mid = first + remaining;
+      auto last = x.end();
+      memcpy(buf_.data() + write_pos_, first, remaining);
+      std::copy(mid, last, std::back_inserter(buf_));
```
shariarriday, line 213: This has been updated as buf_.insert() statement was throwing an error due to a possible bug of gcc. I have found some similar reported issue (https://github.com/google/flatbuffers/issues/8119). Changing the statement to std::copy resolves the issue though, so I went with that.

Neverlord: `insert` is the single most efficient way to copy something into a `vector`. Using a `back_inserter` will repeatedly call `insert` or `push_back`, which has to do a size check each time and potentially has to grow the internal buffer multiple times.

If GCC has a false positive here, it's better to suppress that. In `config.hpp`, we have a set of supression macros. We could add new one over there, e.g.:

```c++
#  define CAF_PUSH_STRINGOP_OVERREAD_WARNING                                   \
    _Pragma("GCC diagnostic push")                                             \
    _Pragma("GCC diagnostic ignored \"-Wstringop-overread\"")
```

(don't forget the MSVC and Clang versions that just do a `push`)

Neverlord (approved):

PR #2021, opened 2025-02-15
Add startup check for socket manager in HTTP client factory

### LINKED ISSUE #1954: caf::detail::beacon is not dereferenced properly when actor_system shuts down
The following example tries to create a scenario where the `actor_system` is shut down when the execution is blocked by `res->first.get()`. This example is not completely deterministic but reproduces the problem often. 

Also, "http://localhost:9880" should be able to handle requests and send responses.

```
#include "caf/net/http/client.hpp"
#include "caf/net/http/with.hpp"
#include "caf/net/middleman.hpp"

#include "caf/actor_system.hpp"
#include "caf/actor_system_config.hpp"
#include "caf/caf_main.hpp"

#include <memory>
#include <utility>

using namespace std::literals;

namespace http = caf::net::http;
namespace ssl = caf::net::ssl;
using namespace caf;

int caf_main(caf::actor_system& sys) {
  std::thread worker([&sys]() {
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
    sys.~actor_system();
  });
  while (true) {
    namespace http = caf::net::http;
    namespace ssl = caf::net::ssl;
    caf::uri uri;
    if (auto err = parse("http://localhost:9880", uri)) {
      sys.println("*** failed to parse URI: {} ", err);
      return EXIT_FAILURE;
    }
    auto res = http::with(sys)
                 .context(ssl::context::enable(false).and_then(
                   ssl::emplace_client(ssl::tls::v1_2)))
                 .connect(uri)
                 .connection_timeout(500ms)
                 .max_retry_count(0)
                 .add_header_fields(std::map<std::string, std::string>{
                   {"content-type", "application/json"}})
                 .request(http::method::post, "{\"test\": \"Message\"}");
    if (res && res->first.get()) {
      continue;
    } else {
      printf("Failed to send request");
      return EXIT_SUCCESS;
    }
  }
}

CAF_MAIN(caf::net::middleman)

```

riemass:
Closes #1954.

Hard to test this since all our multiplexer tests run the `mpx` thread locally. However, I confirmed the code from the ticket running. 
This code does break ABI. I don't know a way of checking if `mpx` is shutting down locally. Comparing the number of running socket managers doesn't work since running the current manager it's delayed. 
I've had many issues with trying to manually run `mng->dispose()` since it's also delayed and run in the `mpx` thread which is no longer running. Then I've tried running `abort` or `handle_error` but those ended in an segmentation fault. 

In the end, checking the boolean from start and manually returning an error is the simplest solution I've found.

### DIFF
--- a/libcaf_net/caf/net/actor_shell.test.cpp
+++ b/libcaf_net/caf/net/actor_shell.test.cpp
@@ -144,7 +144,7 @@ TEST("actor shells can receive messages") {
   auto app = app_t::make(actor{self});
   auto transport = net::octet_stream::transport::make(fd1, std::move(app));
   auto mgr = net::socket_manager::make(&mpx, std::move(transport));
-  mpx.start(mgr);
+  require(mpx.start(mgr));
   fd1.id = net::invalid_socket_id;
   actor hdl;
   self->receive([&hdl](actor& x) { hdl = x; },
@@ -164,7 +164,7 @@ TEST("actor shells can send asynchronous messages") {
   auto app = app_t::make(actor{self});
   auto transport = net::octet_stream::transport::make(fd1, std::move(app));
   auto mgr = net::socket_manager::make(&mpx, std::move(transport));
-  mpx.start(mgr);
+  require(mpx.start(mgr));
   fd1.id = net::invalid_socket_id;
   auto msg = "hello actor shell\n"s;
   auto n = net::write(fd2, as_bytes(make_span(msg)));
@@ -193,7 +193,7 @@ TEST("actor shells can send request messages") {
   auto app = app_t::make(actor{self}, cb);
   auto transport = net::octet_stream::transport::make(fd1, std::move(app));
   auto mgr = net::socket_manager::make(&mpx, std::move(transport));
-  mpx.start(mgr);
+  require(mpx.start(mgr));
   fd1.id = net::invalid_socket_id;
   auto msg = "request"s;
   self->receive([msg](get_atom) { return msg; },
@@ -225,7 +225,7 @@ TEST("actor shells can use flows") {
   auto app = app_t::make(actor{self}, cb);
   auto transport = net::octet_stream::transport::make(fd1, std::move(app));
   auto mgr = net::socket_manager::make(&mpx, std::move(transport));
-  mpx.start(mgr);
+  require(mpx.start(mgr));
   fd1.id = net::invalid_socket_id;
   auto msg = "request"s;
   self->receive([msg](get_atom) { return msg; },
--- a/libcaf_net/caf/net/http/async_client.test.cpp
+++ b/libcaf_net/caf/net/http/async_client.test.cpp
@@ -19,8 +19,6 @@
 
 #include "caf/raise_error.hpp"
 
-#include <iostream>
-
 #ifdef CAF_WINDOWS
 #  include <winsock2.h>
 #else
@@ -115,7 +113,9 @@ struct fixture {
     auto client = net::http::client::make(std::move(app));
     auto transport = net::octet_stream::transport::make(fd2, std::move(client));
     auto mgr = net::socket_manager::make(mpx.get(), std::move(transport));
-    mpx->start(mgr);
+    if (!mpx->start(mgr)) {
+      CAF_RAISE_ERROR(std::logic_error, "failed to start socket manager");
+    }
     fd2.id = net::invalid_socket_id;
     return {res, disposable{mgr}};
   }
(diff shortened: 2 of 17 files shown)


riemass:
The commit was based on main, but I created the PR to `release/2` in the end. I'll rebase in a moment.

libcaf_net/caf/net/multiplexer.cpp
```diff
@@ -221,12 +221,14 @@ class default_multiplexer : public detail::atomic_ref_counted,
 
   // -- thread-safe signaling --------------------------------------------------
 
-  void start(socket_manager_ptr mgr) override {
+  bool start(socket_manager_ptr mgr) override {
     auto lg = log::net::trace("socket = {}", mgr->handle().id);
     if (std::this_thread::get_id() == tid_) {
       do_start(mgr);
+      return !shutting_down_;
```
Neverlord, line 228: We already check the flag in `do_start`. Might as well return properly from that function.

```suggestion
      return do_start(mgr);
```

libcaf_net/caf/net/multiplexer.cpp
```diff
@@ -538,7 +540,7 @@ class default_multiplexer : public detail::atomic_ref_counted,
   /// later via the pollset updater.
   /// @warning assumes ownership of @p ptr.
   template <class T>
-  void write_to_pipe(uint8_t opcode, T* ptr) {
+  ptrdiff_t write_to_pipe(uint8_t opcode, T* ptr) {
```
Neverlord, line 543: We don't really care about the value here. Only whether the write succeeded or not, so I'd return `bool` here as well.

libcaf_net/caf/net/multiplexer.hpp
```diff
@@ -82,7 +82,7 @@ class CAF_NET_EXPORT multiplexer : public async::execution_context {
 
   /// Registers `mgr` for initialization in the multiplexer's thread.
   /// @threadsafe
-  virtual void start(socket_manager_ptr mgr) = 0;
+  virtual bool start(socket_manager_ptr mgr) = 0;
```
Neverlord, line 85: Should definitely add `[[nodiscard]]` here.

Neverlord (changes requested):
Thanks for tracking this down! I'd like to get this fixed sooner, but I also don't see a way to do this without an ABI change.

Just a few nitpicks.

libcaf_net/caf/net/actor_shell.test.cpp
```diff
@@ -144,7 +144,7 @@ TEST("actor shells can receive messages") {
   auto app = app_t::make(actor{self});
   auto transport = net::octet_stream::transport::make(fd1, std::move(app));
   auto mgr = net::socket_manager::make(&mpx, std::move(transport));
-  mpx.start(mgr);
+  std::ignore = mpx.start(mgr);
```
Neverlord, line 147: Why discard the result instead of `check(mpx.start(mgr))` or even `require(mpx.start(mgr))`?

libcaf_net/caf/net/http/async_client.test.cpp
```diff
@@ -115,7 +113,7 @@ struct fixture {
     auto client = net::http::client::make(std::move(app));
     auto transport = net::octet_stream::transport::make(fd2, std::move(client));
     auto mgr = net::socket_manager::make(mpx.get(), std::move(transport));
-    mpx->start(mgr);
+    std::ignore = mpx->start(mgr);
```
Neverlord, line 116: Shouldn't the test abort here instead? This isn't in a test scope, but it could still do something like `if (!...) { throw std::logic_error(...); }`

libcaf_net/caf/net/http/client.test.cpp
```diff
@@ -129,7 +129,7 @@ struct fixture {
     auto client = net::http::client::make(std::move(app));
     auto transport = net::octet_stream::transport::make(fd2, std::move(client));
     auto mgr = net::socket_manager::make(mpx.get(), std::move(transport));
-    mpx->start(mgr);
+    std::ignore = mpx->start(mgr);
```
Neverlord, line 132: Same for the changes in this and the other files: I think the tests should fail if `start` returns `false`.

libcaf_net/caf/net/http/server_factory.cpp
```diff
@@ -175,8 +175,10 @@ expected<disposable> do_start_impl(Config& cfg, Acceptor acc,
                                             cfg.max_connections,
                                             cfg.monitored_actors);
   auto ptr = net::socket_manager::make(cfg.mpx, std::move(impl));
-  cfg.mpx->start(ptr);
-  return expected<disposable>{disposable{std::move(ptr)}};
+  if (cfg.mpx->start(ptr))
+    return expected<disposable>{disposable{std::move(ptr)}};
+  return make_error(sec::disposed,
```
Neverlord, line 180: I don't think `disposed` is the right error code for this situation. Above, we use `logic_error` in a similar case. Applies to all instances where we return `disposed` for this kind of error.

Neverlord (changes requested):

Neverlord (approved):
Good job, thanks. 👍

PR #2022, opened 2025-02-15
Update cirrus FreeBSD pipeline

riemass:
CirrusCI supports the following two images: FreeBSD 14.2 and FreeBSD 15 snap so I had to update our setup, even though FreeBSD didn't go EOL.


Neverlord:
I don't think we need to care about a snapshot release. We don't use unstable releases in CI.

If Cirrus broke the other images, I'd say we just use 14.2 for now and add 15.0 when it comes out at the end of the year.

riemass:
Acknowledged. Please rebase the `release/2` branch afterwards.

Neverlord (approved):

PR #2024, opened 2025-02-24
Make `from_container` work with `std::generator`

### LINKED ISSUE #1968: `observable::from_{callable,container}` do not work well with generator-like APIs

dominiklohmann:
This fixes `from_container` to work with `std::generator`. Additionally, it makes `from_generator` and `from_callable` work with non-copyable callbacks.

I'm opening this pull request primarily for discussion, as this currently breaks the `retry` flow operator when used in combination with `from_container`, as that no longer restarts the iteration from the start of the container. I am uncertain what to do about that. I do personally think that the benefit of this change far outweighs its downsides, but this should definitely be discussed.

Note that this is a breaking change for public-facing APIs, so _iff_ this change makes it should probably land in the `release/2` branch.

See also:
- Fixes #1968

### DIFF
--- a/libcaf_core/caf/flow/gen/from_callable.hpp
+++ b/libcaf_core/caf/flow/gen/from_callable.hpp
@@ -21,7 +21,7 @@ public:
 
   using output_type = detail::unboxed_t<callable_res_t>;
 
-  explicit from_callable(F fn) : fn_(std::move(fn)) {
+  explicit from_callable(F fn) : fn_(std::make_shared<F>(std::move(fn))) {
     // nop
   }
 
@@ -34,7 +34,7 @@ public:
   void pull(size_t n, Step& step, Steps&... steps) {
     for (size_t i = 0; i < n; ++i) {
       if constexpr (boxed_output) {
-        auto val = fn_();
+        auto val = (*fn_)();
         if (!val) {
           if constexpr (detail::is_expected_v<callable_res_t>) {
             if (const auto& err = val.error())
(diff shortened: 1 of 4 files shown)


Neverlord:
I think semantics matter here. A generator is not a container. When producing an `observable` from a container, users should reasonably expect a cold observable. Even if generators use (forward) iterators, they are not containers. They don't *contain* something, they produce values on demand and thus cannot be re-iterated at will.

I can see that it's kinda convenient to plug into the scaffolding for `from_container`, but it breaks the semantics of that operator. It's also not what I would expect as a developer. I would expect that I can pass a generator to `from_generator`. There, I *do* expect a hot observable as a result because once produced a value is gone. Currently, that operator expects something that provides a `pull` member function. However, I think that bit of type inspection should be doable and well worth it to preserve semantics and expectations. Even if you could then (probably) technically pass a vector to `from_generator`, this is asking to turn it into a hot observable.

So long story short, I'm all for supporting generators the way you do, but I think `from_container` is the wrong place. If we can integrate your code into `from_generator` instead, I'm all on board. 👍

dominiklohmann:
I think that's a fair trade-off that you're suggesting, I like that. Those changes shouldn't be too hard to make.

Do you want me to rebase this on top of `release/2`?

dominiklohmann:
I am no longer pursuing this.

PR #2025, opened 2025-02-24
Refactor `with` DSL for WebSocket

Neverlord:
The implementation for the `with` DSL is needlessly complex at the moment, forces a lot of code into headers (making it hard to keep the ABI stable) and is also hard to extend.

This introduces a new approach via an internal `net_config` utility class. The config stores whatever a user can configure for both clients and servers and the DSL then makes sure that users can only use the config for one purpose. This allows us to re-implement the `with` DSL with an opaque config type (that internally inherits from `net_config`), which in turn gives us much more leeway for changing implementation details without breaking the ABI.


riemass:
> The implementation for the `with` DSL is needlessly complex...

What? After I've spent months trying understanding it I have to see it go...

Kind of surprised you didn't split the `net_config` into client and server specific classes? Since it's an internal layer I don't really mind but I'm interested if there was a specific reason for keeping them in a single class?

riemass (approved):

Neverlord:
> What? After I've spent months trying understanding it I have to see it go...

Yeah, it's just too complicated. It's been neat for prototyping, but I don't want to maintain that going forward.

> Kind of surprised you didn't split the `net_config` into client and server specific classes?

Simplicity. 🙂

If we split the class, then we need to have three classes again: one for the generic stuff (what you can configure on `with_t` directly) and then the client and server versions. So we either need to duplicate the generic stuff in client and server again or have the generic stuff as a sub-type... Just isn't worth the hassle IMHO. The user only interacts with high-level DSL and that will make sure that the config results in a valid state anyways.

PR #2029, opened 2025-02-28
Consistently use byte span aliases

Neverlord:


libcaf_core/caf/byte_span.hpp
```diff
@@ -19,6 +19,11 @@ using const_byte_span = span<const std::byte>;
 /// Checks whether the byte span is a valid UTF-8 string.
 CAF_CORE_EXPORT bool is_valid_utf8(const_byte_span) noexcept;
 
+/// Checks whether the byte span is a valid ASCII string, i.e., all bytes are in
+/// the range 0x01 to 0x7F, i.e., does not consider the null terminator as
+/// valid ASCII.
```
riemass, line 24: Why don't we consider the null terminator as valid ASCII?

riemass (approved):
Looks good.

libcaf_core/caf/byte_span.hpp
Neverlord: To me, a null terminator just has a different meaning and isn't really a character. But looks like that's just me, similar functions do consider 0 to be part of ASCII and technically it's also in the standard. So... changed.

riemass (approved):
Great.

Issue #2026, opened 2025-02-27
Remove `caf.net.dsl` classes


Neverlord:
As a proof-of-concept, we have refactored our `with`-DSL for WebSocket (see https://github.com/actor-framework/actor-framework/pull/2025) without the overly complicated `caf.net.dsl` setup. Since that resulted in a lot nicer code (albeit a bit more verbose), it's time to migrate the other `with`-DSLs to the new approach and ditch the `caf.net.dsl` stuff.

riemass:
I wrapped up the transition to the new DSL with the most recent PR. 
Whats left as a followup is a review the error handling to be like in the LP PR, and to add an changelog entry, because now all servers have the `monitor` API call, but before I think only one had the support. Afterwards we can close the ticket.

riemass:
Completed with the last PR.

PR #2031, opened 2025-03-13
Refactor `with` DSL for HTTP

riemass:
Relates #2026.
This PR refactors the `with` DSL for HTTP.


libcaf_net/caf/net/http/server_factory.cpp
```diff
@@ -1,280 +0,0 @@
...
-  });
-}
-
-expected<disposable> server_factory::do_start(dsl::server_config::socket& data,
-                                              push_t push) {
-  return checked_socket(data.take_fd())
-    .and_then(this->with_ssl_acceptor_or_socket([this, &push](auto&& acc) {
-      using acc_t = std::decay_t<decltype(acc)>;
-      return do_start_impl(*config_, std::forward<acc_t>(acc), std::move(push));
-    }));
```
riemass, line 264: I didn't really find the equivalent of `checked_socket` in the port of WebSocket DSL or in the `net_config` base class. Is this an oversight or is the validation of sockets handled in some other way?

libcaf_net/caf/net/http/with.cpp
```diff
@@ -0,0 +1,488 @@
...
+        auth.port = defaults::net::https_default_port;
+      use_ssl = true;
+    } else {
+      auto err = make_error(sec::invalid_argument,
+                            "unsupported URI scheme: expected http or https");
+      return err;
+    }
+    if (use_ssl) {
+      if (!ctx) {
+        if (auto maybe_ctx = context_factory())
```
riemass, line 257: Please note that the factory has a default implementation of `ssl::context::make_client(ssl::tls::v1_2);`. Originally, only HTTP used the context_factory (via `has_lazy_context`) to lazy load the SSL. I would suggest to raise the `context_factory` to the `net_config` base class and port it to the WebSocket DSL as well. Any opinions?

libcaf_net/caf/net/http/with.hpp
```diff
@@ -4,75 +4,302 @@
...
+  /// Sets the optional SSL context factory used to lazily create the SSL
+  /// context when needed by the client. Isn't used when creating servers.
+  /// @param factory The factory that creates the SSL context  for encryption.
+  /// @returns a reference to `*this`.
+  template <typename F>
+  [[nodiscard]] with_t&& context_factory(F&& factory) && {
+    static_assert(std::is_same_v<decltype(factory()), expected<ssl::context>>);
+    set_context_factory(std::forward<F>(factory));
+    return std::move(*this);
   }
```
riemass, line 255: Just a question for placing this DSL member function. Lazy loading the SSL context only makes sense for clients where the context is evaluated based on the protocol (`http://...` vs `https://...`). Servers will start in SSL more based on the presence of a `context`. Hence, this API call should really be forbidden for servers (it won't have any effect if called), but I think it's too much hassle to make a compile error. Is it OK to leave as is (actually, this is the original behavior)?

libcaf_net/caf/net/http/server_factory.cpp
Neverlord: IIRC, `checked_socket` only lifts a socket handle into an `expected` to allow `and_then` chaining. Do you have anything in mind other than checking for `invalid_socket_id` if you say validation of sockets?

libcaf_net/caf/net/http/with.cpp
Neverlord: Sounds good to me. While you're at it, the factory currently uses `std::function`. We use `unique_callback_ptr<...>` for the error callback, so please switch the factory to using that as well to have a consistent setup.

libcaf_net/caf/net/http/with.hpp
Neverlord: All other SSL context setup functions takes place on the "top" level, so I think we should leave it that way. It's clearly documented that the factory does nothing for servers, so I think that's fine.

Neverlord (changes requested):

Neverlord (approved):

Issue #2018, opened 2025-02-13
Allow tests to iterate over pending messages


Neverlord:
Came up from a discussion in chat: it would be nice if the deterministic fixture would allow users to iterate over the pending messages for debugging. Wouldn't be too hard to implement either.

PR #2034, opened 2025-03-14
Allow tests to iterate over pending messages

shariarriday:
closes https://github.com/actor-framework/actor-framework/issues/2018


Neverlord (approved):
I've pulled the dynamic cast out of the loop, looks good otherwise. Thanks! 👍

Issue #1873, opened 2024-05-25
Clean up for version 2.0


Neverlord:
With 2.0, we can tie up a few loose ends from 1.0 (aside from the obvious removal of deprecated features).

- [ ] Remove Timeouts from `behavior`:
CAF 1.0 introduced `set_idle_timeout`, that separates the timeout from the behavior. This allows us to remove the timeout entirely from `behavior`. This will also mean we no longer need `message_handler` and we can move `or_else` to the `behavior` class. The only tricky part here are the `receive` functions in `blocking_actor`, since they will have to pass a timeout separately instead of just constructing a `behavior` with whatever the user has passed (might include a timeout definition with `after(...) >> ...`).
- [ ] Remove Various Mixins:
The new `mail` API made some of our mixins obsolete, namely `sender` and `receiver`. This means we can remove them entirely from the code base.
- [ ] Remove the legacy unit testing API:
We have implemented an all-new unit testing framework in `caf::test` and have ported all of our unit tests to it. We only left the old framework in the code base in 1.0 to give users a time window for migrating to the new API.

Neverlord:
- [x] The `always_quote_strings` in the `stringification_inspector` seems unused (see https://github.com/actor-framework/actor-framework/pull/1960#issuecomment-2466413629) and probably can be removed as well.

riemass:
- [x] Deprecate `lp::frame` in favor of `caf::chunk`. [[1]](https://github.com/actor-framework/actor-framework/blob/fd50435d34bfe4d2dab036005aad4d34b662e238/libcaf_net/caf/net/lp/frame.hpp#L12)

PR #2036, opened 2025-03-22
Remove always_quote_strings

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/1873


libcaf_core/caf/detail/stringification_inspector.hpp
```diff
@@ -138,6 +112,7 @@ class CAF_CORE_EXPORT stringification_inspector
   std::enable_if_t<detail::is_map_like_v<T>, bool>
   builtin_inspect(const T& xs) {
     sep();
```
Neverlord, line 114: Things like `sep()` and `result()` should not be part of the public interface. Please try breaking down the builtin functions using our inspection API. For example, a `map` breaks down into `begin_associative_array(...)` followed by the keys and values and then `end_associative_array()`.

Neverlord (changes requested):

libcaf_core/caf/detail/stringification_inspector.hpp
shariarriday: I have done some refactoring, but  `std::enable_if_t<detail::is_map_like_v<T>, bool>` could not be refactored much as the string output format can not be reproduced with inspection API. `begin_associative_array(...)` and  `end_associative_array()` does not generate output for map type in `{"key" = "value"}` format. Instead it will generate the output in `[["key", "value"]]` format.

Neverlord: In the `json_writer`, we have a stack that stores where we are currently. We could do the same here. Each `begin_*` function would push to the stack and each `end_*` would pop it. Then we can use the context information in the other functions to determine what kind of formatting is currently required. Wanna give it a shot? Let me know if you get stuck or need a bit of scaffolding.

libcaf_core/caf/deep_to_string.test.cpp
```diff
@@ -43,6 +43,10 @@ TEST("pointers and optionals use dereference syntax") {
   check_eq(deep_to_string(static_cast<int*>(nullptr)), "null");
   check_eq(deep_to_string(std::optional<int>{}), "null");
   check_eq(deep_to_string(std::optional<int>{23}), "*23");
+  check_eq(deep_to_string(std::vector<std::optional<int>>{{23}, {24}, {}}),
```
shariarriday, line 46: Added some new tests to ensure that separators are applied properly for `builtin_inspect`

libcaf_core/caf/detail/stringification_inspector.hpp
```diff
@@ -113,6 +89,8 @@ class CAF_CORE_EXPORT stringification_inspector
 
   bool value(std::string_view x);
 
+  bool value(void* x);
```
shariarriday, line 92: This public function has been added to simplify serialization of void types and the public `builtin_inspect` template.

Issue #1997, opened 2025-01-29
Can't get actor system to exit.


alatdneg:
Using system.await_actors_before_shutdown(true);

After updating to CAF 1.0.2, I can no longer use the actor clock, to kill in flight messages.

system.clock().cancel_all();

I've now hit a problem where the system won't shutdown cleanly, sometimes is does though, the app (xStudio) is big..

I've basically no idea how to resolve this, is there a specific list of actions that can cause the system to not shutdown ?

It's tied to actors not exiting obviously, but I'm curious what operations might cause this.

If say I did a mail(atom).send(to_actor). Does that holds a strong ref to the src and dest actors ? And if the dest exits it's message loop before handling the message, will that cause them to not exit ?

Is there a guide on the actor lifetime and this sort of thing ?

cheers!

alatdneg:
Annoyingly, if I enable debug logging in CAF, it exits cleanly most of the time.

Neverlord:
> I've basically no idea how to resolve this, is there a specific list of actions that can cause the system to not shutdown ?

It's actually quite simple. There's a global counter how many actors are currently running. The destructor of the actor system waits for that counter to reach 0 before doing anything.

> If say I did a mail(atom).send(to_actor). Does that holds a strong ref to the src and dest actors ? And if the dest exits it's message loop before handling the message, will that cause them to not exit ?

Yes, that message keeps both the sender and the receiver "alive". The message has a strong reference to the receiver and the scheduler will hold onto a reference to an actor as long as it has at least one message waiting in its mailbox.

However, if an actors goes down, it'll drop all messages from the mailbox and the scheduler will no longer keep a reference to a terminated actor.

> Is there a guide on the actor lifetime and this sort of thing ?

The [Reference Counting](https://actor-framework.readthedocs.io/en/stable/core/ReferenceCounting.html#reference-counting) section in the manual is not exactly a guide, but it has some details.

If the reference count of some actors never drops to 0, it's usually because there are mutual relations. Like a worker holding on to a reference to its supervisor and vice versa. That wouldn't explain why it worked previously with `cancel_all`, though. Is the system isn't stopping at all or is there just something with a long timeout keeping it alive?

> Annoyingly, if I enable debug logging in CAF, it exits cleanly most of the time.

That's how it usually goes...

[...]

PR #2037, opened 2025-04-05
Add a system metric to track running actors by name

tobim:
This change adds an additional actor system metric to track the number of running actors by name.

This is useful both as a general metric as well as when dealing with an actor system that is unable to shut down cleanly. In the second case, this metric can be collected to reveal which actors are remaining and blocking termination.

The function `await_running_count_equal()` now takes an optional timeout value to support the second use-case.

Implementation note:
I opted to add an additional metric instead of extending the existing `running-actors` metric because the number of currently running actor is returned by `inc_running()` and `dec_running()`, and calculating the total based on a gauge family is more costly than accessing a single atomic counter in a contested and potentially hot part of the code.

Preceding discussion: https://github.com/actor-framework/actor-framework/issues/1997#issuecomment-2634935723

### DIFF
--- a/libcaf_core/caf/abstract_actor.cpp
+++ b/libcaf_core/caf/abstract_actor.cpp
@@ -190,7 +190,7 @@ void abstract_actor::register_at_system() {
   if (getf(is_registered_flag))
     return;
   setf(is_registered_flag);
-  [[maybe_unused]] auto count = home_system().registry().inc_running();
+  [[maybe_unused]] auto count = home_system().registry().inc_running(name());
   log::system::debug("actor {} increased running count to {}", id(), count);
 }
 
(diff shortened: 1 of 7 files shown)


Neverlord:
Thanks for taking a stab at it! I agree on the usefulness, but I think having two metrics for the same thing will only lead to confusion. There's no reason to split the metric other than the fact that the registry is currently reading the metric value. I do not believe that implementation details should stop us from doing metrics properly. 🙂

We also cannot merge this to `main` as-is, because it breaks the ABI. Hence, this change will have to wait for CAF 2.0. I've played a bit around with it and will open a followup that adds the label dimension to the existing metric instead.

PR #2040, opened 2025-04-08
Add label dimension to caf.system.running-actors

### LINKED ISSUE #2039: Add label dimension to `caf.system.running-actors`
Having the information available what types of actors are currently running makes debugging shutdown issues much easier. See #1997 and #2037.

Neverlord:
Closes #2039. Relates #2037.

### DIFF
--- a/libcaf_core/caf/actor_system.hpp
+++ b/libcaf_core/caf/actor_system.hpp
@@ -163,9 +163,6 @@ public:
     /// Counts the total number of processed messages.
     telemetry::int_counter* processed_messages;
 
-    /// Tracks the current number of running actors in the system.
-    telemetry::int_gauge* running_actors;
-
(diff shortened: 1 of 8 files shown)


riemass (approved):
LGTM :+1: 
At some point in time it would probably be good to have integration tests set up that would verify the metrics per name.

libcaf_core/caf/actor_system.hpp
dominiklohmann, line 1: The PR that this replaces also had a timeout for how long to wait for actors to exit on shutdown, which was really useful in combination with this newly added metric. Do you plan to add that back, too, to make it easy to print this information on shutdown after some timeout?

dominiklohmann: Awesome. Thanks for upstreaming this! ❤️

Issue #2048, opened 2025-04-13
CLI options cannot set boolean flags to false


Neverlord:
When passing `--caf.logger.console.colored=false` to a CAF application, the parser complains about "invalid argument" instead of setting the parameter. Setting via config file works fine.

PR #2049, opened 2025-04-14
Fix parsing of flags on the command line

Neverlord:
Boolean flags now accept arguments on the command line (#2048). For example, `--foo=true` is now equivalent to `--foo` and `--foo=false` will set the flag to `false`. Short options like `-f` only accept the argument when passing it without a space, e.g., `-ffalse`.

### DIFF
--- a/libcaf_core/caf/config_option_set.test.cpp
+++ b/libcaf_core/caf/config_option_set.test.cpp
@@ -242,6 +242,16 @@ TEST("string parameters") {
   check_eq(read<std::string>({"-vfoobar"}), "foobar");
 }
 
+TEST("boolean parameters") {
+  opts.add<bool>("value,v", "some value");
+  check_eq(read<bool>({"-v"}), true);
+  check_eq(read<bool>({"-vtrue"}), true);
+  check_eq(read<bool>({"-vfalse"}), false);
+  check_eq(read<bool>({"--value"}), true);
+  check_eq(read<bool>({"--value=true"}), true);
+  check_eq(read<bool>({"--value=false"}), false);
+}
+
 TEST("flat CLI options") {
   key = "foo.bar";
   opts.add<std::string>("?foo", "bar,b", "some value");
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/config_option_set.test.cpp
```diff
@@ -242,6 +242,16 @@ TEST("string parameters") {
...
 }
 
+TEST("boolean parameters") {
+  opts.add<bool>("value,v", "some value");
+  check_eq(read<bool>({"-v"}), true);
+  check_eq(read<bool>({"-vtrue"}), true);
+  check_eq(read<bool>({"-vfalse"}), false);
+  check_eq(read<bool>({"--value"}), true);
+  check_eq(read<bool>({"--value=true"}), true);
+  check_eq(read<bool>({"--value=false"}), false);
```
riemass, line 252: We support space separated args as well, like `--value true` and `--value false`. A test would be nice.

riemass (changes requested):
With #1429 the space is also accepted as separator. One of my first tickets.

libcaf_core/caf/config_option_set.test.cpp
Neverlord: That won't work, flags cannot split into two arguments. Neither for short names nor for long names. Passing `--value` will set the parameter to `true`, then CAF will try to evaluate the next parameter as an option name again.

riemass: I don't mind approving it then. I don't really know if this is a feature or a bug.

riemass (approved):

libcaf_core/caf/config_option_set.test.cpp
Neverlord: Hm, half feature half bug? 😅

The general rule is to have flags being `false` by default and then you can turn them on via CLI. Think `--verbose`. That's the design pattern that we've settled on.

The problem is that we have some flags in CAF that predate this approach. It came up with `caf.logger.console.colored`, which is `true` by default. Currently, you can disable it via config file but there is no way to disable it via CLI. Hence I've tagged it as bug. Renaming the flag would be a breaking change, so this solution has to do for now. Not elegant, but at least there's a way to set a flag to `false` via CLI now.

riemass: Thanks, I appreciate the explanation.

PR #2052, opened 2025-04-18
Add timeout for await_running_count_equal

shariarriday:
Relates https://github.com/actor-framework/actor-framework/pull/2040

Adds a timeout for `await_running_count_equal`

### DIFF
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -420,15 +420,20 @@ public:
   }
 
   /// Blocks the caller until running-actors-count becomes `expected`
-  /// (must be either 0 or 1).
-  void await_running_count_equal(size_t expected) const override {
+  /// (must be either 0 or 1) or timeout is reached.
+  void await_running_count_equal(size_t expected,
+                                 timespan timeout = infinite) const override {
     CAF_ASSERT(expected == 0 || expected == 1);
     auto lg = log::core::trace("expected = {}", expected);
     std::unique_lock guard{running_mtx_};
-    while (running() != expected) {
+    const auto pred = [&] {
       log::core::debug("running = {}", running());
-      running_cv_.wait(guard);
-    }
+      return running() == expected;
+    };
+    if (timeout == infinite)
+      running_cv_.wait(guard, pred);
+    else
+      running_cv_.wait_for(guard, timeout, pred);
   }
 
   /// Removes a name mapping.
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/typed_event_based_actor.test.cpp
```diff
@@ -244,7 +244,8 @@ TEST("spawning a typed actor and sending messages") {
   }
   SECTION("run test series with typed_server2") {
     test_typed_spawn(sys.spawn(typed_server2));
-    sys.registry().await_running_count_equal(1);
+    sys.registry().await_running_count_equal(1, 100ms);
```
shariarriday, line 247: I made a small update to this test to ensure that all the cases are covered by tests. Alternatively, we can plan new test for this scenario.

libcaf_core/caf/actor_system.cpp
```diff
@@ -420,15 +420,20 @@ class actor_registry_impl : public actor_registry {
   }
 
   /// Blocks the caller until running-actors-count becomes `expected`
-  /// (must be either 0 or 1).
-  void await_running_count_equal(size_t expected) const override {
+  /// (must be either 0 or 1) or timeout is reached.
+  void await_running_count_equal(size_t expected,
+                                 timespan timeout = infinite) const override {
```
Neverlord, line 425:
```suggestion
                                 timespan timeout) const override {
```

libcaf_core/caf/actor_system.cpp
```diff
@@ -420,15 +420,20 @@ class actor_registry_impl : public actor_registry {
...
     CAF_ASSERT(expected == 0 || expected == 1);
     auto lg = log::core::trace("expected = {}", expected);
     std::unique_lock guard{running_mtx_};
-    while (running() != expected) {
+    const auto pred = [&] {
```
Neverlord, line 429:
```suggestion
    auto pred = [this, &guard] {
```

libcaf_core/caf/typed_event_based_actor.test.cpp
Neverlord, line 247:
```suggestion
    sys.registry().await_running_count_equal(1, 1s);
```

Neverlord: I think that's fine. I'd just pick a longer timeout to be on the safe side.

Neverlord (changes requested):
Just a few nitpicks.

libcaf_core/caf/actor_system.cpp
shariarriday: If we remove the default value in the parameter then we will need to add timeout to the existing usages of `await_running_count_equal`. Should we do that?

Neverlord: Hm, feels redundant on an override but I see why it's needed.

Neverlord (approved):

PR #2053, opened 2025-04-20
Implement HTTP chunked response

### LINKED ISSUE #2038: Support Chunked-Encoding for HTTP client
The CAF HTTP Client unconditionally uses `HTTP/1.1` for requests but does not support Chunked Transfer Encoding. [RFC 9112](https://datatracker.ietf.org/doc/html/rfc9112#name-chunked-transfer-coding) states all recipients must be able to parse and decode this encoding. Unfortunately, this means the client cannot really be used at all in practical scenarios.

riemass:
Closes #2038 

Implemented chunked reading.


libcaf_net/caf/net/http/client.cpp
```diff
@@ -206,10 +206,49 @@ class client_impl : public client {
...
+      consumed += input.size() - remainder.size() + chunk_size + 2;
+      // End of chunk encoded request comes with a zero length chunk.
+      if (chunk_size == 0) {
+        if (!invoke_upper_layer(buffer_))
+          return -1;
+        mode_ = mode::read_header;
+        buffer_.clear();
+        return consumed;
+      }
+      buffer_.reserve(buffer_.size() + chunk_size);
```
Neverlord, line 242:
```suggestion
```

`reserve` is primarily useful for avoiding re-allocations with smaller inserts or things like `emplace_back`. We insert the full range below in one go, so there's no need for a separate `reserve` here.

Neverlord (changes requested):
Nice, that's less code than I've been expecting. Just one nitpick and I would like to see more test coverage. The existing tests are good, but we need additional tests that cover some of the edge cases and errors to merge this with confidence.

libcaf_net/caf/net/http/client.test.cpp
```diff
@@ -450,8 +513,140 @@ SCENARIO("the client receives invalid HTTP responses") {
...
+    WHEN("receiving from an HTTP server") {
+      auto res_promise = async::promise<response_t>{};
+      run_client([](auto*) {}, res_promise, 0x6C);
+      for (auto i = 0u; i < response.size(); i += 2)
+        net::write(fd1, as_bytes(make_span(response).subspan(i, 2)));
+      THEN("the HTTP layer parses the data and calls abort") {
+        check(!res_promise.get_future().get(100ms));
       }
     }
   }
```
riemass, line 652: This is the test that fails on FreeBSD. I did a lot of digging and runs with different parameters/scenarios. What I find is that running the test with `-v5` (which means print trace outputs, set by default in our pipeline for Cirrus with `ctest --output-on-failure`) fails, but when running without the switch, the test succeeds always. 
 
What I suppose happens is that the client, which is run in a background thread, receives the bytes for the second chunk and aborts early, as it sees that the size will be bigger that the maximum allowed. Meanwhile the test thread still does writes to the now closed socket. Manually I confirmed the SIGPIPE occurs when writing the first few bytes of the second chunk. As I don't have the ability to step through the code, this is my best effort at debugging it. Also I don't know why it only takes effect on FreeBSD, Linux should be affected as well.

libcaf_net/caf/net/http/client.cpp
```diff
@@ -206,10 +206,49 @@ class client_impl : public client {
       input.subspan(payload_len_);
       mode_ = mode::read_header;
       return consumed;
+    } else if (mode_ == mode::read_chunks) {
```
Neverlord, line 209: That -1 way down below should basically be unreachable. I'd prefer an explicit "sanity check" here.

```c++
    }
    if (mode_ != mode::read_chunks) {
      abort_and_shutdown("Internal error: invalid mode.");
      return -1;
    }
    // ...
```

libcaf_net/caf/net/http/v1.cpp
```diff
@@ -56,6 +59,33 @@ std::pair<std::string_view, byte_span> split_header(byte_span bytes) {
   }
 }
 
+expected<std::pair<size_t, byte_span>> parse_chunk(byte_span input) {
+  constexpr auto crlf = "\r\n"sv;
+  auto [chunk, remainder] = split_by(to_string_view(input), crlf);
+  if (chunk.size() == input.size()) {
+    // Prevents indefinite octet stream as chunk_size.
+    if (input.size() >= 10)
+      return make_error(sec::protocol_error, "Chunk size part is too long.");
```
Neverlord, line 68: This line isn't covered, but should be easy enough to write a test for.

Neverlord (changes requested):
Tests look much better, thanks! Just two more small requests.

Neverlord (approved):
👍

PR #2054, opened 2025-04-20
Remove `with`  DSL base classes

riemass:
Relates #2026 
Enables monitor API in the `with` DSL for WS servesr. Also fixes some nits I noticed on the go.


riemass:
Updated the PR to remove the old DSL base classes. The only remaining file in the `dsl` directory is `arg.hpp` which is used in the `ssl::context` class. Waiting for the review to see what to do about that.

CHANGELOG.md
```diff
@@ -20,6 +20,19 @@ is based on [Keep a Changelog](https://keepachangelog.com).
...
+- Added `monitor` API to WebSocket and HTTP servers in the `with` DSL (#2026).
+
+### Fixed
+
+- Errors that arise during the `with` DSL setup of servers and clients now
+  properly call `on_error` (#2026).
+
+### Removed
+
+- Removed the with DSL building base classes form `caf/net/dsl/` (#2026).
```
Neverlord, line 34:
```suggestion
```

I don't think that's relevant to users that migrate to CAF 2.0.

libcaf_net/caf/net/web_socket/with.hpp
```diff
@@ -229,50 +240,50 @@ class CAF_NET_EXPORT with_t {
     return std::move(*this);
   }
 
-  /// Creates an `accept_factory` object for the given TCP `port` and
+  /// Creates an `server` object for the given TCP `port` and
```
Neverlord, line 243: That's not creating an actual server. Wording should be more along the lines of `Creates a new server factory object ...`. Same for the occurrences below.

Neverlord (changes requested):
Really glad to see that code go. ✂️

libcaf_net/caf/net/web_socket/with.hpp
riemass: That's the name of the class. Just to double check, you mean just to rephrase the docstring wording, and not actually change the class name? Maybe it wouldn't be a bad idea to change `server` -> `server_factory` and `client` -> `client_factory`.

CHANGELOG.md
riemass: They are not in the `detail` namespace, so my reasoning was that maybe someone implemented their own protocol and DSL using these.

libcaf_net/caf/net/web_socket/with.hpp
Neverlord: Just talking about the documentation. 🙂

The class names are fine IMHO since `server` and `client` are nested in `with`, which makes it clear that they belong to the `with` DSL.

CHANGELOG.md
Neverlord: Hm.. OK then. I think it's very unlikely, but I do see your point.

Neverlord (approved):

Issue #1449, opened 2023-06-19
Enable -Wconversion by default


Neverlord:
After turning on `-Werror` via #1436, the first additional warning we should enable by default is `-Wconversion`.

PR #2057, opened 2025-04-25
Fix -Wconversion warnings

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/1449

### DIFF
--- a/libcaf_core/caf/chrono.cpp
+++ b/libcaf_core/caf/chrono.cpp
@@ -151,7 +151,7 @@ char* print_fractional_component(char* pos, int ns, int precision,
   return pos;
 }
 
-char* print_utf_offset(char* pos, long int offset) {
+char* print_utf_offset(char* pos, long offset) {
   if (offset == 0) {
     *pos++ = 'Z';
     return pos;
@@ -162,7 +162,7 @@ char* print_utf_offset(char* pos, long int offset) {
     *pos++ = '-';
     offset = -offset;
   }
-  auto add_two_digits = [&](long int x) {
+  auto add_two_digits = [&pos](auto x) {
     *pos++ = static_cast<char>((x / 10) + '0');
     *pos++ = static_cast<char>((x % 10) + '0');
   };
--- a/libcaf_core/caf/chrono.cpp
+++ b/libcaf_core/caf/chrono.cpp
@@ -151,7 +151,7 @@ char* print_fractional_component(char* pos, int ns, int precision,
   return pos;
 }
 
-char* print_utf_offset(char* pos, int offset) {
+char* print_utf_offset(char* pos, long int offset) {
   if (offset == 0) {
     *pos++ = 'Z';
     return pos;
@@ -162,7 +162,7 @@ char* print_utf_offset(char* pos, int offset) {
     *pos++ = '-';
     offset = -offset;
   }
-  auto add_two_digits = [&](int x) {
+  auto add_two_digits = [&](long int x) {
     *pos++ = static_cast<char>((x / 10) + '0');
     *pos++ = static_cast<char>((x % 10) + '0');
   };
--- a/libcaf_core/caf/detail/parser/read_floating_point.hpp
+++ b/libcaf_core/caf/detail/parser/read_floating_point.hpp
@@ -156,11 +156,11 @@ void read_floating_point(State& ps, Consumer&& consumer,
   if (exp < 0) {
     for (auto n = -exp; n != 0; n >>= 1, ++i)
       if (n & 0x01)
-        result /= powerTable[i];
+        result /= static_cast<ValueType>(powerTable[i]);
   } else {
     for (auto n = exp; n != 0; n >>= 1, ++i)
       if (n & 0x01)
-        result *= powerTable[i];
+        result *= static_cast<ValueType>(powerTable[i]);
   }
   // 4) Fix sign and call consumer.
   consumer.value(sign == plus ? result : -result);
--- a/libcaf_core/caf/json_builder.test.cpp
+++ b/libcaf_core/caf/json_builder.test.cpp
@@ -143,22 +143,6 @@ TEST("integer") {
   check_eq(val.to_integer(), 42);
 }
 
-TEST("float") {
-  SECTION("value") {
-    check(builder.value(4.2));
-    auto val = builder.seal();
-    check(val.is_double());
-    check_eq(val.to_double(), test::approx{4.2});
-  }
-  SECTION("array") {
-    auto xs = std::vector{4.2f, 4.2f, 4.2f};
-    check(builder.apply(xs));
-    auto val = builder.seal();
-    check(val.is_array());
-    check_eq(printed(val), "[4.2, 4.2, 4.2]"sv);
-  }
-}
-
 TEST("double") {
   check(builder.value(4.2));
   auto val = builder.seal();
--- a/libcaf_core/caf/json_builder.test.cpp
+++ b/libcaf_core/caf/json_builder.test.cpp
@@ -145,10 +145,10 @@ TEST("integer") {
 
 TEST("float") {
   SECTION("value") {
-    check(builder.value(4.2f));
+    check(builder.value(4.2));
     auto val = builder.seal();
     check(val.is_double());
-    check_eq(val.to_double(), test::approx{4.2f});
+    check_eq(val.to_double(), test::approx{4.2});
   }
   SECTION("array") {
     auto xs = std::vector{4.2f, 4.2f, 4.2f};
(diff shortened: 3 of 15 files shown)


libcaf_core/caf/chrono.cpp
```diff
@@ -151,7 +151,7 @@ char* print_fractional_component(char* pos, int ns, int precision,
   return pos;
 }
 
-char* print_utf_offset(char* pos, int offset) {
+char* print_utf_offset(char* pos, long int offset) {
```
Neverlord, line 154:
```suggestion
char* print_utf_offset(char* pos, long offset) {
```

Writing `long int` is a waste of space. No reason to type it, just like no one is actually writing `signed int`. 😉

libcaf_core/caf/chrono.cpp
```diff
@@ -162,7 +162,7 @@ char* print_utf_offset(char* pos, int offset) {
     *pos++ = '-';
     offset = -offset;
   }
-  auto add_two_digits = [&](int x) {
+  auto add_two_digits = [&](long int x) {
```
Neverlord, line 165:
```suggestion
  auto add_two_digits = [&pos](auto x) {
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -267,9 +267,11 @@ class framing_impl : public framing {
...
         || (hdr_.opcode == detail::rfc6455::continuation_frame
             && opcode_ == detail::rfc6455::text_frame))
-      down_->configure_read(receive_policy::up_to(hdr_.payload_len));
+      down_->configure_read(
+        receive_policy::up_to(static_cast<uint32_t>(hdr_.payload_len)));
     else
-      down_->configure_read(receive_policy::exactly(hdr_.payload_len));
+      down_->configure_read(
+        receive_policy::exactly(static_cast<uint32_t>(hdr_.payload_len)));
     return hdr_bytes;
```
Neverlord, line 275: These casts narrow the `payload_len` field, which is 64-bit according to the RFC. Changing our `receive_policy` to `size_t` solves this, but only on 64-bit architectures. So this needs a `if constexpr (sizeof(size_t) < sizeof(uint64_t))` and some bound checking on platforms where `size_t` is smaller.

Neverlord (changes requested):

libcaf_core/caf/telemetry/collector/prometheus.cpp
```diff
@@ -308,7 +308,7 @@ void prometheus::append_histogram_impl(const metric_family* family,
   auto acc = ValueType{0};
   auto index = size_t{0};
   for (; index < buckets.size(); ++index) {
-    acc += buckets[index].count.value();
+    acc += static_cast<ValueType>(buckets[index].count.value());
```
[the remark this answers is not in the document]
shariarriday: The issue is that we are defining `auto acc = ValueType{0};`. We need to static_cast to `ValueType` to ensure that when the values are summed up no warning is shown. Another solution would be change `auto acc = ValueType{0};` into `auto acc = size_t{0};`. As the return type of `buckets[index].count.value()` should always be `size_t`.

libcaf_net/caf/net/http/client.cpp
```diff
@@ -84,7 +84,8 @@ class client_impl : public client {
 
   void request_messages() override {
     if (!down_->is_reading())
-      down_->configure_read(receive_policy::up_to(max_response_size_));
+      down_->configure_read(
+        receive_policy::up_to(static_cast<uint32_t>(max_response_size_)));
```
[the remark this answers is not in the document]
shariarriday: I had considered it. But I decided against it as it would increase the object size. Would like to know your opinion regarding this.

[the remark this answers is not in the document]
Neverlord: Good point. Since we only hold one receive policy object at the transport, I think switching to `size_t` is still cleaner overall and is worth the extra bytes to avoid constant casts.

libcaf_net/caf/net/http/server.cpp
```diff
@@ -74,7 +74,8 @@ class server_impl : public http::server {
 
   void request_messages() override {
     if (!down_->is_reading())
-      down_->configure_read(receive_policy::up_to(max_request_size_));
+      down_->configure_read(
+        receive_policy::up_to(static_cast<uint32_t>(max_request_size_)));
```
Neverlord, line 78: Leftover?

```suggestion
        receive_policy::up_to(max_request_size_));
```

libcaf_core/caf/detail/parser/read_floating_point.hpp
```diff
@@ -156,11 +156,11 @@ void read_floating_point(State& ps, Consumer&& consumer,
   if (exp < 0) {
     for (auto n = -exp; n != 0; n >>= 1, ++i)
       if (n & 0x01)
-        result /= powerTable[i];
+        result /= static_cast<ValueType>(powerTable[i]);
```
Neverlord, line 159: Instead of casting here (and below), why not give `powerTable` the proper type?

libcaf_core/caf/json_builder.test.cpp
```diff
@@ -148,7 +148,7 @@ TEST("float") {
     check(builder.value(4.2f));
```
Neverlord, line 148: This will widen the `float` to a `double` anyways. Let's just remove this `SECTION`. It's redundant with `TEST("double")` below.

libcaf_net/caf/net/lp/framing.cpp
```diff
@@ -85,7 +85,8 @@ class framing_impl : public framing {
         return -1;
       } else {
         log::net::debug("wait for payload of size {}", msg_size);
-        down_->configure_read(receive_policy::exactly(hdr_size + msg_size));
+        down_->configure_read(
+          receive_policy::exactly(static_cast<uint32_t>(hdr_size + msg_size)));
```
Neverlord, line 89:
```suggestion
          receive_policy::exactly(hdr_size + msg_size));
```

libcaf_net/caf/net/lp/framing.cpp
```diff
@@ -94,7 +95,8 @@ class framing_impl : public framing {
         log::net::debug("got message of size {}", msg_size);
         if (up_->consume(msg) >= 0) {
           if (down_->is_reading())
-            down_->configure_read(receive_policy::exactly(hdr_size));
+            down_->configure_read(
+              receive_policy::exactly(static_cast<uint32_t>(hdr_size)));
```
Neverlord, line 99:
```suggestion
            down_->configure_read(receive_policy::exactly(hdr_size));
```

libcaf_net/caf/net/octet_stream/transport.cpp
```diff
@@ -101,6 +101,8 @@ class transport_impl : public transport {
   }
 
   void configure_read(receive_policy rd) override {
+    if constexpr (sizeof(size_t) < sizeof(uint64_t))
+      CAF_ASSERT(rd.max_size <= std::numeric_limits<size_t>::max());
```
Neverlord, line 105:
```suggestion
```

`rd.max_size` has type `size_t` now. This is always true.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -267,9 +267,11 @@ class framing_impl : public framing {
     if (hdr_.opcode == detail::rfc6455::text_frame
```
Neverlord, line 267: Here, we need that `if constexpr (sizeof(size_t) < sizeof(uint64_t))` logic to validate `hdr_.payload_len` before we cast it. But not with an assertion. If the `payload_len` exceeds what we can represent on this platform, we need to call `abort_and_shutdown` with a descriptive error.

Neverlord (changes requested):

libcaf_core/caf/detail/parser/read_floating_point.hpp
shariarriday: `powertable` has values that can only be represented by double (like 1e128, 1e256). The casting is done before division due to it.

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -261,6 +261,15 @@ class framing_impl : public framing {
       abort_and_shutdown(err);
       return -1;
     }
+    if constexpr (sizeof(size_t) < sizeof(uint64_t)) {
+      // If size_t is smaller than uint64_t, we can only handle payloads up to
+      // 2^32-1 bytes.
+      if (hdr_.payload_len > INT32_MAX) {
```
Neverlord:
```suggestion
      if (hdr_.payload_len > std::numeric_limits<size_t>::max()) {
```

libcaf_net/caf/net/web_socket/framing.cpp
```diff
@@ -261,6 +261,15 @@ class framing_impl : public framing {
       abort_and_shutdown(err);
       return -1;
     }
+    if constexpr (sizeof(size_t) < sizeof(uint64_t)) {
+      // If size_t is smaller than uint64_t, we can only handle payloads up to
+      // 2^32-1 bytes.
```
Neverlord:
```suggestion
```

Neverlord (changes requested):
One last nitpick, otherwise LGTM.

PR #2059, opened 2025-04-28
Use localtime_s on all windows platforms

cmburn:
Currently Windows builds fail under MSYS/UCRT64, as localtime_r is not in the standard library. Choosing based on platform instead of compiler would fix this, although I'm unsure if this has any adverse implications for Cygwin?

### DIFF
--- a/libcaf_core/caf/detail/print.cpp
+++ b/libcaf_core/caf/detail/print.cpp
@@ -11,7 +11,7 @@ namespace caf::detail {
 
 size_t print_timestamp(char* buf, size_t buf_size, time_t ts, size_t ms) {
   tm time_buf;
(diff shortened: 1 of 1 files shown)


Neverlord:
Welcome to CAF and thanks for the patch!

MSVC is the only compiler we officially support on Windows. We leave support for Cygwin/MSYS/etc to community patches. So if this fixes an issue for you, I'll merge it as-is. If it breaks something on Cygwin, and someone actually cares about it, someone else will have to iterate on your patch. 🙂

cmburn:
Thanks! I'm leveraging it in a hobby project so depending on my attention span I'll try and fix whatever I come across.

PR #2062, opened 2025-04-29
Add `mail_cache::stash` overload with no arguments

Neverlord:
Picks up a user suggestion in chat to make the API a bit more flexible.


Neverlord:
Adding the `rejected` label here as a reminder that this is a bad idea. Reverted with #2064.

PR #2066, opened 2025-05-01
Allow inheriting from JSON reader/writer

Neverlord:
Overriding how actors are serialized / deserialized is an important customization point. This patch allows specifically overriding this member function on the JSON (de)serializer without opening it up more than necessary.

### DIFF
--- a/libcaf_core/caf/json_reader.cpp
+++ b/libcaf_core/caf/json_reader.cpp
@@ -4,10 +4,11 @@
 
 #include "caf/json_reader.hpp"
 
+#include "caf/actor_control_block.hpp"
+#include "caf/deserializer.hpp"
 #include "caf/detail/assert.hpp"
 #include "caf/detail/bounds_checker.hpp"
 #include "caf/detail/json.hpp"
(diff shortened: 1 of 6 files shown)


libcaf_core/caf/json_reader.cpp
```diff
@@ -746,6 +743,18 @@ class impl : public byte_reader, public internal::fast_pimpl<impl> {
     return false;
   }
 
+  bool value(strong_actor_ptr& ptr) override {
+    // These are customization points for the deserializer. Client code may
+    // inherit from json_reader and override these member functions. Hence, we
+    // need to dispatch to the parent class.
+    return parent_->value(ptr);
```
shariarriday, line 750: Should we consider adding some quick tests to cover serialization/deserialization of actor types?

shariarriday (approved):
Approved with a suggestion regarding adding some new tests.

libcaf_core/caf/json_reader.cpp
Neverlord: Definitely! Done.

PR #2071, opened 2025-05-07
Use consistent error codes for discarded requests

Neverlord:
If an actor terminates, it will discard all messages from its mailbox. Any discarded response message will receive an response message with an error to inform the sender that the request has been discarded. The same happens if a message has been discarded immediately because the mailbox has already been closed.

Previously, CAF has used the exit reason passed to `quit`. However, this makes it hard to differentiate between a discarded request and a request that resulted in an error. Furthermore, this could result in error messages with a default-constructed error code, which semantically makes no sense.


Neverlord:
It's sortof a bugfix (an `error` to a response should never be default-constructed) and a change (no longer use the exit reason when discarding requests). I've settled on filing it under `Changed` in the changelog since CAF will send different error codes from before, so people should be aware of this as not being just a bugfix.

riemass (approved):
I've reviewed the ticket and PR. I completely agree.

PR #2072, opened 2025-05-08
Launch background tasks after multiplexer was run

### LINKED ISSUE #2060: Metrics endpoint hangs indefinitely
Reported by user _polo1250_ on the public chatroom. After setting up the actor application, the `/metrics` endpoint hangs indefinitely when accessing it via `curl`. I've simplified the example further to a bare bones main function with a config file.

main.cpp
```
#include <caf/net/middleman.hpp>

#include <caf/actor_system.hpp>
#include <caf/actor_system_config.hpp>
#include <caf/caf_main.hpp>

#include <atomic>
#include <csignal>
#include <thread>

using namespace caf;

namespace {

std::atomic<bool> shutdown_requested = false;

void set_shutdown_flag(int) {
  shutdown_requested = true;
}

} // namespace

void caf_main(actor_system& sys, const actor_system_config& cfg) {
  signal(SIGINT, set_shutdown_flag);
  signal(SIGTERM, set_shutdown_flag);

  while (!shutdown_requested) {
    std::this_thread::sleep_for(std::chrono::milliseconds(250));
  }
}

CAF_MAIN(caf::net::middleman)

```

config.json
```
{
  "caf": {
    "logger": {
      "console": {
        "verbosity": "trace",
        "format": "[%c:%p] %m"
      },
      "component-levels": {
        "caf.net": "trace",
        "caf.net.http": "trace",
        "caf.net.prometheus": "trace",
        "caf.scheduler": "debug"
      }
    },
    "net": {
      "prometheus-http": {
        "port": 9090,
        "address": "0.0.0.0"
      }
    }
  }
}
```
Link to original message:
https://matrix.to/#/!qNeqsDdXWDBZExYTpF:gitter.im/$09YhAkPG2t855aHsEEQnowRmyvL5Ghc0AtK41qbfaac?via=gitter.im&via=matrix.org&via=the-gdn.net

riemass:
Closes #2060.

The `net::middleman` background tasks, in this case the Prometheus scraper, where run before the multiplexer was run. This means the multiplexer never saw a write event on the 0 socket.

If any kind of server was launched after the initial start, it would send an event to the multiplexer, and it would read the prometheus server data, and start handing it as well. But if no other server was started, then the multiplexer thread will be idle.

Simplest solution, without touching the multiplexer logic, was to reorder the start. Hope this is fine. Alternatively, I could go into the multiplexer details.

### DIFF
--- a/libcaf_net/caf/net/middleman.cpp
+++ b/libcaf_net/caf/net/middleman.cpp
@@ -78,6 +78,7 @@ void launch_prom_server(actor_system& sys, const prom_config& cfg) {
 }
 
 void launch_background_tasks(actor_system& sys) {
+  auto lg = log::net::trace("");
   auto& cfg = sys.config();
   if (auto pcfg = get_as<prom_config>(cfg, "caf.net.prometheus-http")) {
     launch_prom_server(sys, *pcfg);
@@ -111,10 +112,10 @@ middleman::~middleman() {
 void middleman::start() {
   auto fn = [this] {
     mpx_->set_thread_id();
-    launch_background_tasks(sys_);
     mpx_->run();
   };
   mpx_thread_ = sys_.launch_thread("caf.net.mpx", thread_owner::system, fn);
+  launch_background_tasks(sys_);
 }
 
 void middleman::stop() {


riemass:
Stress tested the failed `caf-robot-http-rest` locally. I think this is a pipeline hiccup.

Neverlord:
I'm not sure if I can follow the explanation. If a socket becomes ready, it doesn't matter when you call `poll`/`epoll`. The OS will report the socket as ready regardless of the ordering of the `recv`/`send` calls.

The patch essentially moves the call to `launch_background_tasks` from the multiplexer's thread to (usually) the main thread. So this probably boils down to calling `write_to_pipe(pollset_updater::code::start_manager, ...)` instead of calling `do_start(mgr)`? So the bug might be that we just never read from `updates_` if we add something before ever receiving a socket event?

Neverlord (approved):
We've confirmed that running `apply_updates()` before the first poll also fixes the issue. Launching from outside the thread seems cleaner to me, though. 👍

riemass:
> So the bug might be that we just never read from `updates_` if we add something before ever receiving a socket event?

Yeah it boils down to it. Sorry for explaining it in lazyman terms.

Issue #2073, opened 2025-05-08
Support HTTP `multipart` requests


raxyte:
Currently, there does not seem to be an easy way to send multipart requests from the HTTP client. (can be worked around by manually formatting the request headers and body, I guess. So lower priority than the server.)
For CAF's HTTP server, if you try to send a multipart request, you get `Malformed Request-URI: expected an absolute path.`

Neverlord:
Shouldn't `multipart` be transparent to CAF since it only affects how to parse the payload? That error looks like it's rejecting the first line of the HTTP header. Looking at the [RFC](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2), it looks like we're too strict.

@riemass mind looking into that? The way I read the RFC, I think `request_header::parse` is wrong.

riemass:
Yeah, looking at the code and the RFC we're not compliant. Shouldn't be a hard fix. The error message can only originate from the resource part in the first HTTP line.

Neverlord:
OK, I've added a new ticket for the bug. Let's fix that first.

raxyte:
> Shouldn't `multipart` be transparent to CAF since it only affects how to parse the payload?
That would be a valid approach. Just depends on how much convenience you want to offer to the user.

Neverlord:
What I meant is that CAF shouldn't raise an error if you send a request with multipart. It should be transparent to the code that parses the HTTP request since multipart affects how the payload is encoded (which the request parser shouldn't look at).

Adding API pieces for assembling and parsing multipart payloads definitely seems worthwhile to me.

riemass:
@raxyte just out of curiosity, what was the request target of the HTTP request you sent?

raxyte:
@riemass `curl localhost:8080/hello -F key=val -F k2=v2 -v`

riemass:
Thanks for the provided command. We have a different kind of issue here. We have a bug that would try processing the body of the request again if it contained an '\r\n' EOL sequence. That's were the `Malformed request` log line stems from. Incoming PR will fix that. 

The `net` layer should now delegate the request to the HTTP layer with the body as is. An cosmetics API would be the next step.

PR #2076, opened 2025-05-12
Ensure that HTTP incoming data is parsed only once

riemass:
Relates #2073.

When processing data it was possible to parse the request body again as the first line of a new request. In most cases this would fail immediately so it stayed unnoticed. 
The regression test I come up with isn't very sophisticated and only tests the case where the data is well formed, but it's the easiest approach.


Neverlord (approved):
That's subtle. Thanks for tracking it down!

Issue #2032, opened 2025-03-13
Actors that always have an action scheduled never read messages


dominiklohmann:
This bug was previously discussed in Gitter[^1].

Currently, if an actor schedules an action repeatedly from within that same action, it never reads messages.

One suggested solution from @Neverlord is to introduce a limit to how many actions can be run in one run, similar to the `caf.scheduler.max-throughput`[^2] option which limits how many messages an actor can handle in one run.

[^1]: https://app.gitter.im/#/room/#caf-community:gitter.im/$gNrnQBznMHjjidHLPiyyBbqGNlbzS8uQdN9-vdX1-UY
[^2]: https://github.com/actor-framework/actor-framework/blob/06ed712f97bfdebc2bdf60003fef47be39cf7d3f/libcaf_core/caf/actor_system_config.cpp#L115-L116

dominiklohmann:
@raxyte and I tried our hands at this again earlier today, but it looks like we need to replicate more of the logic than we have so far to write a minimal reproduction.

We can at the moment reliably reproduce this within our code base in a rather complicated and unfortunately very generic actor, but haven't yet managed to do so in a standalone example.

We've now mostly worked around this, but if I can find the time in the coming weeks to get back to this I'll likely build a patched version of CAF with some additional logging so that I can see what's being called in what order.

Can you help us understand the `defaults::max_inline_actions_per_run` value? What counts as a "run"?

Neverlord:
> Can you help us understand the `defaults::max_inline_actions_per_run` value? What counts as a "run"?

Scheduled actors implement the interface `caf::flow::coordinator`. This is what flow operators run on top of. It has a bunch of methods, but the one we care about here is `delay`. Let me get back to delay in a bit, but first a bit of context why we need this machinery in the first place.

If you have a `subscription` object and an `observer` (see also: https://www.interance.io/learning/cpp/guide/flows-part-1 under Protocol between Observables, Observers and Subscriptions), then the subscription will call `on_next` on the `observer` and the `observer` then calls `request` on the subscription object to ask for more data. If the observer always consume the item right away and asks for the next one, you basically have a cycle: `on_next` -> `request(1)` -> `on_next` -> ... if `request` can produce the item on the spot and can call `on_next` again.

Implemented naively, this could blow up the stack worst case or just run a flow basically to completion before doing anything else. Neither would be a good thing.

[...]

PR #2077, opened 2025-05-14
Limit how long an actor may run actions per run

Neverlord:
Instead of adding a limit on how many actions an actor may process in one run, we introduce a new time-based limit. The default for an actor is to stop after 100ms. Users can override this value using the new configuration option `caf.scheduler.run-actions-timeout`.

Relates #2032.


tobim:
I briefly tested this out with our stack but ran into infinite loops when repeatedly re-enqueuing with a `rel_time` of 0 from within the action. So these new semantics introduce a case that the client has to guard against that they didn't have to before.

Neverlord:
@tobim thanks for testing! What function are calling, exactly? `delay_for`/`delay_for_fn`? They go though the actor clock and ultimately push the action to the mailbox, so I think I need a bit more context to understand how this change could cause a regression.

Just always re-scheduling a new action from an action will be an infinite chain, so how does your workflow look like? Just from a high level, so that I can come up with a regression test.

tobim:
> @tobim thanks for testing! What function are calling, exactly? delay_for/delay_for_fn? They go though the actor clock and ultimately push the action to the mailbox, so I think I need a bit more context to understand how this change could cause a regression.

We are using `delay_until_fn`. It some situations it re-enqueues itself with the same `abs_time` indefinitely.

The code is at https://github.com/tenzir/tenzir/blob/65b7c9eacc1e87f9f1d6c569db29826e2b86dec9/libtenzir/src/execution_node.cpp#L45-L52.

Neverlord:
@tobim I've made a small tweak, can you re-run your tests with that version and see if it maybe changes something for the better?

PR #2079, opened 2025-05-20
Fix request-target validation in HTTP

### LINKED ISSUE #2074: HTTP parser rejects valid request lines
Split from #2073 for treating the bug separately from multipart support.

riemass:
Closes #2074.

Tried to follow RFC9112. The system-wide '*' target converted to a '/' request for now, since it doesn't fit our data model containing an uri. 
Also, absolute form has a specific bug. I'll point out in a comment.

### DIFF
--- a/libcaf_net/caf/net/http/request_header.cpp
+++ b/libcaf_net/caf/net/http/request_header.cpp
@@ -14,43 +14,46 @@ namespace {
 constexpr std::string_view eol = "\r\n";
 
 /// Validate the request target part according to RFC9112.
-caf::expected<uri> validate_request_target(http::method method,
-                                           std::string_view request_target) {
+caf::expected<uri> parse_request_target(http::method method,
+                                        std::string_view request_target) {
+  using namespace std::literals;
   if (request_target.empty()) {
     return make_error(sec::invalid_argument,
                       "Malformed Request-URI: request target empty.");
   }
+  expected<uri> res{none};
   if (method == http::method::connect) {
-    if (auto res
-        = make_uri(std::string("nil://") + std::string{request_target})) {
-      return std::move(*res);
-    } else {
-      log::net::debug("Failed to parse CONNECT URI {} -> {}", request_target,
+    if (res = make_uri("nil://"s.append(request_target)); !res) {
+      log::net::debug("Failed to parse CONNECT URI {}: {}.", request_target,
                       res.error());
       return error{sec::invalid_argument, "Malformed CONNECT Request-URI."};
     }
+    if (res->authority().empty()) {
+      log::net::debug("Failed to parse CONNECT URI {}: Authority missing.",
+                      request_target);
+      return error{sec::invalid_argument, "Malformed CONNECT Request-URI."};
+    }
+    return res;
   }
-  std::string uri;
   if (request_target.front() == '/') {
     // The path must form a valid URI when prefixing a scheme. We don't
     // actually care about the scheme, so just use "nil" here for the
     // validation step.
-    uri = "nil:" + std::string{request_target};
+    res = make_uri("nil:"s.append(request_target));
   } else if (starts_with(request_target, "http")) {
-    uri = request_target;
+    res = make_uri(request_target);
   } else if (method == http::method::options
              && request_target == std::string_view{"*"}) {
     log::net::debug("Server-wide options request received. Converting to '/'.");
-    uri = "nil:/";
+    res = make_uri("nil:/");
   }
-  if (auto maybe_uri = make_uri(uri)) {
-    return std::move(*maybe_uri);
-  } else {
-    auto msg = detail::format("Failed to parse URI {} -> {}", request_target,
-                              maybe_uri.error());
-    log::net::debug("{}", msg);
-    return error{sec::invalid_argument, msg};
+  if (res) {
+    return res;
   }
+  auto msg = detail::format("Failed to parse URI {}: {}", request_target,
+                            res.error());
+  log::net::debug("{}", msg);
+  return error{sec::invalid_argument, msg};
 }
 
 } // namespace
@@ -113,10 +116,10 @@ request_header::parse(std::string_view raw) {
     return {status::bad_request, "Invalid HTTP method."};
   }
   auto [uri_str, version] = split_by(first_line_remainder, " ");
-  if (auto maybe_uri = validate_request_target(method_, uri_str); maybe_uri) {
+  if (auto maybe_uri = parse_request_target(method_, uri_str)) {
     uri_ = std::move(*maybe_uri);
   } else {
-    log::net::debug("Failed to parse URI {} -> {}", uri_str, maybe_uri.error());
+    log::net::debug("Failed to parse URI {}: {}.", uri_str, maybe_uri.error());
     raw_.clear();
     return {status::bad_request, "Malformed Request-URI."};
   }
--- a/libcaf_net/caf/net/http/request_header.cpp
+++ b/libcaf_net/caf/net/http/request_header.cpp
@@ -13,6 +13,46 @@ namespace {
 
 constexpr std::string_view eol = "\r\n";
 
+/// Validate the request target part according to RFC9112.
+caf::expected<uri> validate_request_target(http::method method,
+                                           std::string_view request_target) {
+  if (request_target.empty()) {
+    return make_error(sec::invalid_argument,
+                      "Malformed Request-URI: request target empty.");
+  }
+  if (method == http::method::connect) {
+    if (auto res
+        = make_uri(std::string("nil://") + std::string{request_target})) {
+      return std::move(*res);
+    } else {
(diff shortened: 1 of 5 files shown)


libcaf_net/caf/net/http/request_header.test.cpp
```diff
@@ -35,18 +35,33 @@ TEST("parsing a http request") {
...
+      auto request = detail::format(
+        "{} http://example.com/foo/bar HTTP/1.1\r\n\r\n", method_name);
+      net::http::request_header hdr;
+      hdr.parse(request);
+      THEN("the parsed <enum value> request data is valid") {
+        auto expected = block_parameters<uint8_t>();
+        require(hdr.valid());
+        check_eq(static_cast<uint8_t>(hdr.method()), expected);
+        check_eq(hdr.version(), "HTTP/1.1");
+        check_eq(hdr.path(), "foo/bar");
```
riemass, line 62: The `example.com/foo/bar` request target, when read with the uri, would boil down to `foo/bar` expect `/foo/bar" (missing the root backslash). Is it possible to solve this ans stay backwards compatible?

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+/// Validate the request target part according to RFC9112.
+caf::expected<uri> validate_request_target(http::method method,
+                                           std::string_view request_target) {
+  if (request_target.empty()) {
+    return make_error(sec::invalid_argument,
+                      "Malformed Request-URI: request target empty.");
+  }
+  if (method == http::method::connect) {
+    if (auto res
+        = make_uri(std::string("nil://") + std::string{request_target})) {
```
Neverlord, line 25:
```suggestion
        = make_uri("nil://"s + request_target)) {
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+      return std::move(*res);
+    } else {
```
Neverlord, line 27:
```suggestion
    }
```

`else` after return.

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+      log::net::debug("Failed to parse CONNECT URI {} -> {}", request_target,
```
Neverlord, line 28:
```suggestion
      log::net::debug("Failed to parse CONNECT URI {}: {}", request_target,
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+                      res.error());
+      return error{sec::invalid_argument, "Malformed CONNECT Request-URI."};
+    }
+  }
+  std::string uri;
+  if (request_target.front() == '/') {
+    // The path must form a valid URI when prefixing a scheme. We don't
+    // actually care about the scheme, so just use "nil" here for the
+    // validation step.
+    uri = "nil:" + std::string{request_target};
```
Neverlord, line 38:
```suggestion
    uri = "nil:"s + request_target;
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+    uri = request_target;
+  } else if (method == http::method::options
+             && request_target == std::string_view{"*"}) {
+    log::net::debug("Server-wide options request received. Converting to '/'.");
+    uri = "nil:/";
+  }
+  if (auto maybe_uri = make_uri(uri)) {
+    return std::move(*maybe_uri);
+  } else {
+    auto msg = detail::format("Failed to parse URI {} -> {}", request_target,
```
Neverlord, line 49:
```suggestion
    auto msg = detail::format("Failed to parse URI {}: {}", request_target,
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+    if (auto res
+        = make_uri(std::string("nil://") + std::string{request_target})) {
+      return std::move(*res);
+    } else {
+      log::net::debug("Failed to parse CONNECT URI {} -> {}", request_target,
+                      res.error());
+      return error{sec::invalid_argument, "Malformed CONNECT Request-URI."};
+    }
+  }
+  std::string uri;
```
Neverlord, line 33: We don't really need a string here. We can just call `make_uri` in the `if-else` tree directly.

```suggestion
  caf::expected<uri> res;
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -90,6 +112,14 @@ request_header::parse(std::string_view raw) {
     raw_.clear();
     return {status::bad_request, "Invalid HTTP method."};
   }
+  auto [uri_str, version] = split_by(first_line_remainder, " ");
+  if (auto maybe_uri = validate_request_target(method_, uri_str); maybe_uri) {
```
Neverlord, line 116:
```suggestion
  if (auto maybe_uri = validate_request_target(method_, uri_str)) {
```

libcaf_net/caf/net/http/request_header.hpp
```diff
@@ -61,6 +61,12 @@ class CAF_NET_EXPORT request_header : public header {
     return uri_.fragment();
   }
 
+  /// Returns the authority part of the request.
+  /// @note: Only applicable for CONNECT requests.
```
Neverlord, line 65:
```suggestion
  /// @note: Only present for CONNECT requests.
```

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
...
+caf::expected<uri> validate_request_target(http::method method,
+                                           std::string_view request_target) {
+  if (request_target.empty()) {
+    return make_error(sec::invalid_argument,
+                      "Malformed Request-URI: request target empty.");
+  }
+  if (method == http::method::connect) {
+    if (auto res
+        = make_uri(std::string("nil://") + std::string{request_target})) {
+      return std::move(*res);
```
Neverlord, line 26: Shouldn't we check whether an authority is present here?

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -90,6 +112,14 @@ request_header::parse(std::string_view raw) {
     raw_.clear();
     return {status::bad_request, "Invalid HTTP method."};
   }
+  auto [uri_str, version] = split_by(first_line_remainder, " ");
+  if (auto maybe_uri = validate_request_target(method_, uri_str); maybe_uri) {
+    uri_ = std::move(*maybe_uri);
+  } else {
+    log::net::debug("Failed to parse URI {} -> {}", uri_str, maybe_uri.error());
```
Neverlord, line 119:
```suggestion
    log::net::debug("Failed to parse URI {}: {}", uri_str, maybe_uri.error());
```

libcaf_net/caf/net/http/request_header.test.cpp
Neverlord: As far as the URI standard is concerned, `foo:bar` and `foo:/bar` are different. So we don't want to break our implementation for this use case.

For us, I think it matters mostly for the routing. For example in `http_simple_route_base::exec`. If we can have relative paths now, we need to update (and test) our route implementation(s). Here, I would indeed treat a relative path the same. There is nothing a path could be relative to at this level, so path `foo` and `/foo` should be treated the same.

libcaf_net/caf/net/http/request_header.cpp
```diff
@@ -13,6 +13,46 @@ namespace {
 
 constexpr std::string_view eol = "\r\n";
 
+/// Validate the request target part according to RFC9112.
+caf::expected<uri> validate_request_target(http::method method,
```
Neverlord, line 17: For a function named `validate`, I would expect a `bool` or maybe `error` return type. This actually tries to parse its input, so maybe `parse_request_target` instead?

Neverlord (changes requested):

libcaf_net/caf/net/http/request_header.test.cpp
riemass: All right, makes sense to do the change at router level. I'll fix this in a separate commit and update the PR.

libcaf_net/caf/net/http/request_header.cpp
riemass: Can't add `std::string` and `std::string_view`, but `append` seems to work just fine.

libcaf_net/caf/net/http/request_header.cpp
riemass: Local variable `res` is accessed below for logging purposes.

Neverlord (approved):

PR #2089, opened 2025-06-06
Fix nesting same blocks inside BDD outlines

### LINKED ISSUE #1776: Exception when nesting same blocks inside BDD outlines

Neverlord:
Closes #1776.


riemass (approved):
I was dancing around the `nested_` vector and couldn't figure out how to fix this. 
Also, I confirmed locally that this only enables two WHEN blocks, and THEN blocks, correctly, can't be nested.

PR #2091, opened 2025-06-11
Add support for setting SSL SNI hostname

### LINKED ISSUE #2084: Add SNI support for HTTPS requests to virtual hosts
Reported in the public chatroom.

When making HTTPS requests to virtual hosts (e.g., `https://feodotracker.abuse.ch`), the client receives a `421 Misdirected Request` error. The server presents the wrong TLS certificate because it doesn't know which virtual host is being targeted. Adding the `Host` header field to the HTTP request didn't solve the issue, as `caf::net` probably needs to call `SSL_set_tlsext_host_name()` during TLS handshake to enable Server Name Indication (SNI) support, allowing the server to present the correct certificate for the requested domain.

The error returned in the reported example also links to the following documentation: https://www.fastly.com/documentation/guides/concepts/errors/#routing-errors

riemass:
Closes #2084 

No idea how to unit test this. I did try fetching the sites from the ticket and can confirm it works.

### DIFF
--- a/libcaf_net/caf/net/ssl/context.test.cpp
+++ b/libcaf_net/caf/net/ssl/context.test.cpp
@@ -31,6 +31,7 @@ TEST("constructing and setting values in the context object") {
       auto fd_pair = caf::net::make_stream_socket_pair();
       require(fd_pair.has_value());
       auto conn = ctx.new_connection(fd_pair->first);
+      auto g = caf::net::socket_guard{fd_pair->second};
       if (check_has_value(conn)) {
         check_eq(conn->sni_hostname(), host);
       }


libcaf_net/caf/net/ssl/connection.hpp
```diff
@@ -56,6 +56,12 @@ class CAF_NET_EXPORT connection {
   /// @param ret The (negative) result from the preceding call.
   std::string last_error_string(ptrdiff_t ret) const;
 
+  // -- SNI support -----------------------------------------------------------
+
+  /// Sets the SNI hostname for this connection before handshake.
+  /// Must be called before connect().
+  bool set_sni_hostname(const char* hostname) noexcept;
```
Neverlord:
```suggestion
  [[nodiscard]] bool sni_hostname(const char* hostname) noexcept;
```

Also: no getter on the connection object?

libcaf_net/caf/net/ssl/context.hpp
```diff
@@ -235,6 +236,13 @@ class CAF_NET_EXPORT context {
     return use_private_key_file(path.c_str(), file_format);
   }
 
+  /// Sets the SNI (Server Name Indication) hostname for client connections
+  /// created from this context.
+  void sni_hostname(std::string hostname) noexcept;
+
+  /// Returns the optional SNI hostname. Empty if SNI is not configured.
```
Neverlord:
```suggestion
  /// Returns the optional SNI hostname or `nullptr` if not configured.
```

Neverlord (changes requested):
I agree that testing the SNI functionality itself would be tough in a unit test. However, at least we should have basic functionality tests in place. Like calling the getter without prior settter (return `nullptr`) and calling setter and then getter (should return whatever the setter received as argument).

libcaf_net/caf/net/ssl/context.test.cpp
```diff
@@ -10,6 +10,33 @@
...
+    auto host = "feodotracker.abuse.ch"s;
+    ctx.sni_hostname(host);
+    check_eq(ctx.sni_hostname(), host);
+    SECTION("new connections use the selected SNI hostname") {
+      auto fd_pair = caf::net::make_stream_socket_pair();
+      require(fd_pair.has_value());
+      auto conn = ctx.new_connection(fd_pair->first);
+      if (check_has_value(conn)) {
+        check_eq(conn->sni_hostname(), host);
+      }
```
Neverlord: This test is leaking the second socket handle. Please put a socket guard for `fd_pair->second` after the `require` (`first` will be taken care of by the connection object).

Neverlord (changes requested):

libcaf_net/caf/net/ssl/context.test.cpp
riemass: What a mistake. Fixed.

Neverlord (approved):

PR #2103, opened 2025-06-25
Add internal API for adding hidden actors

Neverlord:


riemass (changes requested):
I know this is a simple enough change, but maybe leave a small unit test just to confirm the actor is spawned.
Also, as a followup could we move the printer actor to be a hidden actor?

Neverlord:
Thanks for calling out my laziness! 😉

We already have the daemons API, though. It overlaps with the use case I had in mind for this and it's actually more powerful. So I'll just use the existing API instead.

PR #2108, opened 2025-07-10
Add HTTP multipart reader

Neverlord:
Relates #2073.

Adds a convenience API for reading an HTTP multipart.


libcaf_net/caf/net/http/multipart_reader.hpp
```diff
@@ -0,0 +1,72 @@
...
+    return mime_type_;
+  }
+
+  /// Parses the multipart content and returns the parts.
+  /// @param ok If not `nullptr`, set to `true` if parsing succeeded, `false`
+  ///          otherwise. This allows to distinguish between parsing errors and
+  ///          empty multipart content.
+  /// @returns The parts of the multipart content or an empty vector if parsing
+  ///          failed.
+  std::vector<part> parse(bool* ok = nullptr);
```
riemass, line 44: Why not return an expected of an optional?

libcaf_net/caf/net/http/multipart_reader.hpp
```diff
@@ -0,0 +1,72 @@
...
+  using consume_fn = void (*)(void*, http::header&&, const_byte_span);
+
+  /// Parses multipart content and calls the given function for each part.
+  /// @param fn Function to call for each part
+  /// @param obj User data passed to the function
+  /// @returns true if parsing succeeded, false on any error
+  bool do_parse(consume_fn fn, void* obj);
+
+  /// Provides access to the HTTP header and body.
+  const responder* res_;
```
riemass: We don't need to hold the whole responder in the state, just the body which is a `const_byte_span`. 

This also means that we could have an additional constructor for `multipart_reader` from the header and body, which is what the `http::upper_layer::server` receives in the consume call. Not eliminatory but might be a nice addition.

libcaf_net/caf/net/http/multipart_reader.test.cpp
```diff
@@ -0,0 +1,232 @@
...
+  std::unique_ptr<responder> res;
+  auto setup = [&hdr, &body, &res](std::string_view header,
+                                   std::string_view body_content) {
+    hdr.parse(header);
+    body = body_content;
+    res = std::make_unique<responder>(&hdr, as_bytes(make_span(body)), nullptr);
+  };
+  auto as_ascii = [](const_byte_span bytes) {
+    return std::string_view{reinterpret_cast<const char*>(bytes.data()),
+                            bytes.size()};
```
riemass:
```suggestion
    return to_string_view(bytes);
    };
```

libcaf_net/caf/net/http/multipart_reader.cpp
```diff
@@ -0,0 +1,121 @@
...
+      if (candidate.substr(0, separator.size()) == separator) {
+        return pos;
+      }
+      offset = pos + boundary_delimiter.size();
+    }
+    return std::string_view::npos;
+  };
+  // Parse the payload into parts.
+  std::string_view payload(reinterpret_cast<const char*>(res_->body().data()),
+                           res_->body().size());
```
riemass:
```suggestion
  auto payload = to_string_view(res_->body());
```

riemass (changes requested):
Some nits and suggestions.

libcaf_net/caf/net/http/multipart_reader.hpp
Neverlord: Because I think most of time, people don't care about the distinction between nothing and something unexpected. If they really do care about it, they can pass this flag here or use `for_each` instead.

If you are receiving a multi-form input, you'll most likely iterate over the parts, look at the `Content-Disposition` headers (or something similar) and then check that all required/expected fields are present. Designing the API in this way makes it more convenient to use if you are checking for required fields anyways.

PR #2110, opened 2025-07-11
Add HTTP multipart writer

### LINKED ISSUE #2073: Support HTTP `multipart` requests

Neverlord:
Closes #2073.

Adds a convenience API for writing an HTTP multipart body.

### DIFF
--- a/libcaf_net/caf/net/http/multipart_writer.hpp
+++ b/libcaf_net/caf/net/http/multipart_writer.hpp
@@ -1,0 +1,106 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/main/LICENSE.
+
+#pragma once
+
+#include "caf/byte_buffer.hpp"
+#include "caf/byte_span.hpp"
+#include "caf/detail/net_export.hpp"
+
+#include <string_view>
+
+namespace caf::net::http {
+
+/// A utility class for creating multipart content for HTTP requests.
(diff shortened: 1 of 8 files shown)


libcaf_net/caf/net/http/multipart_writer.hpp
```diff
@@ -0,0 +1,106 @@
...
+  ///                    The function object must be invocable with a
+  ///                    `header_builder&` argument.
+  template <class AddHeadersFn>
+  void append(const_byte_span payload, AddHeadersFn&& add_headers) {
+    using add_headers_t = std::decay_t<AddHeadersFn>;
+    auto fn = [](void* ctx, header_builder& builder) {
+      (*static_cast<add_headers_t*>(ctx))(builder);
+    };
+    do_append(payload, fn, &add_headers);
+  }
```
riemass: Could we rewrite this in terms of capturing the `add_headers` object inside the lambda, and turning the `add_headers_fn` passed into `do_append` into a callback type, similar to what we did with `multipart_reader::for_each`.

riemass (changes requested):
Found a nit. I believe we can avoid the `void*`

libcaf_net/caf/net/http/multipart_writer.hpp
Neverlord: Thanks. I've totally blanked on that one, good thing you caught it!

PR #2114, opened 2025-07-15
Remove `monotonic_buffer_resource` backport

### LINKED ISSUE #2113: Remove `caf::detail::monotonic_buffer_resource`
Looks like `std::pmr::monotonic_buffer_resource` is widely available now, even on Apple's Clang. So it's time to give our backport version the axe.

It looks like it's not quite trivial, because of `caf::detail::monotonic_buffer_resource::allocator`. Using `std::pmr::polymorphic_allocator` instead didn't quite work, because it's not assignable (unlike our own version). So we'll probably have to lift our allocator to a separate class, then drop our own `monotonic_buffer_resource` and once that's done we can see how to get rid of our custom allocator class as well.

Neverlord:
Closes #2113.


libcaf_core/caf/detail/mbr_list.hpp
```diff
@@ -98,7 +97,7 @@ class mbr_list {
 
   using node_type = mbr_list_node<value_type>;
 
-  using allocator_type = monotonic_buffer_resource::allocator<node_type>;
+  using allocator_type = std::pmr::polymorphic_allocator<node_type>;
```
riemass, line 100: I couldn't find where the alias is used?

riemass (changes requested):
Just a nit here.

libcaf_core/caf/detail/mbr_list.hpp
Neverlord: It's used in three places (destructor, `push_back` and `emplace_back`). They didn't show up in the diff because those lines didn't change.

riemass: Resolving, thanks.

riemass (approved):

PR #2117, opened 2025-07-28
Fix memory alignment for actor_control_block

riemass:
Fix undefined behavior that arises from unaligned memory from malloc. An actor_control_block should be aligned to cache line size (64) because of the inherent alignment of the default mailbox.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ is based on [Keep a Changelog](https://keepachangelog.com).
 
 - Errors that arise during the `with` DSL setup of servers and clients now
   properly call `on_error` (#2026).
+- Fix UBSan finding regarding non-aligned memory allocation when spawning
+  actors.
 
 ### Removed
 


libcaf_core/caf/detail/aligned_alloc.hpp
```diff
@@ -0,0 +1,23 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/main/LICENSE.
+
+#pragma once
+
+#include "caf/detail/core_export.hpp"
+#include <cstdlib>
```
Neverlord: If you are including this for `size_t` -> `cstddef`.

Neverlord (changes requested):

libcaf_core/caf/detail/aligned_alloc.hpp
riemass: Yes, here is for `size_t`. The cpp file need it for malloc implementations.  Will fix.

libcaf_core/caf/make_actor.hpp
```diff
@@ -30,6 +31,11 @@ struct make_actor_util {
   }
 };
 
+// A constexpr function to align a value to the next multiple of another value.
+constexpr size_t align_to_next_multiple(size_t value, size_t alignment) {
+  return (value + alignment - 1) / alignment * alignment;
+}
```
Neverlord: That shouldn't be here. It's an internal limitation of `aligned_alloc` on Linux and should be dealt with there.

riemass: I'll wrap it accordingly. 
Note: Requested review before seeing the comment.

riemass:
Ready for review again.

PR #2118, opened 2025-07-30
Enable C++20 and use `requires` instead of `enable_if`

Neverlord:
- CAF now requires C++20 to build
- replace all uses of `enable_if` with `requires`
- use concepts from the STL where it makes the code more readable
- rewrite the `flow::has_impl_include` check to not require `enable_if` while still giving a clear and concise error message to the user

### DIFF
--- a/libcaf_core/caf/config_value_writer.cpp
+++ b/libcaf_core/caf/config_value_writer.cpp
@@ -342,7 +342,8 @@ public:
   }
 
   template <class T>
-  std::enable_if_t<std::is_integral_v<T>, bool> value(T x) {
+    requires std::is_integral_v<T>
+  bool value(T x) {
     if constexpr (std::is_same_v<T, bool>) {
       return push(config_value{x});
     } else if constexpr (std::is_same_v<T, uint64_t>) {
--- a/libcaf_core/caf/detail/type_traits.hpp
+++ b/libcaf_core/caf/detail/type_traits.hpp
@@ -199,12 +199,11 @@ template <class T>
 class is_iterable {
   // this horrible code would just disappear if we had concepts
   template <class C>
-  static bool sfinae(
-    C* cc,
-    // check if 'C::begin()' returns a forward iterator
-    std::enable_if_t<is_forward_iterator_v<decltype(cc->begin())>>* = nullptr,
+    requires is_forward_iterator_v<decltype(std::declval<C*>()->begin())>
+  static bool sfinae(C* cc) {
     // check if begin() and end() both exist and are comparable
-    decltype(cc->begin() != cc->end())* = nullptr);
+    return cc->begin() != cc->end();
+  }
 
   // SFNINAE default
   static void sfinae(void*);
@@ -432,9 +431,6 @@ struct value_type_of<T*> {
 template <class T>
 using value_type_of_t = typename value_type_of<T>::type;
 
-template <class F, class T>
-using is_handler_for_ef = typename std::enable_if<is_handler_for_v<F, T>>::type;
-
 // Checks whether T has a member function named `push_back` that takes an
 // element of type `T::value_type`.
 template <class T>
@@ -495,7 +491,7 @@ CAF_HAS_ALIAS_TRAIT(mapped_type);
 
 CAF_HAS_ALIAS_TRAIT(handle_type);
 
-// -- constexpr functions for use in enable_if & friends -----------------------
+// -- constexpr functions for use in require clauses ---------------------------
 
 /// Checks whether T behaves like `std::map`.
 template <class T>
--- a/libcaf_core/caf/intrusive_ptr.hpp
+++ b/libcaf_core/caf/intrusive_ptr.hpp
@@ -231,29 +231,29 @@ bool operator!=(std::nullptr_t, const intrusive_ptr<T>& x) {
 
 /// @relates intrusive_ptr
 template <class T, class U>
-std::enable_if_t<detail::is_comparable<T*, U*>::value, bool>
-operator==(const intrusive_ptr<T>& lhs, const U* rhs) {
+  requires detail::is_comparable<T*, U*>::value
+bool operator==(const intrusive_ptr<T>& lhs, const U* rhs) {
   return lhs.get() == rhs;
 }
 
 /// @relates intrusive_ptr
 template <class T, class U>
-std::enable_if_t<detail::is_comparable<T*, U*>::value, bool>
-operator==(const T* lhs, const intrusive_ptr<U>& rhs) {
+  requires detail::is_comparable<T*, U*>::value
+bool operator==(const T* lhs, const intrusive_ptr<U>& rhs) {
   return lhs == rhs.get();
 }
 
 /// @relates intrusive_ptr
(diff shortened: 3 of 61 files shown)


libcaf_core/caf/intrusive_ptr.hpp
```diff
@@ -231,45 +231,45 @@ bool operator!=(std::nullptr_t, const intrusive_ptr<T>& x) {
 
 /// @relates intrusive_ptr
 template <class T, class U>
-std::enable_if_t<detail::is_comparable<T*, U*>::value, bool>
-operator==(const intrusive_ptr<T>& lhs, const U* rhs) {
+  requires detail::is_comparable<T*, U*>::value
```
Neverlord, line 234: @riemass FYI, I'm already working on a followup that will clean places where we still use `::value`.

libcaf_core/caf/telemetry/counter.hpp
```diff
@@ -10,6 +10,7 @@
 #include "caf/telemetry/gauge.hpp"
 #include "caf/telemetry/label.hpp"
 
+#include <concepts>
```
riemass: Mind why the include here? From the docs I think that `is_same` lives in type traits, whereby `same_as` is in concepts.

libcaf_core/caf/config_value_writer.cpp
```diff
@@ -342,7 +342,8 @@ class impl : public save_inspector_base<impl>,
   }
 
   template <class T>
```
riemass, line 344:
```suggestion
  template <std::integral T>
```
Also remove requires line and include the concepts library.

libcaf_core/caf/detail/type_traits.hpp
```diff
@@ -199,12 +199,11 @@ template <class T>
 class is_iterable {
   // this horrible code would just disappear if we had concepts
   template <class C>
-  static bool sfinae(
-    C* cc,
-    // check if 'C::begin()' returns a forward iterator
-    std::enable_if_t<is_forward_iterator_v<decltype(cc->begin())>>* = nullptr,
+    requires is_forward_iterator_v<decltype(std::declval<C*>()->begin())>
```
riemass, line 202: There is a `std::forward_iterator` in concepts, could it be used here?

libcaf_core/caf/intrusive_ptr.hpp
riemass: Thanks for the heads up. 
I don't mind the occasional `::value`. I guess we will be moving to things like `equality_comparable_with` in the future.

libcaf_core/caf/log/event.hpp
```diff
@@ -375,11 +371,10 @@ class CAF_CORE_EXPORT event_sender {
...
   template <class SubFieldsInitializer>
-  auto field(std::string_view key, SubFieldsInitializer&& init) //
-    -> std::enable_if_t<
-      std::is_same_v<decltype(init(std::declval<event_fields_builder&>())),
-                     void>,
-      event_sender&&> {
+  event_sender&& field(std::string_view key, SubFieldsInitializer&& init)
+    requires(
+      std::same_as<decltype(init(std::declval<event_fields_builder&>())), void>)
+  {
```
riemass: Couldn't we use `std::invocable<event_fields_builder&>` like in the example above?

libcaf_core/caf/response_handle.hpp
```diff
@@ -88,31 +91,32 @@ class response_handle {
...
+    requires traits::is_non_blocking
+  void then(F f) {
     auto self = self_;
     then(std::move(f), [self](error& err) { self->call_error_handler(err); });
   }
 
-  template <class T>
-  flow::assert_scheduled_actor_hdr_t<flow::single<T>> as_single() && {
+  template <class T, bool = flow::assert_has_impl_include<T>>
+  auto as_single() && {
```
riemass, line 102: Can't follow here all the way through. Is the bool acting as some sort of SFINAE guard?

libcaf_core/caf/scheduled_actor.hpp
```diff
@@ -516,27 +515,20 @@ class CAF_CORE_EXPORT scheduled_actor : public abstract_scheduled_actor,
   /// Utility function that swaps `f` into a temporary before calling it
   /// and restoring `f` only if it has not been replaced by the user.
   template <class F, class... Ts>
-  auto call_handler(F& f, Ts&&... xs) -> std::enable_if_t<
-    !std::is_same_v<decltype(f(std::forward<Ts>(xs)...)), void>,
-    decltype(f(std::forward<Ts>(xs)...))> {
+  auto call_handler(F& f, Ts&&... xs) {
```
riemass: Shoud we add a `requires std::invocable<F, xs...>` for better compile errors?

riemass (changes requested):
Some nits and question here and there. Mostly a very clean transition without breaking the API, so forget what I said earlier in chat. I saw some nice things and look forward to writing code with concepts.

libcaf_core/caf/response_handle.hpp
Neverlord: We actually don't want SFINAE here. Using `enable_if` prior was arguably wrong to begin with. What we want is that the compiler checks whether the proper header was included when instantiating this template and otherwise we'll abort compilation (via `static_assert`).

Our new `flow::assert_has_impl_include<T>` does exactly that. It's templated, so the compiler will only instantiate it once the function is actually being called. And once instantiated, this will either evaluate to some boolean expression (`true`, but we really don't care about the value) or run into the `static_assert` telling the user to include the proper header.

libcaf_core/caf/detail/type_traits.hpp
Neverlord: I should probably have just ignored this file, since we'll replace everything in it with concepts anyways. I just wanted to get rid of all `enable_if` in the first round. Implementing (and using) concepts is round 2. I should've stated that in the PR description.

libcaf_core/caf/config_value_writer.cpp
Neverlord: Absolutely, that's the followup. 🙂

libcaf_core/caf/log/event.hpp
Neverlord: Yep. 👍

libcaf_core/caf/scheduled_actor.hpp
Neverlord: We should. Done.

libcaf_core/caf/telemetry/counter.hpp
Neverlord: Probably some refactoring artifact.

libcaf_core/caf/response_handle.hpp
riemass: Perfect, thanks for explaining.

libcaf_core/caf/config_value_writer.cpp
riemass: I was used in a few places in this PR as well, but makes more sense to go in more detail with a followup.

PR #2121, opened 2025-08-01
Modernize meta-programming utilities

Neverlord:
@riemass here's the promised followup to #2118. Basically a rewrite of our metaprogramming utilities from `type_traits.hpp`, moving everything to concepts. A few notes:

- `is_comparable`: the STL has `equality_comparable_with` but this concept is more strict and the actually doesn't compile with the STL concept. That's something we might pick up at some point to dig deeper, for now I've just re-implemented our old trait class as concept.
- changes to `intrusive/forward_iterator`: turns out this class actually violated the `forward_iterator` concept from the STL, so I've fixed that so we can make use of the concept.
- `callable_trait`: this was the only trait left after re-implementing everything else as concepts. So I've pushed it to its own file. It's also a rather complicated trait, so I think it deserves its own header anyways.


libcaf_core/caf/detail/concepts.hpp
```diff
@@ -0,0 +1,348 @@
...
+/// Checks whether the inspector has an `opaque_value` overload for `T`.
+template <class Inspector, class T>
+concept accepts_opaque_value = requires(Inspector& f, T& x) {
+  { f.opaque_value(x) } -> std::same_as<bool>;
+};
+
+/// Checks whether `T` is primitive, i.e., either an arithmetic type or
+/// convertible to one of STL's string types.
+template <class T, bool IsLoading>
+struct is_builtin_inspector_type_oracle {
```
riemass: Copy-pasted docstring left.

riemass (changes requested):
Found a small docstirng duplicate. Otherwise very nice.

riemass (approved):

PR #2122, opened 2025-08-01
Switch to std::span

Neverlord:
The type `caf::span` became obsolete when switching to C++20. Consequently, we drop our backport version of the class and use `std::span` instead.

We'll leave `span.hpp` in the code base until CAF 3.0, but `caf::span` simply became a deprecated alias for `std::span`. The `make_span` function overloads have been deprecated as well and are no longer used in CAF.


libcaf_net/caf/net/http/v1.cpp
```diff
@@ -53,7 +53,7 @@ std::pair<std::string_view, byte_span> split_header(byte_span bytes) {
   } else {
     auto offset = static_cast<size_t>(std::distance(bytes.begin(), i));
     offset += end_of_header.size();
-    return {std::string_view{reinterpret_cast<const char*>(bytes.begin()),
+    return {std::string_view{reinterpret_cast<const char*>(bytes.data()),
```
riemass, line 56: We have the `to_string_view` function for this. 
I grepped the codebase for `reinterpret_cast<const char*>` and it seems we have this on multiple places, so it might be a good followup to clean it up.

PR #2128, opened 2025-08-11
Introduce new `error_code_enum` concept

Neverlord:
Get rid of the legacy-style `is_error_code_enum_v` and use a concept instead.


libcaf_core/caf/is_error_code_enum.hpp
```diff
@@ -17,7 +15,7 @@ struct is_error_code_enum {
 
 /// @relates is_error_code_enum
 template <class T>
-constexpr bool is_error_code_enum_v = is_error_code_enum<T>::value;
+concept error_code_enum = is_error_code_enum<T>::value;
```
riemass: Looks so nice. Just as a reminder, the header should be renamed.

Issue #2009, opened 2025-02-06
Implement fan_out_request on the new mail API


Neverlord:
Before we can remove the `requester` mixin (see #1873), we first need to re-implement `fan_out_request` for the new mail API.

riemass:
Checklist to track the progress:
- [x] event based fan out request,
- [x] event based scheduled fan out request,
- [ ] ~~blocking fan out request~~,
- [x] deprecate the mixin based fan out request and update the example.

PR #2132, opened 2025-09-12
Implement blocking fan_out_request

riemass:
Implemented the blocking stuff. 
There was a small issue I've had with `do_receive`, since it receives a single message. It behaved weirdly with `timespan`s, so I switched to using `time_point`s. Alternatively, I could've implemented a new `receive_cond` that awaits `n` messages, but then I'd have to move from `abstract_blocking_actor*` to the concrete `blocking_actor*` implementation. 
Relates #2009.

### DIFF
--- a/libcaf_core/caf/blocking_fan_out_request.test.cpp
+++ b/libcaf_core/caf/blocking_fan_out_request.test.cpp
@@ -1,0 +1,816 @@
+// This file is part of CAF, the C++ Actor Framework. See the file LICENSE in
+// the main distribution directory for license terms and copyright or visit
+// https://github.com/actor-framework/actor-framework/blob/main/LICENSE.
+
+#include "caf/test/test.hpp"
+
+#include "caf/actor_system.hpp"
+#include "caf/actor_system_config.hpp"
+#include "caf/blocking_actor.hpp"
+#include "caf/blocking_mail.hpp"
+#include "caf/event_based_actor.hpp"
+#include "caf/scoped_actor.hpp"
+#include "caf/typed_actor.hpp"
+#include "caf/typed_event_based_actor.hpp"
+
+#include <iostream>
+#include <numeric>
+
+using namespace caf;
+using namespace std::literals;
+
+namespace {
+
+struct config : actor_system_config {
+  config() {
+    set("caf.scheduler.max-threads", 2u);
+  }
+};
+
+struct fixture {
+  config cfg;
+  actor_system sys{cfg};
+};
+
+template <typename Fn>
+actor make_server(caf::actor_system& sys, Fn fn) {
+  auto sf = [fn]() -> behavior {
+    return {
+      [fn](int x, int y) { return fn(x, y); },
+    };
+  };
+  return sys.spawn(sf);
+}
+
+// Typed actor interfaces for the fan_out_request test.
+using typed_worker_actor = typed_actor<result<int>(int, int)>;
+using typed_worker_behavior = typed_worker_actor::behavior_type;
+using typed_worker_two_values_actor = typed_actor<result<int, int>(int, int)>;
+using typed_worker_two_values_behavior
+  = typed_worker_two_values_actor::behavior_type;
+using typed_worker_void_actor = typed_actor<result<void>(int, int)>;
+using typed_worker_void_behavior = typed_worker_void_actor::behavior_type;
+
+template <typename Fn>
+typed_worker_actor make_typed_server(caf::actor_system& sys, Fn fn) {
+  auto sf = [fn]() -> typed_worker_behavior {
+    return {[fn](int x, int y) { return fn(x, y); }};
+  };
+  return sys.spawn(sf);
+}
+
+template <typename Fn>
+typed_worker_two_values_actor
+make_typed_server_two_values(caf::actor_system& sys, Fn fn) {
+  auto sf = [fn]() -> typed_worker_two_values_behavior {
+    return {[fn](int x, int y) -> result<int, int> { return fn(x, y); }};
+  };
+  return sys.spawn(sf);
+}
+
+template <typename Fn>
+typed_worker_void_actor make_typed_server_void(caf::actor_system& sys, Fn fn) {
+  auto sf = [fn]() -> typed_worker_void_behavior {
+    return {[fn](int x, int y) -> result<void> {
+      fn(x, y);
+      return {};
+    }};
+  };
+  return sys.spawn(sf);
+}
+
+} // namespace
+
+WITH_FIXTURE(fixture) {
+
+TEST("fan_out_request with a single result") {
+  scoped_actor self{sys};
+  std::vector<actor> workers{
+    make_server(sys, [](int x, int y) { return x + y; }),
+    make_server(sys, [](int x, int y) { return x + y; }),
+    make_server(sys, [](int x, int y) { return x + y; }),
+  };
+  auto sum = std::make_shared<int>(0);
+  auto err = std::make_shared<error>();
+  SECTION("receive with policy select_all") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_all_tag)
+      .receive(
+        [this, sum](std::vector<int> results) {
+          for (auto result : results)
+            check_eq(result, 3);
+          *sum = std::accumulate(results.begin(), results.end(), 0);
+        },
+        [err](error& e) { *err = std::move(e); });
+    check_eq(*sum, 9);
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_all") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_all_tag)
+                 .receive<int>();
+    if (check_has_value(res)) {
+      check_eq(res->size(), 3u);
+      for (const auto& val : *res) {
+        check_eq(val, 3);
+      }
+    }
+  }
+  SECTION("receive with policy select_any") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([sum](int result) { *sum = result; },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*sum, 3);
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive<int>();
+    check_eq(res, 3);
+  }
+}
+
+TEST("fan_out_request with void result") {
+  scoped_actor self{sys};
+  std::vector<actor> workers{
+    make_server(sys, [](int, int) {}),
+    make_server(sys, [](int, int) {}),
+    make_server(sys, [](int, int) {}),
+  };
+  auto ran = std::make_shared<bool>(false);
+  auto err = std::make_shared<error>();
+  SECTION("receive with policy select_all") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_all_tag)
+      .receive([ran]() { *ran = true; },
+               [err](error& e) { *err = std::move(e); });
+    check(*ran);
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_all") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_all_tag)
+                 .receive();
+    check_has_value(res);
+  }
+  SECTION("receive with policy select_any") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([ran]() { *ran = true; },
+               [err](error& e) { *err = std::move(e); });
+    check(*ran);
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive();
+    check_has_value(res);
+  }
+}
+
+TEST("fan_out_request with multiple results") {
+  scoped_actor self{sys};
+  std::vector<actor> workers{
+    make_server(sys, [](int x, int y) { return make_message(y, x); }),
+    make_server(sys, [](int x, int y) { return make_message(y, x); }),
+    make_server(sys, [](int x, int y) { return make_message(y, x); }),
+  };
+  auto swapped_values = std::make_shared<std::vector<std::pair<int, int>>>();
+  auto single_result = std::make_shared<std::pair<int, int>>();
+  auto err = std::make_shared<error>();
+  SECTION("receive with policy select_all") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_all_tag)
+      .receive(
+        [swapped_values](std::vector<std::tuple<int, int>> results) {
+          for (auto result : results) {
+            swapped_values->emplace_back(std::get<0>(result),
+                                         std::get<1>(result));
+          }
+        },
+        [err](error& e) { *err = std::move(e); });
+    check_eq(swapped_values->size(), 3u);
+    for (const auto& pair : *swapped_values) {
+      check_eq(pair.first, 2);
+      check_eq(pair.second, 1);
+    }
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_all") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_all_tag)
+                 .receive<int, int>();
+    if (check_has_value(res)) {
+      check_eq(res->size(), 3u);
+      for (const auto& pair : *res) {
+        check_eq(std::get<0>(pair), 2);
+        check_eq(std::get<1>(pair), 1);
+      }
+    }
+  }
+  SECTION("receive with policy select_any") {
+    self->mail(3, 5)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive(
+        [single_result](int first, int second) {
+          *single_result = std::make_pair(first, second);
+        },
+        [err](error& e) { *err = std::move(e); });
+    check_eq(single_result->first, 5);
+    check_eq(single_result->second, 3);
+    check_eq(*err, error{});
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(3, 5)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive<int, int>();
+    if (check_has_value(res)) {
+      check_eq(std::get<0>(*res), 5);
+      check_eq(std::get<1>(*res), 3);
+    }
+  }
+}
+
+TEST("fan_out_request with type mismatch") {
+  scoped_actor self{sys};
+  std::vector<actor> workers{
+    make_server(sys, [](int x, int y) { return std::to_string(x + y); }),
+    make_server(sys, [](int x, int y) { return std::to_string(x + y); }),
+    make_server(sys, [](int x, int y) { return std::to_string(x + y); }),
+  };
+  auto err = std::make_shared<error>();
+  SECTION("receive with policy select_all") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_all_tag)
+      .receive(
+        [this](std::vector<int> results) {
+          fail("expected an error, got: {}", results);
+        },
+        [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::unexpected_response));
+  }
+  SECTION("receive with policy select_any") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([this](
+                 int results) { fail("expected an error, got: {}", results); },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::unexpected_response));
+  }
+}
+
+TEST("fan_out_request with timeout") {
+  scoped_actor self{sys};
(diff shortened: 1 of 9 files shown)


libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,726 @@
...
+    make_server(sys, [](int x, int y) { return x + y; }),
+    make_server(sys, [](int x, int y) { return x + y; }),
+  };
+  auto sum = std::make_shared<int>(0);
+  auto err = std::make_shared<error>();
+  SECTION("receive with policy select_all") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_all_tag)
+      .receive(
+        [=, this](std::vector<int> results) {
```
Neverlord: Please avoid default captures. We recently cleaned up a lot of those and don't want to re-introduce them.

```suggestion
        [this, sum](std::vector<int> results) {
```

libcaf_core/caf/blocking_fan_out_response_handle.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+  /// Deadline for the response.
+  actor_clock::time_point deadline;
+};
+
+/// This helper class identifies an expected response message and enables
+/// `request(...).receive(...)`.
+template <class Policy, class... Results>
+class blocking_fan_out_response_handle {
+public:
```
Neverlord, line 90:
```C++
  static constexpr bool is_select_all = ...;

  static constexpr bool is_select_any = ...;

  static_assert(is_select_all || is_select_any);
```

Then use `requires` to dispatch instead of using `if constexpr` inside the function. Makes it IMHO easier to follow what's going than two big if-else branches.

libcaf_core/caf/blocking_fan_out_response_handle.hpp
```diff
@@ -0,0 +1,276 @@
...
+          bld.set_value(std::move(args));
+        },
+        [&bld](error& err) { bld.set_error(std::move(err)); });
+      return std::move(bld.result);
+    } else {
+      detail::expected_builder<Results...> bld;
+      std::move(*this).receive(
+        [&bld](Results... args) { bld.set_value(std::move(args)...); },
+        [&bld](error& err) { bld.set_error(std::move(err)); });
+      return std::move(bld.result);
```
Neverlord: I'm not convinced that we need `expected_builder`. In fact, I'd like to see that go.

```suggestion
      expected<Results...> result;
      std::move(*this).receive(
        [&bld](Results&... args) { result.emplace(std::move(args)...); },
        [&bld](error& err) { result = std::move(err)); });
      return result;
```

The only special case we would need to worry about is constructing `expected<void>`. That can be done in a utility function, though:

```c++
  static auto make_expected() {
    if constexpr (sizeof...(Ts) == 0) {
      return expected<void>{};
    }
    ...
  }
```

Neverlord (changes requested):

libcaf_core/caf/blocking_fan_out_response_handle.hpp
riemass: Another case is that having multiple results requires an tuple<Results...>, which the builder solves. But I'll refactor to account for it.

riemass: I still need to remove the expected builder from `blocking_response_handle`, but since that class still uses a specialization for <message> that will be a follow-up.

riemass:
Aside the review comments I added the dynamically typed receive member.

libcaf_core/caf/blocking_fan_out_response_handle.hpp
```diff
@@ -0,0 +1,369 @@
...
+  template <class... Ts>
+  auto receive() &&
+    requires is_dynamically_typed
+  {
+    return std::move(*this).template do_receive<Ts...>();
+  }
+
+private:
+  template<class... Ts>
+  auto do_receive() &&
```
Neverlord, line 193: Please rebase and then merge these together like you did in #2175 before I do another review round.

riemass: Sure, I did for `select_any`. Merging together the `select_all` would be a bit harder because of the message handler signature and vector-tuple wrapper. I think any meta-programming utilities I add wouldn't benefit the readability.

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+  actor_system sys{cfg};
+};
+
+template <typename Fn>
+actor make_server(caf::actor_system& sys, Fn fn) {
+  auto sf = [&]() -> behavior {
+    return {
+      [&](int x, int y) { return fn(x, y); },
+    };
+  };
```
Neverlord: UB: the references are long gone if the handler runs.

```suggestion
  auto sf = [fn]() -> behavior {
    return {
      [fn](int x, int y) { return fn(x, y); },
    };
  };
```

Same in the typed version.

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive<int, int>();
+    if (check_has_value(res)) {
+      check_eq(std::get<0>(*res), 5);
+      check_eq(std::get<1>(*res), 3);
+    }
+  }
+}
+
+TEST("send fan_out_request messages with invalid setups") {
```
Neverlord:
```suggestion
TEST("fan_out_request with type mismatch") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+        [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::unexpected_response));
+  }
+  SECTION("receive with policy select_any") {
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([this](
+                 int results) { fail("expected an error, got: {}", results); },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::all_requests_failed));
```
Neverlord: Why do we get a generic error here? Can we get the error from the last attempt instead? Do the non-blocking versions behave the same way?

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+    }};
+  };
+  return sys.spawn(sf);
+}
+
+} // namespace
+
+WITH_FIXTURE(fixture) {
+
+TEST("send fan_out_request messages that return a result") {
```
Neverlord:
```suggestion
TEST("fan_out_request with a single result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive<int>();
+    check_eq(res, 3);
+  }
+}
+
+TEST("send fan_out_request messages with void result") {
```
Neverlord:
```suggestion
TEST("fan_out_request with void result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive();
+    check_has_value(res);
+  }
+}
+
+TEST("send fan_out_request messages that returns two values") {
```
Neverlord:
```suggestion
TEST("fan_out_request with multiple results") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([this](
+                 int results) { fail("expected an error, got: {}", results); },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::all_requests_failed));
+  }
+}
+
+TEST("timeout a fan_out_request") {
```
Neverlord:
```suggestion
TEST("fan_out_request with timeout") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+      .receive([sum](int result) { *sum = result; },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::all_requests_failed));
+  }
+  for (const auto& worker : workers) {
+    self->send_exit(worker, exit_reason::user_shutdown);
+  }
+}
+
+TEST("send fan_out_request messages that return a result using typed actors") {
```
Neverlord:
```suggestion
TEST("typed fan_out_request with a single result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive();
+    check_eq(res, 3);
+  }
+}
+
+TEST(
+  "send fan_out_request messages that return two values using typed actors") {
```
Neverlord:
```suggestion
TEST("typed fan_out_request with multiple results") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive();
+    if (check_has_value(res)) {
+      check_eq(std::get<0>(*res), 5);
+      check_eq(std::get<1>(*res), 3);
+    }
+  }
+}
+
+TEST("send fan_out_request messages with void result using typed actors") {
```
Neverlord:
```suggestion
TEST("typed fan_out_request with void result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+  }
+  SECTION("receive to expected with policy select_any") {
+    auto res = self->mail(1, 2)
+                 .fan_out_request(workers, 1s, policy::select_any_tag)
+                 .receive();
+    check_has_value(res);
+  }
+}
+
+TEST("send fan_out_request messages with invalid setups using typed actors") {
```
Neverlord:
```suggestion
TEST("typed fan_out_request with error responses") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+    self->mail(1, 2)
+      .fan_out_request(error_workers, 1s, policy::select_any_tag)
+      .receive([this](
+                 int results) { fail("expected an error, got: {}", results); },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*err, make_error(sec::all_requests_failed));
+  }
+}
+
+TEST("send delayed fan_out_request messages that return a result") {
```
Neverlord:
```suggestion
TEST("delayed fan_out_request with a single result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+      .delay(100ms)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([sum](int result) { *sum = result; },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*sum, 3);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("send delayed fan_out_request messages with void result") {
```
Neverlord:
```suggestion
TEST("delayed fan_out_request with void result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+      .delay(100ms)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([ran]() { *ran = true; },
+               [err](error& e) { *err = std::move(e); });
+    check(*ran);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("send delayed fan_out_request messages that return two values") {
```
Neverlord:
```suggestion
TEST("delayed fan_out_request with multiple results") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+        },
+        [err](error& e) { *err = std::move(e); });
+    check_eq(single_result->first, 5);
+    check_eq(single_result->second, 3);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("send delayed fan_out_request messages that return a result using typed "
+     "actors") {
```
Neverlord:
```suggestion
TEST("typed delayed fan_out_request with a single result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([sum](int result) { *sum = result; },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*sum, 3);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("send delayed fan_out_request messages that return two values using typed "
+     "actors") {
```
Neverlord:
```suggestion
TEST("typed delayed fan_out_request with multiple results") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+                                             policy::select_any_tag)
+                            .receive();
+    check_has_value(swapped_values);
+    check_eq(std::get<0>(*swapped_values), 5);
+    check_eq(std::get<1>(*swapped_values), 3);
+  }
+}
+
+TEST(
+  "send delayed fan_out_request messages with void result using typed actors") {
```
Neverlord:
```suggestion
TEST("typed delayed fan_out_request with void result") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+      .delay(100ms)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([ran]() { *ran = true; },
+               [err](error& e) { *err = std::move(e); });
+    check(*ran);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("fan_out_request with mixed results - one worker returns error") {
```
Neverlord:
```suggestion
TEST("fan_out_request with one worker returning an error") {
```

libcaf_core/caf/blocking_fan_out_request.test.cpp
```diff
@@ -0,0 +1,820 @@
...
+    self->mail(1, 2)
+      .fan_out_request(workers, 1s, policy::select_any_tag)
+      .receive([sum](int result) { *sum = result; },
+               [err](error& e) { *err = std::move(e); });
+    check_eq(*sum, 3);
+    check_eq(*err, error{});
+  }
+}
+
+TEST("timeout a delayed fan_out_request") {
```
Neverlord:
```suggestion
TEST("delayed fan_out_request with timeout") {
```

libcaf_core/caf/blocking_fan_out_response_handle.hpp
```diff
@@ -0,0 +1,346 @@
...
+         g{std::forward<OnError>(on_error)}](error& err) mutable {
+          auto lg = log::core::trace("pending = {}", *pending);
+          if (*pending > 0) {
+            timeout.dispose();
+            *pending = 0;
+            g(err);
+          }
+        };
+    auto bhvr = behavior{std::move(helper), std::move(error_handler)};
+    for (const auto& mid : state_.mids) {
```
Neverlord: Message IDs are just integers, so we can just use values:

```suggestion
    for (auto mid : state_.mids) {
```

libcaf_core/caf/blocking_fan_out_response_handle.hpp
Neverlord: Agree, not worth another metaprogramming utility as a one-off.

libcaf_core/caf/blocking_response_handle.hpp
```diff
@@ -66,6 +67,48 @@ template <class Result>
 using blocking_delayed_response_handle_t =
   typename blocking_delayed_response_handle_oracle<Result>::type;
 
+template <class... Ts>
+struct expected_builder;
```
Neverlord: Probably forgot to delete this? Probably added in an earlier commit, so please remove + squash.

Neverlord (changes requested):

libcaf_core/caf/blocking_fan_out_request.test.cpp
riemass: It's the same error handling we had in the old implementation. It's trivial to propagate the last error. I'll check the regular implementation and make a followup.

riemass: Solved in one commit for both implementations since it was a two line change.

libcaf_core/caf/blocking_response_handle.hpp
riemass: Done. Force squashed with last comment and force pushed.

libcaf_core/caf/blocking_fan_out_response_handle.hpp
```diff
@@ -0,0 +1,346 @@
...
+          }
+        };
+    auto bhvr = behavior{std::move(result_handler), std::move(error_handler)};
+    for (const auto& mid : state_.mids) {
+      auto now = std::chrono::steady_clock::now();
+      auto remaining = state_.deadline > now
+                         ? (state_.deadline - now)
+                         : std::chrono::steady_clock::duration::zero();
+      state_.self->do_receive(mid, bhvr, remaining);
+    }
```
Neverlord, line 175: This loop kinda defeats the point of `select_any`, because it will wait until all responses arrive before returning. It should return immediately at the first (valid) response. Of course we can't just return here after the first match because that would "leak" the other responses. I think the only way to do this cleanly is to add a list of "to-be-ignored" message IDs to the blocking actor. Then it silently discards the response for that response once it arrives. Feels a bit hacky, but I couldn't come up with a better idea.

The `select_all` case also needs to return early if it receives an error. Same logic: mark the message IDs to be discarded.

riemass: Thanks for the review. Of course that is a problem. The reason is that I wanted to have the implementation not modify the `abstract_blocking_actor*` interface, but as you suggested above, this will have to change the state of the blocking actor inevitably, so I'll try refactoring it with the suggestions.

riemass: I have a [`in_flight` disposable ](https://github.com/actor-framework/actor-framework/pull/2132/changes#diff-9d15f7b81e3fb057df9a73a1517ed1589ca6ca4ef42b6c3f8e6df2ea90278c12R80)in the `state_`. It's a [composite disposable created with](https://github.com/actor-framework/actor-framework/pull/2132/changes#diff-3ad181645d1bfa7aeff44e516eef7a0697f8bb19bc31fdc2cdb85c1f8a907c53R143) all the disposables returned from `self->request_response_timeout` in one case, and `clock.schedule_message` in the other case. Could returning early and disposing all remaining messages via this disposable be a simpler way to fix this?

Neverlord:
> I have a [`in_flight` disposable ](https://github.com/actor-framework/actor-framework/pull/2132/changes#diff-9d15f7b81e3fb057df9a73a1517ed1589ca6ca4ef42b6c3f8e6df2ea90278c12R80)in the `state_`. It's a [composite disposable created with](https://github.com/actor-framework/actor-framework/pull/2132/changes#diff-3ad181645d1bfa7aeff44e516eef7a0697f8bb19bc31fdc2cdb85c1f8a907c53R143) all the disposables returned from `self->request_response_timeout` in one case, and `clock.schedule_message` in the other case. Could returning early and disposing all remaining messages via this disposable be a simpler way to fix this?

The disposable will only cancel the timeouts.

The loop itself is problematic. It will receive those things in order. This means if the first message ID just happens to arrive late (or never and the timeout triggers), you'll wait that entire time, even if another message ID has long arrived and the function could return.

The other issue is that blocking actors have a `stash_`, where it will keep messages available until they get processed. If we return early here and handle only a single one of the messages, we'll need a way to tell the blocking actor that it has to actively discard the other responses. Otherwise, we'll keep those forever, i.e., have a memory leak.

So this needs to receive messages in a "one of these message IDs please" way and then tell the actor to actively dispose the other responses once they arrive.

Thinking about the timeouts, this will actually use the actor clock, right? In other words, aren't we potentially leaking the timeouts already if they arrive in the mailbox before we call `dispose`? Response messages will sit in the mailbox forever until they get picked up via their message ID.

Seems like we're opening a whole new can of worms with this. Let's put that on hold for now and discuss it again at another time. Maybe we just leave that out for blocking actors after all if it's too much of a pain to implement.

PR #2136, opened 2025-09-30
Implement `as_single` for the new mail API

Neverlord:
Adds the missing `as_single` to the new response handles.

---

> [!NOTE]
> Adds as_single to mail response handles to return flow::single, updates tests (including error handling) and changelog.
> 
> - **Core/API**:
>   - Add `event_based_response_handle::as_single` (typed and dynamically typed variants) to convert mail responses to `flow::single<...>`.
>   - Introduce internal type deduction (`event_based_response_handle_single_t`) for `as_single` return types.
> - **Tests**:
>   - Add coverage for `.as_single` with valid and error responses in `libcaf_core/caf/event_based_mail.test.cpp`.
>   - Rename test section label to `.as_observable` (from `.to_observable`).
>   - Expand `libcaf_core/caf/flow/single.test.cpp` to subscribe directly to `single`, verify value delivery and error propagation via `set_error`.
> - **Docs/Changelog**:
>   - Note support for `as_single` on new `mail` API in `CHANGELOG.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8e011ee6bdb33dbfd60eb150a976feba9bb78998. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/libcaf_core/caf/event_based_response_handle.hpp
+++ b/libcaf_core/caf/event_based_response_handle.hpp
@@ -79,6 +79,28 @@ template <class... Ts>
 using event_based_response_handle_res_t =
   typename event_based_response_handle_res<Ts...>::type;
 
+template <class...>
+struct event_based_response_handle_single_oracle;
+
+template <class Dummy>
+struct event_based_response_handle_single_oracle<Dummy> {
+  using type = flow::single<unit_t>;
+};
+
+template <class Dummy, class T>
+struct event_based_response_handle_single_oracle<Dummy, T> {
+  using type = flow::single<T>;
+};
+
+template <class Dummy, class T0, class T1, class... Ts>
+struct event_based_response_handle_single_oracle<Dummy, T0, T1, Ts...> {
+  using type = flow::single<cow_tuple<T0, T1, Ts...>>;
+};
+
+template <class... Ts>
+using event_based_response_handle_single_t =
+  typename event_based_response_handle_single_oracle<Ts...>::type;
+
 } // namespace caf::detail
 
 namespace caf {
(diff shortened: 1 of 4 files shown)


libcaf_core/caf/event_based_response_handle.hpp
```diff
@@ -146,6 +168,17 @@ class event_based_response_handle {
...
+  flow::assert_scheduled_actor_hdr_t<
+    detail::event_based_response_handle_single_t<Dummy, Results...>>
+  as_single() && {
+    auto cell = state_.self->template response_to_flow_cell<Results...>(
+      state_.mid, std::move(state_.pending_timeout));
+    using cell_t = typename decltype(cell)::value_type;
+    using val_t = typename cell_t::output_type;
+    return flow::single<val_t>{std::move(cell)};
+  }
+
```
riemass, line 181: What is the purpose of Dummy here? Is it so that the user can write both `as_single<void>` and `as_single<>()`, or is there something I'm missing?

Neverlord: We need the `Dummy` to turn this into a template. This will instantiate this method only once called. Otherwise, the compiler would need to have the full definition of `flow::single` available right away (which we want to avoid to have users include `scheduled_actor/flow.hpp` only when needed).

It's a template parameter with a default, so technically you can call it as `as_single()`, `as_single<>()`, and `as_single<void>()` and it will all do the same thing.

riemass: Thanks, forgot about that perspective.
EDIT: I do think we already have that problem since as_observable is unguarded by a template.

PR #2137, opened 2025-10-01
Add a new interface for connectables

Neverlord:
The current `caf::flow::connectable` hard-codes the internal type to `op::publish`. This has been the first, and so far the only, connectable type that we have implemented. However, before we can implement additional types, we first need to untie `connectable` from `op::publish`.


riemass (approved):
Everything seems all right. I find the member function name `ref_count` a bit odd, but it seems that is a common name in the `flow` library. 
Approving, ofc after rebasing to fix the pipeline.

libcaf_core/caf/flow/observable.hpp
```diff
@@ -142,9 +138,9 @@ class connectable {
     return fn(std::move(*this));
   }
 
-  template <class... Ts>
-  disposable subscribe(Ts&&... xs) {
-    return as_observable().subscribe(std::forward<Ts>(xs)...);
+  template <class Subscriber>
+  disposable subscribe(Subscriber&& arg) {
+    return as_observable().subscribe(std::forward<Subscriber>(arg));
```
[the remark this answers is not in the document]
riemass: I was wondering about this as well. Through grep I see we only ever used subscribe with one parameter. Was this parameter pack a feature we had but never used?

libcaf_core/caf/async/consumer_adapter.hpp
```diff
@@ -74,7 +74,7 @@ class consumer_adapter {
       }
     }
 
-    const error& abort_reason() const noexcept {
+    const error& abort_reason() const {
```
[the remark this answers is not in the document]
Neverlord: That function never should have been marked `noexcept` since it allocates memory. Breaking change for 2.0 is OK.

libcaf_core/caf/async/consumer_adapter.hpp
```diff
@@ -163,7 +163,7 @@ class consumer_adapter {
     }
   }
 
-  error abort_reason() const noexcept {
+  error abort_reason() const {
```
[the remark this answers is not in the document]
Neverlord: That function never should have been marked noexcept since it allocates memory. Breaking change for 2.0 is OK.

libcaf_core/caf/flow/op/auto_connect.hpp
```diff
@@ -0,0 +1,377 @@
...
+    auto src = std::get<connectable_ptr>(maybe_source);
+    conn = src->connect();
+    if (!conn) {
+      if (can_connect()) {
+        // If we reach here, it means that the source returned a disposed or
+        // invalid subscription but did not call `on_error`.
+        maybe_source = make_error(sec::invalid_observable);
+      }
+      return false;
+    }
```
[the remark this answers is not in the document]
Neverlord: We only ever call this function after checking for `can_connect`, so the `variant` cannot hold anything else here.

libcaf_core/caf/flow/op/auto_connect.hpp
```diff
@@ -0,0 +1,377 @@
...
+  coordinator* parent() const noexcept override {
+    return parent_;
+  }
+
+  disposable subscribe(observer<T> what) override {
+    // Utility function to fail or complete the subscription immediately in case
+    // the source is no longer connectable.
+    auto short_circuit = [this, &what] {
+      if (auto& err = std::get<error>(state_->maybe_source)) {
+        return super::fail_subscription(what, err);
```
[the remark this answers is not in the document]
Neverlord: That lambda is only called if the `variant` holds an error.

libcaf_core/caf/detail/counted_disposable.hpp
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include "caf/detail/core_export.hpp"
+#include "caf/disposable.hpp"
+#include "caf/ref_counted.hpp"
+
+namespace caf::detail {
+
+/// Decorates another disposable and creates "nested" disposables. When the last
+/// nested disposable is disposed, the parent disposable is also disposed.
```
riemass, line 16:
```suggestion
/// nested disposable is disposed, the parent disposable is also disposed.
/// @note: this construct is not thread-safe. 
```
I'd see someone suspecting this to be thread safe, so better leave a note.

libcaf_core/caf/flow/observable.hpp
```diff
@@ -109,26 +103,17 @@ class connectable {
...
-      pimpl_->auto_connect_threshold(subscriber_threshold);
-      pimpl_->auto_disconnect(true);
-      return observable<T>{std::move(pimpl_)};
-    } else {
-      auto ptr = make_counted<op::publish<T>>(parent(), pimpl_);
-      ptr->auto_connect_threshold(subscriber_threshold);
-      ptr->auto_disconnect(true);
-      return observable<T>{ptr};
+  /// @note Produces an error if `subscriber_threshold` is 0.
+  observable<T> ref_count(size_t subscriber_threshold = 1) {
```
riemass, line 107: Will ditching the `rvalue` overload have some significant performance issues? It would move the `pimpl_` if I'm correct.

libcaf_core/caf/flow/observable.hpp
```diff
@@ -84,23 +87,14 @@ class connectable {
...
-      return observable<T>{ptr};
+  /// @note Connects immediately if `subscriber_threshold` is 0.
+  observable<T> auto_connect(size_t subscriber_threshold = 1) {
+    auto ptr = make_counted<op::auto_connect<T>>(pimpl_->parent(),
+                                                 subscriber_threshold, pimpl_);
+    if (subscriber_threshold == 0) {
+      ptr->state().connect([](auto&) {});
     }
+    return observable<T>{std::move(ptr)};
   }
```
riemass, line 98: Can we evade the allocation if the threshold is zero?

Neverlord: No, we connect it right away, but we still need to `observable` object for observers to subscribe to and the state for remembering that we have connected the input observable.

libcaf_core/caf/flow/observable.hpp
Neverlord: Setting up the initial pipeline by chaining operators is done once. That's not something that would run in a performance-critical code path. I think saving that one extra increase/decrease on the reference count is not worth the code duplication. Otherwise, we would need to overload all of the operators in `observable<T>` as well.

libcaf_core/caf/detail/counted_disposable.hpp
Neverlord: I just didn't think of it. I don't think there's a reason to not make it thread-safe. Just boils down to making the count atomic. Done.

PR #2139, opened 2025-10-03
Switch main CI from Jenkins to GH actions

Neverlord:
Our current CI setup is quite involved and spans Jenkins, Cirrus and GitHub. This cleans up our CI by dropping Jenkins entirely.

I've also cut down on the number of jobs we run. Our goal with our CI setup was to expose the code base to a range of compilers. But we ended up with some redundancy that doesn't really add value IMHO. So my though process here is that we cut down our Linux platforms on CI to 3:
- Ubuntu, simply because it's the default on GH.
- Debian (oldest still officially maintained release), because it usually is the most conservative distro and will have an "old" compiler.
- Fedora (only latest), because it generally runs new stuff and will have a more recent compiler.

Instead of doing various combinations of release and debug builds, I've cut down on those as well and streamlined the GitHub actions in the process.


Neverlord:
I've also removed all of the Docker files from this repo. The ones we're still using have moved to https://github.com/actor-framework/continuous-integration. We build them in the other repo and then push the ready-to-use images to GHCR.

Neverlord:
@riemass as you can see, I've cut down on the number of build images we maintain to two. ✂️

I think for the ongoing development that setup should be a good balance between compiler coverage, build times, cost and maintenance effort.

I was thinking if we ever do want to bring back a wider coverage, then we probably should put extra jobs into a separate pipeline that we run only via manual trigger before cutting releases. That can then run all the things. Not really sure if we need that, though. So far, I don't think running stuff like Alpine or Alma has really added value.

.github/workflows/debug-builds.yml
```diff
@@ -0,0 +1,77 @@
...
+jobs:
+  windows:
+    runs-on: windows-2022
+    steps:
+      - uses: actions/checkout@v4
+      - uses: egor-tensin/vs-shell@v2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+          cache: 'pip'
```
riemass, line 22: It's only used here to setup python in Windows, but we might benefit from caching pip in the same way on `ubuntu-latest`. After some research this would speedup subsequent workflows (the first push to the branch would download pip, any subsequent push to this branch would use the cached pip index and download very fast). 

https://github.com/actions/setup-python#caching-packages-dependencies

PR #2142, opened 2025-10-06
Fix alignment on 32bit ARM architecture

allspark:
on 32bit ARM architecture the size of `std::size_t` and pointers is only 32bit but the aligned of some types is still 64bit. 

adding `alignas(max_align_t)` to the raw storage ensures that it starts accordingly.

### DIFF
--- a/libcaf_core/caf/async/batch.hpp
+++ b/libcaf_core/caf/async/batch.hpp
@@ -194,7 +194,7 @@ private:
     type_id_t item_type_;
     size_t item_size_;
     size_t size_;
-    std::byte storage_[];
+    alignas(max_align_t) std::byte storage_[];
   };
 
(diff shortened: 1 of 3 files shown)


libcaf_core/caf/async/batch.hpp
```diff
@@ -173,13 +173,17 @@ class CAF_CORE_EXPORT batch {
       return size_;
     }
 
+    std::byte* storage() noexcept {
+      return storage_;
+    }
+
```
Neverlord: Please remove this again. This didn't exist on purpose because it grants writable access to the buffer. Once created, the data of a batch is supposed to be immutable. The `batch` has friend access to implement efficient deserialization. The class is also `private` to `batch` at the moment, but it is still written in a way that we could make it public without changes if there's a use case for it.

PR #2147, opened 2025-10-11
Add new functions for operating on SPSC buffers

Neverlord:
The new `consume_on` function on consumer resources enables actors to consume items from a consumer resource without turning it into a flow.

The new `produce_on` function on producer resources is the symmetric counterpart to `consume_on`.

These additions allow developers to get a more "direct" way to operate on SPSC buffers whenever flows are not a good fit.

This needs to land in CAF 1.2. Hence, it's still on C++17.

### DIFF
--- a/libcaf_core/caf/async/spsc_buffer.hpp
+++ b/libcaf_core/caf/async/spsc_buffer.hpp
@@ -483,7 +483,7 @@ private:
 
   using on_wakeup_t = shared_callback_ptr<on_wakeup_signature>;
 
-  /// Stores the current state of the consumer.
+  /// Guards access to the state of the consumer.
   mutable std::mutex mtx_;
 
   /// The actor that owns this consumer.
@@ -776,7 +776,7 @@ private:
 
   using on_cancel_t = shared_callback_ptr<on_cancel_signature>;
 
-  /// Stores the current state of the consumer.
+  /// Guards access to the state of the consumer.
   mutable std::mutex mtx_;
 
   /// The actor that owns this consumer.


libcaf_core/caf/async/spsc_buffer.hpp
```diff
@@ -359,6 +367,139 @@ struct resource_ctrl : ref_counted {
...
+    ptr->deref();
+  }
+
+private:
+  using buffer_ptr_t = spsc_buffer_ptr<T>;
+
+  using on_wakeup_t = shared_callback_ptr<on_wakeup_signature>;
+
+  /// Stores the current state of the consumer.
+  mutable std::mutex mtx_;
```
riemass:
```suggestion
  /// Guards access to the state of the consumer.
  mutable std::mutex mtx_;
```

libcaf_core/caf/async/spsc_buffer.test.cpp
```diff
@@ -480,6 +513,170 @@ SCENARIO("consumer resources may be converted to flows only once") {
...
+        auto buf = rd.try_open();
+        require_ne(buf, nullptr);
+        auto con = make_counted<async::mock_consumer>();
+        auto buf_guard = detail::scope_guard([buf]() noexcept {
+#ifdef CAF_ENABLE_EXCEPTIONS
+          try {
+            buf->cancel();
+          } catch (...) {
+            fprintf(stderr, "failed to cancel buffer subscription! %s:%d\n",
+                    __FILE__, __LINE__);
```
riemass, line 659: Why not a `log::test::error(...)` here?

libcaf_core/caf/async/spsc_buffer.hpp
```diff
@@ -443,6 +607,199 @@ class consumer_resource {
...
+
+private:
+  using buffer_ptr_t = spsc_buffer_ptr<T>;
+
+  using on_demand_t = shared_callback_ptr<on_demand_signature>;
+
+  using on_cancel_t = shared_callback_ptr<on_cancel_signature>;
+
+  /// Stores the current state of the consumer.
+  mutable std::mutex mtx_;
```
riemass:
```suggestion
  /// Guards access to the state of the consumer.
  mutable std::mutex mtx_;
```

libcaf_core/caf/async/spsc_buffer.test.cpp
Neverlord: Because this is a `noexcept` block. Our logging functions can throw exceptions.

PR #2154, opened 2025-10-25
Reduce complexity of the response promises

Neverlord:
- redesign `as_single` and `as_observable` on requests
- use a consistent setup for response handles
- remove the `requester` mixin

@riemass there are two TODO items left in the code. I think rebasing #2132 and fixing it there would be a good idea. I didn't look into the blocking promises since you've already started doing it.

Relates #2009.
Relates #1873.

### DIFF
--- a/libcaf_core/caf/event_based_fan_out_response_handle.hpp
+++ b/libcaf_core/caf/event_based_fan_out_response_handle.hpp
@@ -9,13 +9,13 @@
 #include "caf/detail/response_type_check.hpp"
 #include "caf/disposable.hpp"
 #include "caf/fwd.hpp"
+#include "caf/log/core.hpp"
 #include "caf/message_id.hpp"
 #include "caf/policy/select_all.hpp"
 #include "caf/policy/select_all_tag.hpp"
 #include "caf/policy/select_any.hpp"
 #include "caf/policy/select_any_tag.hpp"
 #include "caf/type_list.hpp"
-#include "caf/log/core.hpp"
 
 #include <type_traits>
 
--- a/libcaf_core/caf/scheduled_actor/flow.hpp
+++ b/libcaf_core/caf/scheduled_actor/flow.hpp
@@ -56,7 +56,7 @@ struct response_to_flow_helper<type_list<>> {
         }
       }
     };
-  };
+  }
 };
 
 template <class T>
@@ -80,7 +80,7 @@ struct response_to_flow_helper<type_list<T>> {
         }
       }
     };
-  };
+  }
 };
 
 template <class T1, class T2, class... Ts>
@@ -106,7 +106,7 @@ struct response_to_flow_helper<type_list<T1, T2, Ts...>> {
         }
       }
     };
-  };
+  }
 };
 
 template <class... Ts>
--- a/libcaf_io/caf/io/typed_broker.hpp
+++ b/libcaf_io/caf/io/typed_broker.hpp
@@ -119,7 +119,7 @@ public:
   connection_handle add_tcp_scribe(network::native_socket fd) {
     static_assert(std::is_convertible_v<actor_hdl, connection_handler>,
                   "Cannot add scribe: broker misses required handlers");
-    return super::add_tcp_scribe(fd);
+    return super::add_scribe(fd);
   }
 
   expected<std::pair<accept_handle, uint16_t>>
@@ -130,10 +130,10 @@ public:
     return super::add_tcp_doorman(port, in, reuse_addr);
   }
 
-  expected<accept_handle> add_tcp_doorman(network::native_socket fd) {
+  accept_handle add_tcp_doorman(network::native_socket fd) {
     static_assert(std::is_convertible_v<actor_hdl, accept_handler>,
                   "Cannot add doorman: broker misses required handlers");
-    return super::add_tcp_doorman(fd);
+    return super::add_doorman(fd);
   }
 
   explicit typed_broker(actor_config& cfg) : super(cfg) {
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -20,7 +20,7 @@ jobs:
       - name: clang-format
         uses: jidicula/clang-format-action@v4.11.0
         with:
-          clang-format-version: '19'
+          clang-format-version: '20'
 
 
   spell-checking:
--- a/libcaf_core/caf/anon_mail.hpp
+++ b/libcaf_core/caf/anon_mail.hpp
@@ -73,7 +73,8 @@ public:
 
   /// Tags the message as urgent, i.e., sends it with high priority.
   [[nodiscard]] auto urgent() &&
-    requires(Priority == message_priority::normal) {
+    requires(Priority == message_priority::normal)
+  {
     using result_t = anon_mail_t<message_priority::high, Args...>;
     return result_t{std::move(content_)};
   }
--- a/libcaf_core/caf/async_mail.hpp
+++ b/libcaf_core/caf/async_mail.hpp
@@ -171,7 +171,8 @@ public:
 
   /// Tags the message as urgent, i.e., sends it with high priority.
   [[nodiscard]] auto urgent() &&
-    requires(Priority == message_priority::normal) {
+    requires(Priority == message_priority::normal)
+  {
     using result_t = async_mail_t<message_priority::high, Trait, Args...>;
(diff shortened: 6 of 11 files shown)


libcaf_core/caf/typed_response_promise.test.cpp
```diff
@@ -127,23 +127,24 @@ SCENARIO("response promises allow delaying of response messages") {
...
+      //                 });
+      //     return result;
+      //   };
+      //   THEN("clients receive an empty response from the dispatcher") {
+      //     expect<ok_atom>().from(self).to(hdl);
+      //     expect<ok_atom>().from(hdl).to(adder_hdl);
+      //     dispatch_message();
+      //     check(fetch_result().empty());
+      //   }
+      // }
```
riemass, line 147: This is the followup for me. :+1:

libcaf_io/caf/io/typed_broker.hpp
```diff
@@ -120,7 +119,7 @@ class typed_broker
   connection_handle add_tcp_scribe(network::native_socket fd) {
     static_assert(std::is_convertible_v<actor_hdl, connection_handler>,
                   "Cannot add scribe: broker misses required handlers");
-    return super::add_tcp_scribe(fd);
+    return super::add_scribe(fd);
```
riemass, line 122: Why the change here? `add_tcp_scribe` to `add_scribe`? Same question for `add_doorman` below.

libcaf_core/caf/event_based_mail.test.cpp
```diff
@@ -605,9 +605,7 @@ TEST("send fan_out_request messages that return a result") {
         .fan_out_request(workers, infinite, policy::select_all_tag)
         .as_observable<int>()
         .do_on_error([err](const error& x) { *err = x; })
-        .for_each([sum](std::vector<int> results) {
-          *sum = std::accumulate(results.begin(), results.end(), 0);
-        });
+        .for_each([sum](int value) { *sum += value; });
```
riemass, line 608: Looks much better.

libcaf_core/caf/event_based_mail.test.cpp
```diff
@@ -1159,13 +1151,9 @@ TEST("send fan_out_request messages that return two swapped values using typed "
...
             .as_observable()
             .do_on_error([err](const error& x) { *err = x; })
-            .for_each(
-              [swapped_values](std::vector<caf::cow_tuple<int, int>> results) {
-                swapped_values->clear();
-                for (auto result : results) {
-                  swapped_values->emplace_back(get<0>(result), get<1>(result));
-                }
-              });
+            .for_each([swapped_values](const caf::cow_tuple<int, int>& result) {
```
riemass, line 1154: An interface question: now that we no longer have a `vector`, would a lambda accepting two `int`s make more sense? 
```suggestion
            .for_each([swapped_values](int first, int second) {
```
I saw that the `select_any` uses cow_tuples as well, so I suppose that's how flows transfer them, and it would be a breaking change. Just want to hear your comment on this.

libcaf_core/caf/event_based_fan_out_response_handle.hpp
```diff
@@ -341,26 +278,30 @@ class event_based_fan_out_delayed_response_handle {
     return std::move(pending_request);
   }
 
-  auto as_single() &&                                                  //
-    requires(!std::same_as<type_list<Results...>, type_list<message>>) //
-  { return std::move(decorated).as_single(); }
+  auto as_single() &&
```
riemass, line 281: You'd think we need a template here as well, but look... Strange in many ways.

libcaf_core/caf/event_based_fan_out_response_handle.hpp
```diff
@@ -260,46 +218,25 @@ class event_based_fan_out_response_handle {
...
+
   using decorated_type
     = event_based_fan_out_response_handle<Policy, Results...>;
 
+  // -- constants --------------------------------------------------------------
+
+  static constexpr bool is_dynamically_typed
+    = std::is_same_v<type_list<Results...>, type_list<message>>;
+
+  static constexpr bool is_statically_typed = !is_dynamically_typed;
```
riemass, line 238: Could be nice to have a templated version in the `metaprogramming` header. Will this be covered in the current PR you're working on? If not I'll transfer it as a followup when I pick up the blocking fan out.

libcaf_core/caf/response_promise.test.cpp
```diff
@@ -122,23 +122,24 @@ SCENARIO("response promises allow delaying of response messages") {
...
+      //                 });
+      //     return result;
+      //   };
+      //   THEN("clients receive an empty response from the dispatcher") {
+      //     expect<ok_atom>().from(self).to(hdl);
+      //     expect<ok_atom>().from(hdl).to(adder_hdl);
+      //     dispatch_message();
+      //     check(fetch_result().empty());
+      //   }
+      // }
```
riemass, line 142: :+1:

libcaf_core/caf/scheduled_actor/flow.hpp
```diff
@@ -26,33 +30,230 @@ struct has_impl_include<scheduled_actor> {
...
+template <class TypeToken, class State>
+  requires flow::assert_has_impl_include<State>
+auto abstract_scheduled_actor::response_to_single(TypeToken,
+                                                  const State& state) {
+  using helper_type = detail::response_to_flow_helper<TypeToken>;
+  using value_type = typename helper_type::value_type;
+  auto cell = make_counted<flow::op::cell<value_type>>(flow_context());
+  if constexpr (State::is_fan_out) {
+    // For fan-out responses, we need to add the handler for each message ID.
+    // This assumes that the fan-out request uses the select any policy. The
```
riemass, line 129: So you can't call as_single with a select all policy, or am I missing something here? Couldn't spot anything in the `response_handle`.

libcaf_core/caf/event_based_response_handle.hpp
```diff
@@ -127,77 +137,44 @@ class event_based_response_handle {
...
+    requires is_dynamically_typed
   auto as_observable() && {
-    auto cell = state_.self->template response_to_flow_cell<Ts...>(
-      state_.mid, std::move(state_.pending_timeout));
-    using cell_t = typename decltype(cell)::value_type;
-    using val_t = typename cell_t::output_type;
-    return flow::single<val_t>{std::move(cell)}.as_observable();
+    // Note: templatized so we need the definition of response_to_observable
+    //       only when instantiating this function
+    using self_type = detail::left<abstract_scheduled_actor, T>;
```
riemass, line 175: C++ can be strange in many ways.

libcaf_io/caf/io/typed_broker.hpp
Neverlord: There's simply no `add_tcp_scribe` on `super`. Don't know why it compiled before. 🤷‍♂️

libcaf_core/caf/event_based_mail.test.cpp
Neverlord: Flows only have one `value_type`. Automagically unboxing tuples to have `for_each` basically "split" the input would be doable, but I think the payoff vs. implementation complexity isn't really there.

libcaf_core/caf/event_based_fan_out_response_handle.hpp
Neverlord: Well, we're inside a template class. So... 😅

libcaf_core/caf/event_based_fan_out_response_handle.hpp
Neverlord: You mean pulling the `std::is_same_v<type_list<Results...>, type_list<message>>;` out? Hm, haven't thought about that. Also not sure if it feels worth it, although I can see the argument for reducing code duplication.

libcaf_core/caf/scheduled_actor/flow.hpp
Neverlord: If you select all... then you don't have a single value anymore, no? If the user wants to combine all values into one again, they can just call `to_vector` on the `observable`.

libcaf_core/caf/event_based_response_handle.hpp
Neverlord: Well, it's basically type obfuscation so the compiler will look the other way until a user actually instantiates the function.

PR #2157, opened 2025-10-26
Remove the deprecated monitored spawn option

Neverlord:
Relates #1873.


riemass (approved):
Approved. Windows initially failed because choco server was down (couldn't pull OpenSSL dependency).

PR #2159, opened 2025-10-26
Improve diagnostics when using typed actors

### LINKED ISSUE #2158: Use concepts to simplify `typed_actor` and friends
Our `typed_actor` implementation and it's supporting classes like `typed_behavior` currently need two template specializations. Now that we have switched to C++20, we can do better and also improve compile-time errors for users.

Neverlord:
Introduce a new `requires` clause to `typed_actor` and supporting classes that checks validity of all message handler signatures.

Closes #2158.

### DIFF
--- a/libcaf_core/caf/typed_actor.hpp
+++ b/libcaf_core/caf/typed_actor.hpp
@@ -24,28 +24,21 @@
 
 namespace caf {
 
-template <class...>
-class typed_actor;
-
 /// Identifies a statically typed actor.
-template <class TraitOrSignature>
-class typed_actor<TraitOrSignature>
-  : detail::comparable<typed_actor<TraitOrSignature>>,
-    detail::comparable<typed_actor<TraitOrSignature>, actor>,
-    detail::comparable<typed_actor<TraitOrSignature>, actor_addr>,
-    detail::comparable<typed_actor<TraitOrSignature>, strong_actor_ptr> {
+template <class... Ts>
+  requires typed_actor_pack<Ts...>
+class typed_actor : detail::comparable<typed_actor<Ts...>>,
+                    detail::comparable<typed_actor<Ts...>, actor>,
+                    detail::comparable<typed_actor<Ts...>, actor_addr>,
+                    detail::comparable<typed_actor<Ts...>, strong_actor_ptr> {
 public:
   // -- member types -----------------------------------------------------------
 
   /// Stores the template parameter pack.
-  using trait = detail::to_statically_typed_trait_t<TraitOrSignature>;
+  using trait = detail::to_statically_typed_trait_t<Ts...>;
 
   /// Stores the template parameter pack.
   using signatures = typename trait::signatures;
-
-  static_assert(detail::are_signatures_normalized_v<signatures>,
-                "Function signatures must be normalized to the format "
-                "'result<Out...>(In...)', no qualifiers or references allowed");
 
   // -- friends ----------------------------------------------------------------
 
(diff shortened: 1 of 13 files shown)


libcaf_core/caf/typed_actor.hpp
```diff
@@ -121,10 +111,10 @@ class typed_actor<TraitOrSignature>
   typed_actor& operator=(typed_actor&&) = default;
   typed_actor& operator=(const typed_actor&) = default;
 
-  template <class... Ts>
+  template <class... Us>
     requires detail::tl_subset_of_v<signatures,
-                                    typename typed_actor<Ts...>::signatures>
-  typed_actor(const typed_actor<Ts...>& other) : ptr_(other.ptr_) {
+                                    typename typed_actor<Us...>::signatures>
+  typed_actor(const typed_actor<Us...>& other) : ptr_(other.ptr()) {
```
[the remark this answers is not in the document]
Neverlord: We can't access that member since it's private. We could add a `friend` declaration here, but I think it's overkill just to avoid a getter.

libcaf_core/caf/typed_actor_pack.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+template <class... Ts>
+struct typed_actor_trait_checker<type_list<Ts...>> {
+  // The only use case for this template is to be used in the
+  // `message_handler_signature` concept. If this template is instantiated,
+  // we want to "deep check" the signatures as well to raise compile-time errors
+  // early in case the trait contains any invalid signatures. If we would simply
+  // assign the result of that to `value` instead of using `static_assert`,
+  // the user would not see which signature is invalid.
+  static_assert((message_handler_signature<Ts> && ...));
```
riemass, line 52:
```suggestion
  static_assert((message_handler_signature<Ts> && ...), "message handlers must have signatures in the form `result<T>(Args...)`");
```
A message might be useful.

libcaf_core/caf/typed_actor_pack.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+/// Checks whether `Ts...` is a valid template parameter pack for `typed_actor`.
+template <class... Ts>
+concept typed_actor_pack =
+  // Either we have exactly one trait argument.
+  (sizeof...(Ts) == 1
+   && (detail::typed_actor_trait_checker<typename Ts::signatures>::value
+       && ...))
+  // Or we have at least one message handler signature.
+  || (sizeof...(Ts) > 0 && (message_handler_signature<Ts> && ...));
```
riemass, line 68: Do we still plant to deprecate giving the message handler signatures directly?

riemass (changes requested):

libcaf_core/caf/typed_actor_pack.hpp
Neverlord: That might be counter-productive, actually. The compiler will give an exact report which type has failed the concept check, which will give the user exactly the source location of where things went wrong plus the type that was wrong. I'm not sure if providing a message here would suppress that and give a generic error instead.

libcaf_core/caf/typed_actor_pack.hpp
Neverlord: With the concept-based checks in place, I lean towards just continuing to support both. We would save only a handful of code lines now anyways and the biggest gripe I had with the previous iteration is gone (namely the specializations on `typed_behavior` and `typed_actor`).

libcaf_core/caf/typed_actor_pack.hpp
riemass: Ok, nothing against leaving it as is then.

PR #2180, opened 2025-11-26
Make `caf::expected` compatible with `std::expected`

### LINKED ISSUE #2165: Make `caf::expected` compatible with `std::expected`
We want to phase out our `caf::expected` eventually and implement it as:

```c++
template <class T>
using expected = std::expected<T, caf::error>;
```

However, our version is API incompatible with `std::expected` at the moment. With the STL version, the error case requires constructing the value via `caf::unexpected`. For example:

```c++
// Currently legal: constructing caf::expected directly from caf::error
caf::expected<int> foo() {
  return caf::make_error(caf::sec::logic_error);
}

// STL-compatible API:
caf::expected<int> foo() {
  return caf::unexpected(caf::make_error(caf::sec::logic_error));
}
```

We should use CAF 2.0 to add the necessary scaffolding for making our interface compatible to the STL version and deprecate all non-standard conversions. Then we can switch to `std::expected` in a future CAF release.

riemass:
This PR prepares the switch from `caf::expected` to `std::expected`. 
It introduces an `caf::unexpected` class. Deprecates construction and comparison with `caf::error` in favor of `unexpected`. 
Also, `transform_or` becomes `transform_error`, and `or_else` with a void-returning function is not supported in std, so it's split out and deprecated.
I've had trouble disabling the `expected.test.cpp` inside the file (with #ifndef) so I disabled it at CMake level.
I've tried isolating commits that make change with the commits that update deprecated stuff to make it easier to review. 
Closes #2165.


CMakeLists.txt
```diff
@@ -84,6 +79,13 @@ set(CAF_SANITIZERS "" CACHE STRING
 set(CAF_BUILD_INFO_FILE_PATH "" CACHE FILEPATH
   "Optional path for writing CMake and compiler version information")
 
+# -- CAF options that depend on others -----------------------------------------
+
+cmake_dependent_option(CAF_ENABLE_OPENSSL_MODULE "Build OpenSSL module" ON
+                       "CAF_ENABLE_IO_MODULE" OFF)
+cmake_dependent_option(CAF_USE_STD_EXPECTED "Enable std::expected support" OFF
+                       "CAF_CXX_VERSION STREQUAL 23" OFF)
```
Neverlord: Should check for >= 23 instead.

riemass:
Should I modify the ci runners to have a fedora with `std::expected` and `c++23`?

libcaf_core/caf/expected.hpp
```diff
@@ -37,7 +36,51 @@ auto expected_from_fn(F&& f, Ts&&... xs) {
...
+
+  caf::error& error() noexcept {
+    return reason;
+  }
+
+  const caf::error& error() const noexcept {
+    return reason;
+  }
+
+  void swap(unexpected& other) {
```
Neverlord:
```suggestion
  void swap(unexpected& other) noexcept {
```

libcaf_core/caf/expected.hpp
```diff
@@ -37,7 +36,51 @@ auto expected_from_fn(F&& f, Ts&&... xs) {
...
+    using std::swap;
+    swap(reason, other.reason);
+  }
+
+  bool operator==(const unexpected& other) const {
+    return reason == other.reason;
+  }
+
```
Neverlord: Should be a free function and `noexcept`.

libcaf_core/caf/expected.hpp
```diff
@@ -37,7 +36,51 @@ auto expected_from_fn(F&& f, Ts&&... xs) {
...
+#else
+
+namespace caf {
+
+/// Represents an unexpected value to be stored in caf::expected.
+class unexpected {
+public:
+  unexpected(caf::error err) : reason{std::move(err)} {
+  }
```
Neverlord: Lacks constructor overloads from the STL, in particular the `std::in_place_t` version.

libcaf_core/caf/expected.hpp
```diff
@@ -69,16 +112,16 @@ class expected {
...
-    requires(std::convertible_to<U, T> || error_code_enum<U>)
-  expected(U x) {
-    if constexpr (std::convertible_to<U, T>) {
-      has_value_ = true;
-      new (std::addressof(value_)) T(std::move(x));
-    } else {
-      has_value_ = false;
-      new (std::addressof(error_)) caf::error(x);
-    }
+  template <std::convertible_to<T> U>
```
Neverlord: Maybe we should restrict using `is_constructible` instead since we're calling the constructor?

libcaf_core/caf/config_value.cpp
```diff
@@ -245,14 +245,14 @@ expected<bool> config_value::to_boolean() const {
           if (auto j = x.find("value"); j != x.end()) {
             return j->second.to_boolean();
           }
-          return format_to_error(sec::conversion_failed,
-                                 "missing value for object of type {}", tn);
+          return caf::unexpected{format_to_error(
```
Neverlord: We have this a lot now. Let's introduce `format_to_unexpected` to cover this use case with less noise.

libcaf_core/caf/config_value.hpp
```diff
@@ -393,8 +395,8 @@ expected<T> get_as(const config_value& x, inspector_access_type::empty) {
   if (x.can_convert_to_dictionary())
     return T{};
   else
-    return make_error(sec::conversion_failed,
-                      "invalid element type: expected a dictionary");
+    return caf::unexpected{make_error(
```
Neverlord: This pattern is also repeated dozens of times. Should go into a new `make_unexpected` helper function.

libcaf_core/caf/config_value.test.cpp
```diff
@@ -171,16 +171,26 @@ SCENARIO("get_as can convert config values to boolean") {
...
-        check_eq(get_as<bool>(cv_empty_list), sec::conversion_failed);
-        check_eq(get_as<bool>(cv_empty_dict), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{0}), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{1}), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{0.f}), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{1.f}), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{""s}), sec::conversion_failed);
-        check_eq(get_as<bool>(config_value{"1"s}), sec::conversion_failed);
+        check_eq(get_as<bool>(cv_null),
+                 caf::unexpected{make_error(sec::conversion_failed)});
```
Neverlord:
```suggestion
        check_eq(get_as<bool>(cv_null),
                 caf::unexpected{sec::conversion_failed});
```

According to [cppref](https://en.cppreference.com/w/cpp/utility/expected/operator_cmp.html), we can compare `expected<T> x` to `unexpected<E> y` as long as `x.error() == y.error()` is well-formed.

libcaf_core/caf/expected.hpp
```diff
@@ -37,7 +36,51 @@ auto expected_from_fn(F&& f, Ts&&... xs) {
...
+#ifdef CAF_USE_STD_EXPECTED
+
+#  include <expected>
+
+namespace caf {
+
+using unexpected = std::unexpected<error>;
+
+template <typename T>
+using expected = std::expected<T, error>;
```
Neverlord:
```suggestion
template <class T>
using unexpected = std::unexpected<T>;

template <class T>
using expected = std::expected<T, error>;
```

libcaf_core/caf/expected.hpp
```diff
@@ -37,7 +36,51 @@ auto expected_from_fn(F&& f, Ts&&... xs) {
...
+
+} // namespace caf
+
+#else
+
+namespace caf {
+
+/// Represents an unexpected value to be stored in caf::expected.
+class unexpected {
```
Neverlord, line 69:
```suggestion
template <class T>
class unexpected {
```

libcaf_core/caf/expected.hpp
```diff
@@ -429,7 +503,7 @@ class expected {
     if (has_value())
       return detail::expected_from_fn(std::forward<F>(f), value_);
     else
-      return expected<res_t>{error_};
+      return expected<res_t>{caf::unexpected{error_}};
```
Neverlord: Using [std::unexpect_t](https://en.cppreference.com/w/cpp/utility/expected/unexpect_t.html) here would make this a lot less noisy IMO.

```suggestion
      return expected<res_t>{unexpect, error_};
```

This would cut down on a lot noise by itself by reducing the amount of times we construct `unexpected{...}` objects only to pass them to an `expected` for construction.

libcaf_core/caf/expected.hpp
```diff
@@ -642,13 +771,21 @@ class expected<void> {
   // -- constructors, destructors, and assignment operators --------------------
 
   template <error_code_enum Enum>
-  expected(Enum x) : error_(std::in_place, x) {
+  [[deprecated("Use constructor with caf::unexpected instead")]]
```
Neverlord:
```suggestion
  [[deprecated("construct using unexpect or from an unexpected instead")]]
```

libcaf_core/caf/expected.hpp
```diff
@@ -642,13 +771,21 @@ class expected<void> {
...
+  expected(Enum x)
+    : error_(std::in_place, x) {
+    // nop
+  }
+
+  [[deprecated("Use constructor with caf::unexpected instead")]]
```
Neverlord:
```suggestion
  [[deprecated("construct using unexpect or from an unexpected instead")]]
```

libcaf_core/caf/expected.hpp
```diff
@@ -938,6 +1106,25 @@ inline bool operator!=(const expected<void>& x, const expected<void>& y) {
...
+    caf::error e;
+    if (!f.object(x).fields(f.field("error", e)))
+      return false;
+    x = caf::unexpected{std::move(e)};
+    return true;
+  } else {
+    return f.object(x).fields(f.field("error", x.error()));
+  }
+}
+
```
Neverlord: This shouldn't exist. The `unexpected` is a pure helper to construct something else, not something we would actively store somewhere.

```suggestion
```

Neverlord (changes requested):

libcaf_core/caf/config_value.test.cpp
riemass: The lack of `explicit` in the error from Enum ctor is making this look weird, but this is gonna come in handy since the other review item is making the unexpected generic.

libcaf_core/caf/expected.hpp
riemass: I hear you. However, `std::unexpect_t` is a c++23 feature which means we need a back-port in the form of `caf::unexpect_t`, and that would make the transition a bit harder later on.

libcaf_core/caf/actor_system.cpp
```diff
@@ -991,16 +991,18 @@ actor_system::dyn_spawn_impl(const std::string& name, message& args,
     "name = {}, args = {}, check_interface = {}, expected_ifs = {}", name, args,
     check_interface, expected_ifs);
   if (name.empty())
-    return sec::invalid_argument;
+    return expected<strong_actor_ptr>(unexpect, sec::invalid_argument);
```
Neverlord: This code is not DRY (repeats the return type multiple times):

```suggestion
    return {unexpect, sec::invalid_argument};
```

libcaf_core/caf/actor_system.hpp
```diff
@@ -465,7 +465,7 @@ class CAF_CORE_EXPORT actor_system {
     }
     auto res = dyn_spawn_impl(name, args, ctx, check_interface, expected_ifs);
     if (!res)
-      return std::move(res.error());
+      return expected<Handle>{unexpect, std::move(res.error())};
```
Neverlord:
```suggestion
      return {unexpect, std::move(res.error())};
```

libcaf_core/caf/actor_system_config.cpp
```diff
@@ -556,7 +557,7 @@ actor_system_config::parse_config(std::istream& source,
                                   const config_option_set& opts) {
   settings result;
   if (auto err = parse_config(source, opts, result); err.valid())
-    return err;
+    return expected<settings>{unexpect, err};
```
Neverlord:
```suggestion
    return {unexpect, err};
```

libcaf_core/caf/expected.hpp
```diff
@@ -938,6 +1160,12 @@ inline bool operator!=(const expected<void>& x, const expected<void>& y) {
   return !(x == y);
 }
 
+} // namespace caf
+
+#endif
+
+namespace caf {
```
Neverlord: Please remove this block, i.e., the `to_string` functions. I suspect we need those in the unit test framework, but we should add support for `expected` in the same way we support `std::optional`. Adding free function overloads for STL types (which `caf::expected` is when setting the appropriate CMake option) is asking for trouble down the line. It might be that just removing them is enough, didn't check.

Neverlord:
I'll put this on hold for now. MSVC still has `void unexpected()` in the global namespace even tough this should no longer exist since C++17. Of course it's messing with `caf::unexpected`. Badly. It's either always qualifying `unexpected`, avoiding/removing it altogether or come up with a better idea.

libcaf_core/caf/expected.hpp
riemass: Removed `to_string`. Added the `stringification_inspector` overload for `expected<T>`. Not entirely happy with how the serialization for the error case looks - I tired separately calling `append` and/or `save` for `"!"` and `error()` and always end up with a nested `sep()` call resulting in the format `!, runtime_error`. The provided solution works but as an extra allocation.

Neverlord (approved):
Thanks for bringing this over the finishing line! 👍

PR #2182, opened 2025-11-28
Modernize use of find, find_if, etc.

Neverlord:


riemass (dismissed):

libcaf_test/caf/test/fixture/deterministic.cpp
```diff
@@ -251,14 +252,15 @@ class deterministic_actor_clock : public actor_clock {
...
+    auto i = std::ranges::find_if(actions, is_not_disposed);
     if (i == actions.end())
       runnable::current().fail({"no pending timeout found", loc});
     return i->first;
   }
 
   /// Returns the time of the last pending timeout.
   time_point last_timeout(const detail::source_location& loc) {
+    // Note: Using reverse iterators, so keeping iterator-based version
     auto i = std::find_if(actions.rbegin(), actions.rend(), is_not_disposed);
```
shariarriday, line 264: Can we consider the following as we are looking to modernize and use range everywhere?

```suggestion
    auto r_actions = actions | std::views::reverse;
    auto i = std::ranges::find_if(r_actions, is_not_disposed);
    if (i == r_actions.end())
```

Neverlord: That's certainly something we can consider. For now, I kept it simple by sticking to the new overloads from `<algorithm>` and avoided including `<ranges>`. One step at a time, basically.

PR #2194, opened 2025-12-06
Update Cirrus to FreeBSD 14.3

riemass:


riemass:
Renamed the directory from `freebsd-14` to just `freebsd` because the initial commit was enabling FreeBSD 15.0. Also, the config setup didn't really change even from FreeBSD 13. @Neverlord  Since we agreed on just FreeBSD 14.3 for now, if you mind the change I can roll that back.

Neverlord (approved):

PR #2222, opened 2026-01-07
Handle HTTP chunked requests

### LINKED ISSUE #2205: Support Chunked-Encoding for the HTTP server
See https://github.com/actor-framework/actor-framework/blob/a3a63e69844408afe444f0a965ca627674a891c0/libcaf_net/caf/net/http/server.cpp#L212-L217

riemass:
Closes #2205 
Organized chunked requests in server and router. The server will have callbacks for each chunk. 
The router aggregates the chunks and routes them as a normal request. 
Side-quest: added `.devcontainer/build` to .gitignore. Better than accidentally pushing the build directory.

### DIFF
--- a/libcaf_net/caf/net/http/server.test.cpp
+++ b/libcaf_net/caf/net/http/server.test.cpp
@@ -371,7 +371,7 @@ SCENARIO("the server receives the largest accepted chunked HTTP request") {
       THEN("the HTTP layer parses the chunked request and calls the "
            "application layer") {
         auto maybe_res = res_promise.get_future().get(1s);
-        log::test::error("THIS IS Error: {}", maybe_res);
+
         require(maybe_res.has_value());
         check_eq(maybe_res->hdr.method(), net::http::method::post);
         check_eq(maybe_res->hdr.version(), "HTTP/1.1");
--- a/libcaf_net/caf/net/http/upper_layer.hpp
+++ b/libcaf_net/caf/net/http/upper_layer.hpp
@@ -43,8 +43,8 @@ public:
 
   /// Consumes a chunk of the message.
   /// @param payload The payload of the received chunk.
-  /// @returns The number of consumed bytes or a negative value to signal an
-  ///          error.
+  /// @returns An error if the chunk cannot be consumed, or an empty error on
+  ///          success.
   virtual error consume_chunk(const_byte_span payload) = 0;
 
   /// Ends a chunked message.


riemass:
@Neverlord When the router combines the chunks and delegates the request, the header field `Transfer-Encoding: chunked` will still be present, and there will be no `Content-Length` header. Should I leave as-is or maybe modify the headers? I'm more concerned with the absence of the `Content-Length` header.

libcaf_net/caf/net/http/router.cpp
```diff
@@ -101,4 +101,25 @@ ptrdiff_t router::consume(const request_header& hdr, const_byte_span payload) {
...
 
+error router::begin_chunked_message(const net::http::request_header& hdr) {
+  hdr_ = hdr;
+  return error{};
+}
+
+ptrdiff_t router::consume_chunk(const_byte_span body) {
+  CAF_ASSERT(hdr_.valid());
+  body_.insert(body_.end(), body.begin(), body.end());
+  return body.size();
```
Neverlord:
```suggestion
  return static_cast<ptrdiff_t>(body.size());
```

libcaf_net/caf/net/http/server.cpp
```diff
@@ -185,6 +185,12 @@ class server_impl : public http::server {
             // Transition to the next mode.
             if (hdr_.chunked_transfer_encoding()) {
               mode_ = mode::read_chunks;
+              if (auto err = up_->begin_chunked_message(hdr_); err.valid()) {
+                write_response(status::internal_server_error,
+                               "Failed to initiate chunked message.");
```
Neverlord:
```suggestion
                               "Failed to initiate chunked transfer.");
```

libcaf_net/caf/net/http/server.cpp
```diff
@@ -220,11 +226,66 @@ class server_impl : public http::server {
...
+            }
+            received_chunks_size_ = 0;
+            mode_ = mode::read_header;
+            return consumed;
+          }
+          if (up_->consume_chunk(remainder.subspan(0, chunk_size)) < 0) {
+            // TODO write response
+            abort(sec::runtime_error, "error while consuming HTTP chunk");
+            return -1;
+          }
```
Neverlord: Looking at this, maybe `consume_chunk` should also just return `error`. Usually we return the number of bytes consumed but this doesn't really make sense here. Either the upper layer consumes the chunk or not. There's nothing this layer could reasonably do if the upper layer only consumes half of the chunk.

libcaf_net/caf/net/http/server.cpp
```diff
@@ -220,11 +226,66 @@ class server_impl : public http::server {
...
+            write_response(status::payload_too_large,
+                           "Payload exceeds maximum size.");
+            abort(sec::protocol_error, "payload exceeds maximum size");
+            return -1;
+          }
+          if (remainder.size() < chunk_size + 2) {
+            // Configure the policy for the next call to consume.
+            // Await at least chunk line length + chunk_size bytes + crlf.
+            const auto least = input.size() - remainder.size() + chunk_size + 2;
+            const auto most = max_request_size_ - received_chunks_size_;
```
Neverlord: Isn't this missing the +2 for CLRF?

libcaf_net/caf/net/http/server.test.cpp
```diff
@@ -28,6 +28,7 @@ namespace {
 struct response_t {
   net::http::request_header hdr;
   byte_buffer payload;
+  std::vector<byte_buffer> chunked_payload;
```
Neverlord:
```suggestion
  std::vector<byte_buffer> payload_chunks;
```

libcaf_net/caf/net/http/server.test.cpp
```diff
@@ -282,21 +300,195 @@ SCENARIO("the client sends a multipart HTTP request") {
...
+        check(std::ranges::equal(maybe_res->chunked_payload[0],
+                                 to_const_byte_span("Hello, world!"sv)));
+        check(std::ranges::equal(maybe_res->chunked_payload[1],
+                                 to_const_byte_span("Developer Network"sv)));
+      }
+    }
+  }
+}
+
+SCENARIO("the server handles edge cases for chunked HTTP requests") {
```
Neverlord: This is way to broad and tells me nothing. What's this scenario actually about? Some tests are about sending `bad_request` back to the client on errors while other tests seem unrelated. Probably should be split up into different scenarios.

Neverlord:
> @Neverlord When the router combines the chunks and delegates the request, the header field `Transfer-Encoding: chunked` will still be present, and there will be no `Content-Length` header. Should I leave as-is or maybe modify the headers? I'm more concerned with the absence of the `Content-Length` header.

The header should reflect what the client has communicated. What use case do you have in mind where I would look at the `Content-Length` header field even though I get the payload in a ready-to-use buffer from CAF?

libcaf_net/caf/net/http/server.cpp
riemass: Doesn't, really. I'll refactor it into an error. Maybe the `end_chunked_message` could return the number of consumed bytes, but that info wouldn't really matter to the lower layer.

libcaf_net/caf/net/http/server.cpp
riemass: When I think about it, you're right. The 2 bytes aren't copied to the payload, which means we can receive two more bytes here.

riemass: This line had a bug where in edge cases the `most` was less that `least` - for failing to account for the line holding the next chunk size.

libcaf_net/caf/net/http/server.test.cpp
riemass: Refactored each GIVEN into a separate TEST. The blocks where too small for a scenario IMHO.

riemass:
> The header should reflect what the client has communicated. What use case do you have in mind where I would look at the `Content-Length` header field even though I get the payload in a ready-to-use buffer from CAF?

We're on the same page the. I can't figure a scenario where someone unconditionally reads the `Content-Length` header while having the payload buffer.

Neverlord (approved):

PR #2225, opened 2026-01-12
Simplify handling of the hidden flag

Neverlord:
Handle the `hidden` flag centrally at the `abstract_actor` to avoid it leaking into `launch` functions that each actor type has to implement independently.

### DIFF
--- a/libcaf_core/caf/abstract_actor.cpp
+++ b/libcaf_core/caf/abstract_actor.cpp
@@ -14,7 +14,6 @@
 #include "caf/default_attachable.hpp"
 #include "caf/detail/assert.hpp"
 #include "caf/log/core.hpp"
-#include "caf/log/system.hpp"
 #include "caf/mailbox_element.hpp"
 #include "caf/system_messages.hpp"
 
@@ -171,8 +170,7 @@ bool abstract_actor::cleanup(error&& reason, scheduler* sched) {
   for (attachable* i = head.get(); i != nullptr; i = i->next.get())
     i->actor_exited(fail_state_, sched);
   if (getf(is_registered_flag)) {
-    auto count = home_system().registry().dec_running();
-    log::system::debug("actor {} decreased running count to {}", id(), count);
+    home_system().dec_running_actors_count(id());
   }
   on_cleanup(fail_state_);
   return true;
--- a/libcaf_core/caf/actor_system.cpp
+++ b/libcaf_core/caf/actor_system.cpp
@@ -215,41 +215,6 @@ public:
     }
   }
 
-  size_t inc_running() override {
-    return ++running_;
-  }
-
-  size_t dec_running() override {
-    auto new_val = --running_;
-    if (new_val <= 1) {
-      std::unique_lock guard{running_mtx_};
-      running_cv_.notify_all();
-    }
-    return new_val;
-  }
-
-  /// Returns the number of currently running actors.
-  size_t running() const override {
-    return running_.load();
-  }
-
-  /// Blocks the caller until running-actors-count becomes `expected`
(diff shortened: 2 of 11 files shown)


libcaf_core/caf/blocking_actor.cpp
```diff
@@ -127,15 +126,10 @@ class blocking_actor_runner : public resumable {
...
-    [[maybe_unused]] auto id = self_->id();
     auto& sys = self_->system();
-    intrusive_ptr_release(self_->ctrl());
     sys.release_private_thread(thread_);
-    if (!hidden_) {
-      [[maybe_unused]] auto count = sys.registry().dec_running();
-      log::system::debug("actor {} decreased running count to {}", id, count);
-    }
+    self_->cleanup(std::move(rsn), ctx);
+    intrusive_ptr_release(self_->ctrl());
```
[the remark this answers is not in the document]
Neverlord: Releasing the thread simply adds it to the list of free threads again. If a thread receives a new job in the brief moment between releasing the thread and cleaning up, that's fine. Simply means the new job is delayed a little bit.

libcaf_core/caf/actor_system.test.cpp
```diff
@@ -22,32 +22,28 @@ TEST("spawn_inactive creates an actor without launching it") {
...
       self->become([](int) {});
       check_eq(self->ctrl()->strong_refs, 1u); // 1 ref by launch
       strong_self = actor{self};
     }
-    check_eq(sys.registry().running(), baseline + 1);
+    check_eq(sys.registry().running(), 1u);
     strong_self = nullptr;
     sys.await_all_actors_done();
-    check_eq(sys.registry().running(), baseline);
+    check_eq(sys.registry().running(), 0u);
```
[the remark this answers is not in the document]
Neverlord: I don't remember why we have added the baselines here. All system-level actors are always hidden, so only user-spawned actors should ever increase this count and a baseline is unnecessary.

libcaf_core/caf/actor_system.cpp
```diff
@@ -589,6 +551,38 @@ class actor_system::impl {
...
+    if (count <= 1) {
+      std::unique_lock guard{running_actors_mtx};
+      running_actors_cv.notify_all();
+    }
+    return count;
+  }
+
+  void await_running_actors_count_equal(size_t expected,
+                                        timespan timeout = infinite) {
+    CAF_ASSERT(expected == 0 || expected == 1);
```
riemass, line 572: Glad to have the assert there, otherwise it could be misused.

libcaf_core/caf/abstract_actor.cpp
```diff
@@ -166,32 +165,17 @@ bool abstract_actor::cleanup(error&& reason, scheduler* sched) {
...
     return false;
-  log::core::debug("cleanup: id = {}, node = {}, fail-state = {}", id(), node(),
-                   fail_state_);
+  log::core::debug("actor {} cleans up with reason {}", id(), fail_state_);
   // send exit messages
   for (attachable* i = head.get(); i != nullptr; i = i->next.get())
     i->actor_exited(fail_state_, sched);
-  unregister_from_system();
+  if (getf(is_registered_flag)) {
+    home_system().dec_running_actors_count(id());
```
riemass, line 173: The original code also had `unsetf(is_registered_flag);`. I suppose this is not an issue since we're cleaning up?

Neverlord: Cleanup already ensures that it will only run once via the `terminated` flag. I've removed clearing the `registered` flag because this is the only place now where we call `dec_running_actors_count` from and it is already taken care of that this happens only once. So clearing the flag is basically just a dead write.

Issue #2226, opened 2026-01-13
HTTP server blocks incoming connections indefinitely after hitting the `max-connections` threshold


Neverlord:
When hitting the `max-connections` threshold, the HTTP server stops accepting incoming connections. That is supposed to happen. However, it seems like it never resumes to accept new connections again after accepted connections shut down.

PR #2228, opened 2026-01-14
Backport max-connections fix to the CAF 1 series

Neverlord:
Backports #2227 to the `release/1` branch for inclusion in CAF 1.2.

Relates #2226.

### DIFF
--- a/libcaf_net/caf/net/dsl/server_config.hpp
+++ b/libcaf_net/caf/net/dsl/server_config.hpp
@@ -79,7 +79,7 @@ public:
     }
   };
 
-  using socket_t = server_config_tag<error>;
+  using socket_t = server_config_tag<socket>;
 
   static constexpr auto socket_v = socket_t{};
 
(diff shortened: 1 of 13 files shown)


libcaf_net/caf/net/dsl/server_config.hpp
```diff
@@ -79,7 +79,7 @@ class server_config {
     }
   };
 
-  using socket_t = server_config_tag<error>;
+  using socket_t = server_config_tag<socket>;
```
riemass, line 83: Is this an accident?

Neverlord: No, the alias is wrong. I've reverted it for now. Seems like no one calls the DSL with an accept token, at least no one complained yet. I left this alias and just fixed `has_accept.hpp` to keep the changes minimal.

Neverlord: Nevermind, the alias is used in several places, just patching `has_accept.hpp` isn't enough. Let's keep that fix and I'll add a changelog entry for it.

Issue #2233, opened 2026-01-21
Get rid of `actor_config::init_fun` and `local_actor::initial_behavior_fac_`


Neverlord:
Those two are remnants of the past. We use them to implement spawning actors from function objects, but they do this in a very counterintuitive way. Basically, setting `init_fun` on the `actor_config` will initialize `initial_behavior_fac_`. If the latter is non-null when launching an actor, it will bypass everything else and create its initial behavior from that factory. We have better tools available in CAF today to spawn function objects with less black magic and it also locks us into supporting `init_fun` for all actors, even if it makes no sense. And it already makes no sense in blocking actors: they will call the factory and then silently discard the `behavior`.

PR #2234, opened 2026-01-21
Avoid using init_fun for spawning function objects

Neverlord:
Relates #2233.

### DIFF
--- a/libcaf_core/caf/stateful_actor.hpp
+++ b/libcaf_core/caf/stateful_actor.hpp
@@ -4,12 +4,16 @@
 
 #pragma once
 
+#include "caf/actor_traits.hpp"
+#include "caf/behavior.hpp"
+#include "caf/callback.hpp"
 #include "caf/detail/concepts.hpp"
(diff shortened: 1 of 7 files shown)


libcaf_core/caf/stateful_actor.hpp
```diff
@@ -95,15 +124,75 @@ namespace caf::detail {
...
+    }
+  }
+
+  behavior_type make_behavior() {
+    if (fn) {
+      auto res = (*fn)(self);
+      fn = nullptr;
+      return res;
+    }
+    return behavior_type{unsafe_behavior_init};
```
riemass, line 191: This behavior can only happen if we call `make_behavior` twice (by accident), or is the blocking actor behavior covered in this case as well?

Neverlord: Shouldn't happen, but also shouldn't crash the program if a user calls it again for some reason (we reset the init function after calling it).

PR #2238, opened 2026-01-22
fix websocket handshake response headers check

zhongzichang:
HTTP header names are not case-sensitive according to HTTP specifications (RFC 9110 and prior RFCs). However, the original code used conditional statements like `if (field == “Upgrade”)`, which caused handshake failures with certain servers.


Neverlord:
Thank you very much for the patch! 🙏

PR #2242, opened 2026-01-24
Add new panic function for formatted fatal errors

Neverlord:
@riemass as per your suggestion. 🙂


Neverlord:
@riemass ah, in case you wonder: my first attempt was to add support for `detail::format` directly to `critical`, but that didn't go well. We use `critical` in a couple headers. Pulling in `caf/detail/format` causes some nasty cyclic dependencies.

PR #2243, opened 2026-01-25
fix: call message_builder::append_tuple with index_sequence

allspark:


Neverlord:
Thank you for the fix! 🙏

We'll also backport it for inclusion in CAF 1.2.

PR #2255, opened 2026-01-28
Revert "Avoid using init_fun for spawning function objects"

Neverlord:
Reverts actor-framework/actor-framework#2234

Apparently, this broke things for users even though all of our tests pass. Since the exact cause is unclear and we have no reproducer: reverting for now and we'll revisit this at some later time.


riemass (approved):
Best way to go forward. Should we make a ticket to revisit it when we get more time?

Neverlord:
> Best way to go forward. Should we make a ticket to revisit it when we get more time?

The original ticket is #2233. Since the PRs are linked, I think it should be easy to pick up from there.

Issue #2115, opened 2025-07-17
Address spawning detached actors inside the deterministic test fixture


riemass:
Currently, the deterministic test fixture does not handle detached actors, which can easily lead to confusing test failures or unexpected behavior. 

To avoid this pitfall, CAF should either immediately fail the test (with a clear error message) if a detached actor is spawned under the deterministic fixture. Alternatively, we could try spawning such actors as attached instead. This approach should be configurable as it could introduce blocking inside the scheduler.

This will help users avoid subtle bugs and make the framework’s behavior more predictable during testing.

PR #2259, opened 2026-01-30
Split the logger into two interfaces

Neverlord:
Having the `start` and `stop` member functions on the public `logger` interface semantically makes no sense and they are only there because the `actor_system` needs them. We "hide" those methods by making them private and then tag the `actor_system` as a friend. However, just having an internal interface that adds those methods that the actor system then uses separates concerns much better.

Furthermore, the `init` method could be removed entirely by simply initializing all required state in the constructor directly.

Relates #2115. However, the relation is not obvious: this is the first PR in a series of PRs.
1. Break the `friend` requirement between `actor_system` and `logger` (this PR).
2. Have a proper interface for the PIMPL in `actor_system`.
3. Allow constructing the `actor_system` with a custom PIMPL, then have the deterministic fixture implement its own specialization and use that. By doing that, the new specialization can intercept "detaching" of actors. And more, which will allow us to clean up some of the current hackery around the deterministic fixture setup.


libcaf_core/caf/logger.test.cpp
```diff
@@ -41,7 +42,7 @@ TEST("line_builder appends strings with space separators") {
 TEST("logger make returns a non-null intrusive_ptr") {
```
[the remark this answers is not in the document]
Neverlord: Done.

libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -0,0 +1,636 @@
...
+  // of file_filter_ and console_filter_ if both outputs are enabled.
+  std::vector<std::string> global_filter_;
+
+  // Filters events by component name for file output.
+  std::vector<std::string> file_filter_;
+
+  // Filters events by component name for console output.
+  std::vector<std::string> console_filter_;
+
+  // Identifies the thread that called `asynchronous_logger::start`.
```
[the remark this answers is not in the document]
Neverlord: The comment clearly states `asynchronous_logger::start`.

libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -0,0 +1,636 @@
...
+    if (cfg_.verbosity != log::level::quiet) {
+      if (cfg_.file_verbosity > log::level::quiet
+          && cfg_.console_verbosity > log::level::quiet) {
+        read_filter(file_filter_, "caf.logger.file.excluded-components");
+        read_filter(console_filter_, "caf.logger.console.excluded-components");
+        std::sort(file_filter_.begin(), file_filter_.end());
+        std::sort(console_filter_.begin(), console_filter_.end());
+        std::set_intersection(file_filter_.begin(), file_filter_.end(),
+                              console_filter_.begin(), console_filter_.end(),
+                              std::back_inserter(global_filter_));
```
shariarriday, line 232: A possible small improvement: we are considering the intersection of the proposed filter here. And `accepts` ensure that common components are excluded. That would mean that we can consider removing those common components from file and console filter to reduce duplicate checks.

Neverlord: Mind picking that up as a followup?

shariarriday: Sure, I will create a PR for it.

PR #2261, opened 2026-01-31
Deprecate caf::io::network utilities

### LINKED ISSUE #2258: Deprecate `caf::io::network` utilities
The classes under `caf::io::network` are long superseded by cleaner APIs in `caf::net`. However, we should still go through a deprecation cycle and then revisit https://github.com/actor-framework/actor-framework/pull/2257 for CAF 3.0.

Neverlord:
Closes #2258.


Neverlord:
Maybe needs compiler-specific macros to export + deprecate in one. Looks like compilers are picky about ordering here.

PR #2264, opened 2026-02-04
Remove duplicate filters

shariarriday:
Small improvement for filter logic as discussed in https://github.com/actor-framework/actor-framework/pull/2259#discussion_r2751675547

### DIFF
--- a/libcaf_core/caf/detail/asynchronous_logger.cpp
+++ b/libcaf_core/caf/detail/asynchronous_logger.cpp
@@ -230,16 +230,12 @@ public:
         std::set_intersection(file_filter_.begin(), file_filter_.end(),
                               console_filter_.begin(), console_filter_.end(),
                               std::back_inserter(global_filter_));
-        std::erase_if(file_filter_,
-                      [&global = global_filter_](const auto& filter) {
-                        return std::ranges::binary_search(global.begin(),
-                                                          global.end(), filter);
-                      });
-        std::erase_if(console_filter_,
-                      [&global = global_filter_](const auto& filter) {
-                        return std::ranges::binary_search(global.begin(),
-                                                          global.end(), filter);
-                      });
+        std::erase_if(file_filter_, [this](const auto& filter) {
+          return std::ranges::binary_search(global_filter_, filter);
+        });
+        std::erase_if(console_filter_, [this](const auto& filter) {
+          return std::ranges::binary_search(global_filter_, filter);
+        });
       } else if (cfg_.file_verbosity > log::level::quiet) {
         read_filter(file_filter_, "caf.logger.file.excluded-components");
         global_filter_ = file_filter_;
--- a/robot/logging/output.robot
+++ b/robot/logging/output.robot
@@ -74,8 +74,8 @@ Run File Logger
     ...    --caf.logger.file.verbosity\=${verbosity}
     ...    --api\=${api}
 
-Run Combined Logger
-    [Arguments]    ${console_excluded_components}   ${file_excluded_components}   ${api}
+Run Console And File Logger
+    [Arguments]    ${console_excluded_components}   ${file_excluded_components}
     Remove File    app.out
     Remove File    app.log
     Run Process    ${BINARY_PATH}
@@ -84,10 +84,13 @@ Run Combined Logger
     ...    --caf.logger.file.verbosity\=debug
     ...    --caf.logger.console.excluded-components\=${console_excluded_components}
     ...    --caf.logger.file.excluded-components\=${file_excluded_components}
-    ...    --api\=${api}
     ...    stderr=app.out
     ${console_output}=    Get File    app.out
-    ${file_output}=    Get File    app.log
+    TRY
+        ${file_output}=    Get File    app.log
+    EXCEPT
+        ${file_output}=    Set Variable    ${EMPTY}
+    END
     RETURN        ${console_output}     ${file_output}
 
 *** Test Cases ***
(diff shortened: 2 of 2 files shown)


libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -230,6 +231,13 @@ class default_logger : public detail::asynchronous_logger,
         std::set_intersection(file_filter_.begin(), file_filter_.end(),
                               console_filter_.begin(), console_filter_.end(),
                               std::back_inserter(global_filter_));
+        std::erase_if(file_filter_, [&global = global_filter_](const auto& filter) {
+          return std::binary_search(global.begin(), global.end(), filter);
+        });
+        std::erase_if(console_filter_, [&global = global_filter_](
+                                         const auto& filter) {
+          return std::binary_search(global.begin(), global.end(), filter);
```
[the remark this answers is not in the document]
shariarriday: Some new unit test has been added to verify filter logic. However, we can consider adding some robot tests to verify that components are excluded properly. 
@Neverlord please let me know your thoughts on updating robot test for the logger.

libcaf_core/caf/logger.test.cpp
```diff
@@ -101,3 +101,56 @@ SCENARIO("logger accepts returns false for levels above verbosity") {
     }
   }
 }
+
+SCENARIO("logger accepts returns false when component is excluded for file and "
+         "console") {
```
Neverlord:
```suggestion
SCENARIO("loggers reject excluded components") {
```

libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -230,6 +230,14 @@ class default_logger : public detail::asynchronous_logger,
         std::set_intersection(file_filter_.begin(), file_filter_.end(),
                               console_filter_.begin(), console_filter_.end(),
                               std::back_inserter(global_filter_));
+        std::erase_if(file_filter_, [&global = global_filter_](
+                                      const auto& filter) {
+          return std::binary_search(global.begin(), global.end(), filter);
```
Neverlord: We recently made a pass to get rid of the iterator-based algorithms where we could. Please don't bring them back for no reason.

```suggestion
        std::erase_if(file_filter_, [&global_filter_](const auto& entry) {
          return std::ranges::binary_search(global_filter, entry);
```

Same below.

libcaf_core/caf/logger.test.cpp
```diff
@@ -101,3 +101,56 @@ SCENARIO("logger accepts returns false for levels above verbosity") {
...
+  GIVEN("an actor system with excluded core component for file and console") {
+    actor_system_config cfg;
+    cfg.set("caf.logger.console.excluded-components", "[\"core\"]");
```
Neverlord: Escaped strings are terribly noisy. IIRC, you can omit the quotes: `"[core]"`. Otherwise use raw string literals.

```
    cfg.set("caf.logger.console.excluded-components", "[\"core\"]");
```

libcaf_core/caf/detail/asynchronous_logger.cpp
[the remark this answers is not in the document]
Neverlord: Should be tested in robot/logging/output.robot. Doesn't need new baselines, only tests re-using the existing ones.

libcaf_core/caf/logger.test.cpp
shariarriday: Sadly, `[core]` does not convert to a string_list. I have updated the test to use raw string literals.

libcaf_core/caf/detail/asynchronous_logger.cpp
shariarriday: Thanks, I missed it. Updated now.

libcaf_core/caf/logger.test.cpp
shariarriday: Test name has been updated.

libcaf_core/caf/detail/asynchronous_logger.cpp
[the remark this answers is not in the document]
shariarriday: New robot test has been added utilizing the same baselines.

libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -230,6 +230,16 @@ class default_logger : public detail::asynchronous_logger,
...
                               std::back_inserter(global_filter_));
+        std::erase_if(file_filter_,
+                      [&global = global_filter_](const auto& filter) {
+                        return std::ranges::binary_search(global.begin(),
+                                                          global.end(), filter);
+                      });
+        std::erase_if(console_filter_,
+                      [&global = global_filter_](const auto& filter) {
+                        return std::ranges::binary_search(global.begin(),
+                                                          global.end(), filter);
```
Neverlord: You have updated to the `ranges` version but why are we still passing in iterator pairs? That's completely missing the point.

robot/logging/output.robot
```diff
@@ -142,6 +162,24 @@ The console logger prints a subset when setting the log level to ERROR
...
+        ${console}    ${file}=    Run Combined Logger    ${console_excluded}    ${file_excluded}    ${api}
+        IF    ${console_receives}
+            Should Be Equal As Strings    ${DEBUG_BASELINE}    ${console}    formatter=repr
+            Should Be Empty    ${file}
+        ELSE
+            Should Be Equal As Strings    ${DEBUG_BASELINE}    ${file}    formatter=repr
+            Should Be Empty    ${console}
+        END
+    END
+
```
Neverlord, line 185: The `legacy` cases can go, they don't exercise different code paths here.

Not a fan of the for loop. There are three cases that we need to test:
1. `caf.core` excluded for console and file loggin: output is empty for both
2. `caf.core` excluded for console but not file logging: only console output is empty
3. `caf.core` excluded for file logging but not console logging: only file output is empty

Having three different tests makes it much easier to grasp what went wrong where if tests fail.

shariarriday: `base.cfg` already lists `caf.core` as excluded component and only `app` component is included in current set of tests. The robot test has been updated to exclude the `app` component in different scenarios.

libcaf_core/caf/detail/asynchronous_logger.cpp
```diff
@@ -230,6 +230,14 @@ class default_logger : public detail::asynchronous_logger,
         std::set_intersection(file_filter_.begin(), file_filter_.end(),
                               console_filter_.begin(), console_filter_.end(),
                               std::back_inserter(global_filter_));
+        std::erase_if(file_filter_,
+                      [&global = global_filter_](const auto& filter) {
+                        return std::ranges::binary_search(global, filter);
+                      });
+        std::erase_if(console_filter_,
+                      [&global = global_filter_](const auto& filter) {
+                        return std::ranges::binary_search(global, filter);
```
Neverlord: Why the complex lambda captures and not just like this?

```suggestion
        std::erase_if(file_filter_,
                      [this](const auto& filter) {
                        return std::ranges::binary_search(global_filter_, filter);
                      });
        std::erase_if(console_filter_,
                      [this](const auto& filter) {
                        return std::ranges::binary_search(global_filter_, filter);
```

robot/logging/output.robot
```diff
@@ -70,6 +74,25 @@ Run File Logger
     ...    --caf.logger.file.verbosity\=${verbosity}
     ...    --api\=${api}
 
+Run Combined Logger
```
Neverlord: What is a combined logger? There are multiple loggers running here. Maybe like this?


```suggestion
Run Console And File Logger
```

robot/logging/output.robot
```diff
@@ -142,6 +165,24 @@ The console logger prints a subset when setting the log level to ERROR
     ${found}=    Run Console Logger  error  default
     Should Be Equal As Strings  ${ERROR_BASELINE}  ${found}  formatter=repr
 
+The logger rejects logs for file and console when all components are excluded
```
Neverlord:
```suggestion
The logger rejects file and console logging for excluded components
```

robot/logging/output.robot
```diff
@@ -142,6 +165,24 @@ The console logger prints a subset when setting the log level to ERROR
...
+    [Tags]    logging
+    ${console}    ${file}=    Run Combined Logger    ${ALL_EXCLUDED_COMPONENT}    ${ALL_EXCLUDED_COMPONENT}
+    Should Be Empty    ${console}
+    Should Be Empty    ${file}
+
+The logger rejects logs for file when all components are excluded but not for console
```
Neverlord:
```suggestion
The logger rejects file logging for excluded components
```

robot/logging/output.robot
```diff
@@ -142,6 +165,24 @@ The console logger prints a subset when setting the log level to ERROR
...
+    [Tags]    logging
+    ${console}    ${file}=    Run Combined Logger    ${COMMON_EXCLUDED_COMPONENT}    ${ALL_EXCLUDED_COMPONENT}
+    Should Be Equal As Strings    ${DEBUG_BASELINE}    ${console}    formatter=repr
+    Should Be Empty    ${file}
+
+The logger rejects logs for console when all components are excluded but not for file
```
Neverlord:
```suggestion
The logger rejects console logging for excluded components
```

PR #2272, opened 2026-02-20
Drop multiplexer dependency to the middleman

Neverlord:
Nothing in CAF ever calls `multiplexer::owner`, so the dependency seems accidental and/or is an oversight from earlier iterations.

It's *technically* public API, but the multiplexer really isn't a user facing API. And the fix is simply calling `system().network_manager()` instead, so it's trivial.


riemass (approved):
I'm all for these changes. Hope nobody complains there was no deprecation cycle.

Neverlord:
> I'm all for these changes. Hope nobody complains there was no deprecation cycle.

If someone does: it's my fault. 😉

I'd be very surprised, though.

PR #2277, opened 2026-02-26
Give external access to message_rejected metrics

riemass:
Give access to `message_rejected` through `actor_system_access`.

### DIFF
--- a/libcaf_core/caf/actor_system.hpp
+++ b/libcaf_core/caf/actor_system.hpp
@@ -105,11 +105,7 @@ public:
   friend class actor_pool;
   friend class blocking_actor;
   friend class detail::actor_system_access;
-  friend class detail::response_promise_state;
   friend class local_actor;
-  friend class net::abstract_actor_shell;
-  friend class net::middleman;
-  friend class scheduled_actor;
 
   template <class>
   friend class actor_from_state_t;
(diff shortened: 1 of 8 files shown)


libcaf_core/caf/detail/response_promise_state.cpp
```diff
@@ -41,7 +42,8 @@ void response_promise_state::deliver_impl(message msg) {
     auto element = make_mailbox_element(self, id.response_id(), std::move(msg));
     source->enqueue(std::move(element), selfptr->context());
   } else {
-    selfptr->home_system().message_rejected(nullptr);
+    detail::actor_system_access{selfptr->home_system()}.message_rejected(
+      nullptr);
```
Neverlord: Can't we remove the `friend` declaration now?

riemass: You're right.

libcaf_core/caf/actor_system.hpp
```diff
@@ -105,11 +105,7 @@ class CAF_CORE_EXPORT actor_system {
   friend class actor_pool;
   friend class blocking_actor;
   friend class detail::actor_system_access;
-  friend class detail::response_promise_state;
   friend class local_actor;
-  friend class net::abstract_actor_shell;
-  friend class net::middleman;
-  friend class scheduled_actor;
```
Neverlord, line 109: In C++ land, having fewer friends is actually a good thing. 😄

PR #2279, opened 2026-02-27
Convert http::router to an interface

### LINKED ISSUE #2223: Convert `http::router` to an interface
Convert `http::router` to an interface and hidden implementation class, similar to what's done with `http::server`. This will help ABI stability in the future.

riemass:
Closes #2223.

### DIFF
--- a/libcaf_net/caf/net/http/router.hpp
+++ b/libcaf_net/caf/net/http/router.hpp
@@ -54,13 +54,6 @@ public:
   void shutdown(const error& err);
 
   virtual void abort_and_shutdown(const error& err) = 0;
-
-protected:
-  /// Creates a request from the given components. Only the router
-  /// implementation may construct requests.
-  static request make_request(request_header hdr, std::vector<std::byte> body,
-                              async::promise<response> prom,
-                              detail::connection_guard_ptr conn_guard);
 };
 
 } // namespace caf::net::http
--- a/libcaf_net/caf/net/http/request.hpp
+++ b/libcaf_net/caf/net/http/request.hpp
@@ -25,9 +25,11 @@ namespace caf::net::http {
 /// promise for the HTTP response.
 class CAF_NET_EXPORT request {
 public:
-  friend class router;
-
   class impl;
+
+  request(request_header hdr, std::vector<std::byte> body,
+          async::promise<response> prom,
+          detail::connection_guard_ptr conn_guard);
 
   request() noexcept = default;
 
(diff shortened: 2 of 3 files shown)


libcaf_net/caf/net/http/router.hpp
```diff
@@ -30,86 +26,42 @@ class CAF_NET_EXPORT router : public upper_layer::server {
...
 
   // -- factories --------------------------------------------------------------
 
   static std::unique_ptr<router> make(std::vector<route_ptr> routes);
 
-  static std::unique_ptr<router> make(std::vector<route_ptr> routes,
-                                      detail::connection_guard_ptr guard);
+  static std::unique_ptr<router>
+  make(std::vector<route_ptr> routes,
+       ::caf::detail::connection_guard_ptr guard);
```
Neverlord: Why did you make that change?

```suggestion
       detail::connection_guard_ptr guard);
```

libcaf_net/caf/net/http/router.hpp
```diff
@@ -30,86 +26,42 @@ class CAF_NET_EXPORT router : public upper_layer::server {
...
+  virtual void abort_and_shutdown(const error& err) = 0;
 
-  /// Tracks connection lifetime for outstanding requests.
-  detail::connection_guard_ptr guard_;
+protected:
+  /// Creates a request from the given components. Only the router
+  /// implementation may construct requests.
+  static request make_request(request_header hdr, std::vector<std::byte> body,
+                              async::promise<response> prom,
+                              detail::connection_guard_ptr conn_guard);
```
Neverlord: Why is this on the interface if only the implementation calls it?

libcaf_net/caf/net/http/router.hpp
riemass: Mistake. Fixed.

libcaf_net/caf/net/http/router.hpp
riemass: The ctor for `request` is private by design and the `router` is declared as a friend of `request`. However, after the refactoring I can't declare the `router_impl` as a friend as it's a hidden implementation class. So I bridged it like this.

Neverlord: I see. With the previous version, we could just add a `friend` declaration and had a bit of extra encapsulation with no additional overhead. However, I don't think it is worth going through the extra factory function here. Let's just move the constructor up to the public section and drop the `friend` declaration (as well as this new factory). The constructor takes an argument from the `detail` namespace, so it's already clear that this isn't meant for public use.

riemass: The headers work with raw bytes, so that's why I was hesitant with making the ctor public, but it's the cleanest way to go forward.

PR #2293, opened 2026-03-16
Fix max throughput handling in detached actors

Neverlord:
Reported in chat (thanks, @lava!): if a detached actors hits the max-throughput parameter, it currently "escapes" to the regular scheduler. This fix makes sure that code path doesn't run for detached actors to make sure they stay in the private thread.


Neverlord:
> Is there a way to put a regression test in place? As these are detached actors, I understand this is hard, so still approving.

Will be a followup: #2296.

Issue #2298, opened 2026-03-20
Outline tests execute the first test twice and skip the final test


shariarriday:
The following test is not executed correctly in the current testing framework. 

```
const auto expected = std::vector<int>{3, 6, 13};

OUTLINE("comparing numbers") {
  GIVEN("the numbers <x> and <y>") {
    auto [x, y] = block_parameters<double, double>();
    WHEN("adding both numbers") {
      auto result = x + y;
      THEN("the sum: <sum> should be in expected") {
        const auto index = block_parameters_index();
        auto sum = block_parameters<double>();
        check_eq(result, caf::test::approx{sum});
        check_eq(result, expected[index]);
      }
    }
  }
  EXAMPLES = R"(
    |   x |   y | sum |
    |   1 |   2 |   3 |
    | 2.5 | 3.5 |   6 |
    | 9.5 | 3.5 |  13 |
  )";
}
```

The expectation is that the test will pass after executing each line in the example once. But, the first line is executed twice and the last line is not executed at all. Which fails all of the test except the first one.

PR #2302, opened 2026-03-21
Update testing framework to run all examples

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/2298

A new test has been added to `outline.test` to verify that all of the examples have been run.

### DIFF
--- a/libcaf_test/caf/test/runnable_with_examples.cpp
+++ b/libcaf_test/caf/test/runnable_with_examples.cpp
@@ -110,13 +110,11 @@ void runnable_with_examples::run_next_test_branch_init() {
                                       test_location());
     }
   }
-  ctx.parameters = ctx.example_parameters[ctx.example_id];
   auto& ptr = ctx.steps[std::make_pair(0, ctx.example_id)];
-  auto guard = detail::scope_guard{[&ctx, &ptr]() noexcept {
-    if (!ptr->can_run() && ctx.example_id + 1 < ctx.example_parameters.size()) {
-      ++ctx.example_id;
-    }
-  }};
+  if (!ptr->can_run() && ctx.example_id + 1 < ctx.example_parameters.size()) {
+    ++ctx.example_id;
+  }
+  ctx.parameters = ctx.example_parameters[ctx.example_id];
 }
 
 } // namespace caf::test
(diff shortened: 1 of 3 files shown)


libcaf_test/caf/test/outline.test.cpp
```diff
@@ -95,3 +97,31 @@ OUTLINE("an outline with multiple when blocks") {
...
+  EXAMPLES = R"(
+    |   x |   y | sum |
+    |   1 |   2 |   3 |
+    | 2.5 | 3.5 |   6 |
+    | 9.5 | 3.5 |  13 |
+  )";
+}
+
+TEST("outline completes all tests") {
+  check(expected.empty());
```
Neverlord: A test that requires that another test ran first... I can see what your thought process was, but that's just too brittle.

Please remove all the global state and side effects here. A way to do this reliable is to run this test from a Robot test with `-s outline -t 'comparing numbers' -v 4` and add some recognizable output here that we can search for.

Something like this:

```c++
auto sum = block_parameters<double>();
if (check_eq(result, caf::test::approx{sum})) {
  log::test::debug("{} + {} = {}", x, y, sum);
}
```

In the Robot test, we then can check that this line appears once for each example given.

Neverlord: Actually, we don't need to add a new test here at all. The Robot test can just run `eating cucumber` or some other existing test if you add a recognizable output to it.

shariarriday: Thanks, I was already thinking about removing only the final test and replacing it with a robot test. But, running a robot test on existing test and ensuring that all example runs by comparing a recognizable output will be the best course of action. I will update tests accordingly.

libcaf_test/caf/test/runnable_with_examples.cpp
```diff
@@ -110,12 +110,12 @@ void runnable_with_examples::run_next_test_branch_init() {
                                       test_location());
     }
   }
-  ctx.parameters = ctx.example_parameters[ctx.example_id];
   auto& ptr = ctx.steps[std::make_pair(0, ctx.example_id)];
   auto guard = detail::scope_guard{[&ctx, &ptr]() noexcept {
```
Neverlord: Looks like a refactoring artifact. This scope guard sits at the end of the function, so it will be called immediately anyways.

```suggestion
```

Issue #2307, opened 2026-03-27
Clean up our reference counting implementations


Neverlord:
We have the class `ref_counted`, which serves as a base class for reference counted objects. However, we can't use that base class in interface classes because that would suffer from the diamond problem. As a workaround in CAF, we have several interface types that implement their own reference counting methods, namely:

- `ref_consumer`
- `ref_coordinated`
- `ref_disposable`
- `ref_execution_context`
- `ref_listener`
- `ref_logger`
- `ref_mailbox`
- `ref_producer`
- `ref_resumable`

Plus the corresponding `deref`, of course. The usual pattern is that those types add some suffix to `ref` and then the implementation class pulls in `ref_counted` and overrides the virtual ref counting methods using `ref()` and `deref()`.

Because we have multiple names for our reference counting methods, we also have 41 matches for `intrusive_ptr_add_ref` for types that need to be compatible with `intrusive_ptr`.

To cut a long story short, reference counting in CAF has gotten quite convoluted over the years and is in dire need of streamlining since it forces us to write way too much boilerplate code. Additionally, we have multiple implementations for reference counting since not all types can use `ref_counted`. For example, `chunk::data` cannot use inheritance and thus has to roll its own implementations for `ref` and `deref`.

So as a first step, we should rethink how we implement our reference counting and have an approach that works with types like `chunk::data` as well.

After that, we should introduce a new base type such as `abstract_ref_counted` that we can use wherever an implementation might inherit it multiple times. That will get rid of all of the suffixed `ref_` and `deref_` methods.

Last but not least, `intrusive_ptr` should first look if a type implements `intrusive_ptr_add_ref` and `intrusive_ptr_release`. We have a few instances where we forward-declare them to make `intrusive_ptr<T>` work with only a forward declaration of `T`, even when used as a member variable. If the free function does not exist, it should try to call `ref()` and `deref()` on the type. This change will eliminate the need for almost all of the overloads of the free function pair.

PR #2308, opened 2026-03-27
Implement new `atomic_ref_count` class

Neverlord:
The inheritance-based approach with `atomic_ref_counted` and `plain_ref_counted` did not work in all cases. For example, `chunk::data` cannot use inheritance and thus had to roll its own reference counting implementation.

The new class `atomic_ref_count` is designed to be used as a member variable and works in all cases where we need an intrusive reference count. This approach simplifies the design and makes sure that we only have a single implementation.

This is the first part for #2307.


libcaf_core/caf/detail/memory_interface.hpp
```diff
@@ -0,0 +1,45 @@
...
+  /// Indicates that the class uses `new` and `delete`. This is the default.
+  new_and_delete,
+  /// Indicates that the class uses `malloc` and `free`.
+  malloc_and_free,
+};
+
+/// Checks whether a class has a static `memory_interface` member constant.
+template <class T>
+concept has_memory_interface = requires(T) {
+  { T::memory_interface } -> std::same_as<memory_interface>;
```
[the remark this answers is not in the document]
Neverlord: Fixed.

libcaf_core/caf/intrusive_cow_ptr.hpp
```diff
@@ -4,34 +4,29 @@
...
+#include "caf/detail/build_config.hpp"
 #include "caf/detail/comparable.hpp"
 #include "caf/intrusive_ptr.hpp"
 
 #include <cstddef>
 #include <cstdint>
+#include <tuple>
 
-namespace caf {
+namespace caf::detail {
```
shariarriday, line 20: Would it be better to move this to a separate file? We are declaring two different namespace in this file.

Neverlord: Those are helpers needed as an implementation detail for this class only. Our convention is putting those helper alongside the implementation since they don't stand on their own.

PR #2310, opened 2026-03-31
Host header should contain explicit non-default port

### LINKED ISSUE #2309: Include explicitly given port number in the HTTP Host header field
Reported in chat by @lava (thanks): 
When using caf's http client with an URL that includes an explicit port (eg. `https://foo.example:4444`) the `Host` header sent by caf does not include the port. This breaks e.g. S3 presigned URLs, where the Host header is part of the url signature.

riemass:
Closes #2309

### DIFF
--- a/libcaf_net/caf/net/http/with.cpp
+++ b/libcaf_net/caf/net/http/with.cpp
@@ -16,8 +16,6 @@
 #include "caf/defaults.hpp"
 #include "caf/detail/connection_acceptor.hpp"
 #include "caf/detail/connection_guard.hpp"
-#include "caf/detail/format.hpp"
-#include "caf/detail/overload.hpp"
 #include "caf/internal/accept_handler.hpp"
 #include "caf/internal/make_transport.hpp"
 #include "caf/internal/net_config.hpp"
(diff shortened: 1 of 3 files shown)


riemass:
This also needs a backport to CAF 1.1.

libcaf_net/caf/net/http/with.hpp
```diff
@@ -337,4 +338,12 @@ class CAF_NET_EXPORT with_t {
   config_ptr config_;
 };
 
+/// Builds the RFC 7230 `Host` header value from @p endpoint. For `http` and
+/// `https`, the host part matches `caf::to_string(uri::authority_type)` (IPv6
+/// literals are bracketed unless IPv4-embedded); the port is omitted when it is
+/// unset, or equal to the default (80 or 443). For other schemes, returns
+/// `authority_type::host_str()`.
+/// @param endpoint URI whose authority is used (typically `http` or `https`).
+CAF_NET_EXPORT std::string host_header_value(const uri& endpoint);
```
Neverlord: A free function that doesn't operator on `with` does not belong here. Also, this will make it part of the public API and the ABI. It's not used anywhere outside of `with.cpp`, so it should live in the anonymous namespace there.

libcaf_net/caf/net/http/with.cpp
```diff
@@ -430,6 +434,43 @@ void with_t::client::do_add_header_field(std::string name, std::string value) {
   config_->fields.insert(std::pair{std::move(name), std::move(value)});
 }
 
+namespace {
+
+/// Hostname or bracketed IP literal for the RFC 7230 Host field (matches the
+/// authority host serialization in `caf::to_string(uri::authority_type)`).
+std::string host_field_hostpart(const uri::authority_type& auth) {
```
Neverlord: Why are we putting in an incomplete copy of `std::string to_string(const uri::authority_type& x)`?

Can't we just use the existing code instead by "filtering" the parts that should be omitted?

```c++
auto auth = endpoint.authority();
if (auth.userinfo) {
  auth.userinfo = std::nullopt; // suppress userinfo in 'Host' header field
}
if (scheme == "http" && auth.port == defaults::net::http_default_port) {
  auth.port = 0; // suppress default HTTP port in 'Host' Header field
}
if (scheme == "https" && auth.port == defaults::net::https_default_port) {
  auth.port = 0; // suppress default HTTPS port in 'Host' Header field
}
auto host_field = to_string(auth);
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -348,3 +349,103 @@ TEST("GH-2226 regression") {
   hdl->dispose();
   check_eq(elog->errors(), std::vector<error>{});
 }
+
+TEST("GH-2309 regression: Host header contains explicit non-default port") {
```
Neverlord, line 355: I'm not particularly fond of this setup. Opening a port and then having the client connect is quite involved. Why not use a connected socket pair and then allow something like this?

```c++
auto res = net::http::with(sys)
           .connect(endpoint, client_fd)
           .connection_timeout(2s)
           .max_retry_count(1)
           .get();
```

Allowing to pass a socket for the HTTP communication would make testing the stack a lot easier.

Then we also don't need to export internal functions only to test them, like in `TEST("host_header_value")` below. Testing that one function doesn't guarantee that the client actually uses that function for generating the `Host` header field.

riemass:
Agree on all review items, I did expose stuff I didn't want only for code coverage. Removed that now. 
Since we can't do the socket_pair in the with API, I'll enable that as a followup, and for now we have a bug fix PR. I did manually check that the new `host_header_value` with `to_string` has the correct output.

Issue #2291, opened 2026-03-16
`http::request` and `http::promise` have no way to set HTTP header fields


Neverlord:
On an `http::responder`, users can explicitly transmit HTTP header fields. However, once converting the responder to a `request` or `promise`, there's no API in place to do that.

heibor:
I have added a test case in the  PR for this issue: https://github.com/actor-framework/actor-framework/pull/2372

PR #2315, opened 2026-04-03
[2291] Add header fields to the http::promise to set the HTTP header fields of the response

heibor:
In the issue[2291], user can explicitly transmit HTTP header fields by using the `respond(status code, std::unordered_map<std::string, std::string> headers, const_byte_span content)` method.

### DIFF
--- a/libcaf_net/caf/net/http/lower_layer.hpp
+++ b/libcaf_net/caf/net/http/lower_layer.hpp
@@ -71,6 +71,10 @@ public:
                      const_byte_span content);
 
   /// @copydoc send_response
+  bool send_response(status code, std::unordered_map<std::string, std::string> headers,
+                     const_byte_span content);
+
+  /// @copydoc send_response
   bool send_response(status code, std::string_view content_type,
                      std::string_view content);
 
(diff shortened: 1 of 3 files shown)


heibor:
This commit can fill some gap in https://github.com/actor-framework/actor-framework/issues/2291

libcaf_net/caf/net/http/lower_layer.hpp
```diff
@@ -70,6 +70,10 @@ class CAF_NET_EXPORT lower_layer::server : public lower_layer {
   bool send_response(status code, std::string_view content_type,
                      const_byte_span content);
 
+  /// @copydoc send_response
+  bool send_response(status code, std::unordered_map<std::string, std::string> headers,
+                     const_byte_span content);
```
Neverlord, line 75: We generally avoid committing to a specific container type in CAF unless we need to. This is not a use case where a hash map will shine, since we basically assemble the header fields once, iterate them, and dispose the container. For this method, `std::span<const std::pair<std::string, std::string>>` is sufficient and client code can use `caf::unordered_flat_map` to put that together.

Neverlord (changes requested):
Thank you for contributing to CAF!

Aside from the compiler error seen on CI and my comment regarding the map type, the new function also needs a test in `responder.test.cpp`.

heibor:
I have rewritten the send_response method with a new parameter in the PR: https://github.com/actor-framework/actor-framework/pull/2372

Neverlord:
Replaced by #2372.

Next time, please update the existing PR instead of opening a new one. 🙂

PR #2372, opened 2026-04-18
[2291] Add a new reponse method on the http::promise

heibor:
On an http::responder, users can explicitly transmit HTTP header fields with the promise.


libcaf_net/caf/net/http/responder.hpp
```diff
@@ -79,6 +79,14 @@ class CAF_NET_EXPORT responder {
       impl_->set_completed();
     }
 
+    /// Sends an HTTP response message to the client. Automatically sets the
+    /// the `Content-Length` header fields.
+    void respond(status code, caf::unordered_flat_map<std::string, std::string> headers,
```
Neverlord:
```suggestion
    void respond(status code, std::span<const std::pair<std::string, std::string>> headers,
```

libcaf_net/caf/net/http/responder.test.cpp
```diff
@@ -222,4 +222,28 @@ TEST("responding with custom header fields with payload") {
...
+TEST("promise responding with custom header fields with payload") {
+  route("/test", [](http::responder& res) {
+    caf::unordered_flat_map<std::string, std::string> headers = {
+      {"Content-Type", "text/plain"},
+      {"X-Custom-Header", "custom-value"},
+    };
+    auto hello = "Hello, world!"sv;
+    auto bytes = as_bytes(std::span{hello});
+    auto prom = std::move(res).to_promise();
+    prom.respond(http::status::ok, headers, bytes);
```
Neverlord: The map itself is not convertible to a `span`, but the `vector` it is storing its elements in is:

```suggestion
    prom.respond(http::status::ok, headers.container(), bytes);
```

Neverlord (changes requested):
Two code places need updating and `clang-format` needs to run on the changes (CI uses version 19). Please also squash your commits.

heibor:
@Neverlord Based on your suggestion, I have resubmitted some modifications to the parameters.

Neverlord:
@heibor are you still planning to work on this or should we close the PR?

PR #2423, opened 2026-06-12
Add examples for multipart reader and writer

### LINKED ISSUE #2111: Implement an example for the new multipart API
Followup to #2073: we have yet to add example code showcasing `multipart_reader` and `multipart_writer`.

Once we have examples, we should also mention the new feature in the manual.

shariarriday:
Closes https://github.com/actor-framework/actor-framework/issues/2111

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -91,6 +91,8 @@ else()
 endif()
 
 add_net_example(http client)
+add_net_example(http multipart-reader)
+add_net_example(http multipart-writer)
 add_net_example(http rest)
 add_net_example(http time-server)
 
--- a/examples/http/multipart-reader.cpp
+++ b/examples/http/multipart-reader.cpp
@@ -1,0 +1,138 @@
+// HTTP server example that accepts multipart/form-data uploads via POST /
+// and parses request parts with caf::net::http::multipart_reader.
+//
+// Usage:
+//   multipart-reader [--port=<port>] [--max-connections=<num>]
+//                    [--max-request-size=<bytes>]
+//                    [--tls.key-file=<path> --tls.cert-file=<path>]
+//
+// For each received part, this example prints all part header fields and the
+// part content to stdout, then responds with HTTP 200. Requests with a
+// non-multipart Content-Type or malformed multipart payload are rejected with
+// HTTP 400.
+
+#include "caf/net/http/multipart_reader.hpp"
+#include "caf/net/http/with.hpp"
+#include "caf/net/middleman.hpp"
+
+#include "caf/actor_system.hpp"
+#include "caf/actor_system_config.hpp"
+#include "caf/byte_span.hpp"
+#include "caf/caf_main.hpp"
+#include "caf/defaults.hpp"
+
+#include <atomic>
+#include <chrono>
+#include <csignal>
+#include <string>
+#include <thread>
+
+using namespace std::literals;
+
+// -- constants ----------------------------------------------------------------
+
+static constexpr uint16_t default_port = 8080;
+
+// -- configuration ------------------------------------------------------------
+
+struct config : caf::actor_system_config {
+  config() {
+    opt_group{custom_options_, "global"} //
+      .add<uint16_t>("port,p", "port to listen for incoming connections")
+      .add<size_t>("max-connections,m", "limit for concurrent clients")
+      .add<size_t>("max-request-size,r", "limit for single request size");
+    opt_group{custom_options_, "tls"} //
+      .add<std::string>("key-file,k", "path to the private key file")
+      .add<std::string>("cert-file,c", "path to the certificate file");
+  }
+
+  caf::settings dump_content() const override {
+    auto result = actor_system_config::dump_content();
+    caf::put_missing(result, "port", default_port);
+    caf::put_missing(result, "max-connections", size_t{128});
+    caf::put_missing(result, "max-request-size",
+                     caf::defaults::net::http_max_request_size);
+    return result;
+  }
+};
+
+// -- main ---------------------------------------------------------------------
+
+namespace {
+
+std::atomic<bool> shutdown_flag;
+
+} // namespace
+
+int caf_main(caf::actor_system& sys, const config& cfg) {
+  namespace ssl = caf::net::ssl;
+  namespace http = caf::net::http;
+  // Do a regular shutdown for CTRL+C and SIGTERM.
+  auto set_shutdown_flag = [](int) { shutdown_flag = true; };
+  signal(SIGTERM, set_shutdown_flag);
+  signal(SIGINT, set_shutdown_flag);
+  // Read the configuration.
+  auto port = caf::get_or(cfg, "port", default_port);
+  auto pem = ssl::format::pem;
+  auto key_file = caf::get_as<std::string>(cfg, "tls.key-file");
+  auto cert_file = caf::get_as<std::string>(cfg, "tls.cert-file");
+  if (key_file.has_value() != cert_file.has_value()) {
+    sys.println("*** inconsistent TLS config: declare neither file or both");
+    return EXIT_FAILURE;
+  }
+  // Open up a TCP port for incoming connections and start the server.
+  auto server
+    = http::with(sys)
+        // Optionally enable TLS.
+        .context(ssl::context::enable(key_file && cert_file)
+                   .and_then(ssl::emplace_server(ssl::tls::v1_2))
+                   .and_then(ssl::use_private_key_file(key_file, pem))
+                   .and_then(ssl::use_certificate_file(cert_file, pem)))
+        .accept(port)
+        .max_connections(caf::get_as<size_t>(cfg, "max-connections"))
+        .max_request_size(caf::get_as<size_t>(cfg, "max-request-size"))
+        .route(
+          "/", http::method::post,
+          [&sys](http::responder& res) {
+            auto value = res.payload();
+            auto& header = res.header();
+            if (!header.has_token("Content-Type", "multipart/form-data")) {
+              sys.println("*** expected multipart/form-data content. Found: {}",
+                          header.field("Content-Type"));
+              res.respond(http::status::bad_request);
+              return;
+            }
+            auto reader = caf::net::http::multipart_reader{header, value};
+            if (!reader.for_each([&sys](http::header&& part_header,
+                                        caf::const_byte_span part_content) {
+                  sys.println("Number of fields: {}", part_header.num_fields());
+                  part_header.for_each_field(
+                    [&sys](std::string_view name, std::string_view value) {
+                      sys.println("{}: {}", name, value);
+                    });
+                  auto val = caf::to_string_view(part_content);
+                  sys.println("Content: {}", val);
+                })) {
+              sys.println("*** failed to parse multipart data");
+              res.respond(http::status::bad_request);
+              return;
+            }
+            res.respond(http::status::ok);
+          })
+        // Launch the server.
+        .start();
+  // Report any error to the user.
+  if (!server) {
+    sys.println("*** unable to run at port {}: {}", port, server.error());
+    return EXIT_FAILURE;
+  }
+  // Wait for CTRL+C or SIGTERM.
+  while (!shutdown_flag) {
+    std::this_thread::sleep_for(250ms);
+  }
+  sys.println("*** shutting down");
+  server->dispose();
+  return EXIT_SUCCESS;
+}
+
+CAF_MAIN(caf::net::middleman)
--- a/examples/http/multipart-reader.cpp
+++ b/examples/http/multipart-reader.cpp
@@ -19,7 +19,6 @@
 #include "caf/actor_system_config.hpp"
 #include "caf/byte_span.hpp"
 #include "caf/caf_main.hpp"
-#include "caf/defaults.hpp"
 
 #include <atomic>
 #include <chrono>
@@ -29,19 +28,16 @@
 
 using namespace std::literals;
 
-// -- constants ----------------------------------------------------------------
-
 static constexpr uint16_t default_port = 8080;
 
-// -- configuration ------------------------------------------------------------
-
+// Custom config for adding command line options for the server.
 struct config : caf::actor_system_config {
   config() {
-    opt_group{custom_options_, "global"} //
+    opt_group{custom_options_, "global"}
       .add<uint16_t>("port,p", "port to listen for incoming connections")
       .add<size_t>("max-connections,m", "limit for concurrent clients")
       .add<size_t>("max-request-size,r", "limit for single request size");
-    opt_group{custom_options_, "tls"} //
+    opt_group{custom_options_, "tls"}
       .add<std::string>("key-file,k", "path to the private key file")
       .add<std::string>("cert-file,c", "path to the certificate file");
   }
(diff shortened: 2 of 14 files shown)


examples/CMakeLists.txt
```diff
@@ -94,6 +94,9 @@ add_net_example(http client)
 add_net_example(http rest)
 add_net_example(http time-server)
 
+add_net_example(multipart reader)
+add_net_example(multipart writer)
```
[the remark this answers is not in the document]
shariarriday: The manual has not been updated yet in this PR. It will be updated once the examples are confirmed.

Neverlord (changes requested):
The examples should be located at `examples/http/multipart-{reader|writer}`.

Especially the writer is a non-trivial example. Users need to read the code to understand what it's doing and how to use it. Please provide an apt description at the top what this actually does and how to use it.

The server checks the content type case sensitive, that's too strict. Please consider something like this:

```diff
diff --git a/examples/multipart/reader.cpp b/examples/multipart/reader.cpp
index 27cc3adba..fe49a01d2 100644
--- a/examples/multipart/reader.cpp
+++ b/examples/multipart/reader.cpp
@@ -9,6 +9,7 @@
 #include "caf/actor_system_config.hpp"
 #include "caf/caf_main.hpp"
 #include "caf/defaults.hpp"
+#include "caf/string_algorithms.hpp"

 #include <chrono>
 #include <csignal>
@@ -56,6 +57,11 @@ void set_shutdown_flag(int) {
   shutdown_flag = 1;
 }

+bool is_multipart_form_data(std::string_view content_type) {
+  auto media_type = caf::split_by(content_type, ";").first;
+  return caf::icase_equal(caf::trim(media_type), "multipart/form-data");
+}
+
 } // namespace

 int caf_main(caf::actor_system& sys, const config& cfg) {
@@ -99,7 +105,7 @@ int caf_main(caf::actor_system& sys, const config& cfg) {
             auto value = res.payload();
             auto& header = res.header();
             const auto content_type = header.field("Content-Type");
-            if (!content_type.starts_with("multipart/form-data")) {
+            if (!is_multipart_form_data(content_type)) {
               sys.println("*** expected multipart/form-data content. Found: {}",
                           content_type);
               res.respond(http::status::bad_request);
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+                     caf::defaults::net::http_max_request_size);
+    return result;
+  }
+};
+
+// -- main ---------------------------------------------------------------------
+
+namespace {
+
+volatile std::sig_atomic_t shutdown_flag = 0;
```
Neverlord: Using `volatile` is an instant red flag. So is using a C API if a better C++ alternative exists.


```suggestion
std::atomic<bool> shutdown_flag;
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+void set_shutdown_flag(int) {
+  shutdown_flag = 1;
+}
```
Neverlord:
```suggestion
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto media_type = caf::split_by(content_type, ";").first;
+  return caf::icase_equal(caf::trim(media_type), "multipart/form-data");
+}
+
+} // namespace
+
+int caf_main(caf::actor_system& sys, const config& cfg) {
+  namespace ssl = caf::net::ssl;
+  namespace http = caf::net::http;
+  // Do a regular shutdown for CTRL+C and SIGTERM.
```
Neverlord, line 58:
```suggestion
  // Do a regular shutdown for CTRL+C and SIGTERM.
  auto set_shutdown_flag = [](int) {
    shutdown_flag = true;
  };
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+  auto pem = ssl::format::pem;
+  auto key_file = caf::get_as<std::string>(cfg, "tls.key-file");
+  auto cert_file = caf::get_as<std::string>(cfg, "tls.cert-file");
+  auto max_connections = caf::get_or(cfg, "max-connections",
+                                     default_max_connections);
+  auto max_request_size = caf::get_or(
+    cfg, "max-request-size", caf::defaults::net::http_max_request_size);
+  auto has_key_file = static_cast<bool>(key_file);
+  auto has_cert_file = static_cast<bool>(cert_file);
+  if (has_key_file != has_cert_file) {
```
Neverlord: What's the point of those casts? IMHO hard to read and can be expressed much more naturally:


```suggestion
  if (key_file.has_value() != cert_file.has_value()) {
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+        .context(ssl::context::enable(key_file && cert_file)
+                   .and_then(ssl::emplace_server(ssl::tls::v1_2))
+                   .and_then(ssl::use_private_key_file(key_file, pem))
+                   .and_then(ssl::use_certificate_file(cert_file, pem)))
+        // Bind to the user-defined port.
+        .accept(port)
+        // Limit how many clients may be connected at any given time.
+        .max_connections(max_connections)
+        // Limit the maximum request size.
+        .max_request_size(max_request_size)
```
Neverlord:
```suggestion
        .accept(port)
        .max_connections(caf::get_as<size_t>(cfg, "max-connections"))
        .max_request_size(caf::get_as<size_t>(cfg, "max-request-size"))
```

Note: won't compile yet. Please add the necessary overloads to the DSL. If `get_as` does not return a value the calls should do nothing (i.e., use the builtin default).

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+  signal(SIGINT, set_shutdown_flag);
+  // Read the configuration.
+  auto port = caf::get_or(cfg, "port", default_port);
+  auto pem = ssl::format::pem;
+  auto key_file = caf::get_as<std::string>(cfg, "tls.key-file");
+  auto cert_file = caf::get_as<std::string>(cfg, "tls.cert-file");
+  auto max_connections = caf::get_or(cfg, "max-connections",
+                                     default_max_connections);
+  auto max_request_size = caf::get_or(
+    cfg, "max-request-size", caf::defaults::net::http_max_request_size);
```
Neverlord: Also used only once.

```suggestion
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include <thread>
+
+using namespace std::literals;
+
+// -- constants ----------------------------------------------------------------
+
+static constexpr uint16_t default_port = 8080;
+
+static constexpr size_t default_max_connections = 128;
+
```
Neverlord:
```suggestion
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+        .max_connections(max_connections)
+        // Limit the maximum request size.
+        .max_request_size(max_request_size)
+        .route(
+          "/", http::method::post,
+          [&sys](http::responder& res) {
+            auto value = res.payload();
+            auto& header = res.header();
+            const auto content_type = header.field("Content-Type");
+            if (!is_multipart_form_data(content_type)) {
```
Neverlord: I know I suggested it initially, but thinking a bit more about it: that should be a builtin feature:


```suggestion
            if (!header.has_token("Content-Type", "multipart/form-data"")) {
```

The new `has_token` should scan the passed field whether it contains the passed token (case insensitive).

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+            }
+            auto reader = caf::net::http::multipart_reader{header, value};
+            bool ok = false;
+            auto parts = reader.parse(&ok);
+            if (!ok) {
+              sys.println("*** failed to parse multipart data");
+              res.respond(http::status::bad_request);
+              return;
+            }
+            for (const auto& part : parts) {
```
Neverlord: The `parse` method is rather low-level, an example should use the high-level API:

```suggestion
            reader.for_each([&sys] {
```

examples/http/multipart-reader.cpp
```diff
@@ -0,0 +1,158 @@
...
+        // Launch the server.
+        .start();
+  // Report any error to the user.
+  if (!server) {
+    sys.println("*** unable to run at port {}: {}", port, server.error());
+    return EXIT_FAILURE;
+  }
+  // Wait for CTRL+C or SIGTERM.
+  while (!shutdown_flag)
+    std::this_thread::sleep_for(250ms);
```
Neverlord:
```suggestion
  while (!shutdown_flag) {
    std::this_thread::sleep_for(250ms);
  }
```

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+struct config : caf::actor_system_config {
+  config() {
+    opt_group{custom_options_, "tls"} //
+      .add<std::string>("ca-file", "CA file for trusted servers");
+  }
+};
+
+namespace {
+
+std::string make_content_type(std::string_view boundary) {
```
Neverlord: Should be a convenience method on the writer.

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+    // Generate the content-type header based on the file extension.
+    auto file_extension = file_path.extension().string();
+    auto file_content_type = "text/plain";
+    if (file_extension == ".jpg" || file_extension == ".jpeg")
+      file_content_type = "image/jpeg";
+    else if (file_extension == ".png")
+      file_content_type = "image/png";
+    else if (file_extension == ".pdf")
+      file_content_type = "application/pdf";
```
Neverlord: This should go into a separate function, e.g., `content_type_by_extension`.

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+      n = fread(buffer, 1, sizeof(buffer), file.get());
+    }
+    if (ferror(file.get()) != 0) {
+      sys.println("*** failed while reading file: {}", file_path.string());
+      return EXIT_FAILURE;
+    }
+    writer.append(file_data, [field_name = std::move(*field_name),
+                              filename = std::move(*filename),
+                              file_content_type](auto& headers) {
+      auto name = std::string{"form-data; name=\""} + field_name
```
Neverlord:
```suggestion
      auto name = "form-data; name=\"" + field_name
```

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+  result.reserve(value.size());
+  for (auto ch : value) {
+    if (ch == '\r' || ch == '\n')
+      return std::nullopt;
+    if (ch == '"' || ch == '\\')
+      result.push_back('\\');
+    result.push_back(ch);
+  }
+  return result;
+}
```
Neverlord: Let's just keep it simple and don't support quoted strings.

```suggestion
bool is_sanitized(std::string_view value) {
  return value.find_first_of("\r\n\"\\") == std::string_view::npos;
}
```

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+      return EXIT_FAILURE;
+    }
+    auto file_path = std::filesystem::path{remainder[i + 1]};
+    auto filename = sanitize_quoted_value(file_path.filename().string());
+    if (!filename) {
+      sys.println("*** invalid filename: contains CR/LF");
+      return EXIT_FAILURE;
+    }
+    std::unique_ptr<FILE, int (*)(FILE*)> file{
+      fopen(file_path.string().c_str(), "rb"), fclose};
```
Neverlord: Reading of the file needs to go into its own function. Use seek to allocate a buffer of the right size upfront.

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+                  .context_factory([ca_file, resource]() {
+                    return ssl::emplace_client(ssl::tls::v1_2)()
+                      .and_then(ssl::load_verify_file_if(ca_file))
+                      .and_then(ssl::use_sni_hostname(resource));
+                  })
+                  .connect(resource)
+                  .retry_delay(1s)
+                  .max_retry_count(5)
+                  .connection_timeout(250ms)
+                  .add_header_field("User-Agent", "CAF-Client")
```
Neverlord:
```suggestion
```

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+                              filename = std::move(*filename),
+                              file_content_type](auto& headers) {
+      auto name = std::string{"form-data; name=\""} + field_name
+                  + "\"; filename=\"" + filename + "\"";
+      headers.add("Content-Disposition", name);
+      headers.add("Content-Type", file_content_type);
+    });
+  }
+  auto payload = writer.finalize();
+  auto content_type = make_content_type(writer.boundary());
```
Neverlord: Those variables are used only once: just pass those values to the DSL methods instead.

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,195 @@
...
+                  .connection_timeout(250ms)
+                  .add_header_field("User-Agent", "CAF-Client")
+                  .add_header_field("Content-Type", content_type)
+                  .request(http::method::post, payload);
+  if (!result) {
+    sys.println("*** Failed to initiate connection: {}", result.error());
+    return EXIT_FAILURE;
+  }
+  sys.spawn([res = result->first](event_based_actor* self) {
+    res.bind_to(self).then(
```
Neverlord: Too complicated, there's no benefit of spinning up actors here. Just use `get` with a timeout on the future.

Neverlord (changes requested):

examples/http/multipart-writer.cpp
```diff
@@ -0,0 +1,186 @@
...
+read_file(const std::filesystem::path& file_path) {
+  using file_ptr = std::unique_ptr<FILE, fcloser>;
+  auto file = file_ptr{fopen(file_path.string().c_str(), "rb")};
+  if (!file)
+    return std::nullopt;
+  if (fseek(file.get(), 0, SEEK_END) != 0)
+    return std::nullopt;
+  auto size = ftell(file.get());
+  if (size <= 0)
+    return std::nullopt;
```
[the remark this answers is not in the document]
Neverlord: I don't think sending a zero-byte payload makes any sense.

libcaf_core/caf/string_algorithms.cpp
```diff
@@ -56,7 +56,7 @@ void split(std::vector<std::string_view>& result, std::string_view str,
   split(result, str, std::string_view{&delim, 1}, keep_all);
 }
 
-std::string_view trim(std::string_view str) {
+std::string_view trim(std::string_view str) noexcept {
   auto non_whitespace = [](char c) { return !isspace(c); };
   if (std::ranges::any_of(str, non_whitespace)) {
```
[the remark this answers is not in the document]
Neverlord: Fixed.

Neverlord (approved):

PR #2431, opened 2026-06-22
Allow lazy template instantiation for typed actors

### LINKED ISSUE #2202: Allow self-referring to typed actor handles in signatures
With our shift to concept-based requirement checking, the following example doesn't work anymore:
```
struct PingPongTrait {
  using signatures = caf::type_list<caf::result<void>(caf::typed_actor<PingPongTrait>)>;
};
```
With the error that the `PingPongTrait` is an incomplete type, and compiler can't access its `signatures` member - when defining the `caf::typed_actor<PingPongTrait>` type in the signature. Our older way from CAF v1.1.0 allowed this, probably through postponing the type check. See #2159.

riemass:
With the `requires` clause typed actor signature checking, it became a compile time error for actors to refer to itself in it's signatures (GH-2202). This commit enables that by moves the signature checking into a static assert instead.

Closes #2202.

### DIFF
--- a/libcaf_core/caf/typed_actor.hpp
+++ b/libcaf_core/caf/typed_actor.hpp
@@ -16,6 +16,7 @@
 #include "caf/intrusive_ptr.hpp"
 #include "caf/stateful_actor.hpp"
 #include "caf/type_id_list.hpp"
+#include "caf/typed_actor_pack.hpp"
 #include "caf/typed_actor_view_base.hpp"
 #include "caf/typed_behavior.hpp"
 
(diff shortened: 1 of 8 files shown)


libcaf_core/caf/typed_actor.hpp
```diff
@@ -27,11 +28,12 @@ namespace caf {
...
 /// Identifies a statically typed actor.
 template <class... Ts>
-  requires typed_actor_pack<Ts...>
 class typed_actor : detail::comparable<typed_actor<Ts...>>,
                     detail::comparable<typed_actor<Ts...>, actor>,
                     detail::comparable<typed_actor<Ts...>, actor_addr>,
                     detail::comparable<typed_actor<Ts...>, strong_actor_ptr> {
+  static_assert(typed_actor_pack<Ts...>);
+
 public:
```
Neverlord: We don't put stuff before the first access modifier. Same for the other places.

```suggestion
public:
  static_assert(typed_actor_pack<Ts...>);

```

Neverlord (changes requested):
Glad that it turned out to be small diff. Just a comment regarding coding style.

Neverlord (approved):

PR #2436, opened 2026-06-27
Verify HTTP server overload responds with HTTP 503

riemass:
Followup to #2236.

### DIFF
--- a/libcaf_net/caf/net/http/with.test.cpp
+++ b/libcaf_net/caf/net/http/with.test.cpp
@@ -4,8 +4,10 @@
 
 #include "caf/net/http/with.hpp"
 
+#include "caf/test/scenario.hpp"
 #include "caf/test/test.hpp"
 
+#include "caf/net/http/response_header.hpp"
 #include "caf/net/middleman.hpp"
 #include "caf/net/socket.hpp"
 #include "caf/net/socket_guard.hpp"
@@ -15,6 +17,8 @@
 
 #include "caf/actor_system.hpp"
 #include "caf/actor_system_config.hpp"
+#include "caf/async/blocking_consumer.hpp"
+#include "caf/defaults.hpp"
 #include "caf/detail/connector.hpp"
 #include "caf/detail/scope_guard.hpp"
 #include "caf/fwd.hpp"
@@ -23,6 +27,7 @@
 #include <chrono>
 #include <future>
 #include <latch>
+#include <source_location>
 #include <thread>
 
 using namespace caf;
@@ -73,6 +78,45 @@ private:
 using error_log_ptr = std::shared_ptr<error_log>;
 
 using latch_ptr = std::shared_ptr<std::latch>;
+
+struct raw_http_response {
+  net::http::response_header header;
+  std::string body;
+};
+
+raw_http_response read_http_response(net::stream_socket fd,
+                                     const std::source_location& loc
+                                     = std::source_location::current()) {
+  auto& self = test::runnable::current();
+  auto require_readable = [&self, fd, loc] {
+    auto err = net::receive_timeout(fd, 1s);
+    self.require(!err.valid(), loc);
+  };
+  std::string raw_header;
+  byte_buffer buf;
+  buf.resize(1);
+  while (raw_header.find("\r\n\r\n") == std::string::npos) {
+    require_readable();
+    auto bytes_read = net::read(fd, buf);
+    self.require_eq(bytes_read, ptrdiff_t{1}, loc);
+    raw_header.push_back(
+      static_cast<char>(std::to_integer<unsigned char>(buf[0])));
+  }
+  raw_http_response result;
+  auto parse_result = result.header.parse(raw_header);
+  self.require_eq(parse_result.first, net::http::status::ok, loc);
+  auto content_length = result.header.content_length().value_or(size_t{0});
+  result.body.reserve(content_length);
+  while (result.body.size() < content_length) {
+    require_readable();
+    buf.resize(content_length - result.body.size());
+    auto bytes_read = net::read(fd, buf);
+    self.require_gt(bytes_read, ptrdiff_t{0}, loc);
+    result.body.append(reinterpret_cast<const char*>(buf.data()),
+                       static_cast<size_t>(bytes_read));
+  }
+  return result;
+}
 
 // std::latch has no timed wait; poll try_wait() until timeout.
 static bool latch_count_down_and_wait_for(latch_ptr lptr,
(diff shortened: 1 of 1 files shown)


libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -370,19 +415,73 @@ TEST("GH-2226 regression") {
...
+        for (auto i = size_t{0}; i < defaults::flow::buffer_size; ++i) {
+          net::http::request req;
+          auto result = consumer.pull(async::delay_errors, req, 1s);
+          require_eq(result, async::read_result::ok);
+          req.respond(net::http::status::ok, "text/plain", "ok");
+        }
+        request_buffer->close();
+        net::http::request dummy;
+        require_eq(consumer.pull(async::delay_errors, dummy, 1s),
+                   async::read_result::stop);
```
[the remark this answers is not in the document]
riemass: This will cause an ASAN memory leak described in #2437.

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
 
 using latch_ptr = std::shared_ptr<std::latch>;
 
+struct raw_http_response {
```
Neverlord, line 82: What's raw about this? It has already been parsed.

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
...
+  };
+  std::string raw_header;
+  byte_buffer buf;
+  buf.resize(1);
+  while (raw_header.find("\r\n\r\n") == std::string::npos) {
+    require_readable();
+    auto bytes_read = net::read(fd, buf);
+    self.require_eq(bytes_read, ptrdiff_t{1}, loc);
+    raw_header.push_back(
+      static_cast<char>(std::to_integer<unsigned char>(buf[0])));
```
Neverlord, line 103:
```suggestion
    raw_header.push_back(std::to_integer<char>(buf[0]));
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
...
+                                     const std::source_location& loc
+                                     = std::source_location::current()) {
+  auto& self = test::runnable::current();
+  auto require_readable = [&self, fd, loc] {
+    auto err = net::receive_timeout(fd, 1s);
+    self.require(!err.valid(), loc);
+  };
+  std::string raw_header;
+  byte_buffer buf;
+  buf.resize(1);
```
Neverlord, line 97:
```suggestion
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
...
+  while (raw_header.find("\r\n\r\n") == std::string::npos) {
+    require_readable();
+    auto bytes_read = net::read(fd, buf);
```
Neverlord, line 100: We don't need a heap-allocated buffer for a single byte.

```suggestion
    auto buf = std::byte{};
    auto bytes_read = net::read(fd, std::span{&buf, 1});
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
...
+    self.require_eq(bytes_read, ptrdiff_t{1}, loc);
```
Neverlord, line 101:
```suggestion
    self.require_eq(bytes_read, 1, loc);
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -74,6 +79,45 @@ using error_log_ptr = std::shared_ptr<error_log>;
...
+    raw_header.push_back(
+      static_cast<char>(std::to_integer<unsigned char>(buf[0])));
+  }
+  raw_http_response result;
+  auto parse_result = result.header.parse(raw_header);
+  self.require_eq(parse_result.first, net::http::status::ok, loc);
+  auto content_length = result.header.content_length().value_or(size_t{0});
```
Neverlord, line 108:
```suggestion
  auto content_length = result.header.content_length().value_or(0u);
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -370,19 +415,73 @@ TEST("GH-2226 regression") {
   check_eq(elog->errors(), std::vector<error>{});
 }
 
+SCENARIO("server responds with 503 unavailable when queue is overloaded") {
```
Neverlord, line 418:
```suggestion
SCENARIO("a server responds with status 503 on queue overloads") {
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -370,19 +415,73 @@ TEST("GH-2226 regression") {
...
+  GIVEN("a HTTP server queueing incoming requests asynchronously") {
+    auto overflow_count = size_t{2};
+    auto [server_fd, client_fd] = unbox(net::make_stream_socket_pair());
+    net::socket_guard client_guard{client_fd};
+    std::optional<async::consumer_resource<net::http::request>> requests;
```
Neverlord, line 423: The resource has a default constructor, can be assigned and has a `valid()` method. No need to wrap this into an optional.

```suggestion
    async::consumer_resource<net::http::request> requests;
```

libcaf_net/caf/net/http/with.test.cpp
```diff
@@ -370,19 +415,73 @@ TEST("GH-2226 regression") {
...
+      }
+      THEN("the two overflow requests receive 503 responses") {
+        for (auto i = size_t{0}; i < overflow_count; ++i) {
+          auto response = read_http_response(client_fd);
+          check_eq(response.header.status(), uint16_t{503});
+          check_eq(response.header.status_text(), "Service Unavailable");
+          check_eq(response.body,
+                   "service unavailable: request could not be queued");
+        }
+      }
```
Neverlord, line 450: Am I reading this wrong? We are writing n+2 requests. The last two should overflow. And then the server is sending the errors first?

Isn't all of this using the same socket? The server must process the requests in order. How else can the client match responses to requests? So I would expect n times a valid response and then 2 errors. Per connection, the server cannot start processing a new request before it has replied to the most recent one.

Neverlord (changes requested):

PR #2439, opened 2026-07-09
Move `make_accept_handler` to the detail namespace and export

riemass:
The intent is to export the `make_accept_handler` function for third-party libraries to use. The symbol is still in the `detail` namespace, so third-parties depending on it should do it at own risk.


Neverlord (approved):
`caf::detail` is of course off limits for third parties. This is for our own use.

Issue #2471, opened 2026-08-22
`caf::flow`: backpressure, lifecycle and documentation defects


shariarriday:
## Bugs

- [ ] Make `interval::request(0)` a no-op — it arms the timer without demand and
  `fire()` underflows `demand_` to `SIZE_MAX`, so the operator emits forever
  (`op/interval.cpp`).
- [ ] Honour `from_external` in `stream_bridge_sub::do_dispose` — deliver
  `on_error(sec::disposed)`, else `release_later()`; today the observer gets no
  terminal event and is never released (`detail/stream_bridge.cpp:142`).
- [ ] Specify the `request(0)` contract on `subscription::request` and apply it
  uniformly — it aborts `ucast`/`mcast` under runtime checks, underflows
  `interval`, and is forwarded blindly by `zip_with`. Specify demand overflow
  behavior at the same time.
- [ ] Make `zip_with` credit each input only up to `demand_ - buf.size()`
  instead of forwarding the full demand to all inputs (`op/zip_with.hpp:89`).
- [ ] Document that `publish` gates upstream credit on its slowest subscriber,
  and consider an opt-in overflow policy for slow subscribers
  (`op/publish.hpp:143`).
- [ ] Document and test `buffer`'s deferral of terminal events while a partial
  batch is pending (`op/buffer.hpp`).

## Documentation — `manual/core/DataFlows.rst`

- [ ] Specify the backpressure contract: who calls `request`, whether
  `request(0)` is legal, whether demand saturates, how lossy operators spend
  demand, and how `cancel` differs from an external `dispose`.
- [ ] Replace `dispose` with `cancel` in the `subscription` description.
- [ ] Correct the `merge` concurrency default — eight applies only to the
  zero-argument form; the variadic form uses `sizeof...(Inputs) + 1`.
- [ ] Fix the "Prefix and Tail" section to describe `prefix_and_tail`, and note
  that it emits nothing when the source ends before `n` items.
- [ ] Replace the stale names `callable_source` and `limit_step` with
  `gen::from_callable` and `step::take`.
- [ ] Classify the APIs as hot, cold, connectable, single-subscribe or
  replaying (`multicaster`, `from_resource`, `cache`, `publish`, `ref_count`).
- [ ] Document the surprising terminal semantics of `element_at`, `to_vector`,
  `on_error_return`, `on_error_return_item` and `reduce`.
- [ ] Cover the missing APIs: the source factories, `zip_with`,
[...]

PR #2475, opened 2026-08-29
Add guard for request handling

shariarriday:
Relates https://github.com/actor-framework/actor-framework/issues/2471


libcaf_core/caf/flow/subscription.hpp
```diff
@@ -133,10 +135,12 @@ class CAF_CORE_EXPORT subscription {
...
   }
 
-  /// Signals demand for @p n more items.
+  /// Signals demand for @p n more items. Passing `0` is a no-op, i.e., it adds
+  /// no demand and has no observable side effects.
   /// @pre `valid()`
   void request(size_t n) {
-    pimpl_->request(n);
+    if (n > 0)
+      pimpl_->request(n);
```
Neverlord, line 142: I don't think we need any other check besides this one, or can you find a counter-example (outside of tests) where we hold a pointer directly instead of going through a `subscription`? As it stands, I think the checks inside the operators are purely redundant. Maybe we can put a `CAF_ASSERT` into the operators.

libcaf_core/caf/flow/subscription.hpp
```diff
@@ -39,7 +39,9 @@ class CAF_CORE_EXPORT subscription {
     /// and thus can clean up the subscription immediately.
     virtual void cancel() = 0;
 
-    /// Signals demand for `n` more items.
+    /// Signals demand for `n` more items. Calling this member function with
+    /// `n == 0` is a no-op: it neither adds demand nor triggers any observable
+    /// side effect such as arming a timer or scheduling work.
```
Neverlord:
```suggestion
    /// Signals demand for `n` more items.
    /// @pre `n > 0`
```

We enforce this in the `request` on the `subscription` object. Hence, we don't need to re-check in every single implementation (except maybe putting a `CAF_ASSERT`).

Neverlord (changes requested):

libcaf_core/caf/flow/op/cell.test.cpp
```diff
@@ -165,4 +165,31 @@ SCENARIO("a failed cell emits zero item") {
   }
 }
 
+SCENARIO("requesting zero items from a cell is a no-op") {
```
Neverlord: Please think about what you are testing here. There's an assert in the cell for `n > 0`. The only reason this doesn't crash is because `subscription` will discard any `request` with `n == 0` and not forward it to the implementation. This test doesn't add any useful coverage. The other new tests probably as well. Please make a pass over the PR to see what's actually worth keeping. Then rebase and squash.

Neverlord (changes requested):

libcaf_core/caf/flow/op/cell.test.cpp
shariarriday: Yes, there are still some pending issues and I did not re-request the PR review for that reason. I am converting the PR to draft for now to avoid confusion.

